PR #51, opened 2015-04-27
Removes old source control bindings and VSTestHost.

zooba:
We don't need the old source control bindings any more, and VSTestHost now has its own repo.


Common/Tests/MockVsTests/MockVsTests.csproj
```diff
@@ -10,10 +10,6 @@
     <RootNamespace>Microsoft.VisualStudioTools.MockVsTests</RootNamespace>
     <AssemblyName>MockVsTests</AssemblyName>
     <FileAlignment>512</FileAlignment>
-    <SccProjectName>SAK</SccProjectName>
```
crwilcox: :+1:

Issue #42, opened 2015-04-23
Azure remote debugging - breaking doesn't work


int19h:
When remotely attached to an Azure website, hitting a breakpoint causes debugger to enter the break mode, but there's nothing in call stack or locals, and the editor says "Code not running".

PR #54, opened 2015-04-27
#42 Azure remote debugging - breaking doesn't work

zooba:
We've started trying to read the source file when listing frames, but we weren't handling the case where we couldn't load the file.


crwilcox:
:+1: Signing off.

PR #58, opened 2015-04-27
Add assembly catalog _after_ verifying that the assembly can be loaded.

zooba:
This fixes test errors where an assembly can't be loaded lazily, which we detect in GetTypes but it's still in the catalog, so MEF fails later on.


crwilcox:
:+1: lgtm

PR #59, opened 2015-04-27
Updates the environment list to use PNGs

zooba:
Updates the environment list to use VS-created PNGs instead of XAML paths. This structure will eventually let us switch to the new image format in VS 2015, but for now we will simply use static raster images.


int19h:
:+1:

crwilcox:
:+1:

PR #61, opened 2015-04-28
Fix #15: [Python view] doesn't work in VS 2015

### LINKED ISSUE #15: [Python view] doesn't work in VS 2015
When expanding [Python View] nodes in debugger, there are no children, and in debug builds there's an assertion failure:

> GetChildren called on an unsupported DkmEvaluationResult.
>   at Microsoft.PythonTools.DkmDebugger.ExpressionEvaluator.GetChildren(DkmEvaluationResult result, DkmWorkList workList, Int32 initialRequestSize, DkmInspectionContext inspectionContext, DkmCompletionRoutine`1 completionRoutine) in C:\TFS\xTVS\Python_Main\Python\Product\Debugger\DkmDebugger\ExpressionEvaluator.cs:line 516
>   at Microsoft.PythonTools.DkmDebugger.LocalComponent.Microsoft.VisualStudio.Debugger.ComponentInterfaces.IDkmLanguageExpressionEvaluator.GetChildren(DkmEvaluationResult result, DkmWorkList workList, Int32 initialRequestSize, DkmInspectionContext inspectionContext, DkmCompletionRoutine`1 completionRoutine) in C:\TFS\xTVS\Python_Main\Python\Product\Debugger\DkmDebugger\LocalComponent.cs:line 423
>   at Microsoft.VisualStudio.Debugger.EntryPoint.IDkmLanguageExpressionEvaluator_GetChildren(IntPtr pvClassInfo, IntPtr Result, IntPtr WorkList, Int32 InitialRequestSize, IntPtr InspectionContext, IntPtr CompletionRoutine)

int19h:
Add a new simplified and streamlined implementation of [Python view] debugger nodes using new custom visualizer features in VS 2015.

Update debugger project to conditionally select the proper implementation depending on targeted VS versions. Use XSLT for .vsdconfigxml files to enable conditional exclusion of elements.

Provide an expression for [C++ view] nodes such that they can be added to the Watch window in VS 2015.


int19h:
Note that these changes require the corresponding fixes on VS side, which are _not_ in VS 2015 RC (but will be in some future VS 2015 build). For Python View it doesn't really matter because it already doesn't work before the change, and still won't work after the change until that new VS build arrives. But for C++ View, this will break what is currently working on 2015 RC. So it might be worth postponing this PR until we have a corresponding VS build.

zooba:
+1

crwilcox:
:+1:

Issue #46, opened 2015-04-24
Debugger fails to run with Stackless


moigagoo:
Using the latest dev version of PTVS, PyPy3 2.4 (32 bit), VS 2015 CTP, Windows 10 TP.

This is the traceback (http://i.imgur.com/X3qNp3F.png):

Traceback (most recent call last):
  File "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\Microsoft\Python Tools for Visual Studio\2.2\visualstudio_py_launcher.py", line 25, in <module>
    import visualstudio_py_debugger
  File "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\Microsoft\Python Tools for Visual Studio\2.2\visualstudio_py_debugger.py", line 65, in <module>
    import stackless
  File "C:\PyPy3\lib_pypy\stackless.py", line 542, in <module>
    _init()
  File "C:\PyPy3\lib_pypy\stackless.py", line 538, in _init
    tasklet._init.__func__(_main_tasklet, label='main')
AttributeError: 'function' object has no attribute '__func__'

Internal error detected. Please copy the above traceback and report at
http://go.microsoft.com/fwlink/?LinkId=293415

Press Enter to close. . .

PR #65, opened 2015-04-30
#46 Debugger fails to run with Stackless

zooba:
Updates debugger for changed API in stackless to follow http://stackless.readthedocs.org/en/latest/library/stackless/debugging.html


int19h:
:+1:

crwilcox:
:+1: Looks good.  Would be nice if GH didn't make a bunch of odd choices on what was edited.

PR #89, opened 2015-05-04
Adding support for UAP based Python applications

juanyaw:
Changes to PTVS to support building ,deploying and debugging Python applications for remote UAP Win10 devices.  Also adding a background service template to support IoT platform.


Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -627,6 +634,8 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
             long tid = stream.ReadInt64();
             string output = stream.ReadString();
 
+            System.Diagnostics.Debug.WriteLine(output);
```
zooba: Deliberate?

Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -1043,16 +1052,14 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
...
-
-                            string newFile = Path.Combine(mapTo, file.Substring(len));
-                            Debug.WriteLine(String.Format("Filename mapped from {0} to {1}", file, newFile));
-                            return newFile;
+                        int len = mapFrom.Length;
+                        if (!mappingInfo[0].EndsWith("\\")) {
+                            len++;
                         }
+
+                        string newFile = Path.Combine(mapTo, file.Substring(len));
```
zooba: I think that `mappingInfo[0]` is being used on line 1056 where it should be `mapFrom`, but probably we should just be using `file.Substring(mapFrom.Length).TrimStart('\\', '/')`. Thoughts, @int19h?

Python/Product/PythonTools/PythonTools/Project/ImportWizard/ProjectCustomizations.cs
```diff
@@ -193,4 +193,28 @@ class GenericWebProjectCustomization : ProjectCustomization {
...
     }
+
+    class UapProjectCustomization : ProjectCustomization {
+        public static readonly ProjectCustomization Instance = new UapProjectCustomization();
+
+        private UapProjectCustomization() { }
+
+        public override string DisplayName {
+            get {
+                return SR.GetString(SR.ImportWizardGenericWebProjectCustomization);
```
zooba: You'll want a new resource string here

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -167,6 +167,19 @@ public PythonProjectNode(IServiceProvider serviceProvider)
...
         }
 
+        protected internal override void SetCurrentConfiguration() {
+            base.SetCurrentConfiguration();
+
+            if (!IsProjectOpened)
+                return;
+
+            if (this.IsAppxPackageableProject()) {
+                EnvDTE.Project automationObject = GetAutomationObject() as EnvDTE.Project;
```
zooba: Don't do a safe cast unless you're going to check the result. InvalidCastExceptions are more helpful than NullReferenceExceptions. (And if this code came from elsewhere, let's take the opportunity to fix it now.)

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
```diff
@@ -0,0 +1,118 @@
+﻿using System;
```
zooba: Copyright header

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
```diff
@@ -0,0 +1,118 @@
...
+                                dkmProcess.AddExceptionTrigger(RemoteDebugExceptionGuid, debugTrigger);
+                            }
+                        }
+                    }
+                }
+
+                return VSConstants.S_OK;
+            } finally {
+                if (pEngine != null && Marshal.IsComObject(pEngine)) {
+                    Marshal.ReleaseComObject(pEngine);
```
zooba: ISTR that we shouldn't explicitly release COM objects... did @DinoV say that?

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
```diff
@@ -0,0 +1,118 @@
...
+                }
+            }
+        }
+
+        void IDkmExceptionTriggerHitNotification.OnExceptionTriggerHit(DkmExceptionTriggerHit hit, DkmEventDescriptorS eventDescriptor) {
+            var remoteProcessTask = default(System.Threading.Tasks.Task);
+
+            using (var evt = new System.Threading.ManualResetEvent(false)) {
+
+                ThreadHelper.Generic.Invoke(() => {
```
zooba: This is a synchronous invocation, so the event is pointless. Switch to InvokeAsync, or if you can get to a service provider from here (with an comment explaining that the global provider is the only option if that's the case) then use our `GetUIThread` extension method. Either of these has better re-entrancy handling than synchronous Invoke.

Python/Product/Uap/Interpreter/PythonUapInterpreter.cs
```diff
@@ -0,0 +1,237 @@
...
+            if (factory == null) {
+                // We have been disposed already, so ignore this event
+                return;
+            }
+
+            _typeDb = factory.GetCurrentDatabase();
+
+            if (_references != null) {
+                _typeDb = _typeDb.Clone();
+                foreach (var reference in _references) {
```
zooba: This whole loop seems to be a nop...

Python/Product/Debugger/Debugger/PythonProcess.cs
int19h: I actually wonder if it wouldn't all be simpler if mapFrom was always forced to end with a slash, and normalized before we even get here.

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
int19h: It's dangerous in general because you don't know whether the RCW that you get is shared with someone else or not, especially if you don't control its creation. We don't do it in our debugger code, and I think it's best to avoid it here as well.

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
int19h: The invoked part has to run before the outer method returns, because at this point the debuggee is paused, and returning will unpause it. So this really needs a synchronous Invoke or something equivalent that will block the caller.

Python/Product/Uap/Interpreter/PythonUapInterpreter.cs
```diff
@@ -0,0 +1,237 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PythonTools.Analysis;
+using Microsoft.PythonTools.Interpreter;
+using Microsoft.VisualStudioTools;
+
+namespace Microsoft.PythonTools.Uap.Interpreter {
+    class PythonUapInterpreter : IPythonInterpreter, IPythonInterpreterWithProjectReferences2, IDisposable {
```
zooba: Other than preventing references (which would be better done by removing `IPythonInterpreterWithProjectReferences2`), what value does this class add over the default interpreter class? If you remove the `MakeInterpreter` override in the factory then you'll get the default for free.

Python/Product/Uap/Interpreter/PythonUapInterpreterFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Microsoft.PythonTools.Uap.Interpreter {
+    class PythonUapInterpreterFactory : PythonInterpreterFactoryWithDatabase {
+        public const string InterpreterGuidString = "{86767848-40B4-4007-8BCC-A3835EDF0E69}";
+        public static readonly Guid InterpreterGuid = new Guid(InterpreterGuidString);
+
+        public PythonUapInterpreterFactory() 
+            : base(
+                  InterpreterGuid, 
+                  "Python UAP",
+                  new InterpreterConfiguration(new Version(3, 5)),
```
zooba: I suspect you'll come to regret hard-coding the version number. At the very least, pass it in from the factory provider, even if you can't currently detect what versions are available.

Python/Product/Uap/Interpreter/PythonUapInterpreterFactoryProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+        public PythonUapInterpreterFactoryProvider() {
+        }
+
+        private void DiscoverFactories() {
+            if (_factories == null) {
+                _factories = new HashSet<IPythonInterpreterFactory>();
+
+                _factories.Add(new PythonUapInterpreterFactory());
+
+                if (InterpreterFactoriesChanged != null) {
```
zooba: Capture `InterpreterFactoriesChanged` into a local before testing and invoking it.

Python/Product/Uap/Project/PythonUapProjectConfig.cs
```diff
@@ -0,0 +1,961 @@
...
+                    return RemoteProcessAttachAsync(
+                         appPackageDebugTarget[0].bstrRemoteMachine,
+                         debugId.ToString(),
+                         debugPort,
+                         sourceDir,
+                         targetDir);
+                };
+
+                int result = debugger.AdviseDebugEventCallback(Debugger.PythonRemoteDebugEvents.Instance);
+                System.Diagnostics.Debug.Assert(result == VSConstants.S_OK, string.Format(System.Globalization.CultureInfo.CurrentCulture, "Failure {0}", result));
```
zooba: What happens if this fails in a release build? We should handle the error case somehow (even throwing an exception that will trigger error reporting is better than silently trying to continue, probably into a deadlock).

Python/Product/Uap/Templates/Projects/BackgroundService/StartupTask.py
```diff
@@ -0,0 +1 @@
+print('Hello World')
```
zooba: Is there anything better we can start users off with? Do they need some sort of loop or some imports? If not, we'd prefer a blank file rather than a useless print statement.

Python/Setup/PythonTools/PythonToolsFiles.proj
```diff
@@ -135,7 +136,11 @@
     <File Include="$(BuildOutputRoot)Binaries\Analyzer\CompletionDB\*.idb">
       <InstallDirectory>CompletionDB</InstallDirectory>
     </File>
-  
+
+    <File Include="*.idb">
+      <InstallDirectory>CompletionDB</InstallDirectory>
+    </File>
+
```
zooba: This section looks like an incorrect merge, as it should be covered by the File element above it.

zooba: Oh I see, you've got new default .idb files. Probably best to just let them be analyser normally, but if they're really necessary, put them in the `Analyzer/CompletionDB` folder.

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
zooba: It also blocks on an event defined just for this purpose. Sticking with the event is better IMHO

zooba:
Also @juanyaw, if you join the Microsoft org then you won't get prompted for a CLA :)

Python/Setup/PythonTools/PythonToolsFiles.proj
juanyaw: Yes, these are analyzed offline since it has to be done inside of an app to get these libraries updated, but I should probably remove this since it was a stop-gap to have intellisense for these libraries.  They should/would be included with whatever package pulled them in.

Python/Product/Debugger/Debugger/PythonProcess.cs
juanyaw: This was a debugging session addition that didn't get removed.  Removing

Python/Product/PythonTools/PythonTools/Project/ImportWizard/ProjectCustomizations.cs
juanyaw: Done

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
juanyaw: Yes, this was copied from code in SharedProject all over the place there :)

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
juanyaw: Done

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
juanyaw: Done

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
juanyaw: The documentation here leads you to think that it is best to Release objects (if I'm reading it right):

https://msdn.microsoft.com/en-us/library/bb145585(v=vs.140).aspx

Python/Product/Uap/Interpreter/PythonUapInterpreter.cs
juanyaw: The default will interrogate the DLL (PYD) on the debugger machine.  Right now, the DLL cannot be loaded outside an app, hence this prevents the dll from being loaded and interrogated.  I believe this is what the default interpreter does.

juanyaw:
Joined Microsoft org and updated based on comments where I could so far

Python/Product/Uap/Interpreter/PythonUapInterpreter.cs
zooba: Adding references doesn't actually do anything other than provide IntelliSense, unless you've got something in your .targets to pull them in (typically they need to be part of a Python environment anyway, so there's little value and much confusion in doing this). Better to prevent adding references at all.

We don't currently, but we ought to hide the References node if the project doesn't support it. Right now you've got a change to omit the node it looks like, but we can generalise that better.

Python/Product/Uap/Debugger/PythonRemoteDebugEvents.cs
zooba: Yeah, that's pretty explicit. They'll be released by the finalizer, but the call should always be coming directly from COM and so we should always have a new ref to release (then again, I just used "should" twice, which probably means we should play it safe and let the finalizer handle it)

Python/Product/Uap/Interpreter/PythonUapInterpreter.cs
juanyaw: That is not exactly right in this case.  References to files end up getting included in the appx package, which is needed for extension modules that are not already included.  This also allows for extension module project references to be included and to be built along with the appx as well.  Even though we are not currently reflecting upon the reference for intellisense because of limitations, it is still a valuable way to include the additional extension modules.  We could just say that you need to just add the file as content, but references seemed to flow as a concept...especially for project references.

Python/Product/Debugger/Debugger/PythonProcess.cs
juanyaw: Just noting here, but this is same code that was already there, there was just a double if statement that I removed.  Unfortunately, it seems to be highlighting the whole block instead of just the duplicate if that was removed.  Your comments are probably still valid though :)

juanyaw:
Now I just noticed that I have to update from UAP to UWP.  In progress

juanyaw:
Is there any other pending action I should take with this review?

zooba:
Nope, you're good. But do keep an eye on our issue tracker for at least a while in case we need to ping you about anything.

PR #95, opened 2015-05-06
Adds lightbulb support for VS 2015.

zooba:
Also improves/simplifies the old smart tags slightly.


int19h:
:+1:

DinoV:
This breaks the Dev14 build because SmartTagTests.cs isn't updated, and it doesn't include test cases for the new lightbulb functionality.

PR #96, opened 2015-05-06
Decorator improvements

zooba:
Fixes a bug for tests not being detected due to skip decorators, allows opting out of full decorator analysis (possibly useful for the test adapter, but I haven't committed to that yet), and removes the IronPython dependency of the main set of analysis tests.


Python/Tests/Analysis/BaseAnalysisTest.cs
```diff
@@ -83,6 +85,14 @@ public BaseAnalysisTest(IPythonInterpreterFactory factory)
             return AnalysisLimits.GetDefaultLimits();
         }
 
+        protected virtual bool SupportPython3 {
```
int19h: This should be "SupportsPython3" per standard "must answer a question" naming rule for bool properties.

int19h:
:+1:

crwilcox:
:+1:  This looks like it will also fix the issues discovering tests with decorators for httpretty etc.?

Issue #99, opened 2015-05-07
Unit Tests named 'runTest' aren't detected by Test Explorer.


hawkerm:
[According to the documentation:](https://docs.python.org/2/library/unittest.html)

> The simplest TestCase subclass will simply override the runTest() method in order to perform specific testing code:

Tests can basically be named **test*** as convention for the runner OR they can override the **runTest** method as mentioned in that fourth line of section 25.3.4.

I can run my test cases successfully in my script through unittest.main() if I call the method **runTest** or **test_whatever** on my TestCase.  But, PTVS only picks up the *test_whatever* in my Test Explorer.

It should pick up the **runTest** method as well and give it the name of its class as the test case name.

Thanks!

hawkerm:
Haven't gotten to the Pull Request part of learning GitHub yet, but looks like in PTVS/Python/Product/TestAdapter/TestAnalyzer.cs, the last function (line 160) should have the search for 'runTest':

```C#
        private static IEnumerable<KeyValuePair<string, IAnalysisSet>> GetTestCaseMembers(
            IPythonProjectEntry entry,
            AnalysisValue classValue
        ) {
            return classValue.GetAllMembers(entry.Analysis.InterpreterContext)
                .Where(v => v.Key.StartsWith("test") || v.Key == "runTest")
                .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
        }
```

Surprised I found this since test was such a generic search term...

crwilcox:
Hey Hawkerm.

You are correct.  The current implementation will only discover methods that match test*.  The docs do seem to indicate though that runTest should also be treated as valid.

I made the change you listed here and added it to our test cases.

I will create a pull request and you can sign off on it.  If you would rather submit a pull request that is fine too.  I wrote some quick instructions below.  

Forking/Pulling:
1) Fork: https://github.com/Microsoft/PTVS/issues/99#fork-destination-box
This will get you a clone of our repo that you have rights to push to.
2) Make the changes in the source.  Add unit tests and run appropriate tests. In this case the TestDiscoverTests.cs file.
4) Commit and Push to your fork like you would any project you have on GitHub
5) Go to https://github.com/Microsoft/PTVS/pulls and create a pull request.

crwilcox:
https://github.com/Microsoft/PTVS/pull/101

PR #101, opened 2015-05-07
#99 'runTest' not discovered

crwilcox:
@hawkerm discovered that we do not match the python docs.  I am
committing his fix on his behalf.  I also have altered our tests to
reflect the change.

### DIFF
--- a/Python/Product/TestAdapter/TestAnalyzer.cs
+++ b/Python/Product/TestAdapter/TestAnalyzer.cs
@@ -162,7 +162,7 @@ namespace Microsoft.PythonTools.TestAdapter {
             AnalysisValue classValue
         ) {
             return classValue.GetAllMembers(entry.Analysis.InterpreterContext)
-                .Where(v => v.Key.StartsWith("test"))
+                .Where(v => v.Key.StartsWith("test") || v.Key.StartsWith("runTest"))
                 .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
         }
     }
--- a/Python/Product/TestAdapter/TestAnalyzer.cs
+++ b/Python/Product/TestAdapter/TestAnalyzer.cs
@@ -157,13 +157,28 @@ namespace Microsoft.PythonTools.TestAdapter {
             }
         }
 
+        /// <summary>
+        /// Get Test Case Members for a class.  If the class has 'test*' tests 
+        /// return those.  If there aren't any 'test*' tests return (if one at 
+        /// all) the runTest overridden method
+        /// </summary>
         private static IEnumerable<KeyValuePair<string, IAnalysisSet>> GetTestCaseMembers(
             IPythonProjectEntry entry,
             AnalysisValue classValue
         ) {
-            return classValue.GetAllMembers(entry.Analysis.InterpreterContext)
-                .Where(v => v.Key.StartsWith("test") || v.Key.Equals("runTest"))
+            var tests = classValue.GetAllMembers(entry.Analysis.InterpreterContext)
+                .Where(v => v.Key.StartsWith("test"))
                 .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
+
+            var runTest = classValue.GetAllMembers(entry.Analysis.InterpreterContext)
+                .Where(v => v.Key.Equals("runTest"))
+                .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
+
+            if (tests.Any()) {
+                return tests;
+            } else {
+                return runTest;
+            }
         }
     }
 }
--- a/Python/Product/TestAdapter/TestAnalyzer.cs
+++ b/Python/Product/TestAdapter/TestAnalyzer.cs
@@ -162,7 +162,7 @@ namespace Microsoft.PythonTools.TestAdapter {
             AnalysisValue classValue
         ) {
             return classValue.GetAllMembers(entry.Analysis.InterpreterContext)
-                .Where(v => v.Key.StartsWith("test") || v.Key.StartsWith("runTest"))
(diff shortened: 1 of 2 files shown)


Python/Product/TestAdapter/TestAnalyzer.cs
```diff
@@ -162,7 +162,7 @@ LocationInfo sourceLocation
             AnalysisValue classValue
         ) {
             return classValue.GetAllMembers(entry.Analysis.InterpreterContext)
-                .Where(v => v.Key.StartsWith("test"))
+                .Where(v => v.Key.StartsWith("test") || v.Key.StartsWith("runTest"))
```
int19h: `StartsWith` is not correct for `runTest` - it must override the method, so it has to use that name exactly, unlike `test_...` pattern.

crwilcox:
FYI:  Wait to merge until @zooba and @hawkerm can look at the changes.

zooba:
If `runTest` is overridden then we should display the class name as the test and pass the class name to the test executor rather than `class.runTest`. From a quick reading of how `unittest` works, `runTest` is not detected as a test method, but having one there may suppress the usual method detection.

Be very sure that the behaviour is consistent between what we show, what we execute/debug, and what the basic command line tools execute (for at least Python 2.7 and 3.2-3.4, 2.6 if possible). I haven't had a chance to test it all, but I'm very sceptical that treating `runTest` as if it were spelled `testrunTest` is the right approach.

int19h:
I experimented with it a little bit from command line, and so long as you use the fully qualified method name to run things (which I believe we do in all cases), `runTest` is no different from any other method. Indeed, you can run any arbitrary method that way, regardless of its name.

When command line is used to run the entire test case class, then the presence of any `test...` method overrides `runTest`, and the latter isn't run at all. So when generating the list of test methods, we might also want to use the same either-or approach.

crwilcox:
My concern is that not discovering it will cause people to ask 'why is runTest not being discovered'.  That being said we should replicate the same as the commandline tools. I will make the changes to make this discover contingent on the discovery of test* tests in the class.

Python/Product/TestAdapter/TestAnalyzer.cs
```diff
@@ -157,13 +157,28 @@ LocationInfo sourceLocation
...
         private static IEnumerable<KeyValuePair<string, IAnalysisSet>> GetTestCaseMembers(
             IPythonProjectEntry entry,
             AnalysisValue classValue
         ) {
-            return classValue.GetAllMembers(entry.Analysis.InterpreterContext)
+            var tests = classValue.GetAllMembers(entry.Analysis.InterpreterContext)
                 .Where(v => v.Key.StartsWith("test"))
                 .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
+
+            var runTest = classValue.GetAllMembers(entry.Analysis.InterpreterContext)
```
int19h: You're calling GetAllMembers twice (and that call won't be lazy, unlike the rest of the query) - might want to reuse that instead.

Python/Product/TestAdapter/TestAnalyzer.cs
```diff
@@ -157,13 +157,28 @@ LocationInfo sourceLocation
...
+                .Where(v => v.Key.Equals("runTest"))
+                .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
```
int19h: This is also repeated now. Query for member type first and stash that away, then do name queries on top of that.

crwilcox: Good point.  I figured the enumeration being delayed that wouldn't make a big difference but I can clean that up a bit.

hawkerm:
Awesome, thanks guys! This looks great. :)

Python/Product/TestAdapter/TestAnalyzer.cs
```diff
@@ -157,13 +157,26 @@ LocationInfo sourceLocation
...
+            var methodFunctions = classValue.GetAllMembers(entry.Analysis.InterpreterContext)
                 .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
+
+            var tests = methodFunctions.Where(v => v.Key.StartsWith("test"));
+            var runTest = methodFunctions.Where(v => v.Key.Equals("runTest"));
+
+            if (tests.Any()) {
+                return tests;
+            } else {
+                return runTest;
```
zooba: In this case, I'd rather return the class object (if we can get its name) than the runTest method. That way you won't see `runTest` in the list of tests - you'll see the test itself (which is an instance of the class, if you've overridden runTest).

crwilcox:
Changes have been made to make the 'runTest' test display by the class name in test explorer which should be more useful.

zooba:
+1

Python/Product/TestAdapter/TestAnalyzer.cs
```diff
@@ -157,13 +160,26 @@ LocationInfo sourceLocation
...
         private static IEnumerable<KeyValuePair<string, IAnalysisSet>> GetTestCaseMembers(
             IPythonProjectEntry entry,
             AnalysisValue classValue
         ) {
-            return classValue.GetAllMembers(entry.Analysis.InterpreterContext)
-                .Where(v => v.Key.StartsWith("test"))
+            var methodFunctions = classValue.GetAllMembers(entry.Analysis.InterpreterContext)
                 .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
+
+            var tests = methodFunctions.Where(v => v.Key.StartsWith("test"));
```
gilbertw: Is it illegal to have both?  What happens if we find something with testSupport but they are using runTest

crwilcox: it is an either or.  test* means runTest won't be run.

PR #104, opened 2015-05-08
Fix #23: Can't debug unit tests on VS2015 RC

### LINKED ISSUE #23: Can't debug unit tests on VS2015 RC
VS 2015 22822.1 D14REL
PTVS 2.2 CL1457733_VS2015_22820

Create python app
Add item, unit test

```
    def test_A(self):
        a = 0
        b = 10
        pass
```

Put a breakpoint on `a = 0`.
Debug test

Occasionally, I get this access violation exception (debugger breaks).

But most of the time, it runs without debugging.  You don't see the "2nd attach" happening at all.  This is worse than with CTP 6 where it would sometimes work as expected (debug and stop at that breakpoint), sometimes get the access violation.

I tried C# and C++ (native) unit test debugging, and haven't had seen any issue yet.

```
System.AccessViolationException was unhandled
Message: An unhandled exception of type 'System.AccessViolationException' occurred in System.ServiceModel.dll
Additional information: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```

int19h:
Fix #23: Can't debug unit tests on VS2015 RC

Use an environment variable to propagate process ID of devenv.exe to the test executor process, instead of relying on a command line switch to the latter that is no longer present in VS 2015.

Do not cache the obtained DTE instance - it does not work when it's passed between threads, and in VS 2015 test executor performs initialization and test running on different threads.


Python/Tests/IronPython/IronPythonAnalysisTest.cs
```diff
@@ -1,517 +1,517 @@
...
-            for (int i = 0; i < len; i++) {
-                Assert.IsTrue(dir[i] is string);
-                result[i] = dir[i] as string;
-            }
-            return result;
-        }
-
-    }
-}
+﻿/* ****************************************************************************
```
int19h: I'm not sure why this shows up as changed. When I have both branches checked out locally and compare the files directly, they're byte for byte identical (i.e. there's no line ending differences or tab differences). Suggestions?

mousetraps: It's probably line endings. Check your autocrlf settings in git - ```git config core.autocrlf```
https://github.com/Microsoft/PTVS/pull/104/files?w=1 will show the changes without whitespace differences.

mousetraps: but unfortunately you can't actually comment in that view :frowning:

mousetraps:
:+1:

huguesv:
I think you can get rid of _dte private member and its only use in Dispose().  Looks good otherwise.

Python/Tests/IronPython/IronPythonAnalysisTest.cs
int19h: That's what I thought initially, but, as noted earlier, diffing the branches locally shows no differences, not even in line endings (I used fc /b).

int19h: Or are you saying that it adjusts the line endings when I check out, and so I don't actually see what's in the branch?

mousetraps: Yes. If you look at the diffs, the [old version](https://github.com/int19h/PTVS/blob/b7d6364a27e6f5ab0959b3693dcfec982360b882/Python/Tests/IronPython/IronPythonAnalysisTest.cs) uses CRLF and the [new version](https://github.com/int19h/PTVS/blob/15c6cbb5ed8f9f12f260062ecd5788b1e96c5c94/Python/Tests/IronPython/IronPythonAnalysisTest.cs) uses LF. This is probably because core.autocrlf is set to 'input' on your machine.

Common/Product/TestAdapter/VisualStudioApp.cs
```diff
@@ -75,6 +67,10 @@ class VisualStudioApp : IDisposable {
         [DllImport("ole32.dll")]
         private static extern int CreateBindCtx(uint reserved, out IBindCtx ppbc);
 
+        public DTE GetDTE() {
+            return GetDTE(_processId);
```
zooba: I believe this can return null - can we check here and get a better exception than a nullref later on? InvalidOp is fine, or make a new one so that we can at least capture the name in crash reports.

zooba:
With the one change (or a good reason not to do it), it's good. Go ahead and merge.

PR #124, opened 2015-05-12
Fixes null reference exception when saving analysis.

zooba:
It's possible (and with a recent change, likely) to have a special callable with no original value. To avoid crashing in this case, we add a check.


zooba:
Another part to this fix coming...

int19h:
:+1:

crwilcox:
:+1:

PR #131, opened 2015-05-12
Fix python tools condition causing build issue

crwilcox:
If Dev14OrLater is not defined (was happening) the old condition causes
a build failure.  Changing this condition to resemble other checks we
have


int19h:
Why is it happening in the first place, though? There should never be a case when it's not set to something. If there is, then that needs to be fixed.

crwilcox:
Going to close this and not merge.

zooba:
It happens when you update the properties file with VS open and don't close/reopen the solution. Then the cached MSBuild data doesn't update to reflect the newly added property.

PR #134, opened 2015-05-13
Use bool instead of string comparison

crwilcox:
While looking through the targets I saw these two lines that could use some touchup.  These changes are basically no-ops but will now fail if we don't have that variable set.  Dev14OrLater should be set everywhere so we would prefer these fail rather than just evaluate in an unexpected way.


int19h:
:+1:

PR #135, opened 2015-05-13
Fix #133: Debugger doesn't handle sys.prefix=='' well

### LINKED ISSUE #133: Debugger doesn't handle sys.prefix=='' well
When `sys.prefix` is set to an empty string (which is what e.g. Sublime Text does for its hosted Python), we treat all Python code in the process as if it was a part of the standard library (because we use `startswith(sys.prefix)` as a check to distinguish standard library and user code). As a result, stepping etc just doesn't work. This is further exacerbated by #132 in attach scenarios, since there's no way to enable stdlib debugging there as a workaround.

int19h:


crwilcox:
:+1:

zooba:
Can we add a test for this case?

PR #138, opened 2015-05-13
Updates skip verification scripts for UWP.

zooba:


int19h:
Can you add this to .gitattributes while you're at it (to enable text diffing for those files)?
```
*.reg diff
```

zooba:
Was just looking into that, but they're UTF-16 with a BOM, so it's not going to work anyway (every reference I've seen so far says that git barfs on anything that isn't ANSI, which was my understanding).

int19h:
They're UTF-16 because that's what regedit exports in by default, but it accepts ASCII as input just fine, so you can just save them as the latter.

zooba:
I think they're still marked as binary because of the previous version and not because of .gitattributes. If it's still an issue we can fix that up later.

int19h:
:+1:

PR #142, opened 2015-05-14
Hide environments

zooba:
Makes it possible to hide environments or make them ineligible to be the default environment.
Makes the UWP environment ineligible to be the default.

### DIFF
--- a/Python/Product/PythonTools/PythonTools.csproj
+++ b/Python/Product/PythonTools/PythonTools.csproj
@@ -271,9 +271,6 @@
     <Compile Include="..\Analysis\IEnumerableExtensions.cs">
       <Link>IEnumerableExtensions.cs</Link>
     </Compile>
-    <Compile Include="..\Analysis\Interpreter\PythonInterpreterFactoryExtensions.cs">
-      <Link>PythonInterpreterFactoryExtensions.cs</Link>
-    </Compile>
     <Compile Include="PythonTools\Repl\IMultipleScopeEvaluator.cs" />
     <Compile Include="PythonTools\Repl\InteractiveWindowProvider.cs" />
     <Compile Include="IPythonToolsToolWindowService.cs" />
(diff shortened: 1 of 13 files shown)


zooba:
@juanyaw may be interested in this too

juanyaw:
Cool, looks good to me.  Thanks for the ordering fix on event suppression as well.

Python/Product/PythonTools/PythonTools/Options/PythonInterpreterOptionsControl.cs
```diff
@@ -60,6 +60,10 @@ public PythonInterpreterOptionsControl()
                 _defaultInterpreter.Items.Clear();
 
                 foreach (var interpreter in _serviceProvider.GetPythonToolsService().InterpreterOptions.Select(x => x.Key).OrderBy(f => f.Description)) {
+                    if (!Microsoft.PythonTools.Interpreter.PythonInterpreterFactoryExtensions.IsUIVisible(interpreter)) {
```
int19h: I'd rather have a `using` + call it as instance method for clarity (and the same for the diff above).

int19h:
:+1:

Python/Product/PythonTools/PythonTools/Options/PythonInterpreterOptionsControl.cs
zooba: There's a naming collision that prevents the extension method from being resolved properly. Looks like I can rename one of the static classes to resolve it though...

Python/Product/VSInterpreters/InterpreterOptionsService.cs
```diff
@@ -154,15 +154,16 @@ sealed class LockInfo : IDisposable {
             try {
                 var store = _settings.GetReadOnlySettingsStore(SettingsScope.Configuration);
                 _providers = LoadProviders(store, serviceProvider);
+
```
crwilcox: nit: extra space

Python/Product/PythonTools/PythonTools.csproj
```diff
@@ -271,9 +271,6 @@
     <Compile Include="..\Analysis\IEnumerableExtensions.cs">
       <Link>IEnumerableExtensions.cs</Link>
     </Compile>
-    <Compile Include="..\Analysis\Interpreter\PythonInterpreterFactoryExtensions.cs">
```
crwilcox: This file still exists.  Any reason to stop compiling it within PythonTools.csproj?

crwilcox:
:+1: 
With fixes/explanations for the above two things

Python/Product/PythonTools/PythonTools.csproj
zooba: It's public in Analysis.dll, so rather than compiling it into multiple DLLs and causing conflicts, I made the one function that was needed public.

PR #143, opened 2015-05-14
Prevents XAML dependencies being updated while iterating over them.

zooba:
Also prevents deadlock when updating object browser.


crwilcox:
:+1:

Issue #34, opened 2015-04-23
XAML designer doesn't work in VS 2015 RC


huguesv:
VS 2015 22823.1 D14REL Enterprise (xlab image)
PTVS 30423.00

Create IronPython WPF app
Double click on .xaml file to open in editor

I get the split view of designer/xaml code but the designer part at the top has a progress bar that never ends.

I don't know if it's that same design host dll issue we've had in the past.

zooba:
I don't see any debug messages or exceptions, and the design host DLL is loaded. This will probably take a bit more investigation, but it's a serious enough regression that we ought to look into it.

zooba:
That commit has fixed us, and there seems to be a module load race condition that will be fixed on the VS side.

PR #144, opened 2015-05-14
#34 XAML designer doesn't work in VS 2015 RC

zooba:
Implements RefType so that the XAML designer no longer aborts initialization when getting our references.

There are other issues due to how the design host handles paths and a possible .NET bug relating to URIs, but this fix should clear up our side of the issue.


huguesv:
lgtm

crwilcox:
:+1:

PR #146, opened 2015-05-14
Prevents exceptions in projection buffers when suggesting missing imports

zooba:
Also fixes the copyright header.


crwilcox:
:+1:

Issue #129, opened 2015-05-12
Python Package icon appears with white background


zooba:
![package icon](https://cloud.githubusercontent.com/assets/1693688/7599163/b7215f16-f8b4-11e4-97d3-1864572dca40.png)

This icon should be coming from the image service and should be adapting to the background colour automatically.

PR #148, opened 2015-05-14
#129 Python Package icon appears with white background

zooba:
Removes failed attempt at correctly handling the background colour.


crwilcox:
:+1:

PR #151, opened 2015-05-14
Improve pip detection

zooba:
Prevents progress bar from getting stuck when pip is not installed.
Improves detection of whether pip is installed so we don't have to check as often.

### DIFF
--- a/Python/Tests/Core/EnvironmentListTests.cs
+++ b/Python/Tests/Core/EnvironmentListTests.cs
@@ -527,11 +527,12 @@ namespace PythonToolsUITests {
                 var environment = list.Environments.Single();
                 var pip = (PipExtensionProvider)list.GetExtensionOrAssert<PipExtensionProvider>(environment);
 
-                Assert.IsFalse(pip.IsPipInstalled().GetAwaiter().GetResult(), "venv should not install pip");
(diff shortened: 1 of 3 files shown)


Python/Tests/Core/EnvironmentListTests.cs
```diff
@@ -527,11 +527,12 @@ public class EnvironmentListTests {
                 var environment = list.Environments.Single();
                 var pip = (PipExtensionProvider)list.GetExtensionOrAssert<PipExtensionProvider>(environment);
 
-                Assert.IsFalse(pip.IsPipInstalled().GetAwaiter().GetResult(), "venv should not install pip");
+                pip.CheckPipInstalledAsync().GetAwaiter().GetResult();
+                Assert.AreEqual(false, pip.IsPipInstalled, "venv should not install pip");
```
crwilcox: Any reason to change these from IsFalse?

crwilcox:
:+1:

Python/Tests/Core/EnvironmentListTests.cs
zooba: `IsPipInstalled` is of type `bool?` and I want error messages to distinguish between `null` and `true`.

PR #154, opened 2015-05-15
Scanning for UWP interpreters and moving CPython UWP targets import

juanyaw:
Moving CPython.targets import into PTVS UWP targets file in lieu of the pyproj
Scanning for UWP interpreters and versions based on installed SDKs

### DIFF
--- a/Python/Product/Uwp/Interpreter/PythonUwpInterpreterFactoryProvider.cs
+++ b/Python/Product/Uwp/Interpreter/PythonUwpInterpreterFactoryProvider.cs
@@ -25,63 +25,72 @@ namespace Microsoft.PythonTools.Uwp.Interpreter {
     class PythonUwpInterpreterFactoryProvider : IPythonInterpreterFactoryProvider {
         private HashSet<IPythonInterpreterFactory> _factories = null;
 
+        private const string PythonUwpSdkTargetsFile = @"DesignTime\CommonConfiguration\Neutral\CPython.targets";
+
         public event EventHandler InterpreterFactoriesChanged;
 
         public PythonUwpInterpreterFactoryProvider() {
         }
 
+        private IDictionary<string, IPythonInterpreterFactory> FindFactoriesFromDirectories(params string[] directories) {
+            var factoryMap = new Dictionary<string, IPythonInterpreterFactory>();
+
+            try {
+                if (directories != null) {
+                    for (int i = 0; i < directories.Length; i++) {
+                        var rootDirectoryInfo = new DirectoryInfo(directories[i]);
+
+                        if (rootDirectoryInfo.Exists) {
+                            foreach (var dirInfo in rootDirectoryInfo.EnumerateDirectories()) {
+                                Version pythonUwpVersion;
+
+                                if (Version.TryParse(dirInfo.Name, out pythonUwpVersion)) {
+                                    var targetsFile = dirInfo.GetFiles(PythonUwpSdkTargetsFile).FirstOrDefault();
+                                    var factoryName = string.Join(" ", rootDirectoryInfo.Name, dirInfo.Name);
+
+                                    if (targetsFile != null) {
+                                        // This will add a new or overwrite factory with new factory found
+                                        // Ordering of the directories means that the last directory specified will 
+                                        // win the battle of conflicting factory names
+                                        factoryMap[factoryName] = new PythonUwpInterpreterFactory(
+                                                new InterpreterConfiguration(
+                                                    dirInfo.FullName,
+                                                    targetsFile.FullName,
+                                                    null,
+                                                    null,
+                                                    null,
+                                                    ProcessorArchitecture.None,
+                                                    pythonUwpVersion,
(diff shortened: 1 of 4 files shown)


Python/Product/Uwp/Interpreter/PythonUwpInterpreterFactoryProvider.cs
```diff
@@ -23,39 +25,68 @@ namespace Microsoft.PythonTools.Uwp.Interpreter {
...
                 _factories = new HashSet<IPythonInterpreterFactory>();
 
-                _factories.Add(new PythonUwpInterpreterFactory(defaultConfig));
-
-                var factoriesChanged = InterpreterFactoriesChanged;
-
-                if (factoriesChanged != null) {
-                    factoriesChanged(this, new EventArgs());
+                if (userSdkInstallDir.Exists) {
+                    foreach (var dirInfo in userSdkInstallDir.EnumerateDirectories()) {
```
zooba: It'll be safer to have this block in its own function (since it's duplicated) and protected with a try/catch. All sorts of random errors can happen when dealing with the file system, and if this code is running off the UI thread (likely) then an error will bring down VS completely

Python/Product/Uwp/Interpreter/PythonUwpInterpreterFactoryProvider.cs
```diff
@@ -23,39 +25,68 @@ namespace Microsoft.PythonTools.Uwp.Interpreter {
...
+                            _factories.Add(
+                                new PythonUwpInterpreterFactory(
+                                    new InterpreterConfiguration(
+                                        dirInfo.FullName,
+                                        targetsFile.FullName,
+                                        null,
+                                        null,
+                                        null,
+                                        ProcessorArchitecture.None,
+                                        new Version(dirInfo.Name),
```
zooba: This is brave, since anyone could make a `CPython.targets` file... `Version.TryParse` would be better, and if it's possible to rename `CPython.targets` to something more unique (even `CPython.UWP.targets` or `CPython.WinRT.targets` would be sufficient) then that'll help protect against rogue packages.

Python/Product/Uwp/Interpreter/PythonUwpInterpreterFactoryProvider.cs
zooba: I also don't see any particular reason not to just iterate over `userSdkInstallDir.EnumerateFiles("CPython.targets", SearchOption.AllDirectories)` here, rather than doing a recursive search within each directory. Combined with ensuring the directory name is a valid version number, that will exclude any top-level .targets files, which would seem to be the only difference.

Python/Product/Uwp/Interpreter/PythonUwpInterpreterFactoryProvider.cs
```diff
@@ -23,39 +25,68 @@ namespace Microsoft.PythonTools.Uwp.Interpreter {
...
-                    null,
-                    null,
-                    ProcessorArchitecture.None,
-                    new Version(PythonVersion),
-                    InterpreterUIMode.CannotBeDefault
-                );
+                var userSdkInstallDir = new DirectoryInfo(Environment.ExpandEnvironmentVariables(@"%LOCALAPPDATA%\Microsoft\VisualStudio\%VISUALSTUDIOVERSION%Exp\Extensions\Microsoft\Python UWP"));
+                var sdkInstallDir = new DirectoryInfo(Environment.ExpandEnvironmentVariables(@"%ProgramFiles(x86)%\Microsoft SDKs\Windows\v10.0\ExtensionSDKs\Python UWP"));
 
                 _factories = new HashSet<IPythonInterpreterFactory>();
```
zooba: Better to do this in the constructor now. Maybe make it a dictionary based on the targets file path so we can avoid re-adding factories every time it's called. (Can't remember off the top of my head whether we require IPythonInterpreterFactory identity to be preserved - I think we do, which means you need to avoid constructing new ones on each call if at all possible.)

juanyaw: This will only be done if _factories is null which won't be case 2nd time around

Python/Product/Uwp/Interpreter/PythonUwpInterpreterFactoryProvider.cs
juanyaw: Moved to own function and check for specific relative path from root folder.  The root folder name is Python UWP which should give enough leads for the search being for a UWP targets file

Python/Product/Uwp/Interpreter/PythonUwpInterpreterFactoryProvider.cs
juanyaw: Fixed to use TryParse

zooba:
Do these extension packages require you to restart VS after installing them?

zooba:
Actually, it doesn't matter, since the provider has to retrigger discovery itself anyway and we currently don't here.

Issue #97, opened 2015-05-06
Trouble 'upgrading' project from PTVS 2.1 to 2.2 RC


hawkerm:
When I loaded my solution from PTVS 2.1 in 2.2 RC the first time I got a 'One-Way' upgrade request notification.  Trying to continue through the prompt, just caused it to pop-up again continually.  

Only after I canceled, and manually reloaded the project from Solution Explorer did it upgrade successfully.

I'm using VS 2013 Update 4.
## Error Info

One-way upgrade
Visual Studio will automatically make functional changes to the following projects in order to open them. You will not be able to open these projects in the version of Visual Studio in which they were originally created.
     - SBA_Serv, "I:\code\SpaceBattleArena\master\SBA_Serv\SBA_Serv.pyproj"

Non-functional changes required
Visual Studio will automatically make non-functional changes to the following projects in order to enable them to open in Visual Studio 2013, Visual Studio 2012, and Visual Studio 2010 SP1. Project behavior will not be impacted.
     - SBA_Serv, "I:\code\SpaceBattleArena\master\SBA_Serv.sln"
## Output Window

I:\code\SpaceBattleArena\master\SBA_Serv\SBA_Serv.pyproj : error  : The application for the project is not installed.

I:\code\SpaceBattleArena\master\SBA_Serv\SBA_Serv.pyproj : error  : Requested operation would result in circular task dependency causing a deadlock.

zooba:
Thanks for the report, I'll check this out during the week.

zooba:
Unfortunately, I haven't been able to reproduce the error, but I suspect it may be because we're returning our own project factory GUID when we should be returning an empty one.

MM1nd:
Unfortunately this happens again with 2.2.2 (probably I'm coming from 2.1 too, have not upgraded in  long time)

zooba:
I'll reopen and we can take a look for 3.0. All the upgrade code exists in the newer version, so it shouldn't matter how old your project was (though that will affect which upgrades we have to perform). sln and pyproj together causing an issue would suggest that it is to do with the factory GUID, since that's the only information that appears in both.

I don't support you have a diff of a project upgrade where it failed initially?

zooba:
No repro on this, and need more information to diagnose.

PR #164, opened 2015-05-15
#97 Trouble 'upgrading' project from PTVS 2.1 to 2.2 RC

zooba:
Changes project factory to return empty GUIDs by default to avoid circular references.


crwilcox:
:+1:

Issue #21, opened 2015-04-20
Types in annotations are not highlighted


zooba:
In

```
def f(t : str = str):
    pass
```

The second `str` is highlighted as a type but the first `str` is not.

PR #173, opened 2015-05-15
#21 Types in annotations are not highlighted

zooba:
Ensures AST walkers will walk parameter decorators.
Also fixes IntelliSense in functions that include lambda functions.


crwilcox:
:+1:

PR #174, opened 2015-05-15
Move SortOrder of IoT UWP template to bottom of list

juanyaw:


crwilcox:
:+1:

zooba:
What about removing the category rollup? People are more likely to find it through the treeview now rather than scrolling down the list, so we may as well keep the list manageable (plus I'd really like to see more IoT templates there over time, and they're not all going to be in the top level).

juanyaw:
Sounds good and done

Python/Product/Uwp/Templates/Projects/BackgroundService/PythonBackgroundService.vstemplate
```diff
@@ -4,8 +4,8 @@
     <Description>A project for creating an Python IoT Background Application</Description>
     <Icon Package="{6dbd7c1e-1f1b-496d-ac7c-c55dae66c783}" ID="402" />
     <ProjectType>Python</ProjectType>
-    <SortOrder>20</SortOrder>
-    <NumberOfParentCategoriesToRollUp>2</NumberOfParentCategoriesToRollUp>
+    <SortOrder>2000</SortOrder>
```
juanyaw: Still leaving this at 2000 in case we eventually do need to do rollups and any subsequent templates can be 2000 based

PR #197, opened 2015-05-18
Don't add empty search paths to the environment.

zjturner:
Previously, the code would construct a key,value pair both consisting
of empty strings, which in turn would cause an empty entry to be
added to the environment block used to launch the process.

This would result in VsShellUtilities.LaunchProcess() failing
internally when launching the interpreter (and it would print
"Exception : Could not launch <python-interpreter>" to the debug
console.

With this patch, we just do nothing if the search paths are empty.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonProjectLaunchProperties.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonProjectLaunchProperties.cs
@@ -305,8 +305,9 @@ namespace Microsoft.PythonTools.Project {
             IServiceProvider provider = null
         ) {
             var pathEnv = project.GetProjectAnalyzer().InterpreterFactory.Configuration.PathEnvironmentVariable;
-            if (pathEnv == null || pathEnv == String.Empty)
+            if (string.IsNullOrEmpty(pathEnv)) {
                 return;
+            }
 
             var paths = new List<string>();
             
--- a/Python/Product/PythonTools/PythonTools/Project/PythonProjectLaunchProperties.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonProjectLaunchProperties.cs
@@ -305,6 +305,9 @@ namespace Microsoft.PythonTools.Project {
             IServiceProvider provider = null
         ) {
             var pathEnv = project.GetProjectAnalyzer().InterpreterFactory.Configuration.PathEnvironmentVariable;
+            if (pathEnv == null || pathEnv == String.Empty)
+                return;
+
             var paths = new List<string>();
             
             string path;


Python/Product/PythonTools/PythonTools/Project/PythonProjectLaunchProperties.cs
```diff
@@ -305,6 +305,9 @@ IProjectLaunchProperties properties
             IServiceProvider provider = null
         ) {
             var pathEnv = project.GetProjectAnalyzer().InterpreterFactory.Configuration.PathEnvironmentVariable;
+            if (pathEnv == null || pathEnv == String.Empty)
```
int19h: `string.IsNullOrEmpty` is the idiomatic C# way to do this check. Also, can you add the curly braces around the body so that it matches the coding style of the rest of the file?

int19h:
I'm somewhat surprised that we can even get into this state. @zooba, shouldn't we default to the variable name being `PYTHONPATH` if it's not specified in interpreter settings?

Python/Product/PythonTools/PythonTools/Project/PythonProjectLaunchProperties.cs
zjturner: When I look at the UI in Tools > Options > Python Tools > Environment Options and I select this interpreter, it does say PYTHONPATH in the box.  So it may very well be a bug that `project.GetProjectAnalyzer().InterpreterFactory.Configuration.PathEnvironmentVariable` is returning String.Empty when the value has been defaulted.  Still might be nice to have this fallback just in case, though.

I don't really understand why this works for the Python-only case (e.g. Enable Native debugging not checked).  I confirmed that the VsDebugInfo was being passed through exactly the same.  It must be an implementation detail of the Native code debug engine.

zjturner:
When I look at the UI in Tools > Options > Python Tools > Environment Options and I select this interpreter, it does say PYTHONPATH in the box.  So it may very well be a bug that `project.GetProjectAnalyzer().InterpreterFactory.Configuration.PathEnvironmentVariable` is returning String.Empty when the value has been defaulted.  Still might be nice to have this fallback just in case, though.

I don't really understand why this works for the Python-only case (e.g. Enable Native debugging not checked).  I confirmed that the VsDebugInfo was being passed through exactly the same.  It must be an implementation detail of the Native code debug engine.

zooba:
It's probably hitting a code path where we only default if it's null, and not NullOrEmpty...

Sure enough, `Analysis\Interpreter\InterpreterFactoryCreator.cs` only coalesces, it doesn't check for an empty string, and `ConfigurationPythonInterpreterFactoryProvider.cs` will pass in an empty string if it's not set.

We should update any locations that do anything like this to default to `CPythonInterpreterFactoryConstants.PathEnvironmentVariableName` (because it's obviously better to use a huge constant where a literal string would suffice...?)

zjturner:
Is that something that should be done instead of this patch or in addition to this patch?  It sounds a little bit more involved, or at least enough that I dont' think I would be able to find all the locations very easily, so maybe you want to make that change?

If you decide not to accept this change, then maybe you could at least add an assert here that `pathEnv` is not null or empty.  Or perhaps even do both (assert but also just exit if it is)

zooba:
This is a good workaround for when the issue arises, but we should still fix the original cause. I've made #211 for this.

Issue #190, opened 2015-05-18
Install from/Generate Requirements.txt commands should be disabled on no interpreters


huguesv:
On a machine with no Python interpreters installed:

Create new console app 
In solution explorer, you'll have "No Interpreters (global default)", select it 
Right-click, Generate Requirements.txt 
You'll get an error

Both Generate and Install commands should be disabled in this case.

From: https://pytools.codeplex.com/workitem/2971
Verified on PTVS 2.2 30515

zooba:
Real bug is that we shouldn't show "No Interpreters" here.

huguesv:
Sorry, I'll add that to the original post.  The machine doesn't have any interpreters installed.

PR #198, opened 2015-05-18
#190 Install from/Generate Requirements.txt commands should be disabl…

zooba:
…ed on no interpreters

Ensures the global default environment is only shown if it is a real environment.


int19h:
:+1:

huguesv:
:+1:

Issue #171, opened 2015-05-15
Django template IntelliSense doesn't understand simple_tag


huguesv:
If you register a tag like this:
```
# mytags.py
import datetime
from django import template

register = template.Library()

@register.simple_tag
def my_current_time(format_string):
    return datetime.datetime.now().strftime(format_string)

```

Then use it:
```
# details.html
{% load mytags %}
{% <check this completion list
```

It doesn't show up in IntelliSense in the template editor.  If you replace `simple_tag` with `tag` or `assignment_tag`, it shows up in completions as expected.

Note: When testing this, it's best to close/reopen the solution after changing the tag code, in order to avoid false positives.  I found that if you make it `tag` or `assignment_tag` then change it to `simple_tag` without reloading, it will still show up in completions.  But it won't after you reload the solution.

https://docs.djangoproject.com/en/dev/howto/custom-template-tags/

From: https://pytools.codeplex.com/workitem/2623
Verified on VS 2015 RC and PTVS 2.2 30515

PR #199, opened 2015-05-18
#171 Django template IntelliSense doesn't understand simple_tag

zooba:
Adds specialization for simple_tag
Also clears tags/filters that came from a file whenever that file is reanalyzed


int19h:
:+1:

huguesv:
Please add a simple_tag and assignment_tag to DjangoAnalysisTestApp\test_render\views.py and adjust test to look for them.

Issue #159, opened 2015-05-15
Cannot delete custom environment via tools options


huguesv:
PTVS 2.2 30515.00

Add a new custom environment (via tools options or env window) 
In tools options, select that env and click Remove Environment 
Notice that it gets removed from the dropdown 
Click OK to accept changes and leave tools options 

Look at environment window, the removed env is still there. 
Go back to tools options, and it's still there.

Note: You can successfully remove an environment via the environments window.

https://pytools.codeplex.com/workitem/2981

PR #204, opened 2015-05-18
#159 Cannot delete custom environment via tools options

zooba:
Fixes environments not being deleted
Fixes settings being changed if you remove an environment and then cancel the change
Fixes virtual environments appearing in Tools-Options


int19h:
:+1:

crwilcox:
:+1:

Issue #155, opened 2015-05-15
Don't auto complete identifiers on first list/tuple element


zooba:
It could be the start of a comprehension, and it's more annoying to have it complete than to have to complete manually.

We should be able to have a "don't force completion" mode where the list appears and filters as normal, but the usual commit characters don't commit unless the user interacts directly with the drop down. This is what we already do for method overrides.

PR #208, opened 2015-05-18
#155 Don't auto complete identifiers on first list/tuple element

zooba:
Adds support to FuzzyCompletionSet to opt-out of committing by default.
Opts-out of committing by default for single element collections in case they are going to become a comprehension.

(The existing tests cover this case fine.)


int19h:
:+1:

crwilcox:
:+1:

PR #224, opened 2015-05-20
Changes running document table to marshal to the UI thread (fixing a …

DinoV:
Changes running document table to marshal to the UI thread (fixing a bunch of failing tests)
Adds multi-select support
Decorates tests using mock framework as mock tests
Makes command routing more like VS


zooba:
+1

PR #230, opened 2015-05-21
Add Python Engine guid to EXCEPTION_INFO object in AD7::GetException()

jsschultz:
Add Python Engine guid to EXCEPTION_INFO object in AD7 call to GetException.

This completes the exception object for the VS Debugger UI to properly process changes to this exception when the user modifies its break settings via the exception thrown dialog.

This change fixes an issue where a user changes the break when thrown setting for a thrown exception via the Exception Assistant (or Exception Thrown Dialog) and it does nothing. The debugger ui relies on the exception's guidType to find the exception changed. Without a guid, the debugger does nothing.


zooba:
Really? That's all it takes? I wish we'd figured that out ages ago...

Thanks for the fix :)

PR #231, opened 2015-05-21
Parallelises parser stdlib test to reduce running time.

zooba:
Fixes detection of too many interpreters when testing.

This change reduced the test running time to less than 1 minute on my machine for the first run (mostly IO time, with the detection issue hurting that too), and down to 10 seconds on subsequent runs.


int19h:
:+1:

Issue #225, opened 2015-05-20
Error while editing


huguesv:
VS 2013 Update 4
PTVS 2.2 Dev 2015-05-19

Activity log error:
```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
 at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
 at System.Collections.Generic.Dictionary`2.ValueCollection.Enumerator.MoveNext()
 at System.Linq.Enumerable.<SelectManyIterator>d__1`2.MoveNext()
 at Microsoft.PythonTools.Analysis.Analyzer.FunctionScope.<GetMergedVariables>d__12.MoveNext()
 at Microsoft.PythonTools.Analysis.Analyzer.InterpreterScope.GetMergedVariableTypes(String name)
 at Microsoft.PythonTools.Analysis.Analyzer.ExpressionEvaluator.LookupAnalysisSetByName(Node node, String name, Boolean addRef)
 at Microsoft.PythonTools.Analysis.Analyzer.ExpressionEvaluator.EvaluateName(ExpressionEvaluator ee, Node node)
 at Microsoft.PythonTools.Analysis.Analyzer.ExpressionEvaluator.EvaluateWorker(Node node)
 at Microsoft.PythonTools.Analysis.ModuleAnalysis.GetMembers(String exprText, SourceLocation location, GetMemberOptions options)
 at Microsoft.PythonTools.Intellisense.NormalCompletionAnalysis.GetCompletions(IGlyphService glyphService)
 at Microsoft.PythonTools.Intellisense.CompletionSource.AugmentCompletionSession(ICompletionSession session, IList`1 completionSets)
 at Microsoft.VisualStudio.Language.Intellisense.Implementation.CompletionSession.Start()
 at Microsoft.VisualStudio.Language.Intellisense.Implementation.CompletionBroker.TriggerCompletion(ITextView textView)
 at Microsoft.PythonTools.Intellisense.IntellisenseController.TriggerCompletionSession(Boolean completeWord)
 at Microsoft.PythonTools.Intellisense.IntellisenseController.HandleChar(Char ch)
 at Microsoft.PythonTools.Intellisense.IntellisenseController.Exec(Guid& pguidCmdGroup, UInt32 nCmdID, UInt32 nCmdexecopt, IntPtr pvaIn, IntPtr pvaOut)
[...]

PR #232, opened 2015-05-21
#225 Error while editing

zooba:
Prevents direct access to dictionaries inside scopes and adds locking to all indirect access.
Adds locking to reference collections.

Not sure yet exactly what the perf impact will be, but there doesn't seem to be a cheaper way to avoid the race conditions we currently suffer from - suggestions welcome.

Test to follow (lots of cross-module analysis running on separate threads for a while and seeing whether we crash)

### DIFF
--- a/Python/Product/Analysis/Analysis.csproj
+++ b/Python/Product/Analysis/Analysis.csproj
@@ -77,7 +77,9 @@
     <Compile Include="..\PythonTools\PythonToolsInstallPath.cs">
       <Link>PythonToolsInstallPath.cs</Link>
     </Compile>
+    <Compile Include="AnalysisDictionary.cs" />
     <Compile Include="AnalysisExtensions.cs" />
+    <Compile Include="AnalysisHashSet.cs" />
     <Compile Include="AnalysisLimits.cs" />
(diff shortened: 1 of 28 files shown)


DinoV:
For NTVS I solved this w/ a thread-safe AnalysisDictionary class which you could re-use for these dictionaries.  At the very least it would be worth comparing the perf of the 2 and going with whichever one is better.

zooba:
I think in this case AnalysisDictionary would perform better, but still suffer from some of the race conditions because it doesn't handle multiple writers. I could be misremembering some of the crashes caused by this, but fairly sure at least one was a case where we had multiple analysis threads somehow.

DinoV:
We shouldn't ever have multiple analysis threads working on the same analysis though...  If that's happening we should track down and fix that bug.  But we can easily end up with multiple analysis threads if we have multiple projects, or 1 project and some loose Python files.

zooba:
You're right. I ported over the dictionary and hashset and it cleared most things up, then I added some fairly tight locks to handle the rest.

DinoV:
LGTM

PR #233, opened 2015-05-21
Fixing TcpTransport to not use IPHostEntry if specified address is al…

juanyaw:
…ready an IP address

### DIFF
--- a/Python/Product/Debugger/Debugger/Transports/TcpTransport.cs
+++ b/Python/Product/Debugger/Debugger/Transports/TcpTransport.cs
@@ -35,18 +35,10 @@ namespace Microsoft.PythonTools.Debugger.Transports {
             }
 
             // PTVSD is using AF_INET by default, so lets make sure to try the IPv4 address in lieu of IPv6 address
-            var clientIp = uri.Host;
-            var ip = default(IPAddress);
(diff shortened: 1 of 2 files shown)


juanyaw:
#234

zooba:
LGTM, but I'd like @int19h to weigh in too.

int19h:
I'm actually curious why we need this at all. Per MSDN docs for `TcpClient`, even when IPv6 is enabled, it will try that first, but will fall back to IPv4 anyway. Or is this just so that we don't even waste time trying IPv6?

Anyway, it should be better to use the `TcpClient` constructor that takes `AddressFamily` instead, explicitly specifying `AddressFamily.InterNetwork` there, and then call `Connect` passing in the hostname and the port.

juanyaw:
That actually is exactly the issue.  It works eventually, but what ends up happening is that it takes an extra 30-90 seconds to fallback when trying the IPv6 address(es).  This eats up a great amount of time when trying to get that initial connect for us.  I'll look at the alternative way of doing this that you mention

juanyaw:
@int19h It looks like the way you suggested works great and looks much better.  Thanks!

int19h:
:+1:

PR #238, opened 2015-05-22
Fix null reference exceptions on background task when looking for mis…

zooba:
…sing imports.


zooba:
First reviewer can merge this

PR #246, opened 2015-05-25
Ensures missing imports calculate the target text on the UI thread.

zooba:
Fixes suggestions appearing when local variables are defined.


int19h:
:+1:

Issue #247, opened 2015-05-25
Parse and analyse async/await syntax


zooba:
[PEP 492](https://www.python.org/dev/peps/pep-0492/) was accepted and the changes are in [3.5.0b1](https://www.python.org/downloads/release/python-350b1/)

At the very least, we need to be able to parse and analyse it

PR #250, opened 2015-05-25
#247 Parse and analyse async/await syntax

zooba:
Adds support for async/await syntax.
Fixes analysis of user-defined __enter__, __iter__ and __next__/next methods.
Fixes some analysis tests
Fixes AnalysisHashSet to correctly update unioned types.


int19h:
Can you also make `AwaitExpression` non-side-effect free in `ProximityExpressionWalker`?

int19h:
Might also need an override in `ExpressionCompletionWalker`.

zooba:
Done. I also need to update the method extractor to correctly extract await statements and generate async def.

zooba:
Actually, going to file that as a separate issue so I can defer it temporarily. Need to work on some higher priorities.

int19h:
:+1:

PR #260, opened 2015-05-27
#132 #259

### LINKED ISSUE #132: Debugger options are ignored when attaching
"Tools -> Options -> Python Tools -> Debugging -> Enable debugging of the Python standard library" is only respected when the process is launched via VS. When attaching (either directly or via ptvsd), the setting is ignored altogether.

### LINKED ISSUE #259: Fully qualified function names in Call Stack are computed incorrectly if there is a source mismatch
If the source that is parsed to compute fully qualified names doesn't match what's being debugged (e.g. because it was edited while debugging), the names can be reported incorrectly. There's no general fix for this unless we can somehow detect those changes when inspecting the code object (e.g. if code objects start cashing hash of the source from which they have been created), but we can at least compare the computed name vs `co_name` in the frame, and fall back to the latter if they don't match.

int19h:
Fix #132: Debugger options are ignored when attaching
Propagate all global debug options to the debug engine whenever it is created in a uniform way (in all attach and launch scenarios), and from there to the debugger script.

Use bitflags instead of textual switches to propagate flags to the debugger script for easier parsing and future extensibility.

Bonus feature: ability to specify additional debug options (e.g. enable Django debugging) in the URL when attaching remotely via ptvsd.

Add tests for output redirection for launch and attach scenarios to test flag propagation.

Fix #259: Fully qualified function names in Call Stack are computed incorrectly if there is a source mismatch

If the computed fully qualified name does not correspond to the unqualified function name reported by the runtime, fall back to the latter.


zooba:
Why is a bitfield any better than a string? On both sides splitting at semicolons and putting into a set seems easier to read, maintain and extend. (I'll grant that it gets messy when you add extra values into the string, but we don't have to allow that.)

int19h:
Mainly it's because we'd need to have the same parsing code duplicated at least twice (for C# and Python). Given that on both sides it's still end up as a bunch of bits for ease of querying anyway, converting it to and from seems unnecessary. Especially since it's not a pipeline where human intervention is expected at any point - the only case where you'd want to actually inspect the values midway, or replace them, is when debugging, and it's not really hard to handle the bitflags then (and I don't think the marginal increase in difficulty of debugging here is worth the complexity).

int19h:
Hm... or are you suggesting that we use a set instead of an enum internally on both sides, and not just in transit? I'd rather not do that on the Python side just to avoid changes throughout the code everywhere (which is why I kept the existing bool flags instead of replacing them with the bitfield). On the C# side, a flags enum is simply more idiomatic and easier to use, IMO. But they don't really have to match, since they're going to be stringified and then parsed as they flow through, anyway, so either side can use the more convenient representation.

int19h:
How about replacing the enum with a property bag class that would encapsulate all debug options (not just flags)? And passing it around stringified. Still needs two parsers, though.

zooba:
It doesn't need a new class. Using a flags enum on the C# side is fine, use `Enum.GetValueNames()` to serialize, send it as semicolon separated value, and the Python side can switch to `DEBUG_OPTIONS = set(options.split(';'))` and (e.g.) `'BreakOnSystemExit' in DEBUG_OPTIONS` everywhere.

Debugging our own debugger is an important use case :)

int19h:
What I mean is, we've got a bunch of other options there that aren't boolean (and might get more in the future). This includes interpreter options, directory mapping stuff, language version etc.

Since I'm doing a significant refactoring here anyway, it might be beneficial to aggregate all of them in a single class that can then be passed around, and knows how to serialize/deserialize itself.

zooba:
I guess so, though I'm a little concerned about the huge range of classes we have for this kind of stuff...

Go ahead. We'll thrash out the details once we have some actual code to talk about.

int19h:
Can you clarify? Aside from PythonDebugOptions (which would be subsumed by this change), what else do we have?

zooba:
More of a general comment. We have more classes on the UI side, and there are 3-5 different classes that contribute settings towards a debug session from the project. etc.

int19h:
After giving it a try, I think I'll stick to your suggestion for the time being (BTW, it seems that Enum.ToString already handles [Flags] nicely, so it's even easier). If it's all in one class, and given that we're setting it in the engine from two different points now, it'd need to not just capture the options themselves, but also capture option changes (and applying them one after another). So for each property it'd need to have an "is this set?" flag along with the actual value, and things just get messy from there. I'll need to think about how to make it all prettier in long term, but for now flags and sets it is.

Python/Product/Attacher/PyAttacher.cs
```diff
@@ -466,12 +470,17 @@ internal sealed class DebugAttach : IDisposable {
...
-                            for (int i = 0; i < guid.Length; i++) {
-                                viewStream.WriteByte((byte)guid[i]);
+
+                            string szDebugId = debugId.ToString().PadRight(64, '\0');
+                            for (int i = 0; i < szDebugId.Length; i++) {
+                                viewStream.WriteByte((byte)szDebugId[i]);
+                            }
+
+                            for (int i = 0; i < debugOptions.Length; i++) {
+                                viewStream.WriteByte((byte)debugOptions[i]);
```
zooba: Encoding.GetBytes()?

Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -585,18 +614,40 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
...
+
+                case DebugOptionsMetric:
+                    {
+                        if (_engineCreated) {
+                            Debug.Fail(DebugOptionsMetric + " metric can only be sent immediately in response to IDebugEngineCreateEvent2.");
+                            return VSConstants.E_FAIL;
+                        }
+
+                        var options = varValue as string;
+                        if (options != null) {
```
zooba: StringOrEmpty?

Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -659,92 +710,31 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
...
+            Send(new AD7CustomEvent(VsPackageMessage.SetDebugOptions, this), AD7CustomEvent.IID, null, null);
+
+            // If this is a windowed application, there's no console to wait on, so disable those flags if they were set.
+            if (_debugOptions.HasFlag(PythonDebugOptions.IsWindowsApplication)) {
+                _debugOptions &= ~(PythonDebugOptions.WaitOnNormalExit | PythonDebugOptions.WaitOnAbnormalExit);
+            }
 
             Guid processId;
-            if (attachRunning && Guid.TryParse(exe, out processId)) {
+            if (_debugOptions.HasFlag(PythonDebugOptions.AttachRunning) && Guid.TryParse(exe, out processId)) {
```
zooba: What if the flag says attach and the guid is invalid? Probably shouldn't fall back to launching in that case.

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteProcess.cs
```diff
@@ -136,9 +137,20 @@ private PythonRemoteProcess(int pid, Uri uri, PythonLanguageVersion langVer)
...
+                // ptvsd attach (this is currently the only way to enable some flags when attaching, e.g. Django debugging).
+                // ParseQueryString will automatically concat them all into a single value using commas.
+                var queryParts = HttpUtility.ParseQueryString(uri.Query);
+                debugOptions = queryParts[AD7Engine.DebugOptionsKey];
+            }
+
             var stream = Connect(uri, warnAboutAuthenticationErrors);
             try {
                 stream.Write(AttachCommandBytes);
+                stream.WriteString(debugOptions);
```
zooba: This is writing an int as a string, correct? Because we put the options into the URI above as an int?

Python/Product/Debugger/DkmDebugger/RemoteComponent.cs
```diff
@@ -302,7 +303,7 @@ internal class RaiseExceptionRequest : MessageBase<RaiseExceptionRequest> {
         string IDkmExceptionFormatter.GetDescription(DkmExceptionInformation exception) {
             var em = exception.Process.GetOrCreateDataItem(() => new ExceptionManager(exception.Process));
             return em.GetDescription(exception);
-        }
+            }
```
zooba: ?

Python/Product/Debugger/Guids.cs
```diff
@@ -14,7 +14,7 @@
 
 using System;
 
-namespace Microsoft.PythonTools.DkmDebugger {
+namespace Microsoft.PythonTools.Debugger {
```
zooba: These are public APIs (unless they're new in 2.2, which I don't think they are), so we can't rename the namespace unfortunately.

Python/Product/Debugger/VsPackageMessage.cs
```diff
@@ -12,10 +12,11 @@
  *
  * ***************************************************************************/
 
-namespace Microsoft.PythonTools.DkmDebugger {
+namespace Microsoft.PythonTools.Debugger {
```
zooba: Ditto. Adding new enum members is okay, but changing the namespace is a breaking change.

Python/Product/PythonTools/PythonTools/Debugger/CustomDebuggerEventHandler.cs
```diff
@@ -83,5 +84,28 @@ public class CustomDebuggerEventHandler : IVsCustomDebuggerEventHandler110 {
...
+                options.Append(";" + AD7Engine.RedirectOutputSetting + "=True");
+            }
+            if (pyService.DebuggerOptions.BreakOnSystemExitZero) {
+                options.Append(";" + AD7Engine.BreakSystemExitZero + "=True");
+            }
+            if (pyService.DebuggerOptions.DebugStdLib) {
+                options.Append(";" + AD7Engine.DebugStdLib + "=True");
+            }
+
+            engine.SetMetric(AD7Engine.DebugOptionsMetric, options.ToString());
```
zooba: Can we replace this options string with our new flags string?

Python/Product/PythonTools/visualstudio_py_debugger.py
```diff
@@ -2155,12 +2152,33 @@ def attach_process(port_num, debug_id, report = False, block = False):
...
+        # If one or more of the prefixes are empty, we can't reliably distinguish stdlib
+        # from user code, so override stdlib-only mode and allow to debug everything.
+        DEBUG_STDLIB = True
+    else:
+        DEBUG_STDLIB = 'DebugStdLib' in debug_options
+
+    def _excepthook(exc_type, exc_value, exc_tb):
+        # Display the exception and wait on exit
+        if exc_type is SystemExit:
+            if ('WaitOnAbnormalExit' in debug_options and exc_value.code != 0) or ('WaitOnNormalExit' in debug_options and exc_value.code == 0):
```
zooba: It's probably fine, but I think I'd rather capture the result of `'WaitOn(Abn|N)ormalExit' in debug_options` outside this function, just in case `debug_options` changes/disappears during execution.

Python/Product/Attacher/PyAttacher.cs
int19h: I'm going to use Encoding.ASCII here, because in practice that is what this is anyway, and using UTF8 will make length validation when parsing command line arguments more complicated.

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteProcess.cs
int19h: My bad. We really need at least one test that actually does use remote attach (and validates the flags). I'll add it.

Python/Product/PythonTools/PythonTools/Debugger/CustomDebuggerEventHandler.cs
int19h: I didn't do it because PythonDebugOptions is internal, but looks like we already have [InternalsVisibleTo] for it, so it's not an issue. I'll make the change.

int19h: On the other hand, this is kinda sorta public API also, because it's on the boundary between Debugger and PythonTools, SetMetric is public, and so is the metric name and option names (exposed as public constants on AD7Engine). Thoughts?

int19h: Ah, one more reason to keep it intact. The current protocol allows for the flexibility of either overwriting an existing flag, or leaving it as is, and we actually use that to merge the values that flow via LaunchSuspended and via SetMetric.

zooba: Sounds like a reason to continue to parse the existing string but to stop generating it, but leaving it alone is also a good option. Maybe add a comment pointing the next person to where new options should be added?

int19h: In the future I'd like to get back to revamping this whole thing such that all parts that deal with options (not just flags) or pass them around work with the same class that knows how to merge instances, stringify and parse. It's just too big of a change to take at this point. I'll leave a TODO comment along these lines.

int19h:
Addressed all review feedback, added a test for ptvsd attach, and various other fixes for issues that I've ran into while doing that :)

int19h:
Ping (I'd like to get this in before the release is locked down further).

Python/Product/PythonTools/visualstudio_py_debugger.py
```diff
@@ -2468,7 +2465,7 @@ def _excepthook(exc_type, exc_value, exc_tb):
             write_bytes(conn, LAST)
         last_ack_event.wait(5)
 
-    if wait_on_exit:
+    if wait_on_normal_exit in debug_options:
```
zooba: Can't see all the scopes in between, but `wait_on_normal_exit` here will be a bool or a name error, and not in the set.

zooba:
+1 with the one fix.

PR #264, opened 2015-05-27
Rename invalid Latin-1 filenames

### LINKED ISSUE #184: Clone fails due to Unicode character?
I am trying to clone PTVS with the VS2015RC Github integration ("Open in Visual Studio").

However I get a failure to clone:
Error encountered while cloning the remote repository: Failed to stat file 'C:/Github/PTVS/Python/Tests/TestData/UnicodePath�.sln': No mapping for the Unicode character exists in the target multi-byte code page.

ethomson:
As mentioned in #184 you have some files in your repository whose names are not UTF-8.  This is not valid.  These files appear to be Latin-1 (`\344` is `ä` in Latin-1.)

I have renamed them to be UTF-8 strings, using `ä` (`U+00e4`).

Fixes #184


ethomson:
I'm getting Connect bugs assigned to me because of this issue, GitHub is getting support requests for this issue, and so is the LibGit2Sharp project.

I'm a Microsoft FTE and do not believe that I require a CLA.

Thanks for your attention to this matter!

int19h:
It seems that it doesn't fix the problems entirely, at least on VS (libgit2sharp?) side of things. I tried pulling your branch and then opening it in VS. Here's what git says about it:

image

And here's what VS says about it:

image

I think that, to avoid all these headaches, we'd rather just switch to not having files with Unicode names, and have the test generate them as needed at runtime. I'll take care of it.

But it is still worth figuring out what's going on with the above two screenshots on VS side of things...

zooba:
I also had trouble finding the renamed files from our tests. Changing the test to generate the file as necessary seems a better approach.

ethomson:
Oops, that was my bad, I missed renaming the `Program.py` and left its folder name as Latin-1.  Fixed that in the PR, thanks!

int19h:
Thanks, this works great!

ethomson:
Cool, thanks!  :dancers:

PR #266, opened 2015-05-27
VS2015 images

zooba:
Finishes the update to use image monikers throughout the project system on VS 2015. You now need to explicitly suppress Obsolete warnings to continue to use image lists.

Custom images can be added to the set of available monikers where necessary - see `IVsImageService2` for the APIs (this isn't necessary for PTVS as all of our images are included).


int19h:
:+1:

zooba:
@mousetraps is probably interested in this too...

crwilcox:
:+1: with comments

int19h:
:+1:

PR #271, opened 2015-05-28
Partially fix #239

### LINKED ISSUE #239: Override the default "Find" button in Attach to Process dialog for remote debugging
The default behavior of the button is to display Windows network machines. When Python remote debugging transport is selected, it's useless, and the fact that it's still there is confusing to the users.

We can override this by implementing `IDebugPortPicker`, and associating it with our port supplier. At a minimum, we should disable the button completely. Better yet, replace it with a MRU list of addresses.

int19h:
Suppress "Find" button in Attach to Process when using the ptvsd port supplier (#239).

Change user-visible strings for debug engine and port supplier name to be clearer and better conform to how the standard VS engines and suppliers are named.


zooba:
+1

Issue #272, opened 2015-05-29
await missing from completions


huguesv:
PTVS.Dev.2015-05-28.VS.2015

async is there, but await is missing

zooba:
It's there, but it doesn't show up as the first keyword on the line. Typing `x = aw...` will display it

PR #277, opened 2015-05-29
#272 await missing from completions

zooba:
Fixes statement range calculation to correctly handle empty statements.
Allows "await" to appear as a statement keyword.


int19h:
:+1:

Issue #256, opened 2015-05-26
Cannot extract async method


zooba:
#247 adds support for async/await syntax, but Refactor/Extract Method currently can't extract properly.

Should be able to generate `async def` and `await` statements when necessary.

zooba:
Also need to fix up the verbatim parsing of `async def`.

PR #278, opened 2015-05-29
#256 Cannot extract async method

zooba:
Fixes verbatim parsing of async def to preserve correct whitespace
Ensures extracted method bodies that use await generate async functions

Also fixes IronPythonDatabaseTest


int19h:
:+1:

PR #281, opened 2015-05-29
Makes copying test files more likely to succeed.

zooba:


gilbertw:
I just bumped into this as well.  Thanks for fixing!!

Issue #263, opened 2015-05-27
[Interpreter] Invalid interpreter when sharing the .pyproj file in VCS


sonictk:
Hi all:

Not sure if I'm doing things wrong here; I have a virtualenv for a project that is being shared in the git repo in order to make it easy for me to transfer my work around workstations (at least, if they have PyCharm/Wing installed).

I'm trying out PTVS for my workflow, and I'm having an odd issue where if I load the solution on another computer, save it and commit it to VCS, the next time I open it on a different machine, I get an *(unavailable)* error and I need to re-add the virtualenv again.

In the diff, it looks like this might be the issue:

![Image](http://i.imgur.com/K1m0xpT.png)

If this is the wrong approach to take for syncing PTVS projects across VCS, I'd appreciate it if anyone could enlighten me on what I could do to solve this issue.

Thanks!

zooba:
If you're adding the virtual environment as a custom environment, then unfortunately yes, it doesn't transfer well between machines. There are some pretty serious architectural changes we'd need to make to support this properly, so it hasn't happened for fear of destabilizing the product and breaking third-parties who use the APIs involved.

The better way to do a virtual environment is to add it to your project directly. If you right click the Environments node in Solution Explorer then you can add an existing virtual environment. This will do a better job of handling the layered libraries too, as well as being portable between machines (provided the same base interpreter is used for the virtual environment and the relative path from project to environment is the same). This feature is really designed to handle project-specific environments, so it isn't necessarily going to be perfect if you want to share a virtual environment between projects.

One other option is to add a `requirements.txt` file to your project. You can generate or update these by right-clicking the environment, and then on a new machine you can create a new virtual environment and we will install whatever is listed for you. It's an extra manual step, but it is going to avoid basically all of the problems you'll face trying to move virtual environments between machines (venv hardcodes some absolute paths, for example, and there's little we can do to work around these).

sonictk:
@zooba Hi:

> The better way to do a virtual environment is to add it to your project directly. If you right click the Environments node in Solution Explorer then you can add an existing virtual environment. 

[...]

PR #282, opened 2015-05-29
#263 Invalid interpreter when sharing the .pyproj file in VCS

zooba:
Adds fallback to detection by path names when a virtual environment's base interpreter cannot be found by ID.
Moves some virtualenv related functions into DerivedInterpreterFactory.

### DIFF
--- a/Python/Product/VSInterpreters/DerivedInterpreterFactory.cs
+++ b/Python/Product/VSInterpreters/DerivedInterpreterFactory.cs
@@ -19,6 +19,7 @@ using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading;
+using Microsoft.VisualStudioTools;
 
 namespace Microsoft.PythonTools.Interpreter {
     class DerivedInterpreterFactory : PythonInterpreterFactoryWithDatabase {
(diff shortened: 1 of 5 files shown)


Python/Product/VSInterpreters/DerivedInterpreterFactory.cs
```diff
@@ -266,5 +267,81 @@ InterpreterFactoryCreationOptions options
...
+            if (!Directory.Exists(prefixPath)) {
+                return null;
+            }
+
+            var cfgFile = Path.Combine(prefixPath, "pyvenv.cfg");
+            if (File.Exists(cfgFile)) {
+                try {
+                    var lines = File.ReadAllLines(cfgFile);
+                    basePath = lines
+                        .Select(line => Regex.Match(line, "^home *= *(?<path>.+)$", RegexOptions.IgnoreCase))
```
int19h: Should this rather be \s* ?

int19h:
:+1:

Python/Product/VSInterpreters/DerivedInterpreterFactory.cs
zooba: Strictly it should be `"[ \t]+"`, according to the CPython sources. I'm okay with being more forgiving here though.

Issue #269, opened 2015-05-28
Need to use a unique guid for the AD7ProgramProvider


gilbertw:
We are colliding with NTVS on the use of {ff3e23a2-da7e-4fa7-af47-6edede4e922e}

int19h:
Should also check other CLSIDs used by debugger - engine, port supplier etc.

PR #283, opened 2015-05-29
#269 - Need unique guids in the debugger

gilbertw:
NTVS was forked from ptvs base and duplicated the guids.  This fixes the collisions by moving to new guids


int19h:
This is going to conflict with my changes in https://github.com/Microsoft/PTVS/pull/260, where I moved Guids.cs up one level from DkmDebugger. Instead of creating a new file, can you just do the same rename, so that these can be properly merged?

Issue #214, opened 2015-05-19
NullReferenceException in debug REPL evaluator tests


zooba:
```
Initialization method PythonToolsTests.DebugReplEvaluatorTests.TestInit threw exception. System.NullReferenceException: System.NullReferenceException: Object reference not set to an instance of an object..
    at Microsoft.PythonTools.Repl.BasePythonReplEvaluator.GetInteractiveCommands(IServiceProvider serviceProvider, IInteractiveWindow window, IInteractiveEvaluator eval) in C:\Source\Python\Product\PythonTools\PythonTools\Repl\BasePythonReplEvaluator.cs:line 1712
   at Microsoft.PythonTools.Repl.PythonDebugReplEvaluator.InitializeAsync() in C:\Source\Python\Product\PythonTools\PythonTools\Repl\PythonDebugReplEvaluator.cs:line 564
   at TestUtilities.Mocks.MockReplWindow..ctor(IInteractiveEvaluator eval, String contentType) in C:\Source\Common\Tests\Utilities\Mocks\MockReplWindow.cs:line 58
   at PythonToolsTests.DebugReplEvaluatorTests.TestInit() in C:\Source\Python\Tests\Core\DebugReplEvaluatorTests.cs:line 72
```

PR #285, opened 2015-05-29
Issue 214

zooba:
#214 NullReferenceException in debug REPL evaluator tests
Adds more mocks to support REPL tests on VS 2015
Fixes debug repl not initializing its child evaluators.

### DIFF
--- a/Common/Tests/Utilities/Mocks/MockReplWindow.cs
+++ b/Common/Tests/Utilities/Mocks/MockReplWindow.cs
@@ -43,6 +43,8 @@ namespace TestUtilities.Mocks {
         private readonly string _contentType;
 
 #if DEV14_OR_LATER
+        private PropertyCollection _properties;
+
         public event EventHandler<SubmissionBufferAddedEventArgs> SubmissionBufferAdded {
             add {
             }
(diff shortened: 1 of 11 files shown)


Common/Tests/Utilities/Mocks/MockReplWindow.cs
```diff
@@ -320,9 +329,9 @@ public class MockReplWindow : IReplWindow {
         #endregion
     }
 
-#if DEV14_OR_LATER
     public static class ReplEvalExtensions {
-        public static Task<ExecutionResult> Initialize(this IReplEvaluator self, IReplWindow window) {
+#if DEV14_OR_LATER
+        public static Task<ExecutionResult> _Initialize(this IReplEvaluator self, IReplWindow window) {
```
int19h: Why underscore?

zooba: To avoid conflicting with the real Initialize method when the class implements it, since then we skip the extension method. Having a unique name for tests makes it more obvious that we're doing something strange here (aka. supporting VS 2013 and 2015 with the same code)

int19h:
:+1:

PR #311, opened 2015-06-02
Fix #301: AttachWithOutputRedirection test asserting from incorrect t…

### LINKED ISSUE #301: AttachWithOutputRedirection test asserting from incorrect thread
```
Description: The process was terminated due to an unhandled exception.
Exception Info: Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException
Stack:
   at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.HandleFailure(System.String, System.String)
   at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.AreEqual[[System.__Canon, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]](System.__Canon, System.__Canon, System.String, System.Object[])
   at DebuggerTests.DebuggerTests+<>c__DisplayClass96_0.<AttachWithOutputRedirection>b__0(System.Object, Microsoft.PythonTools.Debugger.OutputEventArgs)
   at Microsoft.PythonTools.Debugger.PythonProcess.HandleDebuggerOutput(System.IO.Stream)
   at Microsoft.PythonTools.Debugger.PythonProcess.DebugEventThread()
   at System.Threading.ThreadHelper.ThreadStart_Context(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
   at System.Threading.ThreadHelper.ThreadStart()
```

int19h:
…hread

Rewrote the test to only capture output while running, and assert afterwards.

Fixed bug exposed by the updated test with Python 3.4 and 3.5 where directly attaching to the process that hasn't imported threading module yet would cause an assert in that module to fail on process shutdown because it incorrectly identified a debugger helper thread as a main thread.

### DIFF
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -13,6 +13,13 @@
  # ###########################################################################
 
 from __future__ import with_statement
+
+# This module MUST NOT import threading in global scope. This is because in a direct (non-ptvsd)
+# attach scenario, it is loaded on the injected debugger attach thread, and if threading module
+# hasn't been loaded already, it will assume that the thread on which it is being loaded is the
+# main thread. This will cause issues when the thread goes away after attach completes.
+_threading = None
+
 import sys
 import ctypes
 try:
@@ -1526,12 +1533,11 @@ class Thread(object):
                     write_object(conn, type_obj, safe_repr_obj, hex_repr_obj, type_name, obj_len)
 
     def enum_thread_frames_locally(self):
-        global threading
-        if threading is None:
+        global _threading
+        if _threading is None:
             import threading
-        self.send_frame_list(self.get_frame_list(), getattr(threading.currentThread(), 'name', 'Python Thread'))
-
-threading = None
+            _threading = threading
+        self.send_frame_list(self.get_frame_list(), getattr(_threading.currentThread(), 'name', 'Python Thread'))
 
 class Module(object):
     """tracks information about a loaded module"""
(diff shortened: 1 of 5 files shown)


Python/Product/PythonTools/visualstudio_py_debugger.py
```diff
@@ -2127,16 +2133,23 @@ def write_object(conn, obj_type, obj_repr, hex_repr, type_name, obj_len, flags =
...
+    # on an ephemeral debugger attach thread that will go away shortly. We don't need to hot-patch it in that case
+    # anyway, because it will pick up the new thread.start_new_thread that we have set above when it's imported.
     global threading
-    if threading is None:
-        # we need to patch threading._start_new_thread so that 
-        # we pick up new threads in the attach case when threading
-        # is already imported.
+    if threading is None and 'threading' in sys.modules:
         import threading
         threading._start_new_thread = thread_creator
```
zooba: I don't remember this code...

Fairly sure using `global ... import` like this is an implementation detail and not guaranteed (the docs for 2.7 explicitly say "import statement binds the module name in the local namespace" and only assignments refer to the global statement).

It looks like we can replace it with `sys.modules.get('threading')`, and if the initial import of threading is that risky then we should probably do that everywhere.

Python/Product/PythonTools/visualstudio_py_repl.py
```diff
@@ -14,6 +14,11 @@
 
 from __future__ import with_statement
 
+# This module MUST NOT import threading in global scope. This is because in a direct (non-ptvsd)
+# attach scenario, it is loaded on the injected debugger attach thread, and if threading module
+# hasn't been loaded already, it will assume that the thread on which it is being loaded is the
+# main thread. This will cause issues when the thread goes away after attach completes.
+
```
zooba: Remind me - why does debug attach need to import the repl module? Is it just an accidental transitive dependency?

Python/Tests/TestData/DebuggerProject/AttachOutput.py
```diff
@@ -1,6 +1,8 @@
 import sys
 
-while type(sys.stdout).__name__ != '_DebuggerOutput': pass
+attached = False
+while not attached:
+    pass
```
zooba: Maybe a comment explaining how this works or where to look? Just in case someone comes along and wants to reuse this .py file for another test.

Python/Product/PythonTools/visualstudio_py_debugger.py
int19h: We only use it in one other place in this file, in `enum_thread_frames_locally`, to retrieve the thread name. But there we actually need it, I think. And it's a function called often enough that doing an import every time is probably too costly, so caching it as a global is desirable.

This could be rewritten as import + explicit assignment, though.

Python/Product/PythonTools/visualstudio_py_repl.py
int19h: To implement the "connect REPL" command for the debug REPL. Unfortunately, that cannot just import the module locally when and where it needs it, because in direct attach it's not on the import path - so PyDebugAttach has to manually load and inject it (and util).

zooba:
+1 Merge whenever you like

int19h:
I messed up a rebase in my branch, so I'm going to redo this - will need a new PR.

PR #316, opened 2015-06-04
Fixes memory leaks in unit tests.

zooba:
The main causes were:
* TaskProvider.Worker not being correctly terminated
* CodeWindowManager having static caches
* MockVsServiceProvider not disposing services (not necessary within VS, but we need it for tests)
* Events not being unhooked and unnecessary closures

*Please* carefully check the new logic in `TaskProvider` for me. It deserves very thorough review at this point.


Python/Product/PythonTools/PythonTools/Intellisense/TaskProvider.cs
```diff
@@ -446,44 +481,59 @@ abstract class TaskProvider : IVsTaskProvider, IDisposable {
...
             } catch (OperationCanceledException) {
-                // Expected when the timeout expires
-            } catch (ObjectDisposedException ex) {
-                // We have been disposed.
-                Debug.Assert(
-                    ex.ObjectName == "BlockingCollection",
-                    "Handled ObjectDisposedException for the wrong type"
-                );
-                return;
+                // Occurs if we accidentally call Apply() on
```
int19h: On what?

int19h:
:+1:

Python/Product/PythonTools/PythonTools/Intellisense/TaskProvider.cs
zooba: AbortMessage. I was halfway through the comment and decided to rename the message, then never finished the comment.

Issue #315, opened 2015-06-04
Not calling Done() after calculating Navigate To items


zooba:
After we have calculated items for the Navigate To list, we need to call Done() (regardless of success or failure). Otherwise we interfere with other languages.

Looks like we're also not checking that we only look into Python projects, so we should add that too.

PR #317, opened 2015-06-04
#315 Not calling Done() after calculating Navigate To items

zooba:
Ensures we always call Done() after vising library nodes.
Also adds check to ensure we only visit nodes from Python projects
Improves handling of tasks and cancellation tokens.


int19h:
:+1:

Issue #318, opened 2015-06-04
InterpreterPlaceholder instances are making it into saved options


zooba:
We should never be saving InterpreterPlaceholder factories into the shared settings, as these will crash when used as factories.

For example:
```
Level . . . Symbol - InstructionPointer - FunctionOffset - SourceLine - SourceFile

0 . . . microsoft_pythontools!Microsoft.PythonTools.Options.InterpreterPlaceholder.get_Configuration - 0x00000000 - 0x0000001D - - <<Failure related >>
1 . . . microsoft_pythontools_analysis_ni!Microsoft.PythonTools.Interpreter.PythonInterpreterFactoryExtensions.IsUIVisible - 0x53490000 - 0x00000013 - - 
2 . . . microsoft_pythontools!Microsoft.PythonTools.Options.PythonInteractiveOptionsControl+__c._UpdateInterpreters_b__5_2 - 0x00000000 - 0x0000000D - - 
3 . . . system_core_ni!System.Linq.Enumerable+WhereEnumerableIterator_1[[System.__Canon,_mscorlib]].MoveNext - 0x70490000 - 0x00000084 - - 
4 . . . system_core_ni!System.Linq.Buffer_1[[System.__Canon,_mscorlib]]..ctor - 0x70490000 - 0x0000010C - - 
5 . . . system_core_ni!System.Linq.OrderedEnumerable_1+_GetEnumerator_d__1[[System.__Canon,_mscorlib]].MoveNext - 0x70490000 - 0x0000005E - - 
6 . . . microsoft_pythontools!Microsoft.PythonTools.Options.PythonInteractiveOptionsControl.UpdateInterpreters - 0x00000000 - 0x00000209 - - 
7 . . . microsoft_pythontools!Microsoft.PythonTools.Options.PythonInteractiveOptionsPage.PyService_InteractiveWindowsChanged - 0x00000000 - 0x0000000E - - 
8 . . . microsoft_pythontools!Microsoft.PythonTools.PythonToolsService.RaiseEnvironmentsChanged - 0x00000000 - 0x000000A7 - - 
9 . . . microsoft_pythontools!Microsoft.PythonTools.Options.PythonInterpreterOptionsControl.AddInterpreterClick - 0x00000000 - 0x0000023B - - 
```

zooba:
Alternatively, we should provide a configuration on placeholders rather than throwing. Will try it to see what else breaks...

zooba:
Providing the configuration is the right fix.

PR #319, opened 2015-06-04
#318 InterpreterPlaceholder instances are making it into saved options

zooba:
Adds a default InterpreterConfiguration for placeholders to ensure they can be displayed in the UI correctly.


int19h:
:+1:

PR #320, opened 2015-06-04
Adds extra logging for diagnosing DB out of date issues.

zooba:
I have seen issues on various machines where some packages (typically pip) always have 1-2 files deleted at the end of an analysis run and then display as out of date. This logging will help diagnose these issues in release builds (I'm yet to come across a debug build with the problem).


int19h:
:+1:

Issue #274, opened 2015-05-29
lightbulb and print on python 3


huguesv:
PTVS.Dev.2015-05-28.VS.2015
Python 3.5.0b1 64-bit

On Python 3, we always get a lightbulb on `print`.  Even when you accept it and add the import, the lightbulb doesn't go away.

![print](https://cloud.githubusercontent.com/assets/1696845/7874767/fcd6168a-0568-11e5-84ac-edbcab980824.png)

zooba:
I can't reproduce this right now, but I have seen it in the past. Once all the current PRs go in I'll test again with the merged code - a few things have changed here recently.

PR #321, opened 2015-06-04
#274 lightbuld and print on python 3

zooba:
Fixes existing variable lookup to only use the name and not the entire expression.


int19h:
:+1:

Issue #298, opened 2015-06-02
Incorrect encoding detection


zooba:
(From https://pytools.codeplex.com/workitem/3041)

I have a brand new installation of Python Tools for VIsual Studio 2.2 RC on a clean install of Visual Studio 2015 RC. The system is running Windows 10 Professional build 10140, and has Python 3.4.3 installed from the official Python.org MSI. I checked out a project I contribute to on GitHub using the new built-in Git SCM provider, and used the "existing code" solution template. It imported all the code files correctly, but no matter which file I choose, I receive the following error upon file open (or debug break):

```
---------------------------
Microsoft Visual Studio
---------------------------
File Load

Some bytes have been replaced with the Unicode substitution character while loading file C:\Users\Anna\Source\Repos\PyIRC3\PyIRC\extensions\basetrack.py with ascii encoding. Saving the file will not preserve the original file contents.
---------------------------
OK   
---------------------------
```

Since I am using Python 3.4 as my (only) interpreter, I am not sure why PTVS seems to think it should be using ASCII encoding... I also see no option to change it. Searching the issue tracker, I saw someone with the opposite problem as me, but the issue is almost a year old and I saw no resolution to it. I also saw no options for default encoding in the Python section of Visual Studio's settings.

zooba:
If you make a project (use the From Existing Python Code template) with all the files in it, do you still get the warnings? Possibly we're defaulting to something other than UTF-8 when opening files that aren't part of a project.

FWIW, the copyright symbol is encoded correctly in UTF-8 in those files, so it's definitely our fault. I can't test it right this second, but if someone can confirm that we're treating files in a project differently here, then we should be able to fix it.

awilfox:
OP from the CodePlex issue here.  I apologise if I wasn't clear; when I said I used the "existing code solution template" I meant I used the From Existing Python Code template.  All of the files are contained in the project.

* [Screenshot](http://i.imgur.com/ghs8xxY.png "Error in Visual Studio 2015")
* [Project/solution file](http://foxkit.us/PyIRC_vs2015.zip)

zooba:
Ah you're right, I just missed it while reading. So we've definitely got an issue here - hopefully we can get to it before our next release, but it's looking a little far down the backlog right now.

PR #324, opened 2015-06-04
#298 Incorrect encoding detection

zooba:
Defaults to opening files with UTF-8 unless a coding comment is included.


int19h:
:+1:

Issue #340, opened 2015-06-08
ptvsd.enable_attach causes process to never exit under Python 2.5


int19h:
This breaks the AttachPtvsd test under 2.5 in particular.

zooba:
Just pushed a change to add a timeout to the wait, which keeps the test from breaking.

int19h:
I'll submit a proper fix, since it's literally a one line change. But proper timeout is good as a fallback.

PR #341, opened 2015-06-08
Prevents timeout in AttachPtvsd test (only occurs for Python 2.5)

zooba:
Helps AttachPtvsd test work better when test target already has ptvsd installed.

See #340


int19h:
:+1:

PR #343, opened 2015-06-08
Specify timeout when waiting for process exit in AttachWithOutputRedirection test

int19h:
Use predefined constant for timeout in AttachPtvsd test.

### DIFF
--- a/Python/Tests/DebuggerTests/DebuggerTests.cs
+++ b/Python/Tests/DebuggerTests/DebuggerTests.cs
@@ -2631,7 +2631,7 @@ int main(int argc, char* argv[]) {
                     Assert.IsTrue(frame.ExecuteTextAsync("attached = True").Wait(20000), "Failed to complete evaluation within 20s");
 
                     proc.Resume();
-                    proc.WaitForExit();
+                    WaitForExit(proc);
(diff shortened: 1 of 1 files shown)


Python/Tests/DebuggerTests/DebuggerTests.cs
```diff
@@ -2685,7 +2685,7 @@ void main()
                     Assert.IsTrue(attached.WaitOne(20000), "Failed to attach within 20s");
                     Assert.IsTrue(bpHit.WaitOne(20000), "Failed to hit breakpoint within 20s");
 
-                    p.WaitForExit(10000);
+                    p.WaitForExit(DefaultWaitForExitTimeout);
```
zooba: Any reason not to use WaitForExit here too?

int19h: It won't work because it expects a `PythonProcess`. And the reason why we can't wait on `proc` in this snippet is because it doesn't work for remote processes - you'll get `InvalidOperationException` if you do. This, in turn, is because `PythonProcess.WaitForExit` really just delegates to `Process.WaitForExit` of the process that it wraps; and there's no `Process` instance for remote processes.

zooba:
+1

PR #344, opened 2015-06-08
Adds handling for dialogs appearing when opening projects in tests.

zooba:
Adds test configurations for Express for Web.


int19h:
:+1:

PR #345, opened 2015-06-08
Fix AttachPtvsd test to pass proper arguments to IronPython to allow ptvsd to work.

int19h:
Increase waiting period for process startup in AttachPtvsd test, as the original value is often not long enough.

### DIFF
--- a/Python/Tests/DebuggerTests/DebuggerTests.cs
+++ b/Python/Tests/DebuggerTests/DebuggerTests.cs
@@ -17,6 +17,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Net.Sockets;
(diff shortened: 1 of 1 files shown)


Python/Tests/DebuggerTests/DebuggerTests.cs
```diff
@@ -2650,13 +2651,28 @@ void main()
                 WorkingDirectory = TestData.GetPath(),
                 UseShellExecute = false
             };
+            if (this is DebuggerTestsIpy) {
+                psi.Arguments = "-X:Tracing -X:Frames " + psi.Arguments;
+            }
```
zooba: Would rather this was a virtual property - DebugXArgs or something.

zooba:
+1 with a property or a half-good reason not to bother right now.

gilbertw:
:ok:

Issue #351, opened 2015-06-09
Repl is slow when displaying lots of coloured output


zooba:
If you type an error in pandas when in IPython mode, it can be very slow to display the full traceback.

PR #352, opened 2015-06-09
#351 Repl is slow when displaying lots of coloured output

zooba:
#351 Repl is slow when displaying lots of coloured output
Batches edits to the repl window buffer so that the updates all occur at once.
Removes unnecessary marshalling to the UI thread.
Fixes error when opening interactive window.


zooba:
Would like @DinoV to have a look at this one.

Issue #376, opened 2015-06-09
Green interactive colour is too bright in light themes


zooba:
Make it darker

PR #380, opened 2015-06-10
#376 Green interactive color is too bright in light themes

zooba:
#376 Green interactive color is too bright in light themes
Makes the color used for VS 2013 dimmer.


gilbertw:
:+1:

Issue #362, opened 2015-06-09
__init__ completion too aggressive


DinoV:
```python
class C(object):

    pass
```
Go to add __init__ method and type: "def __init__("

You'll end up with:
```python
class C(object):
    def __init__(self, **kwargs):
        return super().__init__(**kwargs)(
    pass
```
We should probably not commit the completion if the user has typed the full display text.

PR #381, opened 2015-06-10
#362 __init__ completion too aggressive

zooba:
#362 __init__ completion too aggressive
Makes commitByDefault argument optional to avoid overriding settings from the completion set.
Fixes Ctrl+Space throwing exceptions when completing a filtered name.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/FuzzyCompletionSet.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/FuzzyCompletionSet.cs
@@ -235,6 +235,9 @@ namespace Microsoft.PythonTools.Intellisense {
         /// If True, the best item is selected by default. Otherwise, the user
         /// will need to manually select it before committing.
         /// </summary>
+        /// <remarks>
+        /// By default, this is set to <see cref="DefaultCommitByDefault"/>
+        /// </remarks>
         public bool CommitByDefault { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Python/Product/PythonTools/PythonTools/Intellisense/FuzzyCompletionSet.cs
```diff
@@ -221,6 +223,8 @@ public class FuzzyCompletionSet : CompletionSet {
                 }
                 _filteredCompletions.Filter(IsVisible);
             }
+
+            CommitByDefault = DefaultCommitByDefault;
```
gilbertw: Is part of the change missing.  The const bool is only referenced once.  Why not simply "= true" here?

zooba: I *was* comparing the value against the default to see if it had been changed, before switching to the nullable bool. As this is public API, I see no harm in programmatically exposing the default vs. documentation only.

gilbertw: I like the xml documentation which is easier to find.  With the documentation you could call out its behavior as well.  The public const requires they guess its usage unless they have the code open, in which case its no better than true.  I doubt folks will be fishing through the class via iSense or object explorer when they could just crack code.  Perhaps I'm just missing the programmatic experience you're thinking of.

gilbertw:
:+1:

Python/Product/PythonTools/PythonTools/Intellisense/FuzzyCompletionSet.cs
zooba: Considering we don't publish the XML docs anywhere, the only place you'll find them is in IntelliSense or by reading the source code. But I'll add it and refer to the constant to cover all bases.

Issue #374, opened 2015-06-09
Local variables are not initially colorized


MikhailArkhipov:
New empty python file in VS (Python Web project)
Paste

def z(x):
    if x == 0:
        bar()
    else:
        z(x - 1)

note 'x' is colorized. 
Save and close file  
Open it again

Observed: x is black. 

Type space at the end of the file.

Observed: now x is colorized;

PR #382, opened 2015-06-10
Intellisense fixes

### LINKED ISSUE #370: Reverse expression parser needs updating for async
```python
async def bar():
    return 100

async def foo():
    (await bar())
```

Do . after (await bar()) and you get generator completions instead of int completions.

zooba:
#370 Reverse expression parser needs updating for async
Always parses context-dependent keywords as keywords when parsing isolated expressions.
Improves stability of tests.

Fixes tokenizer crash when parsing invalid complex value.

#374 Local variables are not initially colorized
Ensures the event is hooked and classifications are calculated when analysis is complete before classifications are requested.
Improves test stability
(Originally wasn't going to look at #374 for 2.2, but needed the fix for test stability or we wouldn't ever get classifications.)

### DIFF
--- a/Python/Product/Analysis/PythonKeywords.cs
+++ b/Python/Product/Analysis/PythonKeywords.cs
@@ -72,15 +72,15 @@ namespace Microsoft.PythonTools.Analysis {
         public static IEnumerable<string> Expression(PythonLanguageVersion version = PythonLanguageVersion.None) {
             yield return "and";
             yield return "as";
-            if (version == PythonLanguageVersion.None || version >= PythonLanguageVersion.V35) {
+            if (version.IsNone() || version >= PythonLanguageVersion.V35) {
                 yield return "await";
             }
             yield return "else";
-            if (version == PythonLanguageVersion.None || version.Is3x()) {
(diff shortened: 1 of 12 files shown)


Python/Product/Analysis/PythonKeywords.cs
```diff
@@ -72,15 +72,15 @@ public static class PythonKeywords {
         public static IEnumerable<string> Expression(PythonLanguageVersion version = PythonLanguageVersion.None) {
             yield return "and";
             yield return "as";
-            if (version >= PythonLanguageVersion.V35) {
+            if (version == PythonLanguageVersion.None || version >= PythonLanguageVersion.V35) {
```
gilbertw: version.IsNone()

zooba: As in, create a new function for this?

gilbertw: Yes,
static bool IsNone(this PythonLanguageVersion version){
  return version == PythonLanguageVersion.None;
}

int19h:
:+1:

crwilcox:
:+1:  lgtm

PR #383, opened 2015-06-10
Fixes #121 IntelliSense suggestions in exceptions are missing

### LINKED ISSUE #121: IntelliSense suggestions in exceptions are missing
Repro.  There are no suggestions in the ValueError for hello.

``` python
hello = 'hello'

raise ValueError(he
```

zooba:
Fixes #121 IntelliSense suggestions in exceptions are missing
Only keep suggesting exceptions when the grouping is the first item.

(Doesn't meet the bar, but it's such a trivial fix to something new since 2.1 that if people want to convince me to take it I won't argue too hard.)


huguesv:
I'm okay with taking it.

crwilcox:
:+1: lgtm

PR #384, opened 2015-06-10
Fixes #379 Cannot change architecture of a custom environment

### LINKED ISSUE #379: Cannot change architecture of a custom environment
I have WinPython 3.4 64-bit mistakenly added as a custom 32-bit environment. When trying to change this in the Environments window, on the "Configure" tab, I flip Architecture to 64-bit and click "Apply". The tab changes to "Overview", and if I switch back to "Configure" I can see that it shows up as 32-bit again.

zooba:
Fixes #379 Cannot change architecture of a custom environment
Converts architecture string to processor architecture, since the configurable provider expects different strings.


gilbertw:
:+1:

PR #385, opened 2015-06-10
#364 Tools | Options | Text Editor | Python | Formatting | Spacing can't be loaded

### LINKED ISSUE #364: Tools | Options | Text Editor | Python | Formatting | Spacing can't be loaded
VS 2015 RC 23008
New Python Web project
Tools | Options | Text Editor | Python | Formatting | Spacing 

VS shows 'an error loading property page'. Same with Statements and Wrapping. General appears to be OK.

zooba:
Fixes #364 Tools | Options | Text Editor | Python | Formatting | Spacing can't be loaded
Removes folder images from treeview.
Also prevents IntelliSense providers from loading against the preview editor to avoid many assertions.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/SmartTagControllerProvider.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/SmartTagControllerProvider.cs
@@ -22,23 +22,25 @@ using Microsoft.VisualStudio.Utilities;
 namespace Microsoft.PythonTools.Intellisense {
 #if !DEV14_OR_LATER
     [Export(typeof(IIntellisenseControllerProvider)), ContentType(PythonCoreConstants.ContentType), Order]
+    [TextViewRole(PredefinedTextViewRoles.Analyzable)]
+    [TextViewRole(PredefinedTextViewRoles.Editable)]
     class SmartTagControllerProvider : IIntellisenseControllerProvider {
 
-        #region MEF Imports
+    #region MEF Imports
 
         [Import]
         public ISmartTagBroker smartTagBroker = null;
 
-        #endregion
+    #endregion
 
-        #region IIntellisenseControllerProvider Members
+    #region IIntellisenseControllerProvider Members
 
         public IIntellisenseController TryCreateIntellisenseController(ITextView textView,
                                                                        IList<ITextBuffer> subjectBuffers) {
             return SmartTagController.CreateInstance(this.smartTagBroker, textView, subjectBuffers);
         }
 
-        #endregion
+    #endregion
     }
 #endif
 }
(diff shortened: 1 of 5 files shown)


Python/Product/PythonTools/PythonTools/Intellisense/SmartTagControllerProvider.cs
```diff
@@ -22,23 +22,25 @@
 namespace Microsoft.PythonTools.Intellisense {
 #if !DEV14_OR_LATER
     [Export(typeof(IIntellisenseControllerProvider)), ContentType(PythonCoreConstants.ContentType), Order]
+    [TextViewRole(PredefinedTextViewRoles.Analyzable)]
+    [TextViewRole(PredefinedTextViewRoles.Editable)]
     class SmartTagControllerProvider : IIntellisenseControllerProvider {
 
-        #region MEF Imports
+    #region MEF Imports
```
zooba: Removing these regions. We don't need two regions for a three line class...

Python/Product/PythonTools/PythonTools/Options/OptionsTreeView.cs
```diff
@@ -42,28 +39,16 @@ class OptionsTreeView : TreeView {
...
-                    using (var g = Graphics.FromImage(bmp)) {
-                        g.FillRectangle(
-                            Brushes.Magenta,
-                            new Rectangle(0, 0, 16, 16)
-                        );
-                    }
-                    _imageList.Images.Add(bmp);
-                    _imageList.Images.Add(openFolder);
-                    _imageList.Images.Add(closedFolder);
+                _imageList = new ImageList();
```
huguesv: Is it a problem if StateImageList is initialized to null?  If not, we could remove _imageList completely.

zooba: I think we need the empty image to get proper layout - don't really want to dig deep enough to figure out how to get that layout without it, and frankly I'd rather just use a WPF treeview there so we don't have to do the ownerdrawn stuff at all. But that's for a future release.

Right now, _imageList is static and that's how we avoid creating lots of image lists over the life of the process.

huguesv: Oh I missed the static there, yeah, better not change it.  I agree about WPF treeview.  I know the high dpi support on these formatting options pages has been problematic at 200%+ and WPF.  I haven't ported the issues from codeplex as I don't have access to a machine that can reasonably go that high anymore.

huguesv:
lgtm

Python/Product/PythonTools/PythonTools/Options/OptionsTreeView.cs
zooba: My laptop does, and I can confirm that it's still an issue :)

PR #392, opened 2015-06-11
Fix flaky AttachThreadingStartNewThread test - don't resume thread until execution is completed.

int19h:


Python/Tests/DebuggerTests/DebuggerTests.cs
```diff
@@ -1952,10 +1952,12 @@ private enum ExceptionMode {
                     PythonThread bpThread = null;
                     bool wrongLine = false;
                     proc.BreakpointHit += (sender, args) => {
+                        var continueEvent = new AutoResetEvent(true);
```
zooba: Can we scope this a little narrower and use a using()?

Python/Tests/DebuggerTests/DebuggerTests.cs
```diff
@@ -1968,6 +1970,8 @@ private enum ExceptionMode {
                             attached.Set();
                             breakpointHit.Set();
                         }
+
+                        continueEvent.WaitOne(1000);
```
gilbertw: You only wait a second here and then continue.  Would be better to Assert and fail the test if the command didn't run in the amount of time expected

int19h: The tricky part is asserting on the right thread (this callback gets called on the debugger background thread, which the test runner knows nothing about - if assert fails on it, it just kills the test without actually logging the assert properly).

Given that the test will basically just fail anyway if we time out here, it feels that adding a bunch more code here to propagate the assert is not worth it.

gilbertw: I agree not worth the complication.

gilbertw:
:+1:

Python/Tests/DebuggerTests/DebuggerTests.cs
int19h: It seems it's already scoped as narrow as it can possibly be (in the same scope where it's waited on).

I'll add a `using`, but note that pretty much none of these tests here do that for the events that they use. We should probably just clean it all up at some point.

zooba: Surely it can be scoped to just around the four lines of code that run between the Reset() and the wait? We only ever set it in one branch, and it's an auto-reset, so presumably we're waiting 1s in most cases when we don't need to be.

Now's as good a time as any to clean up all the tests.

Python/Tests/DebuggerTests/DebuggerTests.cs
int19h: Actually this is wrong altogether, it shouldn't wait in the debugger callback because that blocks the debugger thread and prevents the execution result from being received - so it was _always_ timing out (but waiting still broke the race condition so I didn't notice). I'll rewrite this to wait on the main thread, which can then properly assert.

Python/Tests/DebuggerTests/DebuggerTests.cs
int19h: As far as scoping goes, I had to move `Continue` out of the callback anyway (see my reply to Rick below). But it can still have a `using` - and I'll add the same for other events in this test. I'll make a separate change to fix other tests to avoid spamming this CR.

int19h:
Addressed feedback.

PR #396, opened 2015-06-12
Fixes #395 Interactive Window DTE commands no longer accept arguments

### LINKED ISSUE #395: Interactive Window DTE commands no longer accept arguments
We should be able to type `Python.Interactive /e:"Python 2.7"` in the command window (or invoke the equivalent via DTE) to open a specific interactive window. This has stopped working in VS 2015.

zooba:
Fixes #395 Interactive Window DTE commands no longer accept arguments
Removes eager (and incorrect) handling of commands in the edit filter and lets the global handler deal with them.
Avoids crashing in IntellisenseController after failing to get the view adapter for a particular view.

(All the repl window tests depend on this, so it's well tested.)


huguesv:
:+1:

PR #398, opened 2015-06-12
Makes PythonEditor wait for analyzer to load before continuing

zooba:
Makes PythonEditor wait for analyzer to load before continuing. This stabilises (at least) the refactor rename tests.


crwilcox:
lgtm :+1:

PR #401, opened 2015-06-12
Updates interactive window tests for VS 2015

zooba:
Updates interactive window tests for VS 2015
Adds some of the interactive window commands that were missing from VS 2013
Adds workaround for Shift+Enter shortcut not working in VS 2015 RC
Fixes null reference error reading from standard input.
Fixes references in all projects
Fixes newlines in shared REPL window code.

Please pay particular attention to the files in `Python\Product` - these fix crashes. The rest are test related and may still have issues that can't be discovered without running the full pass on our test machines.


Python/Product/Django/Django.csproj
```diff
@@ -113,6 +113,31 @@
     </Reference>
     <Reference Include="WindowsBase" />
   </ItemGroup>
+  <Choose>
```
int19h: This is long enough, and repeats often enough, that I wonder if perhaps it should have its own .proj file that everyone can then just import?

Common/Tests/Utilities.UI/UI/ReplWindowProxy.cs
```diff
@@ -103,43 +68,33 @@ >>> int('0b100', base=0)
...
-        internal ReplWindowProxy(PythonVisualStudioApp app, ReplWindow window, ReplWindowProxySettings settings) {
+        internal ReplWindowProxy(VisualStudioApp app, IInteractiveWindow window, ToolWindowPane toolWindow, ReplWindowProxySettings settings) {
             Assert.IsNotNull(app, "app is required");
             Assert.IsNotNull(window, "window is required");
             _app = app;
             _window = window;
+            _toolWindow = toolWindow;
             _settings = settings;
-            _replWindowInfo = _replWindows.GetOrCreateValue(_window);
+            _replWindowInfo = _replWindows.GetOrCreateValue(toolWindow);
```
crwilcox: nit.  you bothered to make _toolWindow.  Should you use it?

Common/Tests/Utilities.UI/UI/ReplWindowProxy.cs
```diff
@@ -185,22 +142,22 @@ internal sealed class ReplWindowProxy : IDisposable {
...
+                result = OpenInteractive(app, settings, useIPython ? "IPython" : "Standard");
                 app = null;
 
-                result.Window.Reset();
-                result.ClearInput();
-
                 for (int retries = 10; retries > 0; --retries) {
-                    result.Window.Reset();
+                    result.Reset();
+                    Thread.Sleep(1000);
```
crwilcox: Is the wait necessary? If so is 1 second the right amount here?

Common/Tests/Utilities.UI/UI/ReplWindowProxy.cs
```diff
@@ -550,40 +479,52 @@ public class SessionHolder<T> : IDisposable where T : IIntellisenseSession {
...
+            // notifications via ReadyForInput.
+            for (int checks = 10; checks > 0; --checks) {
+                if (_replWindowInfo.ReadyForInput.WaitOne(TimeSpan.FromSeconds(0.1))) {
+                    return true;
+                }
+                if (IsCaretInStandardInputRegion) {
+                    return true;
+                }
+            }
+            return _replWindowInfo.ReadyForInput.WaitOne(timeout.Subtract(TimeSpan.FromSeconds(1)));
```
crwilcox: we don't really know that we waited the full second so we might be shorting the caller up to .9 seconds on this timeout if I understand?  Not a big deal really, but worth mentioning.

Python/Product/Django/Django.csproj
zooba: I considered that, but fairly sure VS won't handle it properly. The C# project system doesn't respect all of MSBuild's functionality...

Common/Tests/Utilities.UI/UI/ReplWindowProxy.cs
zooba: If we've waited any less then we've returned already. If we reach the final wait then we've done at least 1 second.

Common/Tests/Utilities.UI/UI/ReplWindowProxy.cs
zooba: That's barely even a nit... _toolWindow gets used elsewhere, but there's no need to use it here.

Common/Tests/Utilities.UI/UI/ReplWindowProxy.cs
zooba: Doesn't look like it's necessary. I was testing something..

Common/Tests/Utilities.UI/UI/ReplWindowProxy.cs
crwilcox: Ah.  Good point.

crwilcox:
:+1:

Python/Product/Django/Django.csproj
int19h: I would actually be surprised if they can handle references inside `Choose`, but not imports. If they get the `Reference` items from MSBuild, it should take care of it. If they parse unevaluated items themselves, they probably aren't handling conditions right, anyway (beyond the most basic stuff like configurations).

int19h:
:+1:

PR #409, opened 2015-06-13
Test fixes

zooba:
Fixes a timeout issue
Removes VSTestHost version specifier, which makes it easier to build against different versions.


int19h:
:+1:

PR #425, opened 2015-06-14
Fixes #407 Interactive options does not load environment list

### LINKED ISSUE #407: Interactive options does not load environment list
When you open Tools, Options, Python Tools, Interactive Windows the list of interpreters is not filled. You have to go to Environment Options for it to load.

zooba:
Fixes #407 Interactive options does not load environment list
Loads options dialog state when creating the Python Tool service.


int19h:
:+1:

PR #426, opened 2015-06-14
Fixes #423 Installer does not remove VS 2013 template

### LINKED ISSUE #423: Installer does not remove VS 2013 template
Apparently we need an update to remove the redirection template from VS 2013

zooba:
Fixes #423 Installer does not remove VS 2013 template
Corrects "elif" to "elseif" in WiX authoring.


zooba:
Why oh why didn't this cause a compilation error :(

int19h:
:+1:

PR #429, opened 2015-06-14
Fixes #427 Deadlock specializing methods while updating modules

### LINKED ISSUE #427: Deadlock specializing methods while updating modules
ProjectAnalyzer.cs uses a _contentsLock lock to protect a number of different things, most of which don't need protection. In particular, OnModulesChanged holds the lock while it reloads modules, which could deadlock.

We should release the lock and asynchronously execute ReloadModulesAsync() rather than blocking.

zooba:
Fixes #427 Deadlock specializing methods while updating modules
Removes _contentsLock from PythonAnalyzer
Adds locking to CPythonInterpreter.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
@@ -94,8 +94,6 @@ namespace Microsoft.PythonTools.Intellisense {
         private readonly PythonToolsService _pyService;
         private readonly IServiceProvider _serviceProvider;
 
-        private readonly object _contentsLock = new object();
-
         internal Task ReloadTask;
 
         internal VsProjectAnalyzer(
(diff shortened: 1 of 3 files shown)


Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -188,19 +186,17 @@ public sealed class VsProjectAnalyzer : IDisposable {
...
-        private void OnModulesChanged(object sender, EventArgs e) {
+        private async void OnModulesChanged(object sender, EventArgs e) {
             Debug.Assert(_pyAnalyzer != null, "Should not have null _pyAnalyzer here");
             if (_pyAnalyzer == null) {
                 return;
             }
 
-            lock (_contentsLock) {
-                _pyAnalyzer.ReloadModulesAsync().WaitAndUnwrapExceptions();
+            await _pyAnalyzer.ReloadModulesAsync();
```
int19h: Given that this is async now, it seems that it'll have to deal with the possibility of _pyAnalyzer being disposed concurrently.

zooba: We had to deal with that possibility anyway (and we don't do a good job :( ) as it was being raised from a background thread already. Fairly sure we aren't any less safe now than we were before, unless we hit a crash where previously we'd have hit the deadlock first.

int19h:
:+1:

PR #434, opened 2015-06-15
Update README.md

zooba:


zooba:
I think I made this in the wrong repo? Not that it matters, but FYI, be careful directly editing files when you have write access :)

README.md
```diff
@@ -1,13 +1,13 @@
...
+Visit our page on [VisualStudio.com](https://www.visualstudio.com/en-us/explore/python-vs) for an overview and videos, and check out our [Getting Started](http://aka.ms/PTVSTutorial) series on YouTube.
 
-Our [documentation](https://github.com/Microsoft/PTVS/wiki) is now on the wiki, and is _mostly_ looking good (feel free to help us out). Also, the [issue tracker](http://github.com/Microsoft/PTVS/issues) here is where we are managing bugs and feature requests, as well as any questions you may have. We are not planning to transfer all our [existing issues](https://pytools.codeplex.com/workitem/list/advanced) from CodePlex, so if you have one there that is really important to you, please recreate it with a link to the original so we don't lose it.
+Our [documentation](https://github.com/Microsoft/PTVS/wiki) is now on the wikigood (feel free to help us improve it), and we manage all of our issues here on the [issue tracker](http://github.com/Microsoft/PTVS/issues) or in the gitter chat linked above. Pull requests and feature suggestions are welcome - see the [Contributing](https://github.com/Microsoft/PTVS/wiki#contributing) section for information.
```
gilbertw: wikigood should be wiki

gilbertw: Drop the "in gitter chat".  We should track issues in github to ensure they get resolved and to ensure history of answers is easily searchable.

gilbertw:
:+1:

README.md
zooba: You're right, though I would rather have initial questions come in that way so we can transfer them if they're valuable. I'll reword.

PR #436, opened 2015-06-15
Some fixes for debugger tests that require compiling C++ code

int19h:
- Add Common7\IDE to PATH to pick up some DLLs for VC10.
- Prepend rather than append new entries to PATH to ensure that they override everything else there.
- Always use Win7 SDK for VC10 rather than Win8 SDK (the latter has headers that don't compile with VC10).

Add extra error logging to debugger tests for child properties to help diagnose test run issues.


zooba:
+1

PR #437, opened 2015-06-15
Fix #411: Cannot attach debugger to interactive window

### LINKED ISSUE #411: Cannot attach debugger to interactive window
From AttachReplTest:

```
  File "...\visualstudio_py_repl.py", line 645, in run_one_command
    self.execute_item()
  File "...\visualstudio_py_repl.py", line 996, in execute_attach_process_work_item
    visualstudio_py_debugger.attach_process(port, debugger_id, report = True, block = True)
TypeError: attach_process() missing 1 required positional argument: 'debug_options'
```

int19h:
Propagate debug options (that make sense in that context) to the REPL, and then from the REPL to the main debugger script, when using $attach command.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Repl/BasePythonReplEvaluator.cs
+++ b/Python/Product/PythonTools/PythonTools/Repl/BasePythonReplEvaluator.cs
@@ -444,10 +444,18 @@ namespace Microsoft.PythonTools.Repl {
                     return "Cannot attach to debugger when already attached.";
                 }
 
+                var pyService = _eval._serviceProvider.GetPythonToolsService();
+
+                // Only pass options that make sense for the REPL.
+                var debugOptions = PythonDebugOptions.None;
+                if (pyService.DebuggerOptions.DebugStdLib) {
+                    debugOptions |= PythonDebugOptions.DebugStdLib;
+                }
+
                 PythonProcess debugProcess;
                 using (new StreamLock(this, throwIfDisconnected: true)) {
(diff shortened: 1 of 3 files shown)


Python/Product/PythonTools/PythonTools/Repl/BasePythonReplEvaluator.cs
```diff
@@ -444,10 +444,18 @@ public DisconnectedException(string message)
                     return "Cannot attach to debugger when already attached.";
                 }
 
+                var pyService = _eval._serviceProvider.GetPythonToolsService();
+
+                // Only pass options that make sense for the REPL.
+                var debugOptions = PythonDebugOptions.None;
+                if (pyService.DebuggerOptions.DebugStdLib) {
+                    debugOptions |= PythonDebugOptions.DebugStdLib;
```
gilbertw: Could future options be interesting to pass.  If so add a comment to PythonDebugOptions letting developers know to update this location when they add a new option

gilbertw:
:+1:

Python/Product/PythonTools/PythonTools/Repl/BasePythonReplEvaluator.cs
int19h: I would assume that anyone adding things to PythonDebugOptions would do a Find All References on it to see all uses, at the minimum. So comments are for cases that are not covered by that (e.g. for stuff in .py files).

zooba:
+1

PR #438, opened 2015-06-15
Fix #419: ErrorListTaskListTests not seeing any items on VS 2013

### LINKED ISSUE #419: ErrorListTaskListTests not seeing any items on VS 2013
Similar error in all related tests on VS 2013
```
Assert.AreEqual failed. Expected:<7>. Actual:<0>. 
   at TestUtilities.UI.VisualStudioApp.WaitForTaskListItems(Type taskListService, Int32 expectedCount, Boolean exactMatch) in c:\RollBuild\Common\Tests\Utilities.UI\UI\VisualStudioApp.cs:line 987
   at PythonToolsUITests.ErrorListTaskListTests.TaskListTest(VisualStudioApp app, Type taskListService, IList`1 expectedItems, Int32[] navigateTo) in c:\RollBuild\Python\Tests\Core.UI\ErrorListTaskListTests.cs:line 88
   at PythonToolsUITests.ErrorListTaskListTests.ErrorList() in c:\RollBuild\Python\Tests\Core.UI\ErrorListTaskListTests.cs:line 135
 ```

int19h:
Increase number of retries while waiting for Error List to asynchronously update before giving up.

### DIFF
--- a/Common/Tests/Utilities.UI/UI/VisualStudioApp.cs
+++ b/Common/Tests/Utilities.UI/UI/VisualStudioApp.cs
@@ -970,7 +970,7 @@ namespace TestUtilities.UI {
                 System.Threading.Thread.Sleep(5000);
             }
 
-            for (int retries = 10; retries > 0; --retries) {
+            for (int retries = 20; retries > 0; --retries) {
                 allItems.Clear();
                 IVsEnumTaskItems items;
                 ErrorHandler.ThrowOnFailure(errorList.EnumTaskItems(out items));


gilbertw:
Why the need to increase now?  Have we become slower to populate this, hardware has become slower (unlikely), tests never really worked, etc....  I'm concerned we may have regressed performance

int19h:
We have just started to pay more attention to intermittent test failures, I think.

zooba:
Yeah, it's probably been failing on and off for a while, but there's been so much noise we haven't noticed. The tests are starting to get better, so these are starting to become more obvious.

FWIW, I expect the vast majority of PTVS work until 2.2 final will be improving tests.

zooba:
+1

PR #439, opened 2015-06-16
Uwp targets update to give error text

juanyaw:
Allows the project to load for UWP projects even if the SDK isn't installed and gives a build error indicating where to get CPython UWP SDK if not installed

### DIFF
--- a/Python/Product/Uwp/Microsoft.PythonTools.Uwp.targets
+++ b/Python/Product/Uwp/Microsoft.PythonTools.Uwp.targets
@@ -7,7 +7,7 @@
   </PropertyGroup>
   
   <Target Name="PtvsUwpImportCheck">
-	<Error Text="Could not find CPython UWP SDK.  Please download from http://go.microsoft.com/fwlink/?LinkID=613495&amp;clcid=0x409" Condition="!Exists($(CPythonTargetsFile))" />
+	<Error Text="Could not find CPython UWP SDK.  Please download from http://go.microsoft.com/fwlink/?LinkID=613495&amp;clcid=0x409 and reload the project after install." Condition="!Exists($(CPythonTargetsFile))" />
   </Target>
 
   <Target Name="CoreCompile" Condition="!Exists($(CPythonTargetsFile))" />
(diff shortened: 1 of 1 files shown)


Python/Product/Uwp/Microsoft.PythonTools.Uwp.targets
```diff
@@ -1,12 +1,18 @@
...
     <CPythonTargetsFile>$(LocalAppData)\Microsoft\VisualStudio\$(VisualStudioVersion)Exp\Extensions\Microsoft\Python UWP\$(InterpreterVersion)\DesignTime\CommonConfiguration\Neutral\CPython.targets</CPythonTargetsFile>
     <CPythonTargetsFile Condition="'$(CPythonTargetsFile)' == '' or !Exists('$(CPythonTargetsFile)')">$(MSBuildProgramFiles32)\Microsoft SDKs\Windows\v10.0\ExtensionSDKs\Python UWP\$(InterpreterVersion)\DesignTime\CommonConfiguration\Neutral\CPython.targets</CPythonTargetsFile>
   </PropertyGroup>
   
-  <Import Project="$(CPythonTargetsFile)" />
+  <Target Name="PtvsUwpImportCheck">
+	<Error Text="Could not find CPython UWP SDK.  Please download from http://go.microsoft.com/fwlink/?LinkID=613495&amp;clcid=0x409 and reload the project after install." Condition="!Exists($(CPythonTargetsFile))" />
+  </Target>
+
+  <Target Name="CoreCompile" Condition="!Exists($(CPythonTargetsFile))" />
```
zooba: The condition here doesn't actually do anything - the target is always defined and optionally run.

zooba: I assume it is overridden by the actual targets file?

zooba:
Once we pull out the unnecessary condition, it looks good to go in.

Python/Product/Uwp/Microsoft.PythonTools.Uwp.targets
juanyaw: I'm trying to verify the Target with the condition when SDK is installed now.  It is just defined in the SDK targets as well and wasn't sure if there would be a problem (assuming you are talking about CoreCompile Target)

zooba: The condition shouldn't matter. If you build with the SDK installed then I expect you'll see a message saying that the target has been replaced. The condition isn't actually used until the target is going to be run, which is meaningless here anyway because it's empty.

PR #451, opened 2015-06-17
Fixes #450 NullReferenceException testing test adapter

### LINKED ISSUE #450: NullReferenceException testing test adapter
```
First-chance exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.PythonTools.TestAdapter.TestExecutor.RunTestCase(IFrameworkHandle frameworkHandle, IRunContext runContext, TestCase test, Dictionary`2 sourceToSettings) in C:\Source\Python\Product\TestAdapter\TestExecutor.cs:line 146
```

Looks like this may also be breaking the entire test run and causing lots of skips.

zooba:
Fixes #450 NullReferenceException testing test adapter
Adds missing null check for _app.

### DIFF
--- a/Python/Product/TestAdapter/TestExecutor.cs
+++ b/Python/Product/TestAdapter/TestExecutor.cs
@@ -145,8 +145,8 @@ namespace Microsoft.PythonTools.TestAdapter {
 
             var testCase = new PythonTestCase(settings, test, debugMode);
 
-            var dte = _app.GetDTE();
-            if (debugMode != PythonDebugMode.None) {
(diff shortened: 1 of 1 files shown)


gilbertw:
:+1:

huguesv:
Will we now just crash later at line 216 when it uses `_app` without checking for null?

zooba:
If `_app` is null then the debug mode is None, so we won't hit the later case (but we still need DTE to find some VS settings if we can get it).

huguesv:
:+1:

Issue #415, opened 2015-06-14
Doc-strings need a stable ordering


zooba:
```
Assert.AreEqual failed. Expected:<help 1

help 2>. Actual:<help 2

help 1>. 
   at AnalysisTests.AnalysisSaveTest.OverloadDocString() in c:\RollBuild\Python\Tests\Analysis\AnalysisSaveTest.cs:line 178
```

Alternatively, the test needs updating to not care about ordering.

zooba:
This is resolved (overloads are sorted by number of parameters)

PR #462, opened 2015-06-18
#415 Doc-strings need a stable ordering

zooba:
#415 Doc-strings need a stable ordering
Updates the test to not care about ordering.

Leaving the issue open so we can look at making the order stable after 2.2.


huguesv:
:+1:

int19h:
:+1:

PR #466, opened 2015-06-18
Fixes #465 Analyzer does not follow .pth files when a library path is provided

### LINKED ISSUE #465: Analyzer does not follow .pth files when a library path is provided
We changed in 2.2 to ask the interpreter for `sys.path` rather than trying to calculate it, but only when no library path is provided on the command line. If a library path is provided, we fall back on the old behaviour.

Except we (i.e. me) forgot to expand .pth files in the fallback code. The `TestPthFiles` test is failing as a result.

zooba:
Fixes #465 Analyzer does not follow .pth files when a library path is provided
Expands path files when handling the fallback case of a library path being passed to the analyzer.

As far as regressions go this one is pretty non-fatal, though it is a break with our old behaviour. It *shouldn't* affect normal use through the IDE, but the point was to keep the old behaviour when expected.


int19h:
:+1:

huguesv:
:+1:

PR #470, opened 2015-06-18
Fixes #469 Virtual environments always refresh DB even when option is disabled

### LINKED ISSUE #469: Virtual environments always refresh DB even when option is disabled
We have "Tools, Options, Python Tools, General, Automatically generate completion DB on first use" to avoid automatically kicking off analysis.

However, virtual environments ignore this option and will always automatically generate.

zooba:
Fixes #469 Virtual environments always refresh DB even when option is disabled
Removes unnecessary call to GenerateDatabase
Updates virtual environment test that exposed this issue


int19h:
:+1:

huguesv:
:+1:

PR #473, opened 2015-06-18
Fixes #472 Execute in REPL should not create a window in QueryStatus

### LINKED ISSUE #472: Execute in REPL should not create a window in QueryStatus
`ExecuteInReplCommand` has a call in its `QueryStatusMethod` to `EnsureReplWindow`. We should not be creating the window in query status.

(It also has an invalid cast on VS 2015 which throws and completely disables the command...)

zooba:
Fixes #472 Execute in REPL should not create a window in QueryStatus
Removes unnecessary call to EnsureReplWindow


int19h:
:+1:

huguesv:
:+1:

PR #477, opened 2015-06-18
Fixes #300 Race condition refreshing environments

### LINKED ISSUE #300: Race condition refreshing environments
Saw this error on one of the test machines:

```
(QTAgent32.exe, PID 18220, Thread 88) AgentProcess:CurrentDomain_UnhandledException: IsTerminating : System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at System.Linq.Enumerable.<SelectManyIterator>d__1`2.MoveNext()
   at System.Linq.Enumerable.<DistinctIterator>d__1`1.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.<ConcatIterator>d__1`1.MoveNext()
   at System.Linq.Enumerable.<ConcatIterator>d__1`1.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.OrderedEnumerable`1.<GetEnumerator>d__1.MoveNext()
   at Microsoft.VisualStudioTools.ObservableCollectionExtensions.Merge[T](ObservableCollection`1 left, IEnumerable`1 right, IEqualityComparer`1 compareId, IComparer`1 compareSortKey) in C:\Source\Common\Product\SharedProject\ObservableCollectionExtensions.cs:line 41
   at Microsoft.PythonTools.EnvironmentsList.ToolWindow.UpdateEnvironments(IPythonInterpreterFactory select) in C:\Source\Python\Product\EnvironmentsList\ToolWindow.xaml.cs:line 297
   at Microsoft.PythonTools.EnvironmentsList.ToolWindow.FirstUpdateEnvironments() in C:\Source\Python\Product\EnvironmentsList\ToolWindow.xaml.cs:line 331
   at System.Windows.Threading.DispatcherOperation.InvokeDelegateCore()
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
```

It may not affect the actual product, since we're running in a test thread here, but we are on the "UI" thread.

Looks like the SelectMany in the trace is InterpreterOptionsService.cs:353, and one of the factory providers is detecting something new while we're running.

zooba:
Fixes #300 Race condition refreshing environments
Adds enumerable implementation for getting the list of interpreters from the service. This enumerable will suppress change events during enumeration and try harder to get a coherent list of factories from the providers.
Adds a race-condition test to flush out threading issues.
Hardens mock classes to avoid causing false positives in the race-condition test. The actual VS classes are either thread-safe or marshal back to the UI thread, but there's no value in forcing that here.

### DIFF
--- a/Python/Product/VSInterpreters/InterpreterOptionsService.cs
+++ b/Python/Product/VSInterpreters/InterpreterOptionsService.cs
@@ -13,6 +13,7 @@
  * ***************************************************************************/
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.ComponentModel.Composition.Hosting;
@@ -49,7 +50,7 @@ namespace Microsoft.PythonTools.Interpreter {
 
         // If this collection exists in the settings provider, no factories will
         // be loaded. This is meant for tests.
-        private const string SuppressFactoryProvidersCollection = @"PythonTools\NoInterpreterFactories";
+        internal const string SuppressFactoryProvidersCollection = @"PythonTools\NoInterpreterFactories";
 
         private const string DefaultInterpreterOptionsCollection = @"PythonTools\Options\Interpreters";
         private const string DefaultInterpreterSetting = "DefaultInterpreter";
@@ -178,9 +179,20 @@ namespace Microsoft.PythonTools.Interpreter {
                 DefaultInterpreter = Interpreters.LastOrDefault(fact => fact.CanBeAutoDefault());
             }
 
-            var evt = InterpretersChanged;
-            if (evt != null) {
-                evt(this, EventArgs.Empty);
+            OnInterpretersChanged();
+        }
+
+        private void OnInterpretersChanged() {
+            try {
+                BeginSuppressInterpretersChangedEvent();
+                for (bool repeat = true; repeat; repeat = _raiseInterpretersChanged, _raiseInterpretersChanged = false) {
+                    var evt = InterpretersChanged;
+                    if (evt != null) {
+                        evt(this, EventArgs.Empty);
+                    }
+                }
+            } finally {
+                EndSuppressInterpretersChangedEvent();
             }
         }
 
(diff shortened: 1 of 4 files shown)


zooba:
FTR, I'm not strongly in favour of taking this. It's a significant change for a race condition that I don't think we've seen in the product itself - only in tests.

Python/Product/VSInterpreters/InterpreterOptionsService.cs
```diff
@@ -589,5 +598,74 @@ IServiceProvider serviceProvider
...
+                }
+                _owner.EndSuppressInterpretersChangedEvent();
+            }
+
+            ~InterpretersEnumerator() {
+                Debug.Fail("Interpreter enumerator should always be disposed");
+                Dispose(false);
+            }
+
+            public IPythonInterpreterFactory Current { get { return _e.Current; } }
```
int19h: Formatting.

int19h:
:+1: on the change, but I agree it might be too big at this point.

Python/Product/VSInterpreters/InterpreterOptionsService.cs
zooba: What's wrong with the formatting?

int19h: I think we only put methods all on a single line when the body is `{}`. And even then there are blank lines to separate them to make it clear they are different.

zooba: We don't have any rule about that. I generally always make trivial (and in this case, automatically generated) property getters a single line.

crwilcox:
:+1:.  Agree this is too much to take.  Though it seems locking on settings could be a good idea.  Would there be any reason to take just that?  This really has only been a problem in tests though.

zooba:
More locking is just going to cause deadlocks, especially since the event handlers that need to take the lock are typically triggered from within code that should already hold the lock. This approach defers all of the events until the lock is released. Maybe there's a better way to approach this, but I'm fairly sure more coarsely-grained locks are not it.

Issue #410, opened 2015-06-13
Finish porting build system


zooba:
A few things left to do for the github move:
- remove TFS operations from BuildRelease.ps1
- ~~remove WiX binaries and replace with [GetWix.ps1](https://github.com/Microsoft/VisualStudio-TestHost/blob/master/Build/GetWix.ps1) (and call from build scripts)~~
- ~~remove other things in Tools~~
- update rolling builds to use a git-enabled system (VSO maybe?)

PR #480, opened 2015-06-18
#410 Finish porting build system

zooba:
#410 Finish porting build system
Removes Tools directory
 * we now download WiX on demand and put it into the output location
 * PowershellEnvironment.ps1 is moved to top level
 * BuildTasks is now built from Common/Tools
 * StyleCop was unused


zooba:
Note that this doesn't quite *finish*, but it's a good start.

int19h:
:+1:

PR #488, opened 2015-06-19
Fixes code analysis warnings.

zooba:
The two lock disposes are the only risky parts of this change at all, and I'm fairly sure they're fine.

I haven't fixed any of the complaints about all the classes that should have IDisposable, only the ones that were already disposable but weren't disposing everything.


int19h:
:+1: , but does it meet the bar?

zooba:
Depends how you interpreter the "FxCop clean" release security requirement... I don't think any of these are on the security-specific list, so probably not.

huguesv:
:+1:

crwilcox:
:+1:

PR #494, opened 2015-06-19
Fix #416: DebugCustomInterpreter test needs message updating

### LINKED ISSUE #416: DebugCustomInterpreter test needs message updating
It's looking for: `Interpreter specified in the project does not exist`

Actual message is: `The project cannot be launched because no Python interpreter is available at "<path>". Please check the Python Environments window and ensure the version of Python is installed and has all settings specified.`

int19h:


huguesv:
:+1:

PR #500, opened 2015-06-22
#497 pip is not installing properly

### LINKED ISSUE #497: pip is not installing properly
Our pip downloader is a little old and creaky and should be improved:

* SSL doesn't always verify properly on Python 2.7.9+
* Python 2.7.10+ and 3.4 could use `ensurepip` instead
* Python 3.0 and 3.1 are no longer compatible and require a different pip version (not sure which one yet)
* Proxy authentication is not supported

In most cases, we should simply download [git-pip.py](https://github.com/pypa/pip/blob/master/contrib/get-pip.py) and run that. This bundles a recent version of pip and uses it to install the latest pip, setuptools and wheel. For unsupported Python versions we should keep using the current fwlink to tar.gz to setup.py approach.

zooba:
Fixes #497 pip is not installing properly
Revises pip_downloader.py to use ensurepip if available, falling back on get-pip.py and then downloading directly.
Prevents execution on Python 3.0, which has no supported version of setuptools.
Adds separate links for Python 3.1 and a urlretrieve() implementation that correctly handles fwlinks.
Adds script restart if we are running on IronPython without -X:Frames

### DIFF
--- a/Python/Product/PythonTools/pip_downloader.py
+++ b/Python/Product/PythonTools/pip_downloader.py
@@ -52,10 +52,7 @@ def install_from_source(setuptools_source, pip_source):
         os.chdir(setuptools_temp_dir)
         print('Downloading setuptools from ' + setuptools_source)
         sys.stdout.flush()
-        if os.path.exists('setuptools.tar.gz'):
-            setuptools_package = 'setuptools.tar.gz'
-        else:
-            setuptools_package, _ = urlretrieve(setuptools_source, 'setuptools.tar.gz')
+        setuptools_package, _ = urlretrieve(setuptools_source, 'setuptools.tar.gz')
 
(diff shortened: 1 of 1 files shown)


crwilcox:
:+1: from me and @int19h

We didn't address this, but does this meet the bar for 2.2?  

I am alright with taking this.

zooba:
It's been driven by external changes, so I don't think we have much of a choice. Some versions of Python are broken without this fix (3.0, 3.1 and IronPython), IntelliSense is unavailable for setuptools without this fix (it currently installs as an egg) and recent Python 2.7, 3.4 and 3.5 can install more securely with this fix (because we install a bundled pip and then use it to upgrade - our current scheme performs no validation at all :( ).

Any other opinions?

PR #509, opened 2015-06-23
Fixes #508 Open Command Prompt here doesn't set PATH correctly

### LINKED ISSUE #508: Open Command Prompt here doesn't set PATH correctly
It should include the prefix path of the active environment and a Scripts subdirectory (if it exists).

zooba:
Fixes #508 Open Command Prompt here doesn't set PATH correctly
Sets the string comparer for environment dictionaries.

I want this for 2.2 because it fixes a regression and is a trivial/safe fix.


int19h:
:+1:

crwilcox:
:+1:

PR #510, opened 2015-06-24
Auto-list identifier fixes

### LINKED ISSUE #504: Should not auto-commit identifiers when comma-separating on LHS
As you type:

```
a, b, c, d = ...
```

The `b`, `c` and `d` will try to auto-commit. We should not do this. It's really annoying.

### LINKED ISSUE #506: Should not be auto-listing identifiers in the `for` part of a complete statement
If you start editing the variable part of a completed `for` statement then it will suggest completions. We should not be doing this.

zooba:
Fixes #504 Should not auto-commit identifiers when comma-separating on LHS
Fixes #506 Should not be auto-listing identifiers in the `for` part of a complete statement
Updates the expression scanner to only look up to the caret and not take into account the rest of the statement.
Prevents recursing into parts of statements we have already skipped. This was in some cases causing the expression on (for example) a LHS to affect completions on a RHS when there's no point even walking the LHS.

(Personally, I will probably disable this feature for my own use without these fixes. That doesn't necessarily mean it meets the bar - the fact that it's a new feature for this release is the only thing that gives it any chance at all. I believe the change is safe.)


int19h:
:+1:

PR #512, opened 2015-06-24
Fixes #482 Fix "VS 2015 required" message in installer

### LINKED ISSUE #482: Fix "VS 2015 required" message in installer
The message currently reads:

```
Visual Studio 2015 is required.  The free Visual Studio Express 2015 for Desktop and Visual Studio Express 2015 for Web editions can be downloaded from http://www.visualstudio.com/.
```

We should be recommending Community edition.

zooba:
Fixes #482 Fix "VS 2015 required" message in installer
Updates message to recommend Community edition.


int19h:
:+1:

PR #515, opened 2015-06-25
Fixes #514 Remove RC hacks

### LINKED ISSUE #514: Remove RC hacks
We had to add some hacks to the code to be able to build against both VS 2015 RC and the latest (internal) builds. We should remove these now.

These include:
* interactive window commands workaround
* suggested action interface implementations

(@int19h, @DinoV, @crwilcox  - any others you can think of?)

zooba:
Fixes #514 Remove RC hacks
Removes the workarounds added to deal with bugs that have since been fixed.


int19h:
:+1:

huguesv:
:+1:

PR #530, opened 2015-06-30
Adding wizard for Python UWP template to pick latest Windows 10 SDK automatically

munyirik:


Python/Product/Uwp/Templates/Projects/BackgroundService/PythonBackgroundService.pyproj
```diff
@@ -55,4 +55,9 @@
 
   <Import Project="$(PtvsTargetsFile)" />
 
+  <PropertyGroup>
+    <TargetPlatformVersion>$winsdkversion$</TargetPlatformVersion>
+    <TargetPlatformMinVersion>$winsdkminversion$</TargetPlatformMinVersion>
+  </PropertyGroup>
+
```
zooba: Do you really want these at the end and not before the targets file is imported? What if you later need to add conditions based on the target version?

munyirik: This is intentional. The version property is actually set to a default value ("7.0") by one of the VS .props or .targets files the python UWP .targets file is importing. I haven't dug into this yet but I didn't want to mess around with changing that for now.

zooba: Sounds like a recipe for breakage... I'd preserve the version at the top of your .targets and restore it at the end. That way you at least keep the opportunity to fix/change it in the future without having to modify user projects (we may have hit a problem like this before... :/ )

munyirik: Ok will make the change.

munyirik: Ok, found the issue. The cpython.targets file (from https://github.com/ms-iot/python) was importing $(VCTargetsPath)\Application Type\Windows Store\Default.props which it doesn't need to. The target version should never be "7.0". By removing that import I can add WindowsTargetPlatformVersion and WindowsTargetPlatformMinVersion at the top of the user projects (this value gets assigned to TargetPlatformVersion and TargetPlatformMinVersion). This will actually match the RTM UWP projects for C# and C++.

zooba:
Thanks. I'm okay with this change, but at this stage I need a review from someone else on my team. Should be able to get you a build tomorrow.

DinoV:
LGTM

int19h:
:+1:

PR #535, opened 2015-07-01
Fix redundant registry values for python exceptions under AD7Metrics\…

jsschultz:
Fix redundant registry values for python exceptions under AD7Metrics\\Exception registry node. 

The ProvideDebugExceptionAttribute Register method was unnecessarily creating python exception default stop settings under the python exception category node, which is used to improve load times of debug engines. 

The intent of having exception stop settings under this node (as opposed to the traditional tree structure used as well) is to specify a default stop setting for the category and then add only the stop settings that are an exception to the category default. This way, when the debug engine loads, it only has to read the minimal number of registry values and does not have to traverse the entire registry hive tree under the exception category node.

This change limits writing exception stop settings to 1) the category default and 2) any exception to the category default.

This does not affect default exception settings shown in the Exception Settings window. Only affects exception settings read when debug engine initially loads (when the Exception Settings window has not loaded yet).

Built PythonTools and verified resultant pkgdef file matches expected value. 
Verified correct behavior by manually testing debugging a python app which raises exception. Debugger breaks on exception as expected.

### DIFF
--- a/Common/Product/SharedProject/ProvideDebugExceptionAttribute.cs
+++ b/Common/Product/SharedProject/ProvideDebugExceptionAttribute.cs
@@ -87,8 +87,15 @@ namespace Microsoft.VisualStudioTools {
             key.SetValue("Code", _code);
             key.SetValue("State", (int)_state);
 
-            string name = _path.LastOrDefault() ?? "*";
-            engineKey.SetValue(name, (int)(_state & DkmValidFlags));
+            // Debug engine load time can be improved by writing the exception category default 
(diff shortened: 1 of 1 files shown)


zooba:
Looks good to me. @int19h...?

jsschultz:
Adding @gregg-miskelly

gregg-miskelly:
LGTM

int19h:
:+1:

zooba:
Thanks!

PR #536, opened 2015-07-01
Some minor fixes

zooba:
Fixes API comment.
Fixes vsixmanifest.
Removes unnecessary reference to IComponentModel.

The last one has been triggering the MEF dialog on open. It'll be triggered by something eventually, but better that it isn't us, especially since we're loading on startup.


crwilcox:
:+1:

int19h:
:+1:

Issue #539, opened 2015-07-02
Publish wfastcgi on PyPI


zooba:
There's very little value in having an MSI for this module. We should just make it a pip installable script with enable/disable functions for setting up IIS (these could even support IIS Express... what a concept!)

PR #540, opened 2015-07-02
#539 Publish wfastcgi on PyPI

zooba:
#539 Publish wfastcgi on PyPI
Moves wfastcgi into its own directory and adds a setup.py so we can eventually push it to PyPI.
Adds enable and disable commands for setting up IIS.
Removes the old installer.

There is still work to be done on the readme, but everything else is there AFAIK.


zooba:
(Contributions for the readme file are welcome.)

int19h:
:+1: 

Is this way of including README.rst as a long description the usual approach for packages? Should we do it for ptvsd as well?

zooba:
It's fairly common, though what I've done is convoluted because I wrote it for 3.4 initially and then added the fallback for earlier versions. For the Azure SDK we just have a one-liner doing the same thing: https://github.com/Azure/azure-sdk-for-python/blob/master/setup.py and I'm simplifying this one now.

It really depends how long the description is and whether you want it to show up on the PyPI page. It probably wouldn't hurt to do it for ptvsd.

Python/Product/WFastCgi/README.rst
```diff
@@ -0,0 +1,179 @@
...
+============
+
+Downloading Package
+-------------------
+
+To install via the Python Package Index (PyPI), type:
+
+.. code:: shell
+
+    pip.exe install wfastcgi
```
crwilcox: Because this is IIS it will be on windows, but I never type .exe after pip.

Python/Product/WFastCgi/README.rst
```diff
@@ -0,0 +1,179 @@
...
+    wfastcgi-enable "C:\Program Files (x86)\IIS Express\appcmd.exe" /apphostconfig:C:\Path\To\applicationhost.config
+
+You can disable wfastcgi in the same configuration file using
+``wfastcgi-disable`` with the same options:
+
+.. code:: shell
+
+    wfastcgi-disable "C:\Program Files (x86)\IIS Express\appcmd.exe" /apphostconfig:C:\Path\To\applicationhost.config
+
+Note that uninstalling ``wfastcgi`` does not unregister the CGI application.
```
crwilcox: I feel like we might want to bold this so people see it called out.

Python/Product/WFastCgi/README.rst
```diff
@@ -0,0 +1,179 @@
...
+variables are used by ``wfastcgi``.
+
+WSGI_HANDLER
+------------
+
+This is a Python name that evaluates to the WSGI application object. It is a
+series of dotted names that are optionally called with no parameters. When
+resolving the handler, the following steps are used:
+
+1. As many names as possibleare loaded using ``import``. The last name is never
```
crwilcox: possible are

crwilcox:
:+1:

huguesv:
What should we do with this page?
https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-python-django-web-app-windows-server/

Once you remove it from WebPI, those instructions will need an update.

If you want to integrate into IIS (not IIS Express), can you omit both the path to appcmd.exe and applicationhost.config file?

zooba:
Step 1 on that page should become `pip install wfastcgi`, `wfastcgi-enable`.
Step 2 should change from `C:\Python34\Scripts\wfastcgi.py` to `C:\Python34\Lib\site-packages\wfastcgi.py` (and mention to copy the value from the `wfastcgi-enable` output, which will always be correct).

Better yet, most of those steps could become "use a web.config that looks like this:"...

WebPI will need an update anyway or it'll still get the 2.1 MSI, but it can do pip/wfastcgi-enable just as easily.

For IIS, you can omit both paths. I'll add an example to make it clear.

zooba:
You can see this now at https://pypi.python.org/pypi/wfastcgi if you like.

leonbrave:
Hi zooba
i have a problem with the wfastcgi-enable;

all start with a Flask/jade project, i want to publish with iis and test in a local network. So the first logical step was install wfastcgi packet  (before i installed all IIS properties on windows 7).

The path where the wfastcgi-enable.exe is allocated is User\documents\VS2013\projects\nameproject\nameproject\env\Scripts and there's the wfastcgi.exe and the enable and disable files.

if i executed with adm privileges the enable shows a problem "ensure user has sufficient privileges and try again" and the route is very LONG.

another things, is normal if i try to publish from VS the project the error output is the "setAcl" error? This is because the wfastcgi isn't set yet?

"Error	1	Web deployment task failed. (A value for the 'setAclUser' setting must be specified when the 'setAcl' provider is used with a physical path.)		0	0	QPX_Test"

maybe i'm very confuse with the steps... of course when i test the project the browser display correctly the project.


on the runserver.py i have to put anything else for IIS configuration?
thanks any advice :)

zooba:
This would be better as a new issue, but to quickly answer:
* wfastcgi is _only_ useful with IIS, so you'll need to install that before enabling (it sounds like we don't have a good error message for this case)
* the "setAcl" error is not related to PTVS or wfastcgi, so we'll need more info to track this down. Can you create a new issue for that?
* runserver.py is an _alternative_ to using wfastcgi - they don't work together at all. It's useful for debugging and testing locally, but not after it's been deployed.

leonbrave:
you're very RIGHT!

i didn't notice i didn't install the wfastcgi.msi, now the IIS is displayed wfastcgi and all the files are located on Python34 folder, nice with that.

I keep tracking the SetAcl Error

thanks zooba

zooba:
The MSI is old - the version on PyPI has some fixes that aren't in the old MSIs, so I'd really suggest making the PyPI package work.

Create a new issue with the error details from `wfastcgi-enable` and we'll try and figure it out.

Issue #538, opened 2015-07-02
Backport critical fixes to 2.1.1


zooba:
Since we're not supporting PTVS 2.2 on VS 2010 and VS 2012, we should make one more update to 2.1 to backport fixes for critical issues.

This will be a very narrow, focused list, but feel free to suggest major issues and we'll consider them and add them to this list:

* various analysis crashes (backport some changes to Analysis.dll)
* IPython 3.0 support

zooba:
We don't want to backport all changes to Analysis.dll - Python 3.5 support won't be added, for example. But the change to AnalysisDictionary is valuable as it prevents a lot of race-conditions.

recife25:
#538 
I am working on Microsoft Visual Studio Ultimate 2012 ??? does the new python tool set works with such an environment????

zooba:
PTVS 2.2 will not run on VS 2012. PTVS 2.1 will continue to run on VS 2010, 2012 and 2013, and we are planning a minor update to 2.1.1 for those people who can't move to VS 2013 or VS 2015.

zooba:
#548 should be backported

PR #541, opened 2015-07-02
#538 Backport analysis fixes to 2.1.1.

zooba:
#538


Python/Product/Django/TemplateParsing/TemplateArtifactCollection.cs
```diff
@@ -29,11 +29,14 @@ public TemplateArtifactCollection()
             LeftSeparator = RightSeparator = "";
         }
 
-        protected override ISensitiveFragmentSeparatorsInfo SeparatorInfo {
-            get {
-                return this;
-            }
-        }
+        // HACK: Does not build properly on Dev12
```
crwilcox: is this interesting for this case?  Also, there is another bit of dead code (analysis dictionary) we might want to remove in both 2.2 and 2.1

crwilcox:
:+1:  Provided all the tests are run and show okay.  I scoured the code and nothing is too alarming.  It really shouldn't be since you just ported some changes to a branch.

Python/Product/Django/TemplateParsing/TemplateArtifactCollection.cs
zooba: Not really, except I was working on this in VS 2013 and so not being able to build was really annoying. This code is ifdef'd out for VS 2010 and VS 2012, so it's irrelevant.

AnalysisDictionary is not dead code - it's used for keeping track of variables. If anything, it's probably our biggest bottleneck (which is why changing it to a mostly thread-safe dictionary from a regular dictionary is such a good thing).

int19h:
:+1:

PR #554, opened 2015-07-07
Fixes #553 IPython interactive window has no colors in errors

### LINKED ISSUE #553: IPython interactive window has no colors in errors
Seems like using extension methods is flakey when the underlying interface changes. In this case, a real WriteError() method was added that broke our extension method.

We should avoid relying on extension methods for critical functionality like this.

zooba:
Fixes #553 IPython interactive window has no colors in errors
Uses AppendEscapedText on VS 2015 instead of relying on extension methods.


int19h:
:+1:

huguesv:
:+1:

PR #555, opened 2015-07-07
Issue 406

### LINKED ISSUE #406: Interactive window images do not zoom
Start an IPython interactive window and plot something. Then zoom in and out in the editor - the image stays the same size.

zooba:
Fixes #406 Interactive window images do not zoom
Replaces the standard intraline adornment with our own one.
Also handles generic XAML as forwards-compatibility preparation.


int19h:
:+1:

huguesv:
:+1:

huguesv:
There are some hard coded colors in the adornment code, we should check it all looks right in all 3 themes.

zooba:
They're hardcoded because the images from matplotlib have hardcoded defaults too (including a transparent background). Forcing a white background on those images is necessary for all themes, and the other colours appear on top of that background.

But I'd checked that it looked fine already.

PR #574, opened 2015-07-09
Fixes image layout in interactive window.

zooba:
Fixes image layout in interactive window.
By adding to the buffer, we ensure that FlushOutput does some work and correctly repositions the caret.


int19h:
:+1:

PR #586, opened 2015-07-10
Fixes #585 Interactive window colours are unreadable in VS 2015

### LINKED ISSUE #585: Interactive window colours are unreadable in VS 2015
VS 2015 apparently does not ship theme files for the interactive window colours, so some colours are unreadable.

We can't sensibly work around this, as it will cause conflicts when VS starts including themes. For PTVS 2.2 we should release a registry file that will add the theme content, and then when VS gets the proper fix we can clearly mark that the registry file is no longer required.

zooba:
Fixes #585 Interactive window colours are unreadable in VS 2015
Ports the REPL theme from the old project for VS 2015.


huguesv:
:+1:

crwilcox:
:+1:

PR #587, opened 2015-07-10
Fixes #584 User prompts do not override IPython prompts

### LINKED ISSUE #584: User prompts do not override IPython prompts
If you specify overridden prompts, they will only actually work if you accept the setting change _after_ starting the REPL window. When you reset the window, the prompts revert to the interpreter's choice.

zooba:
Fixes #584 User prompts do not override IPython prompts
Ensures prompt variables are initialized whenever a new command thread is created.


huguesv:
:+1:

crwilcox:
:+1:

PR #588, opened 2015-07-10
Fixes #577 IPython doc strings, question mark does not work

### LINKED ISSUE #577: IPython doc strings, question mark does not work
It seems, that IPython interactive window does not support printing doc strings via "?", for example in normal ipython following statement:
```
>>numpy.abs?
```
is equivalent to:
```
>>print numpy.abs.__doc__
```
and double question mark "??" shows the source code if it is available. Are there any reasons for absence of this feature?

PTVS 2.2 RC2 
VS 2013 Express for Desktop
IPython 3.2.0

zooba:
Fixes #577 IPython doc strings, question mark does not work
Handles the case where a response through the system channel contains data rather than simple text.

This only occurs on IPython 3.0 and later.

### DIFF
--- a/Python/Product/PythonTools/visualstudio_ipython_repl.py
+++ b/Python/Product/PythonTools/visualstudio_ipython_repl.py
@@ -97,6 +97,18 @@ class VsShellChannel(DefaultHandler, ShellChannel):
         payload = content['payload']
         
         for item in payload:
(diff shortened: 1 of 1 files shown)


huguesv:
:+1:

crwilcox:
:+1:

PR #599, opened 2015-07-15
can not set the color and font of python in vs using PTVS!

genliu777:
while in vc++, i can set the code color and font size, there is no way to set the python code color! i look for it in the " tools--enviroment--fonts and colors " and " tools--python tools" of vs2013, and find nothing to set! 
gg


int19h:
This is an issue, not a pull request - please file it as such.

PR #618, opened 2015-07-18
Fixes #616 Visual Studio runs out of memory

### LINKED ISSUE #616: Visual Studio runs out of memory
VS 2013
PTVS 2.1

Visual studio runs out of memory after I run the following in the IronPython interactive console:

from System import IO
IO.Path.GetInvalidPathChars()

zooba:
Fixes #616 Visual Studio runs out of memory
Moves "escape" update outside of condition to ensure we always advance even when colour codes are invalid.
Adds test for printing all characters.


int19h:
+1, and let's take this for 2.2.

PR #620, opened 2015-07-18
Backport #616

zooba:
Backports the fix for #616 (but not the test - too many changes there and it's easy enough to test manually).


int19h:
:+1:

PR #630, opened 2015-07-20
Fix #604: Breakpoint gets a new child bound breakpoint every time its When Hit action changes

### LINKED ISSUE #604: Breakpoint gets a new child bound breakpoint every time its When Hit action changes
User-reported as follows:

I set a regular breakpoint, on line 732 say:

 image

Now I right click on it in the gutter and add an action. It turned into four:
 
image

This makes things pretty confusing, if I click on the gutter to remove the breakpoint, it only removes one of them.

int19h:
Always return some non-null, non-empty function name for IDebugCodeContext2, to enable debugger to correctly detect and clean up duplicate breakpoints (which it creates itself when "When Hit" action changes).

### DIFF
--- a/Python/Product/Debugger/Debugger/DebugEngine/AD7MemoryAddress.cs
+++ b/Python/Product/Debugger/Debugger/DebugEngine/AD7MemoryAddress.cs
@@ -136,8 +136,8 @@ namespace Microsoft.PythonTools.Debugger.DebugEngine {
                 pinfo[0].dwFields |= enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS;
             }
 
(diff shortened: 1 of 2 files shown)


zooba:
+1, but can we test this somehow?

int19h:
It'll have to be a UI test since it requires interaction with the VS debugging package.

zooba:
I'd expected that, but it's worth it. This is an obscure enough fix (looks like a previously undetected NullReferenceException) that we'll want to know when it changes in VS.

int19h:
It's not actually a NullReferenceException. There's no crash there, it's just that the code path in debugger that compares breakpoints will always treat two breakpoints with a null function name as non-equal, and that can break things in subtle (or not so subtle, as is the case here) ways.

PR #640, opened 2015-07-23
Fix mixed-mode debugging on VS 2015

int19h:
Work around compiler being more aggressive doing COMDAT folding on breakpoint-signal functions in the debugger helper DLL, which breaks mixed-mode debugging on VS 2015.

### DIFF
--- a/Python/Product/DebuggerHelper/dllmain.cpp
+++ b/Python/Product/DebuggerHelper/dllmain.cpp
@@ -23,9 +23,7 @@ extern "C" {
     volatile char isInitialized;
 
     __declspec(dllexport) __declspec(noinline)
-    void OnInitialized() {
-        volatile char dummy = 0;
-    }
+    void OnInitialized() DUMMY_BREAKPOINT_FUNCTION
 }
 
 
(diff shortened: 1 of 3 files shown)


zooba:
Does `/OPT:NOICF` (Properties -> Linker -> Optimization -> Enable COMDAT Folding) not work? CPython has that option enabled so it can safely compare function pointers.

int19h:
It works, but it disables this optimization everywhere, which is overkill (probably doesn't matter much here though). The particular trick that I've used here to suppress it is almost verbatim what's used by CLR in similar circumstances, so I'm confident that it'll remain working.

zooba:
I doubt it matters at all here - the likelihood that we have other identical functions is very low, and the space savings are not going to be significant here either. I'd much prefer a documented linker option than a "trick".

zooba:
Also, can you create a 2.2.1 branch from the v2.2 tag and backport whatever we end up with here? I'm still deciding how best to ship this, but we'll almost certainly need a backports branch so may as well get it started.

int19h:
I'll do a different PR since it's an entirely different change.

Issue #676, opened 2015-08-06
Error getting the 'height' property of PlanarImage


gomiero:
Hi,

In file PTVS/Python/Product/PyKinect/PyKinect/pykinect/nui/structs.py, line 213 (class PlanarImage), the method get of the property 'height' is trying to fetch the  value from a property 'value':

return desc.height.value

Since the desc object is a _NuiSurfaceDesc(), it does not have this property and it raises an error.

The correct line (I made a try and it works), is:

return desc.height

, like the property 'width'.

Sorry if this has been reported or if I made mistakes with the language. I don't speak English well and 
is the first time I report a bug in a public site.

Thanks in advace for any advices or sugestions on how can I get better reports to help open source projects!

zooba:
You've reported it perfectly :)

The next step is to create a pull request. You can find that file in our source code at https://github.com/Microsoft/PTVS/blob/master/Python/Product/PyKinect/PyKinect/pykinect/nui/structs.py and then you'll see a little pencil near the top-right of the code. When you click that, it will let you edit the file directly and you can make the fix. (For bigger, more complicated fixes you should clone the entire repository, but for very simple changes using the web site is fine.)

When you save the changes, it will ask you to create a pull request. When you do that, we will be able to review and merge your fix. Then when we next update PyKinect, it will have your fix in it.

gomiero:
Hi, Steve,

I made the change on the code as you explained.

Very thanks for your help!!!   :-)

[]'s

Alexandre.

zooba:
Great. Now you just need to send a pull request so we can merge it into our code. To do that:

1. Go to the [Pull Requests](https://github.com/Microsoft/PTVS/pulls) page
2. Create a new pull request
3. Click on "compare across forks"
4. Select your own fork (`gomiero/PTVS`) from the third dropdown
5. Select your branch (`patch-1`) from the fourth dropdown
6. Click "Create pull request", add any explanation you like
7. Click "Create pull request" again and you're done!

gomiero:
Hi, Steve,

Done!

I hope to help more on this project and all other projects that I can.
Slowly, and with the help of people like you, I'm sure I can contribute with my experience.

Thank you for your patience and help!!

PR #689, opened 2015-08-06
Correction of bug #676 - incorrect '.value'

gomiero:
Correction of the bug #676: "Error getting the 'height' property of PlanarImage"

Removed the '.value' property from line 213.


zooba:
Thanks!

PR #690, opened 2015-08-06
Fix #598: Start with/without debugging fails if Solution Explorer has never been opened

### LINKED ISSUE #598: Start with/without debugging fails if Solution Explorer has never been opened
If you have Solution Explorer hidden, restart VS, open a standalone Python file (don't open any projects/solutions), and try and run with/without debugging you will get an E_FAIL error.

This is because we try and access the current solution before it exists. `StartScriptCommand.cs` should handle this case and continue as for when the solution is empty.

The workaround is to open Solution Explorer, which will initialize an empty solution.

int19h:
Gracefully handle exceptions from EnvDTE.ActiveSolutionProjects.


zooba:
+1

PR #691, opened 2015-08-06
Fixes #685 NullReferenceException in DjangoAnalyzer.GetArg

### LINKED ISSUE #685: NullReferenceException in DjangoAnalyzer.GetArg
From WER:

```
Level . . . Symbol - InstructionPointer - FunctionOffset - SourceLine - SourceFile
0 . . . microsoft_pythontools_django!Microsoft.PythonTools.Django.Project.DjangoAnalyzer.GetArg - 0x00000000 - 0x0000002C - - <<Failure related >>
1 . . . microsoft_pythontools_django!Microsoft.PythonTools.Django.Project.DjangoAnalyzer.RenderProcessor - 0x00000000 - 0x00000035 - - 
2 . . . microsoft_pythontools_analysis!Microsoft.PythonTools.Analysis.Values.SpecializedCallable.Call - 0x00000000 - 0x00000038 - - 
3 . . . microsoft_pythontools_analysis!Microsoft.PythonTools.Analysis.Analyzer.ExpressionEvaluator.EvaluateCall - 0x00000000 - 0x000000B4 - - 
```

From my reading of the GetArg function, the most likely cause is that the call into `django.shortcuts.render` has an invalid `CallExpression` which results in `namedArgs` (see `EvaluateCall`) being an array with null entries. This occurs when an arg with null `.NameExpression` follows an arg with non-null `.Name`.

We should probably assert in `GetNamedArguments` when `args[I].NameExpression` is null, and replace it with a safe non-null value to avoid crashing callers who assume that all elements are non-null.

zooba:
Fixes #685 NullReferenceException in DjangoAnalyzer.GetArg
Adds assertion and safe fallbacks if null arguments appear in incorrect places in calls.
Adds test for null named arguments.

### DIFF
--- a/Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
+++ b/Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
@@ -312,6 +312,8 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
                     }
 
                     res[i - (args.Count - res.Length)] = (NameExpression)args[i].NameExpression;
+                } else if (res != null) {
+                    res[i - (args.Count - res.Length)] = NameExpression.Empty;
                 }
             }
             return res ?? EmptyNames;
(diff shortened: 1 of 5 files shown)


int19h:
:+1:

PR #695, opened 2015-08-06
Fixes #682 Race condition in AddObjectTypeFixup

### LINKED ISSUE #682: Race condition in AddObjectTypeFixup
The below code probably has a race condition when called from multiple threads. We end up in the `else` block of `AddObjectTypeFixup` which causes us to crash.

To avoid the crash, we should catch `InvalidOperationException` in `AnalysisQueue.Worker` and report it, then not provide any IntelliSense.

```csharp
private void AddObjectTypeFixup(Action<IPythonType> assign) {
    var obj = ObjectType;
    if (obj != null) {
        assign(obj);
    } else if (_objectTypeFixups != null) {
        _objectTypeFixups.Add(assign);
    } else {
        throw new InvalidOperationException("Cannot find builtin type 'object'");
    }
}

private void RunObjectTypeFixups() {
    if (_objectType == null) {
        _objectType = BuiltinModule.GetAnyMember(GetBuiltinTypeName(BuiltinTypeId.Object)) as IPythonType;
        if (_objectType != null) {
            var fixups = _objectTypeFixups;
            _objectTypeFixups = null;
            if (fixups != null) {
                foreach (var assign in fixups) {
                    assign(_objectType);
                }
            }
        }
    }
}

private IPythonType ObjectType {
    get {
        if (_objectType == null) {
            RunObjectTypeFixups();
        }
        return _objectType;
    }
}
```

zooba:
Fixes #682 Race condition in AddObjectTypeFixup
Adds double-checked locking (or equivalent) to fixups lists and object instance.
Improves error handling when an 'object' type cannot be found

Please review the changes to SharedDatabaseState.cs carefully - they took a lot of thought and *should* cover all the potential races for the three variables now.

### DIFF
--- a/Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
+++ b/Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
@@ -184,6 +184,10 @@ namespace Microsoft.PythonTools.Interpreter {
                 lock (fixups) {
                     // Ensure it was not changed while waiting for the lock
                     if (ReferenceEquals(_objectTypeFixups, fixups)) {
+                        // Mutate the list under the lock - later we will read
+                        // from the list after clearing _objectTypeFixups under
+                        // the same lock. If there were a race, we would have
+                        // failed the comparison above.
                         fixups.Add(assign);
                         return;
                     }
@@ -191,6 +195,7 @@ namespace Microsoft.PythonTools.Interpreter {
 
                 // _objectTypeFixups changed, and it only changes to null, so we
                 // should call assign directly
+                Debug.Assert(_objectTypeFixups == null, "_objectTypeFixups was changed to something other than null");
                 obj = ObjectType;
                 if (obj != null) {
                     assign(obj);
@@ -221,11 +226,18 @@ namespace Microsoft.PythonTools.Interpreter {
                 lock (fixups) {
                     if (!ReferenceEquals(_objectTypeFixups, fixups)) {
                         // _objectTypeFixups changed while we waited to lock
+                        // This means someone else now owns the list and will
+                        // run the fixups, so we can just return the new object
+                        // type.
                         return objectType;
                     }
                     _objectTypeFixups = null;
                 }
 
+                // At this point, nobody else has a reference to the list. If
+                // they did (see AddObjectTypeFixup), they just entered the lock
+                // now and discovered that _objectTypeFixups has changed and
+                // will not use the reference they have.
                 foreach (var assign in fixups) {
                     assign(objectType);
                 }
--- a/Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
+++ b/Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
@@ -34,7 +34,7 @@ namespace Microsoft.PythonTools.Interpreter {
 #endif
         private readonly List<IPythonModule> _modulesLoading = new List<IPythonModule>();
         private List<Action> _fixups;
-        private List<Action<IPythonType>> _objectTypeFixups;
+        private List<Action<IPythonType>> _objectTypeFixups = new List<Action<IPythonType>>();
         private readonly Dictionary<string, List<Action<IMember>>> _moduleFixups = new Dictionary<string, List<Action<IMember>>>();
         private string _dbDir;
         private readonly Dictionary<IPythonType, CPythonConstant> _constants = new Dictionary<IPythonType, CPythonConstant>();
@@ -43,7 +43,7 @@ namespace Microsoft.PythonTools.Interpreter {
         private readonly Version _langVersion;
         private readonly List<WeakReference> _corruptListeners = new List<WeakReference>();
         private IBuiltinPythonModule _builtinModule;
-        private IPythonType _objectType;
+        private volatile IPythonType _objectType;
         internal readonly SharedDatabaseState _inner;
 
         internal const string BuiltinName2x = "__builtin__";
@@ -53,7 +53,6 @@ namespace Microsoft.PythonTools.Interpreter {
         public SharedDatabaseState(Version languageVersion) {
             _langVersion = languageVersion;
             _builtinName = (_langVersion.Major == 3) ? BuiltinName3x : BuiltinName2x;
-            _objectTypeFixups = new List<Action<IPythonType>>();
         }
 
         public SharedDatabaseState(Version languageVersion, string databaseDirectory)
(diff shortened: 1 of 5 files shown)


Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
```diff
@@ -177,34 +176,71 @@ internal SharedDatabaseState(Version languageVersion, string databaseDirectory,
...
+            if (fixups != null) {
+                lock (fixups) {
+                    // Ensure it was not changed while waiting for the lock
+                    if (ReferenceEquals(_objectTypeFixups, fixups)) {
+                        fixups.Add(assign);
+                        return;
+                    }
+                }
+
+                // _objectTypeFixups changed, and it only changes to null, so we
```
int19h: Can you add an assert to that effect?

Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
```diff
@@ -177,34 +176,71 @@ internal SharedDatabaseState(Version languageVersion, string databaseDirectory,
...
+                // objectType with whatever got there first
+                objectType = Interlocked.CompareExchange(ref _objectType, newObjectType, null) ?? newObjectType;
+            }
+
+            var fixups = _objectTypeFixups;
+            if (fixups != null) {
+                lock (fixups) {
+                    if (!ReferenceEquals(_objectTypeFixups, fixups)) {
+                        // _objectTypeFixups changed while we waited to lock
+                        return objectType;
```
int19h: Shouldn't this be a retry rather than instant return?

Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
```diff
@@ -177,34 +176,71 @@ internal SharedDatabaseState(Version languageVersion, string databaseDirectory,
...
                     }
+                    _objectTypeFixups = null;
+                }
+
+                foreach (var assign in fixups) {
```
int19h: Given that the other method Adds to the list under a lock (and could potentially stash the pointer away just before you null it out), this looks like it must be under lock also.

Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
```diff
@@ -34,7 +34,7 @@ class SharedDatabaseState : ITypeDatabaseReader {
 #endif
         private readonly List<IPythonModule> _modulesLoading = new List<IPythonModule>();
         private List<Action> _fixups;
-        private List<Action<IPythonType>> _objectTypeFixups;
+        private List<Action<IPythonType>> _objectTypeFixups = new List<Action<IPythonType>>();
```
int19h: It's not entirely clear to me whether the field (as opposed to collection) can be changed from concurrent threads. Code that's writing to this field does so under a lock, but code that's reading it (null checks etc) normally doesn't lock, and so could result in stale reads without a memory barrier. It feels like the latter should be using Volatile.Read, or else the writes don't have to be under lock.

int19h:
Is there a significant perf advantage here from trying to hand-code lock-free concurrency instead of using, say, `ReaderWriterLockSlim`? It feels like this is sufficiently complicated that future changes would have to be done _very_ carefully, and could break things very easily with hard-to-diagnose errors... I'd rather keep it simple if at all possible, so that whoever touches this code doesn't have to reconstruct the entire graph of what writes where when every time.

Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
zooba: It means someone else took the list and cleared the field, so we don't have to worry about running the fixups.

Python/Product/Analysis/Interpreter/SharedDatabaseState.cs
zooba: We've cleared the field inside the lock, and the adds check that the field has not been cleared before adding. Since the check/add is entirely within the lock, we know that nobody else has a reference to the list after we lock/clear. (At least, that's how I think it works. Very hard to reproduce this race...)

zooba:
I'm not sure that using an actual lock will really be any simpler, since we're locking the field rather than the value. Last time I tried to be "good" and use the prewritten libraries it turned out to be much simpler to use locks and Thread.

The reads (should) all follow the same pattern. Read to local -> null check -> lock -> read to local again (abort if changed) -> modify local -> unlock. This ensures that if it's already null (as it is for the majority of the time) then we don't lock.

The sole write to the field (to change it to null) is: read to local -> null check -> lock -> read to local again (abort if changed) -> clear field -> unlock -> read from local.

Since we're locking, memory barriers should be in place already, and the double-checking should ensure that after the lock and the second check nobody else has a reference to the same list or can get a reference to the same list before we unlock.

int19h:
It's the first read that's outside the lock that I'm worried about. Not for the null case, but for the case where it's non-null - since we're not locking before it, it looks like it could get a non-null ref to a not fully initialized object if JIT reorders memory writes accordingly. The classic double-checked locking pattern as used for singletons requires that the first read have a proper barrier for this reason (explicit, or `volatile`, `Volatile.Read` etc).

zooba:
Surely the JIT isn't going to reorder memory writes from before the constructor until after an async method is called from outside the object - is that really something that can happen? I'd have expected an implicit memory barrier between fields and the constructor code and/or after construction anyway.

int19h:
There are no implicit memory barriers before after constructor, so the potential reordered & broken sequence goes like this:

1. Allocate a chunk of memory for the object.
3. Set field that's going to reference the object once it's constructed.
2. Set fields in constructor.

The JIT won't do it (though it doesn't actually guarantee that it won't anywhere in the CLI spec, and I've no idea if it's still true with RyuJIT), but the CPU might, due to register splitting and/or caching. From what I've heard, the underlying x86/x64 memory model is strong enough that this cannot happen there, but elsewhere all bets are off.

http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

(C# volatile is basically equivalent to the new Java 5 volatile)

zooba:
Intuitively (to me at least) that seems impossible, or you'd never be able to use an instance variable from within the constructor. It might be a concern if you publish a reference to another thread as part of its own constructor (and certainly it's a problem if you do that and the class is unsealed), but what you are describing would mean that the following code in isolation is subject to a race condition:

```csharp
class F {
    List<int> MyList = new List<int>();

    public F() {
        MyList.Add(1);
    }
}
```

If instantiating `F` ever failed, it would be reported as a bug in the language and by basically any standard it would be treated as a bug.

zooba:
Maybe the difference here is that this isn't a lazily initialised singleton? I'd expect the second null-check inside the lock to be the bigger issue here.

int19h:
I've re-read the code again very carefully and you're right that this doesn't really apply here, mainly because it only changes from something to null, never the other way around (and otherwise the memory model guarantees that construction will complete and be observable before the method that reads the field gets called, as you've noted). Still, this - or maybe even the entire class - needs a big "here be dragons, if you ever touch one of these fields, carefully review _all_ their existing uses to understand threading" comment somewhere visible.

zooba:
@int19h Is this okay to merge?

int19h:
Yep!

PR #701, opened 2015-08-07
Fix #680: ExecuteInReplCommand does not validate filename

### LINKED ISSUE #680: ExecuteInReplCommand does not validate filename
This command may get a null or empty (probably empty) filename from `TryGetStartupFileAndDirectory`, which causes a crash when passing the value to the repl evaluator on a worker thread.

int19h:
Disable command when filename is null or empty.


gilbertw:
:+1:

PR #703, opened 2015-08-07
Fixes #688 by Handling null _vars

### LINKED ISSUE #688: NullReferenceException in MemberResult.Documentation
From WER:

```
Level . . . Symbol - InstructionPointer - FunctionOffset - SourceLine - SourceFile
0 . . . microsoft_pythontools_analysis!Microsoft.PythonTools.Analysis.MemberResult.get_Documentation - 0x00000000 - 0x00000110 - - <<Failure related >>
1 . . . microsoft_pythontools!Microsoft.PythonTools.Intellisense.CompletionAnalysis+__c__DisplayClass13_0._PythonCompletion_b__0 - 0x00000000 - 0x00000009 - - 
2 . . . microsoft_pythontools!Microsoft.PythonTools.Intellisense.DynamicallyVisibleCompletion.get_Description - 0x00000000 - 0x00000028 - - 
3 . . . microsoft_visualstudio_platform_vseditor_ni!Microsoft.VisualStudio.Language.Intellisense.Implementation.DefaultCompletionSetPresenter.ShowToolTip - 0x533F0000 - 0x000000B6 - - 
4 . . . microsoft_visualstudio_platform_vseditor_ni!Microsoft.VisualStudio.Language.Intellisense.Implementation.DefaultCompletionSetPresenter.OnShowTooltipTimer_Tick - 0x533F0000 - 0x00000028 - - 
5 . . . windowsbase_ni!System.Windows.Threading.DispatcherTimer.FireTick - 0x64D00000 - 0x00000034 - - 
```

It appears that either `_vars` or `_vars()` is somehow null, neither of which ought to be possible, but apparently it is.

crwilcox:
_vars was being set to null in at least on spot (ModuleAnalysis 455).
Now it is set as an empty list in cases where it could come as null.

### DIFF
--- a/Python/Product/Analysis/ModuleAnalysis.cs
+++ b/Python/Product/Analysis/ModuleAnalysis.cs
@@ -451,6 +451,7 @@ namespace Microsoft.PythonTools.Analysis {
                     .Select(s => new MemberResult(s.Name, s.GetMergedAnalysisValues()))
                     .ToList();
             } catch (Exception) {
+                // TODO: log exception
                 Debug.Fail("Failed to find scope. Bad state in analysis");
                 return new[] { new MemberResult("Unknown", new AnalysisValue[] { }) };
             }
--- a/Python/Product/Analysis/ModuleAnalysis.cs
+++ b/Python/Product/Analysis/ModuleAnalysis.cs
@@ -451,8 +451,8 @@ namespace Microsoft.PythonTools.Analysis {
                     .Select(s => new MemberResult(s.Name, s.GetMergedAnalysisValues()))
                     .ToList();
             } catch (Exception) {
-                // TODO: log exception
-                return new[] { new MemberResult("Unknown", null) };
+                Debug.Fail("Failed to find scope. Bad state in analysis");
+                return new[] { new MemberResult("Unknown", new AnalysisValue[] { }) };
             }
         }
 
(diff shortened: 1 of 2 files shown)


crwilcox:
I am working on adding a test for this yet, but this is my proposed patch.

int19h:
Are we sure that this is the only place where it's not handling null? Should the fix rather be removing that one single case where null is used, and replacing it with an empty collection? It looks like the code tries really hard to avoid nulls elsewhere.

crwilcox:
I considered doing that also.  These are all of the constructors.  Granted, you are correct that we could still do it incorrectly.  Any thoughts on doing both?

int19h:
Never mind, having had a look at the actual code diff I'm actually fine with either.

int19h:
:+1:

Python/Product/Analysis/ModuleAnalysis.cs
```diff
@@ -451,8 +451,8 @@ private class ErrorWalker : PythonWalker {
                     .Select(s => new MemberResult(s.Name, s.GetMergedAnalysisValues()))
                     .ToList();
             } catch (Exception) {
-                // TODO: log exception
```
zooba: This TODO should stay here, though we probably need to let the exception propagate to be able to log it from outside Analysis.dll.

zooba:
Go ahead and merge after adding the comment back.

PR #708, opened 2015-08-10
Fix #705: Debugging unit tests in mixed mode skips breakpoints occasionally

### LINKED ISSUE #705: Debugging unit tests in mixed mode skips breakpoints occasionally
The problem is that the piece of code in the test launcher that waits for debugger to attach before proceeding only uses the Win32 flag indicating attach, but doesn't wait until the injected helper thread actually registers the trace function. So it's possible to have a race whereby the test will start running, and even complete, before breakpoint tracing is fully active, on a sufficiently fast machine (or a sufficiently slowed down VS).

int19h:
Add a flag that test launcher now uses to detect when the debugger is actually tracing execution (and not just attached to the process).

### DIFF
--- a/Python/Product/TestAdapter/visualstudio_py_testlauncher.py
+++ b/Python/Product/TestAdapter/visualstudio_py_testlauncher.py
@@ -42,9 +42,18 @@ def main():
         # For mixed-mode attach, there's no ptvsd and hence no wait_for_attach(), 
         # so we have to use Win32 API in a loop to do the same thing.
         from time import sleep
-        from ctypes import windll
+        from ctypes import windll, c_char
         while True:
             if windll.kernel32.IsDebuggerPresent() != 0:
+                break
+            sleep(0.1)
+        try:
+            debugger_helper = windll['Microsoft.PythonTools.Debugger.Helper.x86.dll']
+        except WinError:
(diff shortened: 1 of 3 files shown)


Python/Product/TestAdapter/visualstudio_py_testlauncher.py
```diff
@@ -42,11 +42,20 @@ def main():
...
             sleep(0.1)
+        try:
+            debugger_helper = windll['Microsoft.PythonTools.Debugger.Helper.x86.dll']
+        except WinError:
+            debugger_helper = windll['Microsoft.PythonTools.Debugger.Helper.x64.dll']
+        isTracing = c_char.in_dll(debugger_helper, "isTracing")
+        while True:
+            if isTracing.value != 0:
+                break
+            sleep(0.1)
```
zooba: This is a good example of something that should be in ptvsd after the refactor.

zooba:
+1

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
int19h: Absolutely. This whole thing exists solely because `ptvsd.is_attached` is not universally applicable, and this code (or something similar) will just move there in the end.

PR #717, opened 2015-08-11
Fixes #712: Don't add Unicode BOM by default

### LINKED ISSUE #712: Don't add Unicode BOM by default
PTVS 2.2
Visual Studio 2013

File -> New -> File
Choose Python -> Python Class

type a single word consisting of only ascii characters and hit Save.  The UTF-8 BOM is added.

This does not happen for any other types of files.  Notice that if, instead of hitting Save in the final step, you go to File -> Advanced Save Options, you can see that the default encoding is Unicode UTF-8 (with signature).  But for other types of files including .cpp files, it defaults to Western European.  

This also didn't happen prior to PTVS 2.2 either, so it appears to be a regression.

This is a really big pain point as we have to manually adjust this every time we save a file to get the BOM out.

zooba:
Fixes #712: Don't add Unicode BOM by default
Changes default encoding to be UTF-8 without a signature.

### DIFF
--- a/Python/Product/Analysis/Parsing/Parser.cs
+++ b/Python/Product/Analysis/Parsing/Parser.cs
@@ -4831,7 +4831,7 @@ namespace Microsoft.PythonTools.Parsing {
         /// New in 1.1.
         /// </summary>
         public static Encoding GetEncodingFromStream(Stream stream) {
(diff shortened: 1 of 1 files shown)


int19h:
:+1:

rhuijben:
:+1:

PR #728, opened 2015-08-12
Fixes #669 Log all unhandled exceptions to Event Log

### LINKED ISSUE #669: Log all unhandled exceptions to Event Log
Currently we pop a message box telling people to reproduce the error after starting VS with "/log", but this may not be possible.

We should create an event log on install and log messages to that.

zooba:
Fixes #669 Log all unhandled exceptions to Event Log
Creates an event source and writes errors to it.
Also handles errors better in code paths that I happened to be trying to test.


int19h:
:+1:

PR #744, opened 2015-08-13
Tentatively fixes #739 NullReferenceException

### LINKED ISSUE #739: NullReferenceException in ReverseExpressionParser.GetExpressionRange
When I hover my cursor over a particular function call, it gives me an error message "An exception has occurred. This may be caused by an extension." The problem is, I have no extensions currently installed, and PTVS in both VS 2013 and 2015 give me this error. Here is the attached screenshot:
image

I get the error message while hovering over the line that reads : "reduceFn = ReduceFn(dbinfo=dbinfo, **dbopts)"

zooba:
Tentatively fixes #739 NullReferenceException in ReverseExpressionParser.GetExpressionRange
Ensures that Classifier is always non-null.

I can't see anything else in the function that could potentially be null, so guessing this is it.


gilbertw:
:+1:

int19h:
:+1:

Issue #704, opened 2015-08-07
Generate Requirements.txt on Invalid Environment


crwilcox:
I created an environment, removed the underlying interpreter, then tried to generate requirements.

We may want to disable the button or at least not crash to the activity log.  The activity log didn't seem to have any info, but an assertion for the missing interpreter.

---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
Unhandled exception in ExecGenerateRequirementsTxt (E:\PTVS\Python\Product\PythonTools\PythonTools\Project\PythonProjectNode.cs:1891)

System.AggregateException: One or more errors occurred. ---> System.IO.FileNotFoundException: C:\Python27-64\python.exe

   at Microsoft.PythonTools.PythonInterpreterFactoryRunnableExtensions.ThrowIfNotRunnable(IPythonInterpreterFactory factory, String parameterName) in E:\PTVS\Python\Product\PythonTools\PythonTools\PythonInterpreterFactoryExtensions.cs:line 67

   at Microsoft.PythonTools.Project.Pip.Run(IPythonInterpreterFactory factory, Redirector output, Boolean elevate, String[] cmd) in E:\PTVS\Python\Product\PythonTools\PythonTools\Project\Pip.cs:line 45

   at Microsoft.PythonTools.Project.Pip.<Freeze>d__5.MoveNext() in E:\PTVS\Python\Product\PythonTools\PythonTools\Project\Pip.cs:line 101

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------

PR #766, opened 2015-08-19
#704 Generate Requirements.txt on Invalid Environment

zooba:
#704 Generate Requirements.txt on Invalid Environment
Handles FileNotFoundException when listing packages.
Also disables generate/install requirements.txt commands when not able to install packages.


int19h:
:+1:

PR #767, opened 2015-08-19
Fixes #664 Excute File in Python Interactive -- wrong location and always appears

### LINKED ISSUE #664: Excute File in Python Interactive -- wrong location and always appears
1.       P1: Can we move the location below the "core debugging" items group (e.g. give it its own menu group below the “Profiler section” (or just add it to the Profiler section below the Profiler item)
2.       P2: Only have it appear if the Python editor is open.  A lot of developers who may not be Python users will just choose “install everything” so this will reduce visual clutter for them.

zooba:
Fixes #664 Excute File in Python Interactive -- wrong location and always appears
Moves command to near the bottom of the "execute" group.
Fixes logic to determine when command should be visible.

(Ping @Andrew-MSFT)

### DIFF
--- a/Python/Product/PythonTools/PythonTools.vsct
+++ b/Python/Product/PythonTools/PythonTools.vsct
@@ -904,7 +904,7 @@
       <Parent guid="guidPythonToolsCmdSet" id="CodeFileGroup2"/>
     </CommandPlacement>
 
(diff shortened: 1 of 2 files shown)


int19h:
:+1:

Issue #765, opened 2015-08-19
Debug REPL options pretty broken in VS 2015


zooba:
Changing the prompts simply does not work as we need to support them on the evaluator, but don't. It's also incredibly easy to get into situations where the window is broken - try clearing the REPL while not debugging (and possibly while debugging too). I also spotted a few race conditions in the code just by scanning it without even looking for them.

I fixed a crash in #678 by skipping certain options, but as part of any revision to the REPL windows, we need to spend some time really cleaning this up.

zooba:
No longer relevant as we don't have separate debug REPL options.

Issue #678, opened 2015-08-06
NullReferenceException in SaveSettingsToStorage


zooba:
From WER:

```
Level . . . Symbol - InstructionPointer - FunctionOffset - SourceLine - SourceFile
0 . . . microsoft_pythontools!Microsoft.PythonTools.Options.PythonDebugInteractiveOptionsPage.SaveSettingsToStorage - 0x00000000 - 0x00000169 - - <<Failure related >>
1 . . . microsoft_visualstudio_shell_14_0_ni!Microsoft.VisualStudio.Shell.DialogPage.OnApply - 0x00000000 - 0x0000000B - - 
2 . . . microsoft_visualstudio_shell_14_0_ni!Microsoft.VisualStudio.Shell.DialogPage+DialogSubclass.WndProc - 0x00000000 - 0x000000A7 - - 
```

PR #768, opened 2015-08-19
#678 NullReferenceException in SaveSettingsToStorage

zooba:
#678 NullReferenceException in SaveSettingsToStorage
Skips trying to update debug interactive prompts, since they don't work anyway.
Fixes extension methods to test the cast before using the object.

Filed #765 for a full fix in 3.0


int19h:
:+1:

PR #780, opened 2015-08-21
Ensures dependencies are always built into the correct output directory.

zooba:
Ensures dependencies are always built into the correct output directory.


int19h:
:+1:

PR #786, opened 2015-08-24
Fixes #784 Cannot have invalid module as startup file

### LINKED ISSUE #784: Cannot have invalid module as startup file
Because we convert `StartupPath` into `StartupModule` in `Microsoft.PythonTools.targets`, the startup file has to be a valid module name.

We should continue executing (with a blank `StartupModule`) if the name is not a valid module.

zooba:
Fixes #784 Cannot have invalid module as startup file
Allows creation of $(StartupModule) to fail silently and set an empty string.


int19h:
:+1:

PR #787, opened 2015-08-24
Fixes #48 Move/rename the corresponding *.pyc file at the same time w…

### LINKED ISSUE #48: Move/rename the corresponding *.pyc file at the same time when moving/renaming the *.py file
Thank you for PTVS.

I would like to request that PTVS move/rename the corresponding *.pyc file at the same time when moving/renaming the *.py file.

For example, given the following project structure:
```
package1
    __init__.py
    A.py
package2
    __init__.py
testPackage
    __init__.py
    testA.py
```

testA.py:
```python
import unittest
from package1.A import A

class TestRoutes(unittest.TestCase):

    def testA(self):
        A = A()
        print A.Name

if __name__ == '__main__':
    unittest.main()
```
And then if I move A.py to package2, but don't update the from/import statement, and run the test, I expect the test run to fail with an error like "ImportError: No module named A".

However, it succeeds without error (unexpectedly)!

Even cleaning the solution does not resolve the issue because the *.pyc doesn't get deleted during a clean.

Thank you.

zooba:
Fixes #48 Move/rename the corresponding *.pyc file at the same time when moving/renaming the *.py file
Attempts to remove or rename .pyc/.pyo files when moving the source.
We don't bother with __pycache__ folders because those aren't used to try and import files.
If a .pyc or .pyo file is in the project, we don't touch it, and if there is a conflicting file we also don't change anything.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonFileNode.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonFileNode.cs
@@ -22,6 +22,7 @@ using Microsoft.PythonTools.Intellisense;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudioTools;
 using Microsoft.VisualStudioTools.Project;
+using System.Security;
 #if DEV14_OR_LATER
 using Microsoft.VisualStudio.Imaging;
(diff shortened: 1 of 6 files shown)


int19h:
:+1:

PR #788, opened 2015-08-25
V2.2.1

zooba:
Graft fixes

### DIFF
--- a/Python/Product/EnvironmentsList/PipExtension.xaml.cs
+++ b/Python/Product/EnvironmentsList/PipExtension.xaml.cs
@@ -59,7 +59,8 @@ namespace Microsoft.PythonTools.EnvironmentsList {
         }
 
         private void UninstallPackage_CanExecute(object sender, CanExecuteRoutedEventArgs e) {
-            e.CanExecute = e.Parameter is PipPackageView;
+            e.CanExecute = _provider.CanExecute && e.Parameter is PipPackageView;
+            e.Handled = true;
         }
 
         private async void UninstallPackage_Executed(object sender, ExecutedRoutedEventArgs e) {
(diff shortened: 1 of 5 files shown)


Python/Product/EnvironmentsList/PipExtension.xaml.cs
```diff
@@ -76,21 +77,20 @@ internal sealed partial class PipExtension : UserControl {
         }
 
         private void UpgradePackage_CanExecute(object sender, CanExecuteRoutedEventArgs e) {
+            e.Handled = true;
```
gilbertw: Nothing wrong with this other than its just cleanup.  However we should have policy to not accept cleanup in the 2.2.1 release

Python/Product/EnvironmentsList/PipExtensionProvider.cs
```diff
@@ -143,6 +143,11 @@ public sealed class PipExtensionProvider : IEnvironmentViewExtension {
         internal async Task<IList<PipPackageView>> GetInstalledPackagesAsync() {
             string[] args;
 
+            if (!CanExecute) {
+                // Invalid configuration, so assume no packages
```
gilbertw: Can the user do something to fix the invalid configuration?  If so we should log to let them know what is wrong.  Assuming of course we know what is wrong.

zooba: They didn't enter the path to the environment. Everything is disabled and there are big blank spots next to some icons.

I don't disagree we should help them do the right thing, but this isn't the place to add that. Probably the only option that will work reliably here is to block them at the Create step until they've gotten everything right.

Python/Product/EnvironmentsList/PipExtension.xaml.cs
zooba: I'm inclined to agree, we (mostly me) have looked pretty free with what goes into 2.2.1 so far, though I'm not aware of anything significant other than crash/unhandled error fixes.

We'll get to have an RC for this one, as well as internal testing, so I'm not too concerned yet.

Python/Product/EnvironmentsList/PipExtensionProvider.cs
gilbertw: Yea doing it at Create seems best.  I guess we can't really tell why we can't execute here.  Although maybe CanRun could be the one to output the message.  Anyway I've made an issue to track doing something better to prevent bad environments (#790)

gilbertw:
:+1:

PR #824, opened 2015-09-09
fix build

enricosada:
Trying to build i got this error:

```
System.Management.Automation.PowerShell' does not contain a definition for 'HadErrors' and no extension method 'HadErrors' accepting a first argument of type 'System.Management.Automation.PowerShell' could be found
```

`PowerShell.HadErrors` property was added in powershell v3

Windows 7, no powershell 3 installed -> v 1.0.0.0 used.
After powershell 3 install the build work, because v 3.0.0.0 is used

I dont know, maybe is enough to add powershell v3 as prerequite in the documentation.
At least with this pr i see i need v3.0.0.0

### DIFF
--- a/Common/Tests/Utilities/TestUtilities.csproj
+++ b/Common/Tests/Utilities/TestUtilities.csproj
@@ -42,7 +42,7 @@
     <Reference Include="System" />
     <Reference Include="System.ComponentModel.Composition" />
     <Reference Include="System.Core" />
-    <Reference Include="System.Management.Automation, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+    <Reference Include="System.Management.Automation, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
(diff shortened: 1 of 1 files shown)


zooba:
Thanks! This looks like an appropriate fix to me.

OmerRaviv:
@zooba I just received this error message and did not get any indication that I need to install PowerShell 3 in order to build until I found this issue. Could you please update the Prerequisite section of the Build Instructions to mention that if using Windows 7, PowerShell 3 should be installed? (PowerShell 3 is included in Windows 8)

OmerRaviv:
@zooba Also, it might be worth mentioning on the Prerequisites page that when installing VS the C++ support must be checked (otherwise build fails because cl.exe does not exist)

zooba:
That page should be openly editable if you're logged in. Feel free to add the notes directly - even if they're rough. Or let me know if it isn't editable. I can't actually check.

OmerRaviv:
@zooba Apologies, hadn't noticed that. I just made [those changes](https://github.com/Microsoft/PTVS/wiki/Build-Instructions-for-PTVS/b01d1ec8d124ce3aad3e41b2a96f2a10c8fd3644)

PR #861, opened 2015-10-02
Remove express workaround 2.2.1

zooba:


crwilcox:
:+1:

This looks good but seems to have things from the fix for venvs and such.  as long as all of these go in together that shouldn't be an issue.

int19h:
:+1:

PR #872, opened 2015-10-06
Fixes #810 Referencing several Python extension modules does not work

### LINKED ISSUE #810: Referencing several Python extension Modules (*.pyd) does not work
We have a python Project and using PTVS 2.2 in Visual Studio Pro 2013 (Update 2).

While trying to add several references, Intellisence is working only for one. It seems after adding a new reference the *.idb files in AppData\Local\Python Tools\ReferencesDB\12.0 from the existing ones will be deleted.

I checkt the PTVS sources and found a bug in PythonTypeDataBase.cs -> FindDbFile. I could fix it by moving the part with check for correct file last modified date after the check for the correct module Name (see resulting code in ScreenShot). I tested it with a new build and it works fine in our Environment.

Did you already heared about this bug? Is there a possibility to get it fixed in your next released Version?

Best regards
Karo

changedcode

zooba:
Fixes #810 Referencing several Python extension modules does not work
Compares last modified time against the file referenced by the row, rather than the newly added extension.


int19h:
:+1:

PR #883, opened 2015-10-08
Backed out changeset: e45723cc7580

zooba:


zooba:
Pushed into the same branch, so now this also works around the issue we were hitting today.

Python/Product/PythonTools/PythonTools/Editor/AutoIndent.cs
```diff
@@ -84,7 +84,10 @@ private struct LineInfo {
                 }
 
                 string sline = tokens[tokenIndex].Span.GetText();
-                var lastChar = sline.Length == 0 ? '\0' : sline[sline.Length - 1];
+                char lastChar = '\0';
```
crwilcox: What is the point to this change?

crwilcox:
:+1:

Python/Product/PythonTools/PythonTools/Editor/AutoIndent.cs
zooba: It avoids a Roslyn bug that generates invalid code and causes InvalidProgramExceptions.

PR #895, opened 2015-10-14
Correctly maps indentation request for VS 2015 Update 1.

zooba:


Python/Product/PythonTools/PythonTools/Repl/BasePythonReplEvaluator.cs
```diff
@@ -1783,6 +1785,19 @@ class MemberResults {
...
 
 #if DEV14_OR_LATER
+        internal static Version VSInteractiveVersion {
+            get {
+                if (_vsInteractiveVersion == null) {
+                    _vsInteractiveVersion = typeof(IInteractiveWindow).Assembly
+                        .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute), false)
+                        .OfType<AssemblyInformationalVersionAttribute>()
+                        .Select(a => { Version v; return Version.TryParse(a.InformationalVersion, out v) ? v : null; })
+                        .FirstOrDefault(v => v != null) ?? new Version();
```
crwilcox: Getting the first non null entry should be fine. Because we do the version comparison though, shouldn't we get the greatest version?  Though, I am not sure you can end up with more than one `AssemblyInformationalVersionAttribute` anyway.  If that is the case, really shouldn't we validate there is only one non-null entry?

crwilcox:
:+1:

Python/Product/PythonTools/PythonTools/Repl/BasePythonReplEvaluator.cs
zooba: It's not our job to validate someone else's assembly - this is a best attempt at getting it right, and not-crashing is more important than being perfect.

If it were our own assembly (we wouldn't be doing this... but if we were...) then we'd validate it because we could fix it. There's no value here - we're in dirty hack territory already and can't really ask another team to fix their private attributes on their private assembly.

zooba: Incidentally, PTVS 3.0 won't have this hack. We'll just assume that you're running at least VS 2015 U1.

crwilcox: Good enough for me. I am fine as long as we admit this is a dirty dirty hack for VSU1.

PR #903, opened 2015-10-16
Remove old code

### LINKED ISSUE #718: Remove VS 2013 support
For our next release, we should only support VS 2015 and whatever the next version of VS is (should be approx. one year from now).

This will allow us to further improve stability by removing old code for VS 2013 (and earlier), as well as being able to take advantage of new VS 2015 APIs.

### LINKED ISSUE #720: Remove old code from source tree
We no longer build the Hpc and MpiShim projects, and should remove them completely from the repo.

There are a number of Icons folders that are no longer relevant (especially if #718 happens).

After #718, we can remove the ReplWindow project.

zooba:
Fixes #718 Remove VS 2013 support
Fixes #720 Remove old code from source tree
Adds support for VS "15" and removes support for VS 2010, 2012 and 2013.

**Don't** look at the diff for this - it will likely crash any browser. Instead, add `.diff` to the URL, save it as plain text, then open it in something that understands diff format.


zooba:
Alternatively, pull the changes and compare using your favourite git diff tool:

```
git checkout -b zooba-remove-old-code master
git pull https://github.com/zooba/PTVS.git remove-old-code
```

crwilcox:
+    <PTVSAssemblyVersion Condition="'$(VSTarget)' == '14.0'">3.0.0.2015</PTVSAssemblyVersion>
+    <PTVSAssemblyVersion Condition="'$(VSTarget)' == '15.0'">3.0.0.15</PTVSAssemblyVersion>

Is there a reason for the version number difference with Dev15?

crwilcox:
In Debugger.csproj we used to default to 11 with old version of 4.0.  Now it is 14 with old version 14.  This is probably fine but I am not entirely sure where OldToolsVersion is all used?

crwilcox:
In DebuggerHelper.vcxproj
`<When Condition=" '$(VisualStudioVersion)'=='15.0'  Or '$(TargetVisualStudioVersion)'=='VS145' ">`
This is the only place I found VS145 alongside 15.0. Shouldn't we have VS150 here?  If there is something in VS145 we should still call out VS150.

zooba:
The `VS145` was a typo.

I'm not sure where `OldToolsVersion` is used either, though I suspect it's a VS thing for undoing upgrades. That whole block is there to avoid popups when opening the projects, so I'm cautious about removing anything even if it appears unused.

The assembly version is different because the assemblies may be incompatible.

crwilcox:
I noticed the HPC stuff all stripped out.  Is that unsupported right now?  It just wasn't 100% clear it is necessary with all the other changes being about stripping out dead code. Same question for the diff differences for Repl (I think your merge from master might resolve that) and the removal of the smart tag provider.

In CloudServiceWizard `+#elif DEV15 const string AzureToolsDownload = "about:blank";`.  We could probably make this point to the dev14 go.msft.com link but I am alright with blank.  Do we know that the download will be different?  the dev11 and dev12 ones were, but I don't know the compatibility story now vs then.

crwilcox:
Reading your previous comment about compatibility is a good reason to keep about:blank.

:+1: LGTM

PR #913, opened 2015-10-19
Fix working directory of commands in Microsoft.PythonTools.Django.targets

### LINKED ISSUE #909: Some MSBuild targets use wrong working directory
My Django solution uses the following directory structure:

solution
solution\solution.pyproj
solution\solution.sln
solution\project
solution\project\manage.py
solution\project\project
solution\project\project\settings.py

To be able to run and debug the server I set in the project settings "Startup file" to "project\manage.py" and "Working Directory" to "project". Everything works fine except that the first time I press Ctrl+F5 to run the server, the build fails. If I press a second time Ctrl+F5 the server starts up correctly. After I make any changes to the files I again have to attempt to build two times until it succeeds.

I found out that the reason for the build failure is that PTVS runs "C:\Program Files (x86)\Python27\python.exe" -c "import project.settings as settings; print(settings.STATIC_URL)" in the background but from the wrong working directory. It should run from solution\project according to my "Working Directory" setting like the other Django commands do, but it runs from the solution root and consequently can't find the settings module.

In Microsoft.PythonTools.Django.targets I changed all WorkingDirectory="$(QualifiedProjectHome)" to WorkingDirectory="$(WorkingDirectory)" like in all the other commands in the file and it fixed my problem.

chausner:
Fixes #909

### DIFF
--- a/Python/Product/Django/Microsoft.PythonTools.Django.targets
+++ b/Python/Product/Django/Microsoft.PythonTools.Django.targets
@@ -64,7 +64,7 @@
     <RunPythonCommand Target="import $(DjangoSettingsModule) as settings; print(settings.STATIC_URL)"
                       TargetType="code"
                       ExecuteIn="none"
-                      WorkingDirectory="$(QualifiedProjectHome)"
+                      WorkingDirectory="$(WorkingDirectory)"
                       ConsoleToMSBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="DjangoStaticUrlSetting" />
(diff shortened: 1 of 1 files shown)


crwilcox:
:+1: 
@chausner Thanks for sharing your fix.  I don't think we need you to fill the CLA out for this but if you want to it would be out of the way if you ever contribute things that are more significant.  It doesn't take very long really.

Once @zooba takes a look I am okay merging this.

zooba:
Done. Thanks!

PR #929, opened 2015-10-23
Issue 926

### LINKED ISSUE #926: Cannot open file when coding comment does not match actual encoding
If you create a file like:

```python
# -*- coding: utf-8 -*-
# ©®
```

And save it as ANSI encoding, PTVS will raise a `BadSourceException` on load when it encounters the incorrectly encoded characters. This causes the file to not open.

We should handle the failure more gracefully and replace the invalid characters with `?` or `\uFFFD`.

zooba:
Fixes #926 Cannot open file when coding comment does not match actual encoding
Makes fallback class only return correct number of substitution characters.
Changes default encoding to UTF-8 properly.
Makes GetEncodingFromStream() (only used for the editor detection) return non-throwing encodings.

### DIFF
--- a/Python/Product/Analysis/Parsing/Parser.cs
+++ b/Python/Product/Analysis/Parsing/Parser.cs
@@ -58,7 +58,8 @@ namespace Microsoft.PythonTools.Parsing {
 
         private bool _alwaysAllowContextDependentSyntax;
 
-        private static Encoding _utf8throwing;
+        private static readonly Encoding _utf8throwing = new UTF8Encoding(false, true);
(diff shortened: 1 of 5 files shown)


huguesv:
Were there already failing tests for this? (sorry to be that guy)

zooba:
I was sure we did, but I'm not finding it now. t may not have been failing in this particular case though.

zooba:
Ping @huguesv - is it okay with the test now?

huguesv:
:+1:

PR #930, opened 2015-10-23
Fixes #920 Uninstall on upgrade runs devenv /setup

### LINKED ISSUE #920: Uninstall on upgrade runs devenv /setup
We should skip `devenv /setup` on the uninstall part of the upgrade (assuming we can detect it).

zooba:
Fixes #920 Uninstall on upgrade runs devenv /setup
Prevents running devenv /setup on upgrade.


huguesv:
:+1:

PR #946, opened 2015-11-03
Test priorities

zooba:
Splits up tests such that:
* p0 must pass to complete a CI build (generally < 5min)
* p1 should pass every time (1-2hr)
* p2 is mostly a gap...
* p3 are not reliable and need fixing

Also, test category `Installed` means that PTVS must be installed to run the test. `Interactive` means that UI interaction is necessary. These two categories are not yet 100% correct, but they are good enough for p1 non-UI tests to run.


zooba:
Reviewed by @crwilcox  and @huguesv outside of gh.

PR #963, opened 2015-11-05
Fixes #960 REPL command aliases do not work

### LINKED ISSUE #960: REPL command aliases do not work
With the change for VS 2015's shared REPL, we mishandled command aliases.

zooba:
Fixes #960 REPL command aliases do not work
Adds aliases to list of names returned by Names.


DinoV:
LGTM

PR #964, opened 2015-11-05
Fixes #957 Debug REPL commands don't show up in help

### LINKED ISSUE #957: Debug REPL commands don't show up in help
Type `$help` and you won't see any commands coming from debug REPL.  The aliases don't work, but the main commands work.

zooba:
Fixes #957 Debug REPL commands don't show up in help
Adds role and content type to process evaluators so that correct commands are loaded.


DinoV:
LGTM

PR #969, opened 2015-11-05
Fixes #962 Django shell REPL input 2 characters for every character

### LINKED ISSUE #962: Django shell REPL input 2 characters for every character typed
2.2.31104.01
VS 2015 Update 1

- Create new Django project.
- On project node, right-click and Open Django Shell.
- Type 'import', you'll see 'iimmppoorrtt'

Note: arrow keys and backspace do not repeat.

This makes the Django shell is unusable.

zooba:
Fixes #962 Django shell REPL input 2 characters for every character
Prevents an exception due to null buffer parser.


huguesv:
:+1:

PR #971, opened 2015-11-05
Fixes #967 Open command prompt here confused about selection

### LINKED ISSUE #967: Open command prompt here confused about selection?
I have multiple environments setup in my project, but only the first one is installed.
When I select Command Prompt here on the one that's installed, it complains about one that isn't installed.

command-prompt-env-not-found

Interesting, I closed and reloaded the solution, and one of the Unknown Python 3.4 is bolded.  So it was picking up the one that is active instead of the one that's selected.  I couldn't tell before because the bolding is so unreliable...

zooba:
Fixes #967 Open command prompt here confused about selection
Removes unnecessary path from PATH when overriding the factory.
Prevents extra exception when interpreter path has been overridden and environment is invalid.


huguesv:
:+1:

PR #978, opened 2015-11-07
Fixes crash when opening project with reference to C++ project.

zooba:
Fixes crash when opening project with reference to C++ project.


zooba:
This is #979

int19h:
:+1:

PR #1006, opened 2015-12-01
Issue 732 Interactive entry points

zooba:
This condenses our interactive windows down to a single menu item and a combo box in the window itself.

Per-environment interactive settings are gone for good, but are replaced by per-environment startup scripts that should be more flexible and reliable than before.

In this change, IPython support is gone temporarily. A future change will completely revamp the backend implementation, so there was no value in going to the trouble to add new UI for it now when that will certainly change. (The "execution mode" concept will change somehow, probably to be replaced by startup scripts and/or special window providers.)

Interactive window providers are now an extensibility point, so you can create a separate vsix that provides a customized experience in some way. There's also the start of some movement away from IPythonInterpreterFactory and its GUIDs throughout, which can be easily reverted if we decide not to fix them, but is fine for now.


zooba:
Going to rebase this all back on master and open a new PR.

Issue #1032, opened 2015-12-17
WinError bug


zwing99:
I am pretty sure that visualstudio_py_testlauncher.py has a bug.  It references WinError when it should reference WindowsError. https://docs.python.org/2/library/exceptions.html.

This page: https://docs.python.org/2/library/ctypes.html states that WinError is simply a wrapper to raise a WindowsError.

I use 64bit python and i had to change this in order to make PTVS run my unittests.

crwilcox:
Thanks @zwing99. That seems to make sense. If you want to you can submit a PR for that and we should be able to integrate the change. If you are not interested though, let us know and one of the core team can look into it.

Thanks!

zwing99:
Pull request created #1033

crwilcox:
Merged and closing.

PR #1033, opened 2015-12-17
Should be WindowsError not WinError

zwing99:
This fixes issue when trying to run unit tests on 64bit python in visual studio 2015.

https://docs.python.org/2/library/ctypes.html states that WinError is a wrapper that raises WindowsError which is the builtin python exception that is trying to be trapped.

I have opened a bug #1032, which this pull request should fix.


int19h:
:+1:

crwilcox:
Thanks for the fix @zwing99 !

Issue #1029, opened 2015-12-14
wfastcgi.py: SCRIPT_NAME, PATH_INFO incorrect


adamkerz:
According to both the CGI and WSGI standards (https://www.ietf.org/rfc/rfc3875, https://www.python.org/dev/peps/pep-0333/), `SCRIPT_NAME` should be the path to "application object" (ie. the script) and `PATH_INFO` the "remainder of the URL's path". But IIS seems to be passing BOTH with the FULL URL!?

To replicate, set up an application called `test` within the default website and register the wfastcgi handler for use with it. Go to `http://server/test/this` and both `SERVER_NAME` and `PATH_INFO` will be set to `'test/this'` but they should be `SERVER_NAME='test/'`, `PATH_INFO='this'`.

I don't know enough to call this an IIS bug, but it's certainly a wfastcgi bug as it doesn't conform to the WSGI standard. Is there a way to get IIS to do the right thing without modifying wfastcgi? I can't find it.

Using IIS v8.5.9600.16384 on Windows Server 2012 R2 (6.2 build 9200)

crwilcox:
Are you using https://pypi.python.org/pypi/wfastcgi?
https://github.com/Microsoft/PTVS/wiki/wfastcgi

adamkerz:
Yep, from PyPI, wfastcgi-2.2

zooba:
> Go to  http://server/test/this  and both `SERVER_NAME` and `PATH_INFO` will be set to  'test/this'  but they should be `SERVER_NAME='test/'`, `PATH_INFO='this'`

Should that be `SERVER_NAME='server/'` and `PATH_INFO='test/this'`? Or perhaps `SCRIPT_NAME='test/'`?

If the latter is what you meant, by my reading of PEP 3333 it should actually be `SCRIPT_NAME='<section of path up to the web.config file>'` and `PATH_INFO='<rest of path>'`, as that's the best we can do in terms of identifying the app vs. the path. For all of our recommended configurations/templates, this will result in `SCRIPT_NAME` being empty (which is okay), as we don't suggest non-top-level apps. Also, if you set one up, we don't necessarily have that information when we're parsing the fastcgi packet (I think... could be wrong here).

It'd be great to support handling this properly. Contributions are welcome if you figure out how to do it before we get there.

adamkerz:
Hi guys,

Thanks for taking an active interest and sorry about the terse, slightly typo-ed initial report, I was quite busy that day!

[...]

PR #1037, opened 2015-12-21
Fixing SCRIPT_NAME and PATH_INFO for non-root applications

adamkerz:
Django depends on SCRIPT_NAME and PATH_INFO adhering to PEP333 (http://www.python.org/dev/peps/pep-0333/)

See issue: #1029

### DIFF
--- a/Python/Product/WFastCgi/wfastcgi.py
+++ b/Python/Product/WFastCgi/wfastcgi.py
@@ -803,6 +803,13 @@ def main():
                     record.params['SCRIPT_NAME'] = ''
                     record.params['wsgi.script_name'] = wsgi_encode('')
 
+                # correct SCRIPT_NAME and PATH_INFO if we are told what our SCRIPT_NAME should be
+                if 'SCRIPT_NAME' in os.environ:
+                    record.params['SCRIPT_NAME'] = os.environ['SCRIPT_NAME']
+                    record.params['PATH_INFO'] = record.params['PATH_INFO'][len(record.params['SCRIPT_NAME']):]
+                    record.params['wsgi.script_name'] = wsgi_encode(record.params['SCRIPT_NAME'])
+                    record.params['wsgi.path_info'] = wsgi_encode(record.params['PATH_INFO'])
+
                 # Send each part of the response to FCGI_STDOUT.
                 # Exceptions raised in the handler will be logged by the context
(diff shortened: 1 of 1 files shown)


crwilcox:
:+1:

Python/Product/WFastCgi/wfastcgi.py
```diff
@@ -803,6 +803,13 @@ def main():
                     record.params['SCRIPT_NAME'] = ''
                     record.params['wsgi.script_name'] = wsgi_encode('')
 
+                # correct SCRIPT_NAME and PATH_INFO if we are told what our SCRIPT_NAME should be
+                if 'SCRIPT_NAME' in os.environ:
```
zooba: For safety, I'd prefer to check that `PATH_INFO` starts with `SCRIPT_NAME` as part of this condition.

zooba:
The [old IIS docs](https://msdn.microsoft.com/en-us/library/ms524602(v=vs.90).aspx) seem to suggest that you can set some properties to get the right behaviour within the server. I can't easily confirm right now, and I can't find any more recent references, but maybe this is going to be a better solution? (It may be very .NET centric though... so might not actually help.)

adamkerz:
How about that check - if PATH_INFO starts with SCRIPT_NAME (case insensitive)?

Also, `AllowPathInfoForScriptMappings` was replaced with `allowPathInfo` but when I tried that, it didn't do what it says. The [v7.0 docs](https://msdn.microsoft.com/en-us/library/ms691481%28v=vs.90%29.aspx) say that it's set to do the right thing by default (but it doesn't) and when I set it to true, I think it just cleared SCRIPT_NAME and left PATH_INFO as the full path. (It was a week or two ago that I tried it out). Either way, when setting it to true or false, neither produced the correct SCRIPT_NAME+PATH_INFO behaviour.

adamkerz:
If you figure out how to get IIS to do the right thing on it's own, I'm more than happy to test and use that instead of this slight hack!

zooba:
I'm happy with this. Can't guarantee when we'll make a release for wfastcgi (I'm keen to split it out and rewrite the tests in Python), but it sounds like you're unblocked anyway and there's no rush.

adamkerz:
Correct, I am unblocked so no rush. Thanks for your help on this!

PR #1062, opened 2016-01-19
Added Common and VSCommon assemblies for shared code and resources.

### LINKED ISSUE #745: Create shared assemblies for common code
Currently we have too many shared code files between projects, which causes all sorts of dependency nightmares.

We should add the following DLLs to reduce the burden of dealing with shared sources:
* Microsoft.PythonTools.Resources.dll
 * contains all string resources
* Microsoft.PythonTools.Utilities.dll
 * contains utility functions that require VS dependencies
* Microsoft.PythonTools.VSUtilities.dll
 * contains utility functions that also require VS DLLs

None of these will be considered public API. Mostly I expect them to expose static/extension methods. Methods with an obvious home in our existing DLLs should stay there (e.g. the analysis extensions).

zooba:
Fixes #745


int19h:
:+1:

PR #1071, opened 2016-01-28
Build process fixes

zooba:
Fixes signing certificates.
Disables default VS 2013 build.
Prevents hg users from changing line endings


int19h:
:+1:

PR #1072, opened 2016-01-29
Update Django templates

zooba:
Samples are still in another repo for this branch, so that update will have to be done separately.


huguesv:
:+1: 

Since we generate a few tests in the created project, make sure that they can run.  We had trouble with those before with new django versions.

zooba:
Yep, already checked the tests. We still need the `setup()` call, unfortunately.

PR #1073, opened 2016-01-29
Fixes #175 Django syncdb removal, and migrations support

### LINKED ISSUE #175: Django syncdb removal, and migrations support
If you try running our django syncdb command on django 1.8, you'll see:
```
The syncdb command will be removed in Django 1.9
```

It appears that we should include commands, and refer to migrations (in docs, readme, etc) instead.
https://docs.djangoproject.com/en/1.8/topics/migrations/

zooba:
Fixes #175 Django syncdb removal, and migrations support


huguesv:
:+1:

PR #1074, opened 2016-01-29
Issue 1043

zooba:
Fixes debugging of templates with Django 1.8+
Also fixes template offsets being messed up when line endings are not `\n`.


int19h:
:+1:

Issue #808, opened 2015-09-03
Automatic Brace Completion


moleschott:
Hi guys.
In settings, Automatic brace completion is disabled.It's very annoying thing to close all opening braces. Is there any way to handle this issue.

zooba:
Currently it isn't supported for Python - we need to write the support to enable it. Marking it as an enhancement.

moleschott:
Thanks zooba.

crwilcox:
If you are interested in making a contribution I did this for Node.js Tools.  I am not sure it will be a verbatim copy, but this changeset (and future changes to BraceCompletion.cs) should provide a good idea of how you can use the IVS interface (IBraceCompletionContext) to provide this support for PTVS.

https://github.com/Microsoft/nodejstools/commit/95eb76d875384493dcaa481901e327c0d6c49f56

romanvm:
The same for quotes. It's very annoying to have different experiences for C++/C# (where brace/quote completion works) and Python.

moleschott:
Any improvements on this issue @zooba .

zooba:
Not yet - it won't be in our next update (we've been focusing on crash/bug fixes and avoiding enhancements). We're certainly willing to do it, but it's a matter of having time and balancing it against other features.

(Many enhancements like this get into the product because one of the core developers really wants it. However, I personally don't like this feature so it's unlikely to be me who forces it in :) )

moleschott:
You mean that followers opinions are not important and can be easily ignored. I get that we will always write code in your teams planned style. If you wish something, it will be a gift for us. This is an embarrassing comment @zooba from a Microsoft supported project manager.

zooba:
Not at all, it just means that we have to prioritise everything and this is fairly low priority. But because we accept contributions, if anybody wants to bypass our prioritization, they are more than welcome to and we will accept the change. The point of the "help wanted" tag is to indicate that (a) we probably won't get to this ourselves and (b) we think it's a suitable change for someone to contribute.

Occasionally our own team will bypass the prioritization outside of work hours, but recently I've been the only one who has done that. I'm simply saying that I'm not going to do that for this feature, and so unless someone steps up to contribute it then it isn't going to make it ahead of the other things we need to work on.

int19h:
[...]

PR #1091, opened 2016-02-10
Added Automatic Brace Completion #808

martinzimmermann:
This enables Automatic Brace Completion the solution is very similar to that, that @crwilcox  suggested.

### DIFF
--- a/Python/Product/PythonTools/PythonTools.csproj
+++ b/Python/Product/PythonTools/PythonTools.csproj
@@ -237,6 +237,9 @@
     </Reference>
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="PythonTools\Editor\BraceCompletion\BraceCompletionContext.cs" />
+    <Compile Include="PythonTools\Editor\BraceCompletion\BraceCompletionContextProvider.cs" />
+    <Compile Include="PythonTools\ProvideBraceCompletionAttribute.cs" />
     <Compile Include="PythonTools\Repl\IMultipleScopeEvaluator.cs" />
     <Compile Include="PythonTools\Repl\InlineReplAdornment.cs" />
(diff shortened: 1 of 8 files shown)


zooba:
Looks good! The one bit that bothers me is the helper function in the test to enable/disable brace completion. We used to have code like that all over the place and it is notoriously unreliable.

It should be possible to get at the setting through DTE, as we do in a few other places. When I get a bit of time, I'll see if I can track down the names needed to find the setting.

martinzimmermann:
Jeah, I didn’t like it either. I tried to get it working through DTE but after a few hours I gave up. A MSDN article suggested that the names can be found in the registry, but I wasn’t able to locate the TextEditor -> Python -> General option page. 

If you can point me in the right direction I can also update the test myself.

zooba:
Sorry I took so long to get around to this - things have been very busy here. Because of the conflicts (which weren't real conflict), I've merged this by hand. Hopefully is still has your name on it, but I can't be sure of that (wasn't a regular git merge... git and I don't get along that well :) )

You'll see it appear in our next major update, which is going to be a while away yet (preview builds coming soon though), but congrats on being the first external contributor for it!

martinzimmermann:
I think something went wrong, because it is not associated with my Github account (Github associates commits by commit email as far as I know) however I don’t care as long as it is fixed.

Also I’m very honored to be the first external contributor (I probably should commit another thing to claim that status officially :stuck_out_tongue_winking_eye:)

zooba:
Yeah, I think the email got dropped when I moved it :( Sorry about that.

You're more than welcome to contribute more (though it is only first external contributor _for our next release_ - but there haven't been many in the past, so you're probably still in the first 10 ;) )

PR #1103, opened 2016-02-18
Issue 1101

zooba:


zooba:
Hoping these are the last changes needed for 2.2.3. This screenshot shows what the project menu looks like now.

image

huguesv:
FYI readme.mht for Django Web Project will need to be updated, as it includes a screenshot similar to the one you have above & it needs to refer to the new Django Create Superuser command instead of Django Sync.

PR #1126, opened 2016-03-01
Templates

zooba:
Moves templates into central location and gives them the expected layout. This reduces the amount of MSBuild magic we need.


zooba:
I'm going to make a further adjustment to this and split up templates by category. That will make it much easier to build sub-packages.

zooba:
Still more fixes required

zooba:
Build works fine, so merging this in.

PR #1138, opened 2016-03-10
Django fixes

zooba:
Grafts Django fixes from 2.2.3


huguesv:
That looks fine.  Is the fix for "Add Django App" not working already in this branch?

int19h:
:+1:

zooba:
@huguesv The fix for that was in the installer, which is both irrelevant now (for low-impact install) and also already in the branch (for old-style install).

Issue #1110, opened 2016-02-23
Can't create or load IronPython project


huguesv:
PTVS 3.0

Problem creating a new IronPython project, when IronPython is already installed.  I get a message box:
```
---------------------------
Microsoft Visual Studio
---------------------------
The domain manager specified by the host could not be instantiated.
---------------------------
OK   
---------------------------
```

You can create one if you don't have IronPython installed.  But if I try to load it after installing IronPython, I get the same error in output window.

zooba:
I think this is because of some GAC changes and we don't have enough search paths added to the new appdomain we create for IronPython IntelliSense.

zooba:
Should be fixed with #1200 and #1216

PR #1160, opened 2016-03-17
Temporarily work around #1110 Can't create or load IronPython project

zooba:
Temporarily work around #1110 Can't create or load IronPython project
Exclude IronPython support from builds.
Fixes inconsistencies with other exclusions.


int19h:
:+1:

PR #1162, opened 2016-03-17
Fixes #1161 Cannot find Python install dir when profiling

### LINKED ISSUE #1161: Cannot find Python install dir when profiling
Profiling needs to find the extension installation directory, but with a recent change it finds the wrong directory and fails because the profiling tools are not there.

zooba:
Fixes #1161 Cannot find Python install dir when profiling
Allows callers to PythonToolsInstallPath.GetFile to override the assembly used for locating files.


int19h:
:+1:

PR #1179, opened 2016-03-21
Issue 808

zooba:
Updated PR #1091 to merge, and also to fix how the test was updating the setting.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Editor/BraceCompletion/BraceCompletionContextProvider.cs
+++ b/Python/Product/PythonTools/PythonTools/Editor/BraceCompletion/BraceCompletionContextProvider.cs
@@ -45,12 +45,17 @@ namespace Microsoft.PythonTools.Editor.BraceCompletion {
             Debug.Assert(openingPoint.Position >= 0, "SnapshotPoint.Position should always be zero or positive.");
 
             if (openingPoint.Position > 0) {
-                var classificationSpans = openingPoint.Snapshot.TextBuffer.GetPythonClassifier().GetClassificationSpans(new SnapshotSpan(openingPoint - 1, 1));
+                var classificationSpans = openingPoint.Snapshot.TextBuffer
+                    .GetPythonClassifier()
+                    .GetClassificationSpans(new SnapshotSpan(openingPoint - 1, 1));
+
                 foreach (var span in classificationSpans) {
-                    if (span.ClassificationType.IsOfType("comment"))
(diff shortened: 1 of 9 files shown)


Python/Product/PythonTools/PythonTools/ProvideBraceCompletionAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.PythonTools {
+
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
+    class ProvideBraceCompletionAttribute : RegistrationAttribute {
+        private string _languageName;
+
+        public ProvideBraceCompletionAttribute(string languageName) {
+            _languageName = languageName;
+        }
```
DinoV: Nit: formatting should have a newline between methods

Python/Product/PythonTools/PythonTools/Editor/BraceCompletion/BraceCompletionContextProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+                return false;
+            }
+        }
+
+        private bool IsValidBraceCompletionContext(SnapshotPoint openingPoint) {
+            Debug.Assert(openingPoint.Position >= 0, "SnapshotPoint.Position should always be zero or positive.");
+
+            if (openingPoint.Position > 0) {
+                var classificationSpans = openingPoint.Snapshot.TextBuffer.GetPythonClassifier().GetClassificationSpans(new SnapshotSpan(openingPoint - 1, 1));
+                foreach (var span in classificationSpans) {
```
DinoV: Can we add the check for '' before the current point to avoid making triple quoted strings being painful?

zooba: Afraid not - the brace completion service doesn't even try and trigger. Here's how the sequence of pressing `'` repeatedly looks:

1. Adds first `'` and provisional `'`
2. Overwrites provisional `'`, so now we have `''`
3. Adds third `'` and no brace completion request is made - we have `'''`
4. Repeat step 3 for all subsequent quotes

So when typing `'''`, we don't complete the closing three (we _could_ support that manually through the edit filter, but brace completion works on `char` and not `string`, so I don't think we'd get overwrite support easily), but it's also not especially painful.

PR #1190, opened 2016-03-24
Microbuild

zooba:
Updates build and signing to work with our internal build system.

Significant changes:
* NuGet packages are listed in `PreBuild.ps1` and installed without version numbers in directories
* Forward port of #1186 
* Fixed FXCop issues
* Moved sample projects under Templates folder
* Release builds produce full set of VSIXs, FOPXs and MSIs


Python/Product/Django/Project/DjangoProject.cs
```diff
@@ -57,6 +58,20 @@ partial class DjangoProject :
         public DjangoProject() {
         }
 
+        public void Dispose() {
```
DinoV: Who actually calls this Dispose method?  It's also weird to see the full Dispose pattern with the bool disposing parameter when this doesn't a finalizer that calls it with Dispose(false).

Python/Product/Django/Project/DjangoPropertyPage.cs
```diff
@@ -30,6 +30,16 @@ class DjangoPropertyPage : CommonPropertyPage {
         public DjangoPropertyPage() {
             _control = new DjangoPropertyPageControl(this);
         }
+
```
DinoV: Same thing here, who's calling this?  Also why are we introducing protected members into classes with no subclasses?  Neither of these will ever be called with Dispose(false)

Python/Product/EnvironmentsList/PipExtensionProvider.cs
```diff
@@ -72,6 +73,15 @@ public sealed class PipExtensionProvider : IEnvironmentViewExtension {
             _cache = PipPackageCache.GetCache(_index, _indexName);
         }
 
+        [SuppressMessage("Microsoft.Usage", "CA2213:DisposableFieldsShouldBeDisposed", MessageId = "_pipCancel",
```
DinoV: It doesn't look like EnvironmentView ever disposes of it's extensions, so this is never called?

Python/Product/Profiling/Profiling/BaseHierarchyNode.cs
```diff
@@ -148,5 +149,20 @@ abstract class BaseHierarchyNode : IVsUIHierarchy {
                 ev.OnPropertyChanged(itemid, propid, flags);
             }
         }
+
```
DinoV: This never gets called either?

Python/Product/Uwp/Project/PythonUwpPropertyPage.cs
```diff
@@ -32,6 +32,16 @@ class PythonUwpPropertyPage : CommonPropertyPage {
         public PythonUwpPropertyPage() {
             _control = new PythonUwpPropertyPageControl(this);
         }
+
+        public void Dispose() {
```
DinoV: I'm wondering who calls this one as well, but the protected pattern is also certainly not necessary

Python/Setup/PythonToolsInstaller/PythonToolsInstaller.wixproj
```diff
@@ -77,10 +76,6 @@
       <Name>UwpMsm</Name>
       <Project>{83946E81-7A1B-4D3B-927F-4CD67AC95BE7}</Project>
     </ProjectReference>
-    <ProjectReference Include="$(PTVSPRPath)VsLoggerSetup\VsLogger.wixproj" Condition="'$(IncludeVsLogger)' == 'True'">
```
DinoV: We're removing our private telemetry DLL?  That seems like a step backwards

DinoV:
I'm really just wondering about all these random Dispose methods that are introduced that are never getting called.  If we're adding them they should actually be getting called and we probably shouldn't be using the protected Dispose pattern in places where we don't need it which is most of them.

Python/Setup/PythonToolsInstaller/PythonToolsInstaller.wixproj
zooba: It's using the wrong system, so we're not getting any data from it anyway. When we figure out how to use the new one then we'll add that in.

zooba:
The Dispose methods are added to keep FXCop happy. I can make them not use the full pattern, but that requires sealing all of the classes which is a breaking change (though we're accepting breaking changes for this release, so I guess that's fine - just automatically went for the non-breaking change).

Issue #1210, opened 2016-04-07
Per-user interpreter not found when per-machine one exists


zooba:
I have three Python 3.5 interpreters installed: 64-bit per-user, 32-bit per-machine and 32-bit per user. The last one of these is not listed.

It should either be listed alongside the 32-bit per-machine install, or replace the per-machine one entirely.

Issue #1195, opened 2016-03-28
Intellisense error with class method


lmazuel:
When using a class method, the tooltip highlight does not follow the good parameter:
![image](https://cloud.githubusercontent.com/assets/1050156/14089737/4c4dfb84-f4eb-11e5-9580-c036d2b92264.png)
and then:
![image](https://cloud.githubusercontent.com/assets/1050156/14089752/628eb532-f4eb-11e5-8f25-d1b2b1944bb5.png)

Tested with [GitPython](http://gitpython.readthedocs.org/en/stable/index.html), class method `Repo.clone_from` and PTVS 2.2.31124.00 on VS 14.0.24720.00 Update 1

zooba:
We shouldn't be displaying the `cls` parameter here when you're accessing the method on a class.

Issue #1119, opened 2016-02-25
Cannot rename a file the same name that a missing file in Solution Explorer


lmazuel:
Hi,

Let's suppose I have `A.py` and `B.py` in a folder and I added these two files in my Python project.
I delete `A.py` from the HD using the Windows File Explorer (not PTVS). As expected, the `A.py` file in Solution Explorer in PTVS becomes black with a yellow triangle to indicate the file is missing.
Now I try to rename `B.py` into `A.py` using PTVS. PTVS refuses arguing that:

> A file or folder with name 'A.py' already exists on disk at this location. Please choose another name.

which is not true, the file does not exist.

VS 2015 with PTVS 2.2.31124.00

Issue #923, opened 2015-10-21
Return annotation does not receive any syntax highlighting


brettcannon:
E.g.,:
```python
def foo(arg: t.Iterable) -> t.Iterable: ...
```

The `t.Iterable` for the function parameter gets highlighted, but the exact same type annotation for the return type gets no highlighting.

PR #1219, opened 2016-04-11
Bug fixes

### LINKED ISSUE #110: Fix ClassDef header index
The header index for ClassDef is inconsistent with FunctionDef.  It should point to the ':' and should be one more forward than it is.

Once Fixed fix the region collapsing to not need a + 1

Python/Product/PythonTools/PythonTools/OutliningTaggerProvider.cs  
AddTagIfNecessary(node, node.HeaderIndex + 1, node.Decorators);

DinoV:
#110 Fix ClassDef header index
	Set header to always be the start of the :, always use the header index when generating the outlining tags
#923 Return annotation does not receive any syntax highlighting
	Just need to walk the retur annotation in FunctionDefinition
#1119 Cannot rename a file the same name that a missing file in Solution Explorer
	Make Remove return a bool so we can detect when it fails, check for open document, prompt to close/save,
	and remove the old item if we successfully remove the hierarchy node and the file doesn't get saved
#1195 Intellisense error with class method
	Need to write out that function is class method in saved analysis, and get rid of hacky static descriptor support
#56 Set next statement warning prompt regarding moving to a different function
	We weren’t saving the frame in the constructor, so when we’d go to compare the memory addresses we’d end up comparing by line #
#206 Copy paste item to another project's subfolder doesn't include file in project
	We were going to the current Python project, not consulting the source hierarchy, to check if we were a non-member item
#337 Cannot handle generalised unpacking
	Flush out generalized unpacking support at the parsing level, and add unit tests
#1210 Per-user interpreter not found when per-machine one exists
	Favor per-user interpreters over per-machine


Python/Product/VSInterpreters/CPythonInterpreterFactoryProvider.cs
```diff
@@ -230,77 +221,90 @@ class CPythonInterpreterFactoryProvider : IPythonInterpreterFactoryProvider {
...
+                        RegisterInterpreters(machineFactories, registeredPaths, false, python64, ProcessorArchitecture.Amd64);
+                    }
+                }
+            }
+
+            // Then update our cached state with the lock held.
+            lock (this) {
+                var uniqueIds = userFactories.Select(x => x.Key)
+                       .Concat(machineFactories.Select(x => x.Key))
+                       .Distinct();
```
zooba: Put into a set to avoid multiple evaluations/breakage (if we later change userFactories or machineFactories in the first for loop)

You could also use `.Keys` instead of the `.Select`, right? And `.Union()` instead of `.Select().Distinct()`. Or maybe I'm missing something there.

Python/Product/VSInterpreters/CPythonInterpreterFactoryProvider.cs
```diff
@@ -230,77 +221,90 @@ class CPythonInterpreterFactoryProvider : IPythonInterpreterFactoryProvider {
...
+
+                foreach (var id in uniqueIds) {
+                    InterpreterInformation newInfo;
+                    // User factories with the same ID take precedence over machine factories
```
zooba: We talked at one point about showing both and labelling them differently (though probably only if both exist). Something we want to try? Or save it for later?

Python/Product/PythonTools/PythonTools/OutliningTaggerProvider.cs
```diff
@@ -103,7 +103,7 @@ internal class OutliningTagger : ITagger<IOutliningRegionTag> {
 
                 var tags = await entry.Analyzer.GetOutliningTagsAsync(snapshot);
                 if (tags != null) {
-                    _tags = tags.Concat(ProcessRegionTags(snapshot)).ToArray();
+                    _tags = tags.Concat(ProcessRegionTags(_buffer.CurrentSnapshot)).ToArray();
```
zooba: Why would we want to use a different snapshot here?

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -1862,7 +1865,7 @@ class ExprWalker : PythonWalker {
                 }
 
                 yield return OutliningTaggerProvider.OutliningTagger.GetTagSpan(
-                    currentSnapshot,
+                    translator.TextBuffer.CurrentSnapshot,
```
zooba: I don't really like the idea of this potentially changing while enumerating... why do we want to make this change?

Python/Product/VSInterpreters/CPythonInterpreterFactoryProvider.cs
DinoV: The difficultly there is actually with the IDs themselves...  If we want to do that we need to have that show up in the interpreter ID, so that we'd end up with something like:
Global|PythonCore|3.5|x86|User
 or
Global|PythonCore|3.5|x86|Machine

Or something along those lines.  My current thinking is that will cause more trouble in cross machine scenarios then it's worth enabling users to have both a per-user and per-machine install.

Python/Product/PythonTools/PythonTools/OutliningTaggerProvider.cs
DinoV: The issue here is that the buffer may have been modified during the await call.  So we capture the current snapshot which holds the earliest possible version we need to translate from.  But then we need to apply the changes to the current version, not the earliest possible version.

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
DinoV: Yeah, I'll hoist that to a local at the start of the function, the important thing is that we get the current snapshot when we entered this function

Issue #1136, opened 2016-03-09
No cross-package analysis for site-packages


lmazuel:
Azure-common package contains conditional aliases:
```python
try:
    from msrestazure.azure_exceptions import CloudError
except ImportError:
    raise ImportError("You need to install 'msrestazure' to use this feature")
```

It seems to be a problem for PTVS
![intellisenseproblem](https://cloud.githubusercontent.com/assets/1050156/13623009/2b18c310-e558-11e5-875f-307719d0dced.png)

There is no problem if I import directly the class

zooba:
This is actually because we don't do sophisticated enough cross-module analysis when analyzing installed packages. We do `msrestazure` and `azure` separately, and they don't know anything about each other, which is why the imports don't resolve.

We could probably scan the install metadata and selectively load other packages to handle this. If we try and load everything all the time then analysis will never complete.

lmazuel:
Even if it's not solving the whole problem, if I add a `__all__` statement this can help? Because in that case, you will have the explicit set of stuff I want to export.

zooba:
No, the only way to help is to copy the entire type definition over (or put a stub definition under `if False`).

PR #1222, opened 2016-04-14
Analyzer speed improvements

zooba:
This PR is mostly about improving the analysis browser so it can export diffable versions of DBs.

The analyzer perf improvement comes from deferring all files from `*.test[s].*` packages into a second analysis. Effectively, we let test packages import from their surrounding package, but not feed analysis back into them. This results in significantly less results, most of which appear as fewer parameter types (which are basically unused anyway).

In a majority of cases, these look like improvements (losing incorrect information), while there are some cases where we lose correct information about some attribute types. Given the speed improvement in the initial analysis, this seems to be worthwhile.

We also now ensure that all analyzer packages are loaded when analyzing others, and allow the order to be overridden to ensure certain packages are analyzed first (addresses #1136).


Python/Product/AnalysisBrowser/FunctionView.cs
```diff
@@ -37,29 +38,45 @@ public FunctionView(IModuleContext context, string name, IPythonFunction member,
...
+                }
 
-                            args[i].Add(parameters[i].Name);
+                var args = new List<HashSet<string>>();
+                var types = new List<HashSet<string>>();
+
+                foreach (var overload in _function.Overloads) {
+                    var parameters = overload.GetParameters();
+                    for (int i = 0; i < parameters.Length; ++i) {
+                        while (args.Count <= i) {
```
DinoV: That's this while loop for?  Isn't this the same as args.Add(new HashSet<string>()); with no loop?

zooba: Depends how many overloads we have, how many parameters each has, and what order we get to them in.

Python/Product/AnalysisBrowser/MultipleMemberView.cs
```diff
@@ -44,5 +46,38 @@ class MultipleMemberView : MemberView {
...
+            TextWriter writer,
+            string currentIndent,
+            string indent,
+            Stack<IAnalysisItemView> exportStack,
+            out IEnumerable<IAnalysisItemView> exportChildren
+        ) {
+            var children = new HashSet<string>();
+            foreach (var c in Children) {
+                string displayValue = "";
+                if (c is ClassView) {
```
DinoV: Maybe have a DisplayValue virtual member?

DinoV:
LGTM other than the slightly odd loop

Python/Product/AnalysisBrowser/MultipleMemberView.cs
zooba: Considered that, but I'd still want the type filtering here. e.g. `ModuleRefView.DisplayValue` would have something valuable to display, but we'd still want to exclude it here for the purposes of the export.

zooba:
The loop is basically unavoidable, and I don't really like any of the alternatives I can think of for a transpose operation like that.

PR #1223, opened 2016-04-14
Issue 732

zooba:


Python/Product/PythonTools/PythonTools.vsct
```diff
@@ -286,24 +286,13 @@ permissions and limitations under the License.
...
-      </Button>
-
       <Button guid="guidPythonToolsCmdSet" id="cmdidFillParagraph" priority="0x7f0f" type="Button">
         <Parent guid="guidStdEditor" id="IDG_VS_EDITOR_ADVANCED_CMDS"/>
         <CommandFlag>DynamicVisibility</CommandFlag>
         <CommandFlag>DefaultInvisible</CommandFlag>
         <Strings>
           <ButtonText>&amp;Fill Comment Paragraph</ButtonText>
-          <ToolTipText>TODO: Explain what this command actually does</ToolTipText>
+          <ToolTipText>Reflows the selected paragraph to maximize its width.</ToolTipText>
```
DinoV: Adjust it's width might be better

Python/Product/PythonTools/PythonTools/Commands/ExecuteInReplCommand.cs
```diff
@@ -49,31 +49,14 @@ internal sealed class ExecuteInReplCommand : Command {
...
-            
-            if (project != null && vsProjectContext.IsProjectSpecific(config)) {
-                project.AddActionOnClose(window, BasePythonReplEvaluator.CloseReplWindow);
-            }
+            string replId = project != null ?
+                PythonReplEvaluatorProvider.GetEvaluatorId(project) :
+                PythonReplEvaluatorProvider.GetEvaluatorId(config);
+            var window = provider.OpenOrCreate(replId);
+            project?.AddActionOnClose(window, InteractiveWindowProvider.Close);
+            //if (project != null && project.Interpreters.IsProjectSpecific(factory)) {
```
DinoV: Delete commented out code

Python/Product/PythonTools/PythonTools.vsct
```diff
@@ -286,24 +286,13 @@ permissions and limitations under the License.
         </Strings>
       </Button>
 
-      <Button guid="guidPythonToolsCmdSet" id="cmdidSendToDefiningModule" priority="0x201" type="Button">
-        <Parent guid="guidPythonToolsCmdSet" id="EditMenuGroup"/>
-        <CommandFlag>DynamicVisibility</CommandFlag>
```
DinoV: Why are we losing this feature?

Python/Product/PythonTools/PythonTools/Navigation/CodeWindowManager.cs
```diff
@@ -143,8 +145,8 @@ class CodeWindowManager : IVsCodeWindowManager, IVsCodeWindowEvents {
             var wpfTextView = VsEditorAdaptersFactoryService.GetWpfTextView(vsTextView);
             if (wpfTextView != null) {
                 var factory = ComponentModel.GetService<IEditorOperationsFactoryService>();
-                var editFilter = new EditFilter(wpfTextView, factory.GetEditorOperations(wpfTextView), _serviceProvider);
-                editFilter.AttachKeyboardFilter(vsTextView);
+                EditFilter.GetOrCreate(_serviceProvider, ComponentModel, vsTextView);
+                ReplEditFilter.GetOrCreate(_serviceProvider, ComponentModel, vsTextView);
```
DinoV: Why do we deal with the REPL here?  The code window manager is dealing with windows which are normal code files hosted in the editor, and in particular is handling the navigation tool bar.  This code path shouldn't be hit at all for a REPL window

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
```diff
@@ -30,202 +34,178 @@ namespace Microsoft.PythonTools.Repl {
...
-            // seems to work fine.
-            __VSCREATETOOLWIN creationFlags = 0;
-            if (!roles.Contains("DontPersist")) {
+            if (alwaysCreate) {
                 creationFlags |= __VSCREATETOOLWIN.CTW_fForceCreate;
             }
 
             var replWindow = _windowFactory.Create(GuidList.guidPythonInteractiveWindowGuid, id, title, evaluator, creationFlags);
+            ((ToolWindowPane)replWindow).BitmapImageMoniker = KnownMonikers.PYInteractiveWindow;
+            replWindow.InteractiveWindow.Properties[typeof(IVsInteractiveWindow)] = replWindow;
```
DinoV: Why are we assigning this property?  It seems very strange that we're storing a property under an interface type that we don't own, and I don't see any other cases of typeof(IVsInteractiveWindow) being used as far as searching on github is concerned

Python/Product/PythonTools/PythonTools/Repl/LoadReplCommand.cs
```diff
@@ -39,10 +39,14 @@ class LoadReplCommand : IInteractiveWindowCommand {
...
-            var eval = window.Evaluator as BasePythonReplEvaluator;
-            if (eval != null && eval.CurrentOptions != null) {
-                finder.Search(eval.CurrentOptions.WorkingDirectory);
-                finder.SearchAll(eval.CurrentOptions.SearchPaths, ';');
+            var eval = window.GetPythonEvaluator();
+            if (eval != null) {
+                finder.Search(eval.WorkingDirectory);
+                string searchPaths;
+                // HACK: Should not assume variable name is PYTHONPATH
+                if (eval.EnvironmentVariables.TryGetValue("PYTHONPATH", out searchPaths)) {
```
DinoV: eval.Analyzer.InterpreterFactory.Configuration.PathEnvironmentVariable instead of the hack?

Python/Product/PythonTools/PythonTools/Repl/PythonInteractiveEvaluator.CommandProcessorThread.cs
```diff
@@ -0,0 +1,1017 @@
...
+            if (!string.IsNullOrWhiteSpace(InterpreterArguments)) {
+                args.Add(InterpreterArguments);
+            }
+
+            args.Add(ProcessOutput.QuoteSingleArgument(PythonToolsInstallPath.GetFile("visualstudio_py_repl.py")));
+            args.Add("--port");
+            args.Add(portNum.ToString());
+
+            args.Add("--execution-mode");
+            args.Add("standard");
```
DinoV: We don't have IPython support anymore?

DinoV:
ReplEditFilter.QueryStatus has a bunch of unused branches which can be cleaned up

DinoV:
SelectableReplEvaluator.AttachWindow: There's a bunch of stuff dealing with VsProjectAnalyzer as a property which I don't think is necessary anymore with the associated changes to IntellisenseController.PropagateAnalyzer and Extensions.GetAnalyzer

DinoV:
Does anything get outputted when the user switches evaluators?  I don't see any explicit output, but maybe something occurs when the new evaluator gets started?

Python/Product/PythonTools/PythonTools/Navigation/CodeWindowManager.cs
zooba: The refactoring split EditFilter into two classes, and I left the code to attach them in the original location. It's either not the place to attach edit filters at all (probably true), or it's the right place to attach both. I don't know where the better place would be if one exists.

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
zooba: Not sure, maybe for something that's been removed since? Gone now

Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -1716,7 +1716,7 @@ internal enum EventTriggering {
         /// Return the value of a project property
         /// </summary>
         /// <param name="propertyName">Name of the property to get</param>
-        /// <param name="resetCache">True to avoid using the cache</param>
```
DinoV: It looks like the old comment was correct?

Python/Product/PythonTools/PythonTools/Options/SuppressDialogOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public void Save() {
+            _service.SaveString(SwitchEvaluatorSetting, Category, SwitchEvaluator);
+        }
+
+        public void Reset() {
+            SwitchEvaluator = null;
+        }
+
+        public string SwitchEvaluator { get; set; }
```
DinoV: The const values this gets set to should probably be available over here as well, or this should be an enum or bool?

Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
```diff
@@ -40,24 +41,28 @@ class SendToReplCommand : Command {
...
 
-            PythonReplEvaluator eval = repl.InteractiveWindow.Evaluator as PythonReplEvaluator;
-
-            eval.EnsureConnected();
-            repl.InteractiveWindow.Submit(GetActiveInputs(activeView, eval));
+            var version = analyzer.InterpreterFactory.Configuration.Version.ToLanguageVersion();
 
             repl.Show(true);
+
+            foreach (var code in GetActiveInputs(activeView, version)) {
```
DinoV: Why do we submit each input individually instead of sending them all to SubmitAsync at once?

Common/Product/SharedProject/ProjectNode.cs
zooba: You're right - no idea why I changed that...

Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
zooba: There's a bug in the interactive window where they only process the first element. Haven't checked whether they've fixed it yet, but it definitely shipped in VS 2015 and IIRC U1.

Python/Product/PythonTools/PythonTools/Navigation/CodeWindowManager.cs
DinoV: I'm pretty sure we still don't need this to be created here...  We already still attach this in ReplWindowCreationListener and the CodeWindowManager is never used for a REPL window

Python/Product/PythonTools/PythonTools/Navigation/CodeWindowManager.cs
```diff
@@ -203,6 +205,10 @@ class CodeWindowManager : IVsCodeWindowManager, IVsCodeWindowEvents {
                 if (_curView.Properties.TryGetProperty(typeof(EditFilter), out editFilter) && editFilter != null) {
                     editFilter.DoIdle((IOleComponentManager)_serviceProvider.GetService(typeof(SOleComponentManager)));
                 }
+                ReplEditFilter replEditFilter;
```
DinoV: Why did we gain this?  ReplEditFilter.DoIdle doesn't even do anything

Python/Product/PythonTools/PythonTools/Options/SuppressDialogOptions.cs
zooba: I wanted to avoid creating enums for every possible dialog, and the only UI I ever expect to expose here is "reset all dialogs". So all the possible values are constrained to where the dialog is created/used, and are specific to the dialog (in this case, the values are in `ReplEditFilter` and could be `AlwaysSwitch` or `AlwaysOpenNew`).

Python/Product/PythonTools/PythonTools/Navigation/CodeWindowManager.cs
zooba: Removed

Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
DinoV: Can we just add a comment about that then?

zooba: Yep, already added.

Python/Product/PythonTools/PythonTools/Options/SuppressDialogOptions.cs
zooba: Making this class internal and leaving the API as-is.

Python/Product/PythonTools/PythonTools/Navigation/CodeWindowManager.cs
DinoV: We're still doing this?

---
In reply to: [61307399](https://github.com/Microsoft/PTVS/pull/1223#discussion_r61307399) [](ancestors = 61307399)

PR #1225, opened 2016-04-14
Fixes #1117 Auto indent default is block

### LINKED ISSUE #1117: Auto indent default is block
PTVS 3.0
CPython 3.4

I am using a pre-installed image with PTVS 3.0, and the default indentation was set to Block rather than Smart.  I am 99% sure I didn't change it myself.  I checked that C# default was set to Smart.

Block indenting gives you this:

```
def hello():
print('hello')
```

zooba:
Delays package load until after the shell is initialized.


zooba:
Hold on this - I may have a simplification.

PR #1226, opened 2016-04-15
Bugs

### LINKED ISSUE #186: Quick Info doesn't properly handle `$` magic commands in Interactive Window
Type `$help` in the REPL, then hover over it. A Quick Info tip appears, treating the single character over which the mouse cursor is hovered as an identifier. It should treat the entire command as a single token, and not provide the tip for it at all, since it's not a variable.

### LINKED ISSUE #557: Can't edit Python files in Visual Studio diff window
When comparing files in Visual Studio compare/diff (for example, comparing a local version to one checked in to TFS source control), I can't edit the local version of the file. This works fine for other file types. I can also try to use automatic refactoring tools on the local version, but it has no effect after pressing "Apply".

### LINKED ISSUE #597: Extract Method does not appear on the editor context menu
(Definitely VS 2015, haven't checked VS 2013 but I doubt it's changed there.)

The Extract Method command is no longer visible on the editor context menu.

It appears under Edit -> Refactor, and the keyboard shortcut still works. Users can use Tools -> Customize to add the item to `Editor Context Menus | Code Window` - "Extract Method" is found under Edit in "Add Command".

### LINKED ISSUE #613: Don't show signature help in bases list
If you trigger signature help in the bases list (example below), we show the parameters to `__init__`.

We should not show any signature help here.

Example:
```python
class MyClass(signature help here):
    pass
```

### LINKED ISSUE #792: Find all references on an imported member shows the definition in the reference list
VS Desktop Express 2013, IronPython 2.7.5, PTVS Dev 2015-08-21 VS 2013

This is a minor, slightly untidy cosmetic discrepancy, possibly not worthy of your attention :) and maybe even by design, but if so I don't see why :(

Right-click -> Find All References on an imported item, from a pure Python code module or a .NET assembly, seems to tag the definition line onto the end of the References list in the Find Symbol Results window as well as showing it in the Definitions list. It doesn't do this for anything defined in the current module

DinoV:
Fix #597
Add a Extract Method suggested action and include a preview of the new change
Add a PreviewChangesService for creating a preview text view based upon a set
 of changes

Fix #186
Special case checking for REPL buffers that start with a $ and don't provide 
quick info.  Unfortunately we have to do that because the REPL window input 
kind doesn't change until we have a complete input

Fix #557 
Clean up our BufferParser/AnalysisEntry/TextView tracking so that we can 
support this.  We no longer attach any properties to text buffers at all.  
Instead we always resolve from textview and/or filename into a 
ProjectAnalyzer and then get the buffer from there.  

We now flow TextView's where they're available into our analyzer.  There's
only really 3 spots where we don't have that full context: Outlining 
regions and proximity info via the old IVsLanguageInfo interfaces.

Add support for CommonPackage.GetActiveTextView so that we can find the 
correct active text view for diff windows. This is used all over the place 
to find the text view when we're doing various editing commands like extract
method, find all refs, goto def, etc...

Update Extensions to add various helpers which deal with the caret 
within a projection buffer.  

Fix #792 
Just don't add the reference when creating the variable, we'll add it on assign

Fix #613 
Treat hitting class the same as hitting def


Common/Tests/Utilities.UI/UI/ReplWindowProxy.cs
```diff
@@ -131,6 +133,12 @@ internal sealed class ReplWindowProxy : IDisposable {
             _onDispose.Add(action);
         }
 
+        public VsProjectAnalyzer Analyzer {
+            get {
+                return ((IPythonReplIntellisense)Window.Evaluator).ReplAnalyzer;
+            }
+        }
+
```
zooba: Shouldn't be in the common test project - needs to go in TestUtilities.Python.

Python/Product/PythonTools/PythonTools.csproj
```diff
@@ -719,7 +720,7 @@
     <None Include="Theme\dark.vstheme" />
     <None Include="Theme\make_theme_pkgdef.py" />
     <Content Include="$(PackagesPath)\Newtonsoft.Json\lib\net45\Newtonsoft.Json.dll">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
zooba: Do we really need to copy always? This causes builds to fail if the DLL is still in use (which happened almost immediately to me for some reason)

Python/Product/PythonTools/PythonTools/Commands/RemoveImportsCommand.cs
```diff
@@ -36,14 +36,15 @@ class RemoveImportsCommand : Command {
 
         public override async void DoCommand(object sender, EventArgs args) {
             var view = CommonPackage.GetActiveTextView(_serviceProvider);
-            var analyzer = view.GetAnalyzer(_serviceProvider);
+            var analyzer = view.GetAnalyzerAtCaret(_serviceProvider);
+            var pythonCaret = view.GetPythonCaret();
```
zooba: If we're going to assume that `Value` is valid below, just read it here..

Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -255,23 +255,113 @@ public static class Extensions {
...
+                }
+
+                if (analyzer == null && view != null) {
+                    // We could spin up a new analyzer for non Python projects...
+                    analyzer = view.GetBestAnalyzer(provider);
+                }
+
+                if (analyzer != null) {
+                    entry = analyzer.GetAnalysisEntryFromPath(path);
+                    return entry != null;
```
zooba: This is a good place to use `entry = analyzer?.GetAnalysisEntryFromPath(path)` and reduce some control flow.

Python/Product/PythonTools/PythonTools/Intellisense/CompletionAnalysis.cs
```diff
@@ -103,7 +113,7 @@ public class CompletionAnalysis {
 
         internal AnalysisEntry GetAnalysisEntry() {
             AnalysisEntry entry;
-            if (TextBuffer.TryGetAnalysisEntry(out entry) && entry != null) {
+            if (_view.TryGetAnalysisEntry(TextBuffer, _serviceProvider, out entry) && entry != null) {
```
zooba: Does this mean analysis entries are more closely tied to views than buffers? That seems wrong...

Python/Product/PythonTools/PythonTools/Intellisense/ExtractMethodSuggestedActionSourceProvider.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        public ISuggestedActionsSource CreateSuggestedActionsSource(ITextView textView, ITextBuffer textBuffer) {
+            return new ExtractMethodSuggestedActionSource(this, textView);
+        }
+
+        class ExtractMethodSuggestedActionSource : ISuggestedActionsSource {
+            private readonly ITextView _view;
+            private readonly ExtractMethodSuggestedActionSourceProvider _parent;
+            private bool? _canExtract;
+            private static Guid _telemetryId = new Guid("{0D887D91-1E11-4DC3-92F5-79A5182004C3}");
```
zooba: readonly

Python/Product/PythonTools/PythonTools/Intellisense/ExtractMethodSuggestedActionSourceProvider.cs
```diff
@@ -0,0 +1,206 @@
...
+                if (MethodExtractor.CanExtract(_view) == true) {
+                    return new SuggestedActionSet[] {
+                    new SuggestedActionSet(
+                        new [] {  new SuggestedAction(_parent, _view) },
+                        SuggestedActionSetPriority.High
+                    )
+                };
+                }
+
+                return Array.Empty<SuggestedActionSet>();
```
zooba: `Enumerable.Empty<T>()` (apparently faster due to some caching, but I'd be surprised if it wasn't negligible)

Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
```diff
@@ -85,28 +84,49 @@ internal sealed class IntellisenseController : IIntellisenseController, IOleComm
...
             if (_quickInfoSession != null && !_quickInfoSession.IsDismissed) {
                 _quickInfoSession.Dismiss();
             }
+
             var pt = e.TextPosition.GetPoint(EditorExtensions.IsPythonContent, PositionAffinity.Successor);
             if (pt != null) {
-                var quickInfo = await VsProjectAnalyzer.GetQuickInfoAsync(pt.Value);
-
-                QuickInfoSource.AddQuickInfo(_textView.TextBuffer, quickInfo);
+                if (_textView.Properties.ContainsProperty(typeof(VisualStudio.InteractiveWindow.IInteractiveEvaluator)) &&
```
zooba: There's a `GetInteractiveWindow()` extension method for this IIRC.

Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
```diff
@@ -118,36 +138,33 @@ internal sealed class IntellisenseController : IIntellisenseController, IOleComm
...
             }
 
-            BufferParser existingParser;
-            if (!subjectBuffer.Properties.TryGetProperty(typeof(BufferParser), out existingParser)) {
-                _bufferParser.AddBuffer(subjectBuffer);
-            } else {
-                // already connected to a buffer parser, we should have the same project entry
-                Debug.Assert(_bufferParser._analysis == existingParser._analysis);
+            if (analyzer != null) {
+                analyzer.MonitorTextBufferAsync(subjectBuffer, isTemporaryFile).ContinueWith(task => {
```
zooba: `ContinueWith` does not marshal back to the original thread by default, but we need to if we want `AttachedViews++` to be safe.

Probably more reliable just to lock the parser while incrementing though.

Python/Product/PythonTools/PythonTools/Intellisense/NormalCompletionAnalysis.cs
```diff
@@ -136,7 +137,7 @@ internal NormalCompletionAnalysis(VsProjectAnalyzer analyzer, ITextSnapshot snap
                     }
                 }
 
-                if (pyReplEval != null && _snapshot.TextBuffer.GetAnalyzer(_serviceProvider).ShouldEvaluateForCompletion(text)) {
+                if (pyReplEval != null && pyReplEval.ReplAnalyzer.ShouldEvaluateForCompletion(text)) {
```
zooba: Should `ReplAnalyzer` and `_analyzer` be identical here? Should we assert that somewhere?

Python/Product/PythonTools/PythonTools/Repl/BasePythonReplEvaluator.cs
```diff
@@ -89,12 +89,16 @@ internal abstract class BasePythonReplEvaluator :
         const string ExecuteFileEx_Module = "module";
         const string ExecuteFileEx_Process = "process";
 
+        private readonly string _analysisFilename;
         protected BasePythonReplEvaluator(IServiceProvider serviceProvider, PythonToolsService pyService, PythonReplEvaluatorOptions options) {
```
zooba: Blank line between these

Python/Product/VSInterpreters/InterpreterOptionsService.cs
```diff
@@ -226,7 +226,7 @@ sealed class InterpreterOptionsService : IInterpreterOptionsService {
             using (var key = Registry.CurrentUser.CreateSubKey(collection, true)) {
                 key.SetValue(LibraryPathKey, config.LibraryPath ?? string.Empty);
                 key.SetValue(ArchitectureKey, config.ArchitectureString);
-                key.SetValue(VersionKey, config.Version.ToString());
+                key.SetValue(VersionKey, config.Version?.ToString() ?? "2.7");
```
zooba: Do we really want to keep using "2.7" as the default? Do we have any back-compat dependencies on this?

Python/Tests/TestData/VirtualEnv/env/Scripts/deactivate.bat
```diff
@@ -1,20 +1,20 @@
...
+    set "PYTHONHOME=%_OLD_VIRTUAL_PYTHONHOME%"
+    set _OLD_VIRTUAL_PYTHONHOME=
+)
+
+if defined _OLD_VIRTUAL_PATH (
+    set "PATH=%_OLD_VIRTUAL_PATH%"
+	set _OLD_VIRTUAL_PATH=
+)
+
+:END
```
zooba: We need to figure out which line ending settings are messing with this file...

zooba:
Generally a bit concerned/confused by the analysis entry apparently being linked to the view rather than the buffer. I'd expect no more than one entry per buffer, and the view shouldn't be relevant at all.

Python/Product/PythonTools/PythonTools.csproj
DinoV: VS must of changed this on me :(

Python/Product/PythonTools/PythonTools/Intellisense/CompletionAnalysis.cs
DinoV: Not really, they're determined by some combination of the two.  In general we use the buffers filename grabbed from the ITextDocument property to find the analyzer.  If the file is in a project then the project analyzer will be used and that's that.  

Sometimes the view also influences it though, but we can't just rely upon a property on the buffer.  For the diff view case for example you end up w/ a left view, a right view, and an inline view.  We can use the diff factory to figure out that we're in a diff view, and then if one file is a project file and one is not, we can open both in the project context, with the temporary file that's outside of the project being marked as such so that we'll throw it away.

If I then went and opened that temporary file in a non-diff view we'd throw it into the default analyzer and we'd have 2 different buffer parsers working on it and 2 different analyzers operating on it.  

Likewise REPL window buffers are also directly influenced by their view - in that case we have buffers that have no file name except for the fake filename we create for them.  We treat them all as the same file, so all of the buffers get combined and analyzed as a single AST but parsed individually.  

Other projection buffers should work similarly but it may take some special handling just like it does for other weird views like the diff view.  But there's enough odd combinations where we need to consider both the view and the buffer for actually understanding how something gets analyzed.

Python/Product/PythonTools/PythonTools/Intellisense/ExtractMethodSuggestedActionSourceProvider.cs
DinoV: Says the person who regularly uses .Any() instead of .Count != 0 ;)

Python/Product/PythonTools/PythonTools/Intellisense/NormalCompletionAnalysis.cs
DinoV: It seems like they should, I'll switch to using _analyzer and add the assert

DinoV:
So the important thing to realize is that the analysis entry is no longer "linked" to either the view or the buffer.  Instead it's looked up based upon a combination of the view and the buffer.  The diff view is the primary example - we need to use the difference view to determine the context of the left hand side file when doing a diff.  We need to know that the proper place to analyze it is in the context of the project.  That way when you hover on stuff that's been deleted we know what it is.

Technically we do the same thing with REPL windows and have been doing so for some time.  It's just that we've manually been updating the buffers to say that they're part of the same cohesive set of code.  Now we can effectively say it's the view that's causing these things to be stitched together.  You could imagine some crazy scenario where someone takes a REPL window buffer and pulls it out of context into a different view.  At that point it *should* get a different analysis as it's now in a completely different context.

For the most part though the buffer is still king as it gives us the file name which gives us the analyzer, analysis, and buffer parser to go to.  As long as we're dealing with simple files on disk opened in standard windows there will only be one entry per buffer.

But when we fall out of that scenario we use the view / buffer to get the correct context.  And then it could happen that we could end up with multiple entrys per buffer, and when we need to do an intellisense operation we'll route down and get the right one.

Python/Product/VSInterpreters/InterpreterOptionsService.cs
DinoV: No, but that's the default we've been using.  We should probably add a const somewhere and update all uses of it, but this was just fixing a crash I ran into.

Issue #995, opened 2015-11-23
Can't auto complete methods of value of dict or child class of dict which init by __init__


fan-wenjie:
if input the code in editor
a=[1,2,3]
b=dict(c=a)
b["c"].

a=[1,2,3]
b=dict({'c':a})
b["c"].

class map(dict):
...
a=[1,2,3]
b=map({"c":a})
b["c"].

It show  "No completions"
But if input it in python interactive or 
a=[1,2,3]
b={'c':a}
b["c"].
it can auto complete methods of a

zooba:
I think we're missing handling for the dict constructor. Using `b = {c: a}` should work fine (though this is an area that is _really_ hard to get 100% correct... there are so many dicts that we can very quickly run out of memory if we try to be too thorough).

PR #1227, opened 2016-04-19
Fix #576, #995, and a couple of other small fixes

### LINKED ISSUE #576: IronPython .NET constructors and Intellisense failure
IronPython 2.7.5, PTVS 2.2RC and VS 2013 Desktop

To inherit from a C# class and call a specific base class constructor the IronPython docs say

```
.NET constructors can be overloaded.
 To call a specific base type constructor overload, you need to define a
 __new__  method (not __init__) and call __new__ on the .NET base type.
```

Assuming CSharpTest has multiple constructors and I want to call a specific one this, as recommended, works but gives no Intellisense on c

```
class Child (CSharpTest):
    def __new__(cls):
        instance = CSharpTest.__new__(cls, "one", "two")
        return instance

c = Child()
``` 
  
But if I call the C# constructor directly it also works and I get Intellisense on c.

```
class Child (CSharpTest):
    def __new__(cls):
        instance = CSharpTest("one", "two")
        return instance

c = Child()
``` 
 
It looks like Intellisense can’t track through __new__ to see what the resulting type is. I don't know if this is an IronPython documentation failure (why bother going through  __new__ if calling the constructor directly works) or something that Intellisense doesn't, or can't, do.

Sorry about the formatting  of__ new__ in the last paragraph. It should be double underscore+ "new"+double underscore but I can't see how to override the markdown without putting the text in a code block :(

DinoV:


Python/Product/Analysis/Analyzer/InterpreterScope.cs
```diff
@@ -290,8 +290,11 @@ public InterpreterScope(AnalysisValue av, InterpreterScope outerScope)
             _nodeScopes.Clear();
         }
 
-        public virtual IAnalysisSet AddNodeValue(Node node, IAnalysisSet variable) {
-            return _nodeValues[node] = variable;
+        public virtual IAnalysisSet AddNodeValue(Node node, NodeValueKind kind, IAnalysisSet variable) {
+            NodeValue next;
+            _nodeValues.TryGetValue(node, out next);
+            _nodeValues[node] = new NodeValue(kind, variable, next);
```
zooba: I assume we "never" end up with many entries in here for the same kind, but can we assert that? Could become a memory issue, and certainly could affect Get performance if we end up with 1000s of one kind at the start.

Also, is adding the bottleneck, or is getting? I'd have thought getting, so it may be better to just use a regular list than invent a linked list.

Python/Product/Analysis/Values/FunctionInfo.cs
```diff
@@ -499,23 +499,25 @@ private class StringArrayComparer : IEqualityComparer<string[]> {
             varRef.AddTypes(unit, value);
         }
 
-        public override IAnalysisSet GetMember(Node node, AnalysisUnit unit, string name) {
-            // Must unconditionally call the base implementation of GetMember
-            var ignored = base.GetMember(node, unit, name);
```
zooba: Make sure we haven't lost references as a result of removing this.

Python/Product/Analysis/Values/InstanceInfo.cs
```diff
@@ -149,9 +149,6 @@ internal class InstanceInfo : AnalysisValue, IReferenceableContainer {
         }
 
         public override IAnalysisSet GetMember(Node node, AnalysisUnit unit, string name) {
-            // Must unconditionally call the base implementation of GetMember
-            var ignored = base.GetMember(node, unit, name);
```
zooba: Make sure we haven't lost references as a result of doing this.

Python/Product/Analysis/Values/PartialFunctionInfo.cs
```diff
@@ -102,8 +102,7 @@ class PartialFunctionInfo : AnalysisValue {
         }
 
         public override IAnalysisSet GetMember(Node node, AnalysisUnit unit, string name) {
-            // Must unconditionally call the base implementation of GetMember
-            var res = base.GetMember(node, unit, name);
```
zooba: Make sure we haven't lost references.

Python/Product/Analysis/Values/SuperInfo.cs
```diff
@@ -80,9 +80,6 @@ internal class SuperInfo : AnalysisValue {
         }
 
         public override IAnalysisSet GetMember(Node node, AnalysisUnit unit, string name) {
-            // Must unconditionally call the base implementation of GetMember
-            var ignored = base.GetMember(node, unit, name);
```
zooba: Make sure we haven't lost references.

Python/Product/IronPython/Interpreter/IronPythonInterpreter.cs
```diff
@@ -126,7 +126,8 @@ class IronPythonInterpreter :
         }
 
 #if DEBUG
-        class AssertListenerInitializer {
+        [Serializable]
+        class AssertListenerInitializer : MarshalByRefObject {
```
zooba: Probably shouldn't be both serializable and MarshalByRef. Why change it?

Python/Templates/IronPython/ProjectTemplates/Python/ConsoleProject/ConsoleProject.pyproj
```diff
@@ -24,7 +24,7 @@
     <None Include="$safeprojectname$.py" />
   </ItemGroup>
   <ItemGroup>
-    <InterpreterReference Include="{80659ab7-4d53-4e0c-8588-a766116cbd46}\2.7" />
+    <InterpreterReference Include="IronPython|2.7 32-bit" />
```
zooba: Can we use "-32" as the suffix for 32-bit rather than having a space in there? It _should_ be fine, but I've learned to be wary of spaces in identifiers.

Python/Tests/Analysis/AnalysisTest.cs
```diff
@@ -3761,6 +3761,141 @@ class fob(dict):
...
+
+a = setdict()
+a[42] = 100
+b = a[42]
+
+pass
+";
+            var entry = ProcessText(text);
+            var items = entry.GetTypesByIndex("b", text.IndexOf("pass"));
+            Assert.IsTrue(items.Count() == 0);
```
zooba: `AssertUtil.ContainsExactly(items)` here has the nice side-effect of showing what is in the collection when it does contain something.

Python/Product/IronPython/Interpreter/IronPythonInterpreter.cs
zooba: Misread the diff - I thought this was previously MarshalByRef.

The important thing to ensure here is that the listener is added within the appdomain. My understanding was that Serializable will achieve that (as the listener should be MarshalByRef), but marshalling the initializer will just add the listener in the original appdomain again.

Python/Product/Analysis/Analyzer/InterpreterScope.cs
DinoV: Correct, at most we'll currently end up with 2.  That open happens when we have a subclass of dict.  In NTVS I did something similar but that still had a Dictionary<TKey, TValue> for storing these so the kind was encoded in a struct with the node.  Because this has switched to AnalysisDictionary (presumably to fix a thread safety issue at some point) I didn't want to allocate a reference on all of the gets.  But we're bounded here by the number of NodeValueKind's, and even then the kinds would need to be started being used across many nodes to collide.

Python/Product/Analysis/Values/FunctionInfo.cs
DinoV: The base class implementation doesn't do anything anymore, so at some point these comments just became stale.

Python/Product/IronPython/Interpreter/IronPythonInterpreter.cs
DinoV: I'll just make it serializable, I'm not sure how this ever worked, I just ran into the exception while debugging something else.

Python/Product/Analysis/Analyzer/InterpreterScope.cs
zooba: I'm more worried about hitting a case where we incorrectly keep adding a node value against the same node with the same type, without checking whether it already exists (possibly some time in the future when someone else calls this). Maybe making it a GetOrAddNodeValue() function would address that?

Python/Tests/Analysis/AnalysisTest.cs
```diff
@@ -3860,7 +3860,7 @@ class setdict(dict):
 ";
             var entry = ProcessText(text);
             var items = entry.GetTypesByIndex("b", text.IndexOf("pass"));
-            Assert.IsTrue(items.Count() == 0);
+            AssertUtil.ContainsExactly(items);
```
zooba: Same applies a few more times further down.

Python/Product/Analysis/Analyzer/InterpreterScope.cs
DinoV: That's actually the pattern here...  There's a GetOrMakeNodeValue which everyone uses.  AddNodeValue exists solely for consumption by GetOrMakeNodeValue.  Unfortunately we can't quite get the visibility right on this because derived scopes (comprehension, and is instance) override this and forward it to the real scope.  But if we declare it as protected we can't dispatch to the outer scope one on the base type.

zooba: So is a debug assertion going to help ensure we stick to that pattern?

PR #1274, opened 2016-05-09
Fixes #1272 Ctrl+C, Ctrl+V on empty line pastes in the middle of the line

### LINKED ISSUE #1272: Ctrl+C, Ctrl+V on empty line pastes in the middle of the line
In C#, if you Ctrl+C without a selection it copies the current line. If you then paste that line without a selection, it is inserted preceding the current line.

In Python, when you paste it inserts the line right at the caret. You need to press Home before pasting to get the equivalent of C#.

We should always paste at the start of the line in this scenario.

zooba:
Fixes #1272 Ctrl+C, Ctrl+V on empty line pastes in the middle of the line
Check whether we are pasting a complete line with no selection, and if so move to the start of the line before pasting.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Navigation/EditFilter.cs
+++ b/Python/Product/PythonTools/PythonTools/Navigation/EditFilter.cs
@@ -633,12 +633,20 @@ namespace Microsoft.PythonTools.Language {
             if (pguidCmdGroup == VSConstants.GUID_VSStandardCommandSet97) {
                 switch ((VSConstants.VSStd97CmdID)nCmdID) {
                     case VSConstants.VSStd97CmdID.Paste:
-                        string updated = RemoveReplPrompts(_pyService, Clipboard.GetText(), _textView.Options.GetNewLineCharacter());
+                        string updated = RemoveReplPrompts(
+                            _pyService,
+                            Clipboard.GetText(),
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools/Navigation/EditFilter.cs
```diff
@@ -633,12 +633,20 @@ class NodeEnumerator<T> : IVsEnumNavInfoNodes where T : IVsNavInfoNode {
             if (pguidCmdGroup == VSConstants.GUID_VSStandardCommandSet97) {
                 switch ((VSConstants.VSStd97CmdID)nCmdID) {
                     case VSConstants.VSStd97CmdID.Paste:
-                        string updated = RemoveReplPrompts(_pyService, Clipboard.GetText(), _textView.Options.GetNewLineCharacter());
+                        string updated = RemoveReplPrompts(
```
DinoV: Better would be detecting what the editor is doing here.  There's a "VisualStudioEditorOperationsLineCutCopyClipboardTag" tag which is being added on the copy.  We can detect it and just let the editor handle the paste if it's present.

zooba: Nice, that's much simpler.

int19h:
:+1:

PR #1278, opened 2016-05-10
Fix #753

### LINKED ISSUE #753: Send to Interactive command should work with no text selected
The `Send to Interactive` command (default keyboard shortcut sequence: `CTRL + E, CTRL + E`) should still work, even if no text is selected in the text editor. If no text is selected in the Editor window, the behavior should be to select the line that the cursor is currently on, and send that to the Python Interactive Window, instead of disabling the feature altogether.

image

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell
http://trevorsullivan.net
http://twitter.com/pcgeek86

DinoV:
Adds support for sending to interactive line-by-line

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
+++ b/Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
@@ -67,7 +67,7 @@ namespace Microsoft.PythonTools.Commands {
             if (selection.StreamSelectionSpan.Length > 0) {
                 // Easy, just send the selection to the interactive window.
                 input = activeView.Selection.StreamSelectionSpan.GetText();
-                if (!input.EndsWith("\n")) {
+                if (!input.EndsWith("\n") && !input.EndsWith("\r")) {
                     input += activeView.Options.GetNewLineCharacter();
                 }
                 focusRepl = true;
@@ -96,7 +96,7 @@ namespace Microsoft.PythonTools.Commands {
                     activeView.Caret.PositionChanged += Caret_PositionChanged;
                     activeView.Properties[_executedLastLine] = _executedLastLine;
                 }
-            } else if (repl.InteractiveWindow.CurrentLanguageBuffer.CurrentSnapshot.Length != 0) {
+            } else if ((repl.InteractiveWindow.CurrentLanguageBuffer?.CurrentSnapshot.Length ?? 0) != 0) {
                 // We reached the end of the file but have some text buffered.  Execute it now.
                 input = activeView.Options.GetNewLineCharacter();
             } else {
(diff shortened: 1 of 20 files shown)


Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
```diff
@@ -45,26 +56,76 @@ class SendToReplCommand : Command {
...
-            repl.Show(true);
 
-            var version = analyzer.InterpreterFactory.Configuration.Version.ToLanguageVersion();
+            string input;
+            bool focusRepl = false;
+
+            if (selection.StreamSelectionSpan.Length > 0) {
+                // Easy, just send the selection to the interactive window.
+                input = activeView.Selection.StreamSelectionSpan.GetText();
+                if (!input.EndsWith("\n")) {
```
zooba: Theoretically could just end with `\r`, if someone has chosen that setting.

Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
```diff
@@ -45,26 +56,76 @@ class SendToReplCommand : Command {
...
+                }
 
-            repl.Show(true);
+                if (!moved) {
+                    // There's no where for the caret to go, don't execute the line if 
+                    // we've already executed it.
+                    activeView.Caret.PositionChanged += Caret_PositionChanged;
+                    activeView.Properties[_executedLastLine] = _executedLastLine;
+                }
+            } else if (repl.InteractiveWindow.CurrentLanguageBuffer.CurrentSnapshot.Length != 0) {
```
zooba: `CurrentLanguageBuffer` can be null at weird times - I'd rather not crash because of it. `(repl.InteractiveWindow.CurrentLanguageBuffer?.CurrentSnapshot.Length ?? 0) != 0` should be good enough handling here, and we can Assert in the else block below.

Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
```diff
@@ -45,26 +56,76 @@ class SendToReplCommand : Command {
...
+                // We've hit the end of the current text view and executed everything
+                input = null;
+            }
+
+            if (input != null) {
+                repl.Show(focusRepl);
+
+                InteractiveInputs inputs;
+                if (!repl.InteractiveWindow.Properties.TryGetProperty(typeof(InteractiveInputs), out inputs)) {
+                    repl.InteractiveWindow.Properties[typeof(InteractiveInputs)] = inputs = new InteractiveInputs(repl.InteractiveWindow, _serviceProvider);
```
zooba: IIRC there's a GetOrCreateSingletonProperty` that does this.

zooba:
Few minor comments. If you don't think they're worth fixing, go ahead and merge.

PR #1281, opened 2016-05-11
Fix #1275 and improve editing files w/ odd extensions

### LINKED ISSUE #1275: ObjectDisposedException on buffer change
Received this error in ActivityLog.xml:

```
ObjectDisposedException: Cannot access a disposed object.
at System.Threading.TimerQueueTimer.Change(UInt32 dueTime, UInt32 period)
at System.Threading.Timer.Change(Int32 dueTime, Int32 period)
at Microsoft.PythonTools.Intellisense.BufferParser.Requeue() in PythonTools\Intellisense\BufferParser.cs:line 442
at Microsoft.PythonTools.Intellisense.BufferParser.BufferChangedLowPriority(Object sender, TextContentChangedEventArgs e) in PythonTools\Intellisense\BufferParser.cs:line 431
at Microsoft.VisualStudio.Text.Utilities.GuardedOperations.RaiseEvent[TArgs](Object sender, EventHandler`1 eventHandlers, TArgs args)
```

DinoV:
Fix #1275 - we just need to null out of the buffer parser when we stop monitoring
Also improves some error handling w/ the out of proc analyzer, and improves the experience when opening a file with the Python editor that aren't .py files.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Extensions.cs
+++ b/Python/Product/PythonTools/PythonTools/Extensions.cs
@@ -316,6 +316,10 @@ namespace Microsoft.PythonTools {
             return false;
         }
 
+        internal static AnalysisEntry GetAnalysisEntry(this FileNode node) {
+            return ((PythonProjectNode)node.ProjectMgr).GetAnalyzer().GetAnalysisEntryFromPath(node.Url);
+        }
+
         /// <summary>
         /// Gets the best analyzer for this text view, accounting for things like REPL windows and
         /// difference windows.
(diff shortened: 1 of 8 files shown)


Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -316,6 +316,10 @@ public static class Extensions {
             return false;
         }
 
+        internal static AnalysisEntry GetAnalysisEntry(this FileNode node) {
+            return ((PythonProjectNode)node.ProjectMgr).GetAnalyzer().GetAnalysisEntryFromPath(node.Url);
```
zooba: Is `GetAnalyzer` the one that's guaranteed to never return `null`?

zooba:
@DinoV Just the one query, and you can merge if it's fine.

Python/Product/PythonTools/PythonTools/Extensions.cs
DinoV: Yep!

PR #1294, opened 2016-05-16
Fixes #1285 BreaksExistingExtensionException

### LINKED ISSUE #1285: BreaksExistingExtensionException when attempting to install new Roslyn vsix on Dev 15 Preview
I believe the dependency on "VisualStudio Interactive Copmonents" needs to be bumped to ```[1.0.0.50618,3.0)``` (or at least ```...,2.1)``` :

https://github.com/Microsoft/PTVS/blob/master/Python/Product/PythonTools/source.extension.vsixmanifest

The plan is for the Roslyn version in Dev15 to be 2.0.0.0.

zooba:
Fixes #1285 BreaksExistingExtensionException when attempting to install new Roslyn vsix on Dev 15 Preview

Changes version specifier for interactive components.

### DIFF
--- a/Python/Product/PythonTools/source.extension.vsixmanifest
+++ b/Python/Product/PythonTools/source.extension.vsixmanifest
@@ -21,7 +21,7 @@
   <Dependencies>
     <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" Version="4.6" />
     <Dependency IsRequired="false" Version="[14.0,16.0)" d:Source="Installed" Id="TestWindow.Microsoft.0771d463-d74d-4e95-aac2-39d3c7ec1f97" DisplayName="Test Explorer" />
-    <Dependency d:Source="Installed" Id="1F42C6D0-F876-4AF0-8185-1BEB0A325BB2" DisplayName="VisualStudio Interactive Components" Version="[1.0.0.50618,2.0)" />
+    <Dependency d:Source="Installed" Id="1F42C6D0-F876-4AF0-8185-1BEB0A325BB2" DisplayName="VisualStudio Interactive Components" Version="1.0.0.0" />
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/source.extension.vsixmanifest
```diff
@@ -21,7 +21,7 @@
   <Dependencies>
     <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" Version="4.6" />
     <Dependency IsRequired="false" Version="[14.0,16.0)" d:Source="Installed" Id="TestWindow.Microsoft.0771d463-d74d-4e95-aac2-39d3c7ec1f97" DisplayName="Test Explorer" />
-    <Dependency d:Source="Installed" Id="1F42C6D0-F876-4AF0-8185-1BEB0A325BB2" DisplayName="VisualStudio Interactive Components" Version="[1.0.0.50618,2.0)" />
+    <Dependency d:Source="Installed" Id="1F42C6D0-F876-4AF0-8185-1BEB0A325BB2" DisplayName="VisualStudio Interactive Components" Version="1.0.0.0" />
```
KevinH-MS: Doesn't this need to be a range?  (I thought the version was ```2.0.0.0``` in Dev15)

zooba: If it's not a range, it is treated as a minimum.

zooba: Actually, no, that's special handling just for VS versions. Dependency versions are treated as an exact match.

I'll just remove the version field entirely - we have programmatic handling for breakage between interactive versions already.

KevinH-MS: 👍

PR #1295, opened 2016-05-16
Fixes #268 In Python 2.7 True and False show up in the Autos window. …

### LINKED ISSUE #268: True and False show up in Autos window on 2.7
We show globals if they are referenced, and in 2.7 they are technically globals, so they qualify. But we might want to make a special exception here.

karthiknadig:
…This fix excludes them from showing up.


int19h:
:+1:

PR #1296, opened 2016-05-16
Removes interactive version from vsixmanifest entirely.

zooba:
Removes interactive version from vsixmanifest entirely.
Fixes process object being disposed while still in use on a background thread.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/WebProject/PythonWebLauncher.cs
+++ b/Python/Product/PythonTools/PythonTools/WebProject/PythonWebLauncher.cs
@@ -26,6 +26,7 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
+using System.Threading.Tasks;
 using System.Web;
 using System.Windows.Forms;
 using Microsoft.PythonTools.Debugger;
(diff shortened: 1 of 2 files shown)


Python/Product/PythonTools/PythonTools/WebProject/PythonWebLauncher.cs
```diff
@@ -134,29 +140,41 @@ LaunchConfiguration defaultConfig
...
+
                 OnPortOpenedHandler.CreateHandler(
                     uri.Port,
                     shortCircuitPredicate: shortCircuitPredicate,
                     action: () => {
-                        var web = _serviceProvider.GetService(typeof(SVsWebBrowsingService)) as IVsWebBrowsingService;
-                        if (web == null) {
-                            PythonToolsPackage.OpenWebBrowser(url);
-                            return;
+                        try {
```
DinoV: As long as you're touching this could potentially steal NTVS's code for launching the browser: https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs#L1089  It ends up being a much nicer experience then just opening a browser on our own because the debugger can coordinate these.

zooba: Spent ten minutes on it and it doesn't just drop in nicely, so we can come back to it later. (It also only seems to benefit IE if I'm reading correctly?)

PR #1299, opened 2016-05-17
Fix leaks

### LINKED ISSUE #727: Memory leak in cross-module class instantiation
Given the following two modules, each time `modA` is reanalyzed, memory is leaked until `modB` is reanalyzed.

`modA.py`:
```python
from modB import B

def f():
    x = B()
```

`modB.py`:
```python
class B:
    def __init__(self):
        self.__init__
```

It seems that the call to `modB.B.__init__` from the first analysis sticks around as a dependency of the getattr in `self.__init__`. `modA.f` is a dependency of this call, so it also sticks around, which keeps the entire v1 of `modA`'s analysis in memory.

### LINKED ISSUE #1248: PTVS to do suggest imports in the way C# suggests usings
In Visual Studio in a C# file, if I begin to type a statement like:

`var scn = new SqlConnection(); `

the reference to SqlConnection will be underlined in a red squiggly. If I right-click on it, the context menu contains a Resolve item, that suggests that I need a using statement. If I accept the suggestion, the using statement is added to my program.

If I do the same thing in ptvs, no error (squiggly) is generated and there is no suggestion to import the missing module.

Can this be added?

Similarly, if I have another Python file called Foo containing a class called Foo that I want to use that is part of the same project, and I begin to type:

`f = Foo()`

I'd like ptvs to first note that there is an error (missing ref) and second offer to add 

`from Foo import Foo `

to my import statements.

Can this be done?

DinoV:
Fix #1248 and fix #727


Python/Product/Analysis/Parsing/Ast/PythonAst.cs
```diff
@@ -193,8 +193,8 @@ public sealed class PythonAst : ScopeStatement, ILocationResolver {
             _body.AppendCodeString(res, ast, format);
             res.Append(this.GetExtraVerbatimText(ast));
         }
-
-        #region ILocationResolver Members
+#if FALSE
```
zooba: Remove dead code

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -232,9 +244,9 @@ internal sealed class ProjectEntry : IPythonProjectEntry {
                 string pathPrefix = Path.GetDirectoryName(_filePath) + "\\";
                 var children =
                     from pair in _projectState.ModulesByFilename
-                    // Is the candidate child package in a subdirectory of our package?
+                        // Is the candidate child package in a subdirectory of our package?
```
zooba: Indent?

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -381,6 +381,12 @@ out _analysisProcess
                     }
                     entry.SearchPathEntry = childFile.parent;
                     break;
+                case AP.UnhandledExceptionEvent.Name:
+                    Debug.Fail("Unhandled exception from analyzer");
+                    var exception = (AP.UnhandledExceptionEvent)e.Event;
+
+                    VisualStudio.Shell.ActivityLog.LogError(Strings.ProductTitle, exception.message);
```
zooba: You can use `VsTaskExtensions.ReportUnhandledException` for this (there's a public static non-extension version). This will also Debug.Fail, pop up a dialog, suppress multiple modal reports of the same error, write to the event log, and avoid crashing when the activity log is unavailable.

Python/Product/Analysis/VariableDef.cs
```diff
@@ -85,10 +87,23 @@ abstract class DependentData<TStorageType> where TStorageType : DependencyInfo {
...
+        protected static bool IsVisible(IVersioned accessor, IVersioned declaringScope, IVersioned assigningScope) {
+            return true;
+            /*
+            if (accessor != null && accessor.IsVisible(assigningScope)) {
+                return true;
+            }
+            if (declaringScope != null && declaringScope.IsVisible(assigningScope)) {
+                return true;
+            }
+            return false;*/
```
zooba: Unused code?

zooba:
+1

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
DinoV: The problem is it's based around taken an exceptions and tasks and not a string object, and we had to turn the exception into a string to send it across the wire.

---
In reply to: [63613169](https://github.com/Microsoft/PTVS/pull/1299#discussion_r63613169) [](ancestors = 63613169)

Python/Product/Analysis/VariableDef.cs
DinoV: This is actually called...  We need to figure out what policy we want to enforce here.   This is what NTVS did, but Python will be different.  I'm leaving this here as food for thought

---
In reply to: [63613180](https://github.com/Microsoft/PTVS/pull/1299#discussion_r63613180) [](ancestors = 63613180)

PR #1301, opened 2016-05-17
Fixes #119 Changes to values in the Debug Interactive will refresh ...

### LINKED ISSUE #119: Changes to values in Debug Interactive not reflected on Autos or Locals window
I have a value 'guess' that I alter in the Debug Interactive Window. It is changed, but this change is not reflected in the locals, autos, or watch window. 

 It seems Immediate window changes aren't perfect either. The Autos and Watch windows are updated but locals is not. 
 Immediate Window Changes

karthiknadig:
Changes to values in the Debug Interactive will refresh locals, autos, and watch after every expression eval in REPL


Python/Product/PythonTools/PythonTools.csproj
```diff
@@ -104,6 +104,9 @@
     <ProjectReference2 Include="$(BuildRoot)Python\Product\ProjectWizards\ProjectWizards.csproj" />
     <ProjectReference2 Include="$(BuildRoot)Python\Product\TestAdapter\TestAdapter.csproj" />
     <ProjectReference2 Include="$(BuildRoot)Python\Product\XamlDesignerSupport\XamlDesignerSupport.csproj" />
+    <Reference Include="Microsoft.VisualStudio.Debugger.Interop.14.0, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <EmbedInteropTypes>True</EmbedInteropTypes>
```
karthiknadig: Is it Ok if EmbedInteropTypes is True?

zooba: For the debugger interop assemblies, I think so. It's generally fairly obvious if it isn't going to work.

Check with a Dev15 release whether the `Version=14.0.0.0` part of the name needs to be `Version=$(VSTarget).0.0`

Python/Product/PythonTools/PythonTools/Debugger/DTEDebuggerExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Microsoft.PythonTools.Debugger.DebugEngine;
```
zooba: Need a copyright header. We also sort using with `System.*` first.

Python/Product/PythonTools/PythonTools/Repl/PythonInteractiveEvaluator.cs
```diff
@@ -43,6 +43,7 @@
 using Microsoft.VisualStudio.Utilities;
 using Microsoft.VisualStudioTools;
 using Task = System.Threading.Tasks.Task;
+using Microsoft.PythonTools.Debugger;
```
DinoV: Nit on formatting: usings should be sorted

Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -1481,5 +1481,9 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
             }
             return null;
         }
+
+        internal async void RefreshThreadFrames(long threadId) {
+            await _process.GetThreadFramesAsync(threadId);
+        }
```
DinoV: Because this is async void it's going to be a fire and forget call...  So while we await on the threads coming back here, we don't await on them coming back DTEDebuggerExtensions.  Seems like we need this to return a Task and we need to do a .Wait() over in DTEDebuggerExtensions, and we might need to do a .ConfigureAwait(false) here to make sure we don't deadlock

zooba: At the very least, there should be a `HandleAllExceptions(...).DoNotWait()` on the task (back over in PythonTools project) in case things blow up.

Python/Product/PythonTools/PythonTools/Debugger/DTEDebuggerExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.PythonTools.Debugger
+{
+    public static class DTEDebuggerExtensions
+    {
+        /// <summary> 
+        /// Forces debugger to refresh its variable views (Locals, Autos etc) by re-querying the debug engine. 
+        /// </summary> 
+        /// <param name="debugger"></param> 
+        public static void RefreshVariableViews(this EnvDTE.Debugger debugger) {
+            AD7Engine engine = AD7Engine.GetEngines().SingleOrDefault(target => target.Process != null && target.Process.Id == debugger.CurrentProcess.ProcessID);
```
zooba: `SingleOrDefault` can crash in many interesting ways, as can DTE in general. It would be ideal to avoid DTE completely if possible.

Python/Product/PythonTools/PythonTools/Debugger/DTEDebuggerExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.PythonTools.Debugger
+{
```
zooba: We do opening braces on same line for everything. Otherwise, default C# formatting settings.

Python/Product/PythonTools/PythonTools/Debugger/DTEDebuggerExtensions.cs
int19h: Unfortunately, DTE is the only way to reach the debugger from the outside (i.e. not from within the engine).

Python/Product/PythonTools/PythonTools/Debugger/DTEDebuggerExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// Forces debugger to refresh its variable views (Locals, Autos etc) by re-querying the debug engine. 
+        /// </summary> 
+        /// <param name="debugger"></param> 
+        public static void RefreshVariableViews(this EnvDTE.Debugger debugger) {
+            AD7Engine engine = AD7Engine.GetEngines().SingleOrDefault(target => target.Process != null && target.Process.Id == debugger.CurrentProcess.ProcessID);
+            if (engine != null) {
+                engine.RefreshThreadFrames(debugger.CurrentThread.ID);
+            }
+
+            var vsDebugger = (IDebugRefreshNotification140)ServiceProvider.GlobalProvider.GetService(typeof(SVsShellDebugger));
```
zooba: Using `ServiceProvider.GlobalProvider` makes it much harder to run tests, so we avoid it as much as possible. Perhaps make this an extension method on IServiceProvider instead and get the other services in here?

karthiknadig: Can I make it a part of the Extensions.cs where most of the IServiceProvider extensions are defined?

zooba: Yep, that's fine.

Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -1482,8 +1483,8 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
             return null;
         }
 
-        internal async void RefreshThreadFrames(long threadId) {
-            await _process.GetThreadFramesAsync(threadId);
+        internal async System.Threading.Tasks.Task RefreshThreadFrames(long threadId) {
+            await _process.GetThreadFramesAsync(threadId).ConfigureAwait(false);
```
zooba: `ConfigureAwait(false)` here has no impact as the outer caller will need to do it as well.

In this case, just do `return _process.GetThreadFramesAsync(threadId)` and remove the `async` from the declaration.

Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -737,6 +739,24 @@ public static class Extensions {
...
         }
 
+        internal static SVsShellDebugger GetShellDebugger(this IServiceProvider provider) {
+            return (SVsShellDebugger)provider.GetService(typeof(SVsShellDebugger));
+        }
+
+        internal static async System.Threading.Tasks.Task RefreshVariableViews(this IServiceProvider serviceProvider) {
+            EnvDTE.Debugger debugger = serviceProvider.GetDTE().Debugger;
+            foreach (AD7Engine engine in AD7Engine.GetEngines()) {
+                if (engine.Process != null && engine.Process.Id == debugger.CurrentProcess.ProcessID) {
```
zooba: @int19h - any chance we could match this condition multiple times? Should we break out of the loop once we hit this once?

Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -737,6 +739,24 @@ public static class Extensions {
...
+                    await engine.RefreshThreadFrames(debugger.CurrentThread.ID);
```
zooba: What happens in the parallel watch window? I suspect that we can only modify values on the current thread anyway, but worth checking in case we need to refresh all threads.

Python/Product/PythonTools/PythonTools/Extensions.cs
karthiknadig: It should break after the first match. I will add a break.

int19h: There might be more than one match if we're attached to a remote process that happens to have the same ID. But in this case the first match might not be the right one.

To get the exact instance, it needs to be comparing the transport qualifier (`Process2.TransportQualifier`) vs `PythonRemoteProcess.Uri` in addition to process ID. 

Alternatively, we can just refresh all active Python engines.

Python/Product/PythonTools/PythonTools/Extensions.cs
karthiknadig: It seems to detect that there was a change, but is unable to evaluate it. It gives a button next to the value to refresh it.

Python/Product/PythonTools/PythonTools/Extensions.cs
int19h: To be more specific about the way matching should be done, given a single `EnvDTE.Process2`:

- Check for `Process2.Transport.ID`. If it is our `PythonRemoteDebugPortSupplier`, then it's our remote process. Otherwise, it's either our local process, or a non-Python process.

  * If it's a remote process, find the engine with a matching process ID, for which `AD7Engine.PythonProcess` is `PythonRemoteProcess`, and its `Uri` matches `Process2.TransportQualifier`.

  * Otherwise, find the engine with a matching process ID, for which `AD7Engine.PythonProcess` is _not_ `PythonRemoteProcess`

Python/Product/PythonTools/PythonTools/Extensions.cs
zooba: That sounds fine to me - as long as there's some visual feedback and an action for the user to take.

zooba:
@dinov Have any other concerns? Go ahead and merge if you don't.

Python/Product/Debugger/Debugger/DebugEngine/DebuggerConstants.cs
```diff
@@ -65,6 +65,11 @@ class DebuggerConstants {
             get { return _guidLanguageDjangoTemplate; }
         }
 
+        static private Guid _guidLocalPortSupplier = new Guid("708C1ECA-FF48-11D2-904F-00C04FA302A1");
+        static public Guid guidLocalPortSupplier {
```
int19h: @zooba, are all the new C# 6 features available for use? If so, this can be simplified to ` ... { get } = new Guid(...);`.

int19h:
:+1:

PR #1306, opened 2016-05-18
Replaces Azure Remote Debug functionality with opening the help page.

zooba:
Replaces Azure Remote Debug functionality with opening the help page.


zooba:
Merging because it's the same as #1307.

PR #1311, opened 2016-05-20
Fixes #1309 Removed version number from debug repl initialization mes…

### LINKED ISSUE #1309: Debug Interactive window should not claim to be 2.7
When you open debug interactive it displays:

```
Debug 2.7 interactive window [PTVS 3.0.1000.00-14.0]
Type $help for a list of commands.
>>> 
```

We should prevent the version number from appearing in this case.

karthiknadig:
Removed version number from debug repl initialization message


karthiknadig:
This is how it looks with the change:
```
Debug interactive window [PTVS 3.0.1000.00-14.0]
Type $help for a list of commands.
>>>
```

int19h:
:+1:

PR #1319, opened 2016-05-24
Fixes #107 Added code to show data tip for  __foo variables and members.

### LINKED ISSUE #107: Can't inspect private variables in debugger
When the debugger is running I can't inspect, edit, or access private variables.  

I imagine this is because the debugger isn't doing the transform to add '_ClassName' to the variable?  E.g.

```Python
class MyClass:
    def __init__(self):
         self.__private_var = 5
    def public_method(self):
        print 'public method', self.__private_var

obj = MyClass()
# But if you inspect the object, it's not hard to get to
dir(obj)
['_MyClass__private_var', '__doc__', '__init__', '__module__', 'public_method']
```

karthiknadig:


Python/Product/Analysis/Parsing/Ast/MemberExpression.cs
```diff
@@ -76,7 +76,11 @@ public sealed class MemberExpression : Expression {
             res.Append('.');
             if (!this.IsIncompleteNode(ast)) {
                 res.Append(this.GetSecondWhiteSpaceDefaultNull(ast));
-                res.Append(this.GetVerbatimImage(ast) ?? _name);
+                if (format.UseVerbatimImage) {
+                    res.Append(this.GetVerbatimImage(ast) ?? _name);
```
int19h: @DinoV, can you take a look? I'm not sure what the right thing to do here, actually... it seems like we're being somewhat inconsistent already, in that, if verbatim data is not available (regardless of `format`), we will use the mangled name. So if you parse `class Foo: __bar` and then do `ToCodeString` on it, you'll get `class Foo: __Foo_bar`. This fix essentially relies on that behavior, except it changes it to be consistent with `format`. Is that the best approach?

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -1960,6 +1974,66 @@ class ExprWalker : PythonWalker {
             return new AnalysisVariable(type, location);
         }
 
+        internal static string ExpressionForDataTip(IServiceProvider serviceProvider, ITextView view, SnapshotSpan span) {
```
int19h: I'd rather have this be async, and do the async-to-sync adaptation at the point where it is called from the implementation of IVs* interface. Basically, as close to the boundary as possible, so most of the code is async, and gets all the benefits of that.

Also, I don't recall if we already have a generic helper to take a task and wait for it synchronously in a manner that prevents deadlocking. If we don't, you might want to port this over:

https://github.com/Microsoft/RTVS/blob/db6353d5ebdd12bcfa44b11b4a86bc8175d20777/src/Debugger/Impl/TaskExtensions.cs#L26

zooba: If you _do_ port that over, it needs rewriting to avoid using `ThreadHelper`. We need to pass around `IServiceProvider` instances and use those so that we can mock everything out correctly for tests.

.gitignore
```diff
@@ -20,3 +20,5 @@ Python/Product/Django/Templates/Projects/*/*RoleConfiguration.mht
 Python/Product/Django/Templates/Projects/*/ConfigureCloudService.ps1
 Python/Product/Django/Templates/Projects/*/ps.cmd
 __pycache__/
+/Python/PythonTools.VC.VC.opendb
+/Python/PythonTools.VC.db
```
zooba: Change these to `*.vc.db` and `*.vs.opendb` patterns.

Python/Product/Analysis/Parsing/CodeFormattingOptions.cs
```diff
@@ -236,6 +236,11 @@ public sealed class CodeFormattingOptions {
         [CodeFormattingDefaultValue(true)]
         public bool BreakMultipleStatementsPerLine { get; set; }
 
+        [CodeFormattingExample("foo.__bar", "foo._A__bar")]
+        [CodeFormattingCategory(CodeFormattingCategory.Statements)]
+        [CodeFormattingDescription("UseVerbatimImageShort", "UseVerbatimImageLong")]
+        [CodeFormattingDefaultValue(true)]
+        public bool UseVerbatimImage { get; set; }
```
zooba: This probably shouldn't be exposed to users. Not sure whether we support any attribute for that though.

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
zooba: Also, I'd have thought that all of our threading was either free-threaded or UI-bound - there shouldn't be a need for a new synchronization context. But @DinoV may disagree - I haven't quite caught up with the out-of-proc analyzer threading model.

Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -911,5 +918,32 @@ public static class Extensions {
             }
             return prettyPrinted.ToString().Trim();
         }
+
+        public static void RunSynchronouslyOnUIThread(this System.Threading.Tasks.Task task, IServiceProvider serviceProvider, [CallerMemberNameAttribute] string callerMemberName = "") {
+            if (!Dispatcher.CurrentDispatcher.CheckAccess()) {
+                string message = string.Format("{0} must be called on a UI thread.", callerMemberName);
+                Debug.Fail(message);
+                return;
```
zooba: Is this actually synchronous if we can only call it from the thread it will run on?

Does this differ enough from our `UIThread` class that you can't use that? Should we extend/update that class?

zooba: Actually, can you just explain why it's different from simply awaiting the Task? The default awaiter marshals back to the calling thread.

Python/Product/Analysis/Parsing/CodeFormattingOptions.cs
```diff
@@ -236,6 +236,11 @@ public sealed class CodeFormattingOptions {
         [CodeFormattingDefaultValue(true)]
         public bool BreakMultipleStatementsPerLine { get; set; }
 
+        [CodeFormattingExample("foo.__bar", "foo._A__bar")]
```
DinoV: Does this end up showing up in tools->options->text editor->python->formatting ?  We don't really want it to be there, but we probably don't need all of the attributes if it doesn't.

Python/Product/PythonTools/PythonTools/Extensions.cs
int19h: This is a replacement for `task.Result` rather than `await` - you'd use it where there's an async API, and you need to call it from someplace that has to be sync. In our case it's the data tip service, because it has to implement the VS interface for that stuff, which is sync.

Python/Product/Analysis/Parsing/CodeFormattingOptions.cs
karthiknadig: It does not show up in there.

Python/Product/Analysis/Parsing/Ast/MemberExpression.cs
DinoV: Round tripping is really only going to work in a reasonable way if we parse in verbatim mode, and in that case we'll always have the right verbatim image available if it differs from the real name.

Elsewhere where we deal with this we first do:

            string privatePrefix = GetPrivatePrefixClassName(scope);
            var ast = GetAstFromText(exprText, privatePrefix);

Does that work in this case too?

int19h: We don't actually want to round-trip here. What's needed is to get the actual expression with mangled identifiers (as a string, to send to evaluator), given a span in the original source.

Python/Product/PythonTools/PythonTools/Extensions.cs
zooba: So is the difference from `WaitAndUnwrapExceptions` that is _does_ allow other tasks to interrupt the one being waited? Or that it _doesn't_?

int19h: Yes. Essentially, the difference is that `WaitAndUnwrapExceptions` blocks the thread on which it's called, so if the task that it is waiting on requires posting a continuation to that thread at any point, everything deadlocks. Whereas this VS task factory stuff does something (I'm not sure exactly what it is, but I believe it's the absolute minimum that's needed to process incoming continuations) that lets this all work without deadlocks.

zooba: So in that case, we really want this in VsTaskExtensions as an alternative to `WaitAndUnwrapExceptions` that just waits directly when off the UI thread and does this when on the UI thread. Then anywhere we can pass in a service provider we should, so that we get the right behavior.

Python/Product/Analysis/Parsing/Ast/MemberExpression.cs
DinoV: But we are round tripping :)  I think getting the private prefix first, and then getting the AST with the private prefix will give us the expression w/ the mangled identifiers.  And it's what we're already doing elsewhere.

---
In reply to: [64830525](https://github.com/Microsoft/PTVS/pull/1319#discussion_r64830525) [](ancestors = 64830525)

PR #1323, opened 2016-05-25
Fixes #1251 No completions in virtual env REPL

### LINKED ISSUE #1251: No completions in virtual env REPL
Dev15 Preview 2, Python 2.7, in a virtual env REPL:
```
text = "hello"
tex <ctrl-space>
```
Nothing comes up. I haven't changed the default options.

Edit: While it originally seemed to on global and virtual env, now it only reproduce for virtual env.

zooba:
Fixes #1251 No completions in virtual env REPL
Enables getting the project associated with an interpreter ID so that we can flow settings through to the interactive window.

Also fixes a potential crash on VS shutdown.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Extensions.cs
+++ b/Python/Product/PythonTools/PythonTools/Extensions.cs
@@ -290,8 +290,15 @@ namespace Microsoft.PythonTools {
 
             string path = buffer.GetFilePath();
             if (path != null) {
-                var docTable = provider.GetService(typeof(SVsRunningDocumentTable)) as IVsRunningDocumentTable4;
-                var cookie = docTable.GetDocumentCookie(path);
+                var docTable = (IVsRunningDocumentTable4)provider.GetService(typeof(SVsRunningDocumentTable));
+                var cookie = VSConstants.VSCOOKIE_NIL;
+                try {
+                    cookie = docTable.GetDocumentCookie(path);
+                } catch (ArgumentException) {
+                    // Exception may be raised while VS is shutting down
+                    entry = null;
+                    return false;
+                }
                 VsProjectAnalyzer analyzer = null;
                 if (cookie != VSConstants.VSCOOKIE_NIL) {
                     IVsHierarchy hierarchy;
@@ -486,7 +493,14 @@ namespace Microsoft.PythonTools {
                 IVsHierarchy hierarchy;
                 uint itemid;
                 docTable.GetDocumentHierarchyItem(cookie, out hierarchy, out itemid);
-                return hierarchy.GetProject()?.GetPythonProject();
+                var project = hierarchy.GetProject();
+                if (project != null) {
+                    return project.GetPythonProject();
+                }
+
+                object projectObj;
+                ErrorHandler.ThrowOnFailure(hierarchy.GetProperty(itemid, (int)__VSHPROPID.VSHPROPID_ExtObject, out projectObj));
+                return (projectObj as EnvDTE.Project)?.GetPythonProject();
             }
             return null;
         }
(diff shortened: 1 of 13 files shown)


Python/Product/VSInterpreters/CPythonInterpreterFactoryProvider.cs
```diff
@@ -180,6 +181,13 @@ class CPythonInterpreterFactoryProvider : IPythonInterpreterFactoryProvider {
                     version = new Version(2, 7);
                 }
 
+                try {
+                    var langVer = version.ToLanguageVersion();
+                } catch (InvalidOperationException) {
+                    // Version is not currently supported
+                    return;
+                }
```
zooba: This is needed to prevent a crash when you have Python 3.6 installed (which is currently not supported, but gets through most of the later checks).

Python/Tests/Utilities.Python.Analysis/MockInterpreterOptionsService.cs
```diff
@@ -165,5 +165,9 @@ public class MockInterpreterOptionsService : IInterpreterOptionsService, IInterp
         public string AddConfigurableInterpreter(string name, InterpreterConfiguration config) {
             throw new NotImplementedException();
         }
+
+        public object GetProperty(string id, string propName) {
+            throw new NotImplementedException();
```
DinoV: return null?

Python/Product/PythonTools/PythonTools/Repl/PythonInteractiveEvaluator.cs
```diff
@@ -134,7 +134,17 @@ partial class PythonInteractiveEvaluator :
                 if (config == null) {
                     _analyzer = _serviceProvider.GetPythonToolsService().DefaultAnalyzer;
                 } else {
-                    _analyzer = new VsProjectAnalyzer(_serviceProvider, interpreterService.FindInterpreter(config.Interpreter.Id));
+                    Build.Evaluation.Project projectFile = null;
+                    var moniker = ProjectMoniker ??
```
DinoV: Can everything in here up to GetProjectFromFile(moniker) be factored into a method so it can be used both here and down below?

Python/Tests/Utilities.Python.Analysis/MockInterpreterOptionsService.cs
zooba: Everything else in this mock throws.

Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -486,7 +493,14 @@ public static class Extensions {
                 IVsHierarchy hierarchy;
                 uint itemid;
                 docTable.GetDocumentHierarchyItem(cookie, out hierarchy, out itemid);
-                return hierarchy.GetProject()?.GetPythonProject();
+                var project = hierarchy.GetProject();
+                if (project != null) {
+                    return project.GetPythonProject();
+                }
+
+                object projectObj;
```
DinoV: GetProject is already getting ExtObject from the hierarchy...  So this code just seems to be completely redundant?

Python/Product/PythonTools/PythonTools/Repl/PythonInteractiveEvaluator.cs
```diff
@@ -134,7 +134,17 @@ partial class PythonInteractiveEvaluator :
...
+                        interpreterService.GetProperty(config.Interpreter.Id, "ProjectMoniker") as string;
```
int19h: I'd prefer to have a named constant for this literal, since it's used in more than one place.

int19h:
:+1:

Python/Product/PythonTools/PythonTools/Repl/PythonInteractiveEvaluator.cs
zooba: Brought it back down to 2 places, and if you can suggest a good place for it that doesn't involve exposing knowledge about a particular interpreter configuration provider (and also doesn't imply that only a certain provider can use it, or that all providers need to support it), let me know and I'll make the change after I get back.

PR #1335, opened 2016-06-04
Allow to modify PythonProjectNodeProperties.CommandLineArguments

### LINKED ISSUE #1078: Setting project properties from VS extension fails
I’m working on a free visual studio extension which provides a nice UI to edit command line arguments.
Sadly our extension does not support PTVS projects. We couldn't figure out how to set the project properties.

Is there an API which our extension could use for PTVS projects?

Right now we set the arguments with this [code fragment](https://github.com/MBulli/SmartCommandlineArgs/blob/master/SmartCmdArgs/SmartCmdArgs/CmdArgsPackage.cs#L153) which fails for PTVS projects:

```csharp
foreach (EnvDTE.Configuration config in project.ConfigurationManager)
{
	if (config.Properties != null)
	{
		foreach (EnvDTE.Property prop in config.Properties)
		{
            // CommandLineArguments = Python 
			if (prop.Name == "CommandLineArguments")
			{
				prop.Value = prjCmdArgs;
				break;
			}
		}
	}
}
```

The code doens’t work because `config.Properties` is null. I understand that this behavior is related to issue #86. Please note that we don’t want to set the arguments config-specifc.

I then tried to set the values directly by skipping ConfigurationManager like this:

```csharp
foreach (EnvDTE.Property prop in project.Properties)
{
	// CommandLineArguments = Python 
	if (prop.Name == "CommandLineArguments")
	{
		prop.Value = prjCmdArgs; // Crash: No setter method found
		break;
	}
}
```

But the assignment of `prop.Value` throws an `ArgumentException` "Property set method not found."

```
   at System.Reflection.RuntimePropertyInfo.SetValue(Object obj, Object value, BindingFlags invokeAttr, Binder binder, Object[] index, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.SetValue(Object obj, Object value, Object[] index)
   at Microsoft.VisualStudioTools.Project.Automation.OAProperty.<>c__DisplayClass24_0.<set_Value>b__0()
   at Microsoft.VisualStudioTools.UIThread.Invoke(Action action)
   at Microsoft.VisualStudioTools.Project.Automation.OAProperty.set_Value(Object value)
   at SmartCmdArgs.CmdArgsPackage.UpdateProjectConfiguration() in CmdArgsPackage.cs:line 170
   at SmartCmdArgs.CmdArgsPackage.SolutionEvents_Opened() in CmdArgsPackage.cs:line 190
   at EnvDTE._dispSolutionEvents_SinkHelper.Opened()
```
InnerException is null.

All test are done with VS2015 Update 1 and PTVS 2.2.2.

Cellule:
Added a set method on the PythonProjectNodeProperties.CommandLineArguments to allow extensions to modify this property.

I wanted to add support to Python Projects for https://github.com/MBulli/SmartCommandlineArgs/issues/9.
However, I couldn't find an easy way to modify the CommandLineArguments property without messing with your internal objects.

I don't know if there was a good reason not to put the `set` method, but I figured it might just be because there was no need before.

Closes #1078 

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/ptvs/1335)

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonProjectNodeProperties.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonProjectNodeProperties.cs
@@ -53,6 +53,9 @@ namespace Microsoft.PythonTools.Project {
             get {
                 return this.Node.ProjectMgr.GetProjectProperty(CommonConstants.CommandLineArguments, true);
             }
+            set {
+                this.Node.ProjectMgr.SetProjectProperty(CommonConstants.CommandLineArguments, value);
+            }
         }
 
         /// <summary>


MBulli:
Related issue #1078

PR #1342, opened 2016-06-15
Fix #1255, true division isn't enable when parsing 3.x code

### LINKED ISSUE #1255: IntelliSense does not recognise pathlib objects constructed by '/'
Example:
```python
from pathlib import Path

a = Path('.').absolute()
b = a / 'src'
```
When I do as shown in the snippet above, IntelliSense would not recognise object 'b' as a 'Path' object.

DinoV:


Python/Product/Analysis/Parsing/Parser.cs
```diff
@@ -76,14 +76,14 @@ public class Parser : IDisposable { // TODO: remove IDisposable
             _langVersion = langVersion;
             _verbatim = verbatim;
             _bindReferences = bindRefs;
-            
+
```
DinoV: This is the actual fix, _languageFeatures needs to be assigned before we reset it

Python/Product/Analysis/Values/SequenceBuiltinInstanceInfo.cs
```diff
@@ -87,7 +87,10 @@ public SequenceBuiltinInstanceInfo(BuiltinClassInfo klass, bool sequenceOfSelf,
                     }
                     break;
             }
-            return res ?? base.BinaryOperation(node, unit, operation, rhs);
+            if (res.Count == 0) {
```
DinoV: This fixes the reverse operation, res is never null here

PR #1346, opened 2016-06-16
Adds triple-quotes around exception message so that copy-pasting into…

zooba:
Adds triple-quotes around exception message so that copy-pasting into github looks nicer.


huguesv:
👍

PR #1365, opened 2016-06-22
Update install for VS 15

zooba:
Changes from fopx installer to swix for use with VS "15".
This also requires updating project files to look for targets slightly differently, which is why there are changes to the core product.


Python/Tests/Core/ProjectUpgradeTests.cs
```diff
@@ -291,79 +291,57 @@ out factoryGuid
...
             var factory = new PythonProjectFactory(null);
             var sp = new MockServiceProvider();
-            var shell = new MockVsShell();
-
             sp.Services[typeof(SVsQueryEditQuerySave).GUID] = null;
             sp.Services[typeof(SVsActivityLog).GUID] = new MockActivityLog();
-            sp.Services[typeof(SVsShell).GUID] = shell;
             factory.Site = sp;
 
-            var projectFile = TestData.GetPath(Path.Combine("TestData", "ProjectUpgrade", "WebProjectType.pyproj"));
```
huguesv: WebProjectType.pyproj is no longer used?  If so you could delete the file.

Python/Setup/swix/file_associations.swr
```diff
@@ -0,0 +1,38 @@
...
+    
+    vs.fileAssociation
+        Extension=.pyw
+        ProgId=VisualStudio.py.[InstanceId]
+        DefaultProgramRegistrationPath=$(VSDefaultProgramPath)
+    
+    vs.fileAssociation
+        Extension=.pyproj
+        ProgId=VisualStudio.Launcher.pyproj.[InstanceId]
+        PerceivedType=text
```
huguesv: what does this mean?

Python/Setup/swix/product.json
```diff
@@ -0,0 +1,65 @@
...
+        },
+        "microsoft.PythonTools.IronPython": {
+          "version": "[3.0.0.0, 4.0)"
+        },
+        "microsoft.PythonTools.IronPython.Templates": {
+          "version": "[3.0.0.0, 4.0)"
+        },
+        "microsoft.PythonTools.ML": {
+          "version": "[3.0.0.0, 4.0)"
+        }
```
huguesv: Uwp and Uwp.Templates are missing, is it intentional?

huguesv:
:shipit:

Python/Setup/swix/file_associations.swr
zooba: It means that the items in `HKCR\SystemFileAssociations\text` also apply to this file type, which is wrong in this case. Updating this to be `ContentType=application/xml` as it should be (and moving `PercievedType=text` onto the source file types).

Python/Setup/swix/product.json
zooba: No, fixed.

PR #1369, opened 2016-06-22
Fixes #1365 _ReplOutput has no attribute 'closed'

zooba:
Fixes #1365 _ReplOutput has no attribute 'closed'
Adds closed attribute.
Also makes encoding property into a class attribute.


huguesv:
👍

PR #1396, opened 2016-07-01
Adds component manifests and better structures swix files.

zooba:


Python/Setup/swix/Packages/Microsoft.PythonTools.Profiling.Vsix.swixproj
```diff
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
mousetraps: Why did you rename it to *.vsix.swixproj? I recall running into a strange issue where I was forced to append "vsix", but it screwed up other things down the line (failed to copy the vsix properly). I managed to get around it by moving the MicroBuild.Core targets file declaration towards the end.

Python/Setup/swix/swix.generate.targets
```diff
@@ -49,10 +49,6 @@ package name=$([System.IO.Path]::GetFileNameWithoutExtension($(OutputName))) ver
 <WriteLinesToFile File="$(SwrFile).tmp" Lines="vs.dependencies
 $(_Dependencies)
 " Overwrite="no" Encoding="ASCII" Condition="$(_Dependencies) != ''" />
-
-    <ItemGroup>
-      <Package Include="$(SwrFile)" />
-    </ItemGroup>
   </Target>
 
   <Target Name="_GenerateSwrFromSources" Condition="@(SwrSource) != ''">
```
mousetraps: I find the additional level of abstraction difficult to read through and verify in this case - idk if it'll work for you, but i'll at least send a link to the PR when I'm done implementing the VSIX stuff for NTVS

Python/Setup/swix/Packages/dirs.proj
```diff
@@ -0,0 +1,36 @@
...
+    <ProjectFile Include="Microsoft.PythonTools.IronPython.Templates.Vsix.swixproj" Condition="'$(IncludeIronPython)' == 'True'"/>
+    <ProjectFile Include="Microsoft.PythonTools.ML.Vsix.swixproj"/>
+    <ProjectFile Include="Microsoft.PythonTools.Profiling.Vsix.swixproj"/>
+    <ProjectFile Include="Microsoft.PythonTools.Samples.Vsix.swixproj"/>
+    <ProjectFile Include="Microsoft.PythonTools.Uwp.Vsix.swixproj" Condition="'$(IncludeUwp)' == 'True'"/>
+    <ProjectFile Include="Microsoft.PythonTools.Uwp.Templates.Vsix.swixproj" Condition="'$(IncludeUwp)' == 'True'"/>
+    <ProjectFile Include="Microsoft.PythonTools.Web.Templates.Vsix.swixproj"/>
+    <ProjectFile Include="CPython2x64.Msi.swixproj"/>
+    <ProjectFile Include="CPython3x64.Exe.swixproj"/>
+  </ItemGroup>
```
mousetraps: Did you verify registry keys in the original MSI were properly accounted for?

Python/Setup/swix/Packages/dirs.proj
```diff
@@ -0,0 +1,36 @@
...
+      <Properties>SignType=none</Properties>
+    </ProjectFile>
+  </ItemDefinitionGroup>
+
+  <PropertyGroup>
+    <StopOnFirstFailure>true</StopOnFirstFailure>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectFile Include="Microsoft.PythonTools.Core.Vsix.swixproj"/>
```
mousetraps: No targets file swixproj?

Python/Setup/swix/Packages/Microsoft.PythonTools.Profiling.Vsix.swixproj
zooba: Just to make it clear (to me at least) that this package is the vsix file. e.g. `Microsoft.PythonTools.Profiling` is a "component" (that can be optionally enabled/disabled by users; `<OutputType>manifest</OutputType>` and `vs.package.type component`), while `Microsoft.PythonTools.Profiling.Vsix` is a "VSIX package" (`<OutputType>vsix</OutputType>`). Other packages tend to have `.Exe` or `Msi` at the end when they solely exist for that payload.

(The OutputType and SwixProjectType properties are set in the SetupProjectBefore/After.settings files, since I arranged the projects to be in separate dirs.)

Python/Setup/swix/swix.generate.targets
zooba: If you write out the list of every file into a static swr (and keep it up to date over time) then you don't need any of this. I was too lazy to do that :)

---
In reply to: [69642273](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69642273) [](ancestors = 69642273)

Python/Setup/swix/Packages/dirs.proj
zooba: Yep. There's a static file with file associations defined, and the only other one we had was a version number that is no longer needed.

---
In reply to: [69642494](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69642494) [](ancestors = 69642494)

Python/Setup/swix/Microsoft.PythonTools.vsmanproj
```diff
@@ -10,13 +10,11 @@
...
 
-    <TargetName>Microsoft.PythonTools</TargetName>
-    <OutputPath>$(SetupOutputPath)swix\</OutputPath>
-    <OutDir>$(OutputPath)</OutDir>
+    <TargetName>$(MSBuildProjectName)</TargetName>
   </PropertyGroup>
 
   <ItemGroup>
-    <MergeManifest Include="$(OutputPath)*.json" />
+    <MergeManifest Include="$(OutputPath)Microsoft.*.json" />
```
mousetraps: Are you intentionally including externals here?

Python/Setup/swix/Packages/dirs.proj
zooba: No, they are part of this one, the Django one, and the Uwp one.

---
In reply to: [69642498](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69642498) [](ancestors = 69642498)

Python/Setup/swix/Microsoft.PythonTools.vsmanproj
zooba: Yes. There's a separate vsmanproj for those, so that we can update the externals manifest without having to change the core PTVS manifest.

---
In reply to: [69643035](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69643035) [](ancestors = 69643035)

zooba: But since we currently have Microsoft.VisualStudio.Workload.Python in this repo, it can't be `Microsoft.PythonTools.*.json` like I intended. But maybe we'll move the workload back into the VS repo at some point... not sure yet.

---
In reply to: [69643277](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69643277) [](ancestors = 69643277,69643035)

Python/Setup/swix/Components/CPython2x64.res.swr
```diff
@@ -0,0 +1,6 @@
+use vs
+
+vs.localizedResources
+  vs.localizedResource language=en-us
+                       title="CPython 2.7"
+                       description="Python is a programming language that lets you work more quickly and integrate your systems more effectively. Python 2.7 is a legacy version used in many existing systems."
```
mousetraps: any reason this isn't just included in the swr script?

Python/Setup/swix/swix.generate.targets
mousetraps: Or you could just build a classic vsix and all it takes is a single line to reference it :wink:

Python/Setup/swix/Packages/Microsoft.PythonTools.Profiling.Vsix.swixproj
mousetraps: Okay, I'd just suggest double-checking your output is exactly as you expect it to be then - I was  running into some file corruption when i did this earlier

Python/Setup/swix/Components/CPython2x64.res.swr
zooba: One day we will need to localize this file. I'm assuming it'll be easier if it is following the original format rather than some clever generator.

---
In reply to: [69643432](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69643432) [](ancestors = 69643432)

Python/Setup/swix/Packages/Microsoft.PythonTools.Profiling.Vsix.swixproj
zooba: I have, and it is (after a few attempts...). Managed a successful install earlier today :) But this is the part I was referring to last week when I mentioned the extra level of indirection.

---
In reply to: [69643900](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69643900) [](ancestors = 69643900)

Python/Setup/swix/Packages/dirs.proj
mousetraps: Ahh, gotcha, right. not necessary w/ full file enumeration

Python/Setup/swix/swix.generate.targets
zooba: Except then you can't deploy files into the MSBuild, ProjectTemplates or ItemTemplates directories, and you can't guarantee a stable install directory. So this works out better even though it's a little more work. Most of it is generated from the classic VSIX layout though ;)

---
In reply to: [69643738](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69643738) [](ancestors = 69643738)

Python/Setup/swix/Packages/dirs.proj
zooba: Actually not necessary with multiple generation steps using different base install directories. See lines 23-37 of swix.generate.targets

---
In reply to: [69644061](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69644061) [](ancestors = 69644061)

Python/Setup/swix/Workloads/Microsoft.VisualStudio.Workload.Python.swixproj
```diff
@@ -0,0 +1,28 @@
...
+      <Version>[3.5,3.6)</Version>
+    </Dependency>
+    <Dependency Include="CPython2x64">
+      <Type>Optional</Type>
+      <Version>[2.7,2.8)</Version>
+    </Dependency>
+  </ItemGroup>
+
+  <Import Project="SetupProjectAfter.settings" />
+</Project>
```
mousetraps: i noticed elsewhere you were still relying on devenv /setup - fyi to make sure you also move to the new template manifest format

zooba: That's just the sideload script, and it's to pick up the new pkgdefs and VSCTs. We're already doing templates properly

---
In reply to: [69644524](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69644524) [](ancestors = 69644524)

Python/Setup/swix/Workloads/Microsoft.VisualStudio.Workload.Python.swixproj
```diff
@@ -0,0 +1,28 @@
...
+
+  <ItemGroup>
+    <Package Include="$(MSBuildProjectName).res.swr" />
+
+    <Dependency Include="Microsoft.PythonTools" />
+    <Dependency Include="Microsoft.PythonTools.Web" />
+    <Dependency Include="Microsoft.PythonTools.Samples">
+      <Type>Optional</Type>
+    </Dependency>
+    <Dependency Include="Microsoft.PythonTools.Uwp">
```
mousetraps: Doesn't ptvs also have dependencies on team explorer, web tools, and the like? is there a reason those aren't included in this workload file as hidden dependencies?

zooba: I don't think they're hard dependencies, and I'd rather make them weak dependencies before we drag in a lot of unnecessary stuff. Starting small and working our way up is my plan here

---
In reply to: [69644898](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69644898) [](ancestors = 69644898)

Python/Setup/swix/swix.generate.targets
mousetraps: I get MSBuild (we actually have a separate swixproj to handle files like that), but shouldn't Project / Item templates be deployed properly? Otherwise how does the experimental hive work?

Python/Setup/swix/Workloads/Microsoft.VisualStudio.Workload.Python.swixproj
mousetraps: But they're *eventually* going to be required+hidden, right?

Python/Setup/swix/swix.generate.targets
zooba: That uses VSSDK deployment, not SWIX. We have a project dedicated to deploying templates for the experimental hive with its own vsixmanifest. You can't install the swix packages into the experimental hive.

---
In reply to: [69645463](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69645463) [](ancestors = 69645463)

Python/Setup/swix/Workloads/Microsoft.VisualStudio.Workload.Python.swixproj
zooba: If we can't get away without them, sure. Web tools we might want to defer until someone creates a web project and then prompt, it really depends how big they are. Or they'll be optional+default

---
In reply to: [69645747](https://github.com/Microsoft/PTVS/pull/1396#discussion_r69645747) [](ancestors = 69645747)

mousetraps:
:+1:

PR #1398, opened 2016-07-05
Sync with PHP Tools for Visual Studio

Miloslav:
- Performance improvements for include/exclude operation
- IVsBuildStorageProperty partially supports user project file
- Support for EnvDTE.Projects property
- Minor bug fixes
- VSHPROPID_ProjectDir should return ProjectHome ?


Common/Product/SharedProject/CommonPropertyPage.cs
```diff
@@ -271,12 +271,12 @@ public abstract class CommonPropertyPage : IPropertyPage, IDisposable {
...
-            try {
+            //try {
                 Apply();
                 return VSConstants.S_OK;
-            } catch (Exception e) {
-                return Marshal.GetHRForException(e);
-            }
+            //} catch (Exception e) {
+            //    return Marshal.GetHRForException(e);  // catching exception causes VS to ignore it completely, throw is important!
+            //}
```
zooba: Just go through and delete code rather than commenting it out. If it deserves a comment (like this one does), put it above the code to mention that we're letting exceptions fall through on purpose.

Common/Product/SharedProject/HierarchyNode.cs
```diff
@@ -1172,28 +1172,40 @@ internal abstract class HierarchyNode :
         }
 
         /// <summary>
+        /// Handles the exclude from project command. After operations 
+        /// is completed it refreshes property browser.
+        /// </summary>
+        internal virtual int ExcludeFromProjectWithRefresh() {
```
zooba: If ExcludeFromProject is virtual, does this need to be virtual? What other ways might we want to do this?

Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -952,7 +968,7 @@ internal enum EventTriggering {
                     return this.Caption;
 
                 case __VSHPROPID.VSHPROPID_ProjectDir:
-                    return this.ProjectFolder;
+                    return this.ProjectHome;
```
zooba: Are you sure about this? ProjectDir is supposed to be the directory of the project file. What other components assume that all the source files are beneath this directory?

Common/Product/SharedProject/Wpf/Commands.cs
```diff
@@ -76,7 +76,11 @@ public static class Commands {
             if (!tb.AcceptsReturn) {
                 var path = e.Parameter as string ?? tb.GetValue(TextBox.TextProperty) as string;
                 while (!string.IsNullOrEmpty(path) && !Directory.Exists(path)) {
-                    path = Path.GetDirectoryName(path);
+                    try {
+                        path = Path.GetDirectoryName(path);
+                    } catch {
```
zooba: Catch `ArgumentError` here, otherwise next time we run this through FxCop it'll shout at us :)

zooba:
Nice to hear from you again! These mostly look good - main comment is to just delete the commented out code (and I assume you've already checked all the callers).

Common/Product/SharedProject/ProjectNode.cs
Miloslav: I wasn't sure about this, msdn just says this is a path to project directory. We had few components that assumed this would return ProjectHome. I've changed them to use ProjectHome from EnvDTE.Properties instead. So I'm reverting this back to ProjectFolder.

Common/Product/SharedProject/HierarchyNode.cs
Miloslav: No need for this method to be a virtual. Same with IncludeInProjectWithRefresh.

Common/Product/SharedProject/CommonPropertyPage.cs
Miloslav: Sure. Done

Common/Product/SharedProject/Wpf/Commands.cs
Miloslav: sure: -) done

Miloslav:
Hi Steve! Thank you for reviewing the changes. I've tried to fix all the comments you've mentioned.

zooba:
Thanks! Looks good. Glad to see you're still making progress too

Miloslav:
Thanks! We're also happy PTVS and NTVS are very active :-)

PR #1404, opened 2016-07-06
Interactive window

zooba:
Switches to binary insertion of our own interactive window, based off the common VS one in https://github.com/dotnet/roslyn


huguesv:
:+1:

PR #1405, opened 2016-07-06
Fixes #88 Support code cells

### LINKED ISSUE #88: Support code cells
One thing to keep in mind is that Spyder or MATLAB go one step further: for interactive scripting, you break your script into sections using special comment markup (#%% in Spyder or %% in MATLAB). Then no matter where in the code section the cursor position is, the whole section can be executed with a keyboard shortcut. This has several advantages: it is very fast as there is no need to select a region of code. Also selecting exactly the same region of code repeatedly can be distracting and very boring. In addition, when you select a region of code you could accidentally press a "wrong" key and lose the code! Then you will have to press Undo, etc etc. 

 So I guess a more complete proposal would be to adopt or including a similar feature to Spyder's "Code Cells" (the #%% comment markup) in addition to the "Execute cell" and "Execute and advance to the next cell" command and the respective short-cuts. 

https://pythonhosted.org/spyder/editor.html

We can go beyond REPL support here and also support some visuals for these cells (e.g. adorn them with a border, maybe add some other visual effects to emphasize the boundaries). 

(Migrated from https://pytools.codeplex.com/workitem/3018 and http://pytools.codeplex.com/discussions/581641#post1398987)

zooba:
Fixes #88 Support code cells
Adds support for marking code cells with "#%%", including collapsing them and sending the entire cell to the REPL when pressing Ctrl+Enter on the cell header.
Also fixes some threading issues.

See #88 for a screenshot of how it looks.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
+++ b/Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
@@ -53,7 +53,7 @@ namespace Microsoft.PythonTools.Commands {
         }
 
         private static bool IsCellMarker(ITextSnapshotLine line) {
-            return line.GetText().Trim(' ', '\t').StartsWith("#%%");
+            return OutliningTaggerProvider.OutliningTagger._codeCellRegex.IsMatch(line.GetText());
         }
 
         public override async void DoCommand(object sender, EventArgs args) {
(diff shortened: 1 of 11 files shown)


Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
```diff
@@ -52,6 +52,10 @@ class SendToReplCommand : Command {
             _serviceProvider = serviceProvider;
         }
 
+        private static bool IsCellMarker(ITextSnapshotLine line) {
+            return line.GetText().Trim(' ', '\t').StartsWith("#%%");
```
huguesv: Might have been nicer if it used the same code as the tagger to determine if it's a cell marker, but I guess it's unlikely to be something that causes maintenance issues.

zooba: Not difficult to do, just have to expose the regex more publicly.

huguesv:
👍

PR #1407, opened 2016-07-07
Improves #88 to send the entire cell no matter where the cursor is.

zooba:
Improves #88 to send the entire cell no matter where the cursor is.
Fixes race condition causing FileUpdate requests to fail.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
+++ b/Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
@@ -81,15 +81,19 @@ namespace Microsoft.PythonTools.Commands {
                 ITextSnapshotLine targetLine = snapshot.GetLineFromPosition(selection.Start.Position);
                 var targetSpan = targetLine.Extent;
 
-                if (IsCellMarker(targetLine)) {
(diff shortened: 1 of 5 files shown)


Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
```diff
@@ -81,15 +81,19 @@ class SendToReplCommand : Command {
                 ITextSnapshotLine targetLine = snapshot.GetLineFromPosition(selection.Start.Position);
                 var targetSpan = targetLine.Extent;
 
-                if (IsCellMarker(targetLine)) {
-                    while (targetLine.LineNumber < snapshot.LineCount - 1) {
-                        var nextLine = snapshot.GetLineFromLineNumber(targetLine.LineNumber + 1);
-                        if (IsCellMarker(nextLine)) {
-                            break;
+                for (int lineNo = targetLine.LineNumber; lineNo >= 0; --lineNo) {
```
huguesv: The comment 3 lines above is now slightly out of date, you should mention what this block is doing, and/or extract into a method that makes it clearer.

huguesv:
:+1:

PR #1417, opened 2016-07-12
Change string formatting in pykinect.nui.__init__.py to use `%`

zachpanz88:
Make sure that string formatting is consistent throughout the file, plus this is a better way to format strings.


zooba:
Thanks!

PR #1418, opened 2016-07-12
Remove double negative logic

JacobGorman:
Instead of if != NOT_TRACKED the code now says if == TRACKED which is more logically coherent and easy to understand


zooba:
This isn't actually a double negative: `eTrackingState` could be one of three values, so by checking that it is not `NOT_TRACKED`, we're actually checking that it is either `POSITION_ONLY` or `TRACKED`. So your change could affect the behavior.

@DinoV may have an idea as to which behavior is intended here.

PR #1437, opened 2016-07-15
Vendor info

zooba:
Loads vendor info from registry and displays in environments list.

Looks like this:
image

### DIFF
--- a/Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
+++ b/Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
@@ -307,6 +307,37 @@
                             </Button>
                         </StackPanel>
                     </DataTemplate>
+
+                    <DataTemplate DataType="{x:Type l:EnvironmentView}" x:Key="VendorInformation">
+                        <StackPanel Orientation="Vertical">
+                            <Label x:Name="Vendor">
+                                <TextBlock>
+                                    <Run xml:space="preserve">By</Run>
+                                    <Run Text="{Binding Vendor,Mode=OneWay}" />
+                                </TextBlock>
+                            </Label> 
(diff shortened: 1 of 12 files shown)


Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
```diff
@@ -307,6 +307,37 @@
                             </Button>
                         </StackPanel>
                     </DataTemplate>
+
+                    <DataTemplate DataType="{x:Type l:EnvironmentView}" x:Key="VendorInformation">
+                        <StackPanel Orientation="Vertical">
+                            <Label x:Name="Vendor">
+                                <TextBlock>
+                                    <Run xml:space="preserve">By</Run>
```
huguesv: `xml:space` not needed since there's no space

Python/Tests/Core/EnvironmentListTests.cs
```diff
@@ -657,6 +657,47 @@ public class EnvironmentListTests {
         }
 
         [TestMethod, Priority(1)]
+        [TestCategory("10s")]
```
huguesv: What's the meaning of `10s`?

huguesv:
:+1:

Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
zooba: And I want it to stay that way :)

I'll remove

Python/Tests/Core/EnvironmentListTests.cs
zooba: The test may take over 10 seconds on the test machines.

PR #1448, opened 2016-07-19
Dev15 issues

zooba:


huguesv:
If you haven't already, make sure the Open Interactive Window command doesn't appear on C# project when the Python package is loaded, for example you have a Python and C# project in same solution.
My guess is that your fix only works when Python package is not loaded ;)

huguesv:
if it works then 👍

zooba:
I have. The fix is fine - the code to show the menu was already there, but the vsct changed around it.

PR #1457, opened 2016-07-22
Various fixes

### LINKED ISSUE #241: Automatically add a # when you hit enter in the *middle* of a comment
Example:
```
# this is a really really long line hello world
```
Place the cursor before hello world and hit enter.  I would like to get:
```
# this is a really really long line
# hello world
```
But instead I get:
```
# this is a really really long line
hello world
```

### LINKED ISSUE #968: Project command interactive windows are not associated with the project's environment
Because the window is created and started before CurrentOptions is updated, the window does not know about its interpreter.

This is the underlying cause of #962 and #965

We need to flow through the active environment better so that the association is correct. This is likely to be part of #732.

### LINKED ISSUE #1264: `__main__` does not appear in REPL scopes list
When you start a REPL, the scope is `__main__`. However, once you change scope you can't switch back because it isn't in the list of scopes.

We should ensure that it is always there.

### LINKED ISSUE #1403: Remove hard dependency on code coverage
Since code coverage is not available in community edition, you can no longer build PTVS without Pro (or possibly enterprise).

We should make this dependency looser so that you can still build on community, just without that feature.

### LINKED ISSUE #1435: Cannot upgrade packages in Admin-only installs
When Python is installed for all users, it may be in a location where packages cannot be modified.

We should prompt for admin credentials in this case and run the install elevated.

zooba:
Fixes #1403 Remove hard dependency on code coverage
Fixes #1264 __main__ does not appear in REPL scopes list
Also adds "Open scope in editor" button.
Fixes relative file paths being returned from REPL backend.
Fixes crash when filename is absent.
Makes Ipc.Json.Connection.WriteError more robust for when VS is exiting.
Fixes REPL windows not closing
Fixes #968 Project command interactive windows are not associated with the project's environment
Fixes #1435 Cannot upgrade packages in Admin-only installs
Adds permissions heuristic and improves prompt for elevation.
Adds process runner so that elevated processes can have correct environment and stream output.
Fixes #241 Automatically add comment marker when hitting enter in an existing comment


huguesv:
You'll need to add that RunElevated executable to the setup project(s).

Python/Product/EnvironmentsList/PipExtensionProvider.cs
```diff
@@ -502,15 +484,21 @@ public class PipException : Exception {
             : base(info, context) { }
     }
 
-    public class ValueEventArgs<T> : EventArgs {
-        public T Value { get; set; }
+    public sealed class QueryShouldElevateEventArgs : EventArgs {
+        public bool Cancel { get; set; }
+        public bool Elevate { get; set; }
+        public string TargetDirectory { get; private set; }
```
huguesv: Don't need private set if you only assign in the constructor.  Same for class below.

huguesv:
Thanks for the comments feature.  LGTM.

Python/Product/EnvironmentsList/PipExtensionProvider.cs
zooba: Neat, didn't realize that. Thought it was only readonly fields.

PR #1458, opened 2016-07-22
Fixes #1343 Tooltips do not include signature

### LINKED ISSUE #1343: Improvement in PTVS intellisense
We are looking to extend the extend [Developer Assistant](https://blogs.msdn.microsoft.com/onecode/p/devassistant/) to PTVS . This would help bring code samples in the intellisense for the users as soon as they start typing. 
We rely on the description tooltip to get the full signature of the current selected API, including the class name and function name.  However, in PTVS intellisense, it only has the description of the API.   It does not contain the class name and function name. Could we get these info?.
screenshot

zooba:
Fixes #1343 Tooltips do not include signature
Adds rich description protocol and improves content of descriptions.

### DIFF
--- a/Common/Product/SharedProject/UIThread.cs
+++ b/Common/Product/SharedProject/UIThread.cs
@@ -122,6 +122,45 @@ namespace Microsoft.VisualStudioTools {
         }
 
         /// <summary>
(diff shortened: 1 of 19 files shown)


huguesv:
:shipit:

PR #1472, opened 2016-07-25
Fixes #1249 Highlighting for async is wrong when a decorator is used

### LINKED ISSUE #1249: Highlighting for `async` is wrong when a decorator is used
See screenshot

image

zooba:
Fixes #1249 Highlighting for async is wrong when a decorator is used
Preserves the original start index of a function so it can be used even if a decorator later changes it.


huguesv:
:+1:

PR #1479, opened 2016-07-26
Fixes #1473 Directory/File IntelliSense on multi level folder

### LINKED ISSUE #1473: Directory/File IntelliSense on multi level folder
Once you autocomplete a folder, you'll have something like this (end backslash because it's a folder, with caret before the end quote):
```
os.chdir(r"c:\Program Files (x86)\"
```

To get completions for the next level of folder or files, you have to either:
- type a character (you may just want to browse all, in which case that's not a good option)
- type a space (you'll have to backspace on it, because you don't really want a space there)
- delete the backslash and press ctrl+space

I would have expected ctrl+space to work after the backslash.  Having to remove the one that was automatically added for me was not obvious.

zooba:
Fixes #1473 Directory/File IntelliSense on multi level folder
Updates string literal completions to only operate on string contents
Specialized completion triggering for filenames.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Extensions.cs
+++ b/Python/Product/PythonTools/PythonTools/Extensions.cs
@@ -155,7 +155,13 @@ namespace Microsoft.PythonTools {
             }
 
             if (position > lastToken.Span.Start) {
-                if (lastToken.CanComplete()) {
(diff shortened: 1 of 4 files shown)


zooba:
@huguesv Thoughts?

huguesv:
I'll have to try it out to be able to tell.  Sounds good in theory.  Feel free to check in.

PR #1487, opened 2016-07-27
Fixes analysis crash due to UWP factory provider.

zooba:
FYI @alanch-ms - the call is what triggers the FileNotFoundException (when loading the types required to JIT the function body), so it needed to be one level deeper.


alanch-ms:
Thanks!  I didn't know the exception was being thrown at JIT.  Sorry about that.

PR #1489, opened 2016-07-28
Fixes #1463 Non-Python files are tokenized as Python files after swit…

### LINKED ISSUE #1463: Non-Python files are tokenized as Python files after switching environment
Create a new Django project, then add another environment and switch to it.

All the html and css files start showing Python tokenization errors (and assertions appear in debug builds from the tokenizer). These should not be being tokenized, so something clearly goes off when we start a new analyser.

zooba:
Fixes #1463 Non-Python files are tokenized as Python files after switching environment
Filters file nodes to those that are Python code files.


huguesv:
👍

PR #1493, opened 2016-07-29
Ensures text manager is not loaded before the shell is initialized.

zooba:
Ensures text manager is not loaded before the shell is initialized. (This helps prevent the Tab setting from being incorrect)
Switches to correct parsing of Unicode identifiers.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Extensions.cs
+++ b/Python/Product/PythonTools/PythonTools/Extensions.cs
@@ -51,6 +51,7 @@ using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Utilities;
 using Microsoft.VisualStudioTools;
 using Microsoft.VisualStudioTools.Project;
+using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.PythonTools {
     public static class Extensions {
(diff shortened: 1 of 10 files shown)


Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -826,6 +827,51 @@ public static class Extensions {
...
+                }
+                return VSConstants.S_OK;
+            }
+        }
+
+        internal static Task WaitForShellInitializedAsync(this IServiceProvider provider) {
+            if (provider.IsShellInitialized()) {
+                return Task.FromResult<object>(null);
+            }
+            return new ShellInitializedNotification(provider.GetShell()).Task;
```
huguesv: Is there a risk of waiting forever here if the shell initializes after you've checked, but before you register for the property changes event?

zooba: Yes. We should check once more before returning

Python/Product/PythonTools/PythonToolsService.cs
```diff
@@ -579,13 +561,16 @@ public sealed class PythonToolsService : IDisposable {
 
         internal System.IServiceProvider Site => _container;
 
-        internal LanguagePreferences LangPrefs {
-            get {
-                return _langPrefs;
+        internal LanguagePreferences LangPrefs => _langPrefs.Value;
+
+        internal async Task<LanguagePreferences> GetLangPrefsAsync() {
```
huguesv: I think that deserves a comment.  If I need preferences, how do I know if I'm supposed to call the method or the propery?

huguesv:
:+1:

PR #1495, opened 2016-07-29
Fixes launching interpreter without a startup script.

zooba:


Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
```diff
@@ -423,6 +423,8 @@ sealed class InterpreterListToolWindow : ToolWindowPane {
                 }
             }
 
+            config.LaunchOptions["NeverPauseOnExit"] = "true";
```
huguesv: Feels like we should use a constant, but fine if it's too much trouble because of project dependencies.

huguesv:
:+1:

PR #1500, opened 2016-07-30
Don't execute code if it's a blank string.

### LINKED ISSUE #1491: Interactive window: unexpected EOF while parsing
VS 2015 Update 3
Latest PTVS from master

Open interactive window
Press ENTER
Result:
```
>>> 
  File "<stdin>", line 1
    
    ^
SyntaxError: unexpected EOF while parsing
>>> 
```

huguesv:
Fixes #1491

I haven't run the tests yet.


zooba:
I think I'd rather have the Python code check whether it's empty. That way when (if) we add a backend that can do something with an empty line, it isn't being blocked by VS.

PR #1512, opened 2016-08-02
Adds missing Microsoft.PythonTools.Uwp.Wizards.dll to signing script

zooba:
FYI @alanch-ms - I missed it as well, so hopefully next time we add a new assembly we'll remember between us.


alanch-ms:
Thanks for the info!

PR #1536, opened 2016-08-12
Issue 1469

zooba:
This is a big change to achieve the following goals:
* properly support PEP 514
* centralise package management
* make it possible to change package manager for an interpreter
* fix many issues around environments in the registry
* fix some issues around installed/upgrade package state

Also in here:
* a new `InterpreterArchitecture` class to use rather than `ProcessorArchitecture` (mainly for much easier string conversions, but also to fully control our public API)
* `LibraryPath` is removed from interpreter configuration, since the factory can now handle everything to do with the library path (we already used a set of paths in most cases anyway)
* `DerivedInterpreterFactory` is removed (as the set of paths better handles virtual environments anyway)
* `FullDescription` is removed and we go back to fully specified `Description`
* interpreter ID format has changed (which will require an upgrade step for any prior 3.0 projects)
* registry search is now a separate class (for testing and for use in tests)

It isn't finished yet, though it should be fully functional right now. There are still a few references to the old pip support that need to be changed to use the factory. Feel free to start reviewing, as the overall architecture is probably not going to change dramatically.


Python/Product/Analysis/Interpreter/InterpreterArchitecture.cs
```diff
@@ -0,0 +1,184 @@
...
+
+namespace Microsoft.PythonTools.Interpreter {
+    public struct InterpreterArchitecture : 
+        IFormattable,
+        IEquatable<InterpreterArchitecture>,
+        IComparable<InterpreterArchitecture>
+    {
+        public ProcessorArchitecture Value { get; set; }
+
+        public InterpreterArchitecture(ProcessorArchitecture arch) {
```
DinoV: >ProcessorArchitecture [](start = 39, length = 21)

Why not just define some extension methods on this type, or define a new enum with extension methods?

zooba: The flexible ToString(), mainly. Makes it possible to do string.Format("{0}{1: ()}", name, arch) to get "2.7" or "2.7 (32-bit)"

---
In reply to: [75406677](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75406677) [](ancestors = 75406677)

zooba: I also figured we should own all of our public API, especially given the issues we've had with Roslyn recently. The use of ProcessorArchitecture isn't actually necessary, so I guess this could be a base class with private singleton subclasses?

---
In reply to: [75407051](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75407051) [](ancestors = 75407051,75406677)

DinoV: The only downside is you won't be able to switch on it then...  I was mostly a little surprised because the description made it sounded like ProcessorArchitecture would go away, but I think it's an okay dependency (other than being in an odd namespace).  

---
In reply to: [75503748](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75503748) [](ancestors = 75503748,75407051,75406677)

Python/Product/Analysis/Interpreter/PythonInterpreterFactoryExtensions.cs
```diff
@@ -52,6 +54,19 @@ public static class PythonInterpreterFactoryExtensions {
...
         public static async Task<HashSet<string>> FindModulesAsync(this IPythonInterpreterFactory factory, params string[] moduleNames) {
+            var withPackages = factory as IPackageManager;
+            if (withPackages != null) {
+                var res = new HashSet<string>();
+                foreach (var m in moduleNames) {
+                    if ((await withPackages.GetInstalledPackageAsync(new PackageSpec(m), CancellationToken.None)).IsValid) {
+                        res.Add(m);
+                    }
+                }
+                if (res.Count == moduleNames.Length) {
```
DinoV: >if (res.Count == moduleNames.Length) { [](start = 16, length = 38)

I assume this fnuction will never be called with duplicate module names? :)

Python/Product/Analysis/Interpreter/InterpreterArchitecture.cs
zooba: We don't actually switch anywhere on this. I think the only places we did in the past was to choose a string to display, and that's now encapsulated properly.

---
In reply to: [75513220](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75513220) [](ancestors = 75513220,75503748,75407051,75406677)

Python/Product/Analysis/Interpreter/PythonInterpreterFactoryExtensions.cs
zooba: Sure. That sounds like a good way to avoid bugs 

---
In reply to: [75513879](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75513879) [](ancestors = 75513879)

Python/Product/Analysis/Interpreter/InterpreterArchitecture.cs
DinoV: The singletons sound good to me


---
In reply to: [75514069](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75514069) [](ancestors = 75514069,75513220,75503748,75407051,75406677)

Python/Product/Analysis/Interpreter/Pep440Version.cs
```diff
@@ -20,10 +20,9 @@
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
-using System.Threading.Tasks;
 
-namespace Microsoft.PythonTools.EnvironmentsList {
-    struct Pep440Version : IComparable<Pep440Version>, IEquatable<Pep440Version> {
+namespace Microsoft.PythonTools.Interpreter {
+    public struct Pep440Version : IComparable<Pep440Version>, IEquatable<Pep440Version> {
```
DinoV: >Pep440Version [](start = 18, length = 13)

can we just call this PythonVersionSpec or something a little more friendly?  We can always put in the comment that it's a PEP 440 compatible version...  I just don't see too many people seeking out "Pep..." as type names ;)

Python/Product/Analysis/Interpreter/PythonInterpreterFactoryWithDatabase.cs
```diff
@@ -55,60 +56,78 @@ public class PythonInterpreterFactoryWithDatabase :
...
         public PythonInterpreterFactoryWithDatabase(
             InterpreterConfiguration config,
-            bool watchLibraryForChanges
+            InterpreterFactoryCreationOptions options
         ) {
-            _config = config;
-
-            if (_config == null) {
+            if (config == null) {
                 throw new ArgumentNullException("config");
```
DinoV: >config [](start = 49, length = 6)

nameof(config)

Python/Product/PythonTools/PythonTools/Navigation/EditFilter.cs
```diff
@@ -658,7 +658,10 @@ class NodeEnumerator<T> : IVsEnumNavInfoNodes where T : IVsNavInfoNode {
                             var line = pyPoint.Value.GetContainingLine();
                             var lineText = line.GetText();
```
DinoV: What does this do?  Can we get a comment?  It's a little disturbing that we're looking for a comment charater w/o using the tokenizer as it could be embedded in a string literal

zooba: This is the code that adds the comment marker on the next line when you break a comment on the current line.

With this change, it can only be embedded in a multiline string literal and it's at the start of the line, so I'm inclined to keep the behaviour for now and see whether people notice. But I'll add the comment here

---
In reply to: [75517787](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75517787) [](ancestors = 75517787)

Python/Product/Analysis/Interpreter/Pep440Version.cs
zooba: Renamed to PackageVersion to match PackageSpec

---
In reply to: [75514490](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75514490) [](ancestors = 75514490)

Python/Product/VSInterpreters/PipPackageManager.cs
```diff
@@ -0,0 +1,651 @@
...
+
+            if (allowFileSystemWatchers) {
+                _libWatchers = new List<FileSystemWatcher>();
+                _refreshIsCurrentTrigger = new Timer(RefreshIsCurrentTimer_Elapsed);
+            }
+        }
+
+        public void SetInterpreterFactory(IPythonInterpreterFactory factory) {
+            if (factory == null) {
+                throw new ArgumentNullException("factory");
```
DinoV: >factory [](start = 49, length = 7)

nameof

Python/Product/VSInterpreters/PipPackageManager.cs
```diff
@@ -0,0 +1,651 @@
...
+            }
+            if (!File.Exists(factory.Configuration?.InterpreterPath)) {
+                throw new NotSupportedException();
+            }
+
+            _factory = factory;
+
+            // TODO: Get index from factory
+            //_indexUrl = "https://pypi.python.org/pypi/";
+            _indexUrl = null;
```
DinoV: >_indexUrl [](start = 12, length = 9)

Does this need to happen?  Can we ifdef out this and the other code which uses this because it's not implemented?

Python/Product/VSInterpreters/PipPackageManager.cs
```diff
@@ -0,0 +1,651 @@
...
+        }
+
+        public async Task<bool> InstallAsync(PackageSpec package, IPackageManagerUI ui, CancellationToken cancellationToken) {
+            AbortOnInvalidConfiguration();
+            await AbortIfNotReady(cancellationToken);
+
+            bool success = false;
+            List<string> args;
+
+            if (!SupportsDashMPip) {
```
DinoV: >SupportsDashMPip [](start = 17, length = 16)

Is the reason for this that the 2.7 style doesn't work on 3.x?  Otherwise why not just keep things simple and always do the uglier way given that we need to do it anyway.

Python/Product/VSInterpreters/PipPackageManager.cs
zooba: Index URL is a little weird right now, I'm not sure when we're going to move to pypi.org completely or how that will affect our code for searching PyPI. But I think it's only relevant in one place right now, so I'll pull out the references.

---
In reply to: [75519124](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75519124) [](ancestors = 75519124)

Python/Product/VSInterpreters/PipPackageManager.cs
zooba: Technically "pip.main()" is not a supported entry point, so if they do something to break it I'd rather not break absolutely everywhere.

---
In reply to: [75519467](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75519467) [](ancestors = 75519467)

Python/Product/Analysis/Interpreter/IPackageManagerUI.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Called when an operation completes.
+        /// </summary>
+        /// <param name="operation">
+        /// An operation identifier. This is intended for logging rather than
+        /// user information.
+        /// </param>
+        /// <param name="success">True if the operation was successful.</param>
+        void OnOperationFinished(string operation, bool success);
+
```
DinoV: I assume these can be called from any thread...  We should add a comment to that effect.

zooba: I also clarified that the "On*TextReceieved" text should include trailing newlines, and updated some of the code that was adding extra newlines to it.

---
In reply to: [75521085](https://github.com/Microsoft/PTVS/pull/1536#discussion_r75521085) [](ancestors = 75521085)

zooba:
Signed off by @dinov with the caveat that I need to pay a lot of attention to what breaks and go in and fix it :)

PR #1546, opened 2016-08-16
Fixes #1544 VS 15 Preview 4 test adapter not discovering tests

### LINKED ISSUE #1544: VS 15 Preview 4 test adapter not discovering tests
VS 15 Build 25615.00
Data Science workload

Add New Item, Unit Test
Save project, close and reload solution, restart VS...  I was never able to get the tests detected.  I made sure it wasn't a case of wrong active interpreter (ie. tests run with startup file set to the test & F5).

[Microsoft.VisualStudio.Default.err.txt](https://github.com/Microsoft/PTVS/files/419092/Microsoft.VisualStudio.Default.err.txt)

zooba:
Fixes #1544 VS 15 Preview 4 test adapter not discovering tests
Splits the test executor and discoverer into a separate assembly with fewer dependencies. This should allow it to load without any assemblies from VS.


huguesv:
:shipit:

PR #1548, opened 2016-08-19
ProjectHome should have trailing back slash

Miloslav:
ProjectHome without trailing slash was causing some issues reflecting file system changes to our project system. There are other ways how to fix it, but this one seems to be most efficient.

GetParentFolderForPath or RenameDirectory should use IsSameDirectory instead of IsSamePath


Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -634,7 +634,6 @@ internal enum EventTriggering {
                     projectHome = CommonUtils.GetAbsoluteDirectoryPath(
                         this.ProjectFolder,
                         this.GetProjectProperty(CommonConstants.ProjectHome, resetCache: false));
-                    projectHome = CommonUtils.TrimEndSeparator(projectHome);
```
zooba: This change will likely have much wider effects than the other two. Does it need to be changed or is the IsSameDirectory change sufficient?

zooba:
@Miloslav Thanks for keeping the PR up to date - do you have a reply for my comment? Are you waiting for exactly one year? :)

huguesv:
I'm going to close since it's been a while, but feel free to create a new PR that resolves the conflicts and provides the info requested by @zooba. Thanks!

PR #1552, opened 2016-08-23
Fixes #735 Project template for creating C extensions

### LINKED ISSUE #735: Project template for creating C extensions
Add a new project template that creates the necessary C++ project, with the correct settings for building a C extension to Python 3.5.

This is the next step in our C++ interop support, which started with the introduction of C++ mixed-mode debugging in PTVS 2.1.  Additionally, Python 3.5 (and its extensions) are built with VS 2015, which will be one of the supported VS versions for PTVS 3.0.

Compared to the Win32 Project wizard, it will be a quicker/better experience.  The user won't need to know what type of project, precompiled headers they need to select.  We can also add the correct .h and .lib files, and the output file name.

We should also add a matching Python project to the solution.  A reference to the C++ project will be present, with a entry in search path (that may not be needed?).

We can also generate a correct setup.py that will be used when installing from sources.

We should ensure that any code we generate as part of the template will compile successfully on other operating systems/compilers.

As far as boilerplate code, we probably should be adding some examples of implementing a function/class, and have the matching test code in the Python project.  TBD.

zooba:
Fixes #735 Project template for creating C extensions
Adds native extension module template for VC.


zooba:
@brettcannon Think you could take a look at the C code in here? (You've seen it before, it's unchanged, but this is last chance before it gets checked in properly)

brettcannon:
LGTM

Issue #1107, opened 2016-02-21
[Feature Request] Support Python 3.5 for IIS


Korijn:
The steps on [this page](https://azure.microsoft.com/en-us/documentation/articles/web-sites-python-configure/) don't apply to Python 3.5, apparently, because the subpackage [wfastcgi](https://pypi.python.org/pypi/wfastcgi) is incompatible.

Can this be fixed, or can you explain what the problem is so that someone else might contribute a fix?

zooba:
`wfastcgi` should work fine with Python 3.5 - we just haven't posted an update to it since Python 3.5 released and so the classifiers haven't been updated.

Are you encountering actual problems trying to use it?

gschizas:
The problem is that the default installation folder for Python 3.5 is in `Program Files`. wfastcgi.enable() doesn't properly quote the arguments (which have spaces)

I've fixed it on my PC, I'll be doing a pull request soon.

Korijn:
Is this a priority yet?

gschizas:
It has been live, on production, for some time now.

Korijn:
Awesome. Where do you blog about these things?

gschizas:
I rarely blog :)

zooba:
Yeah, sorry, it was fixed but we didn't make any major announcement. At the very least I should have posted a comment here and closed the issue - I'll try to be better at that next time.

Korijn:
No worries. Thanks!

PR #1563, opened 2016-08-26
Allow IIS working with Python 3.5

gschizas:
When the wfastcgi.py file is in a folder with spaces (default for Python 3.5), `enable` and `disable` don't properly quote the filename, so it gets passed to python.exe unquoted (and only "C:\Program" counts as the filename.


gschizas:
This resolves #1107

zooba:
Thanks! I'll update the published package too.

Issue #1206, opened 2016-04-06
Django: Intellisense named urls from urls.py


KronosKoderS:
It would be nice to have the urls found in urls.py listed as options when completing the url tag:

    {% url '<intellisense>' %}

zooba:
Yes, it would!

rv404674:
hi @zooba , i am a newcomer and would like to take this up. Any suggestions for me. Thanks.

PR #1581, opened 2016-09-01
Django urls intellisense

David-Wobrock:
Hi,

Here is a very first version of django url parsing.
It only proposes the `name` parameter of all `django.conf.urls.url()` calls as a completion to a `{% url %}` template block.
This PR is related to #1206 and if this piece of code is satisfactory, I'll be happy to extend it for handling parameters, named params, namespaces, nested ns, template var conflicts, etc.

And btw, I took the liberty to refactor the template block classes into different files, and also to commit two unchanged files (I guess only the line endings changed, but I have to admit that I didn't investigate more into it. Actually, I had two modified files just after cloning the repo, before even editing anything)

Cheers,
David


zooba:
Just so you're aware, I probably won't get a chance to look at this for at least a week.

Possibly @DinoV might and he can merge it if he's happy, but otherwise the delay is totally my fault.

zooba:
Thanks David, this looks great!

PR #1587, opened 2016-09-02
Fixes #1576 PythonCore default description does not match 3.6 description

### LINKED ISSUE #1576: PythonCore default description does not match 3.6 description
Python 3.6 registers its description as "Python 3.6 (32-bit)" but the default format for PythonCore entries is "Python 32-bit 3.5"

We should update our format to match Python 3.6's, otherwise entries do not sort properly.

zooba:
Fixes #1576 PythonCore default description does not match 3.6 description
Updates default description format.


huguesv:
:+1:

PR #1588, opened 2016-09-02
Fixes #1571 NullReferenceException when clicking "Install Python Package"

### LINKED ISSUE #1571: NullReferenceException when clicking "Install Python Package" without Python Environments window
Clicking "Install Python Package" now opens the environments window and selects the right panel, but if the window is not currently open we have a null reference for the currently selected item.

See `InterpreterListToolWindow.cs`:

```csharp
            var select = envs.Environments.OfType<EnvironmentView>().FirstOrDefault(e => e.Factory == interpreter);
            var ext = extension == null ?
                null :
                select.Extensions.FirstOrDefault(e => e != null && extension.IsEquivalentTo(e.GetType()));
```

`select` may be null in this case, and so it fails. The following line is a null check and the `var ext...` should be inside that.

zooba:
Fixes #1571 NullReferenceException when clicking "Install Python Package" without Python Environments window
Retries the selection operation lazily until the window is loaded.


huguesv:
:+1:

PR #1591, opened 2016-09-02
Fixes #1574: Add enter at end of fully typed word is ignored

### LINKED ISSUE #1574: Add enter at end of fully typed word option is ignored
Turn on the option to "Add new line on enter at end of fully typed word", then bring up completions for a fully typed word and hit enter. No new line is added.

zooba:
Fixes #1574: Add enter at end of fully typed word is ignored
Use applicable span rather than trigger point to determine if full word has been entered.


huguesv:
Looks reasonable

PR #1592, opened 2016-09-02
Improves assertion for missing database check

zooba:
Improves assertion for missing database check, and hopefully prevents it from triggering incorrectly.


huguesv:
:+1:

PR #1597, opened 2016-09-10
fix "Cannot add duplicate collection entry" error

YoungjaeKim:
When this script runs multiple times, it throws error as below.

> ERROR ( message:New application object missing required attributes. Cannot add duplicate collection entry of type 'application' with combined key attributes 'fullPath, arguments' respectively set to 'D:\\Python27\\\\python.exe, "F:\\sitesroot\\0\\bin\\wfastcgi.py"'. )

So, I added a line to remove previous config by following similar solution; http://stackoverflow.com/a/35506641/361100
It tested in Azure Web Role + Flask project.

### DIFF
--- a/Python/Templates/SharedFiles/ConfigureCloudService.ps1
+++ b/Python/Templates/SharedFiles/ConfigureCloudService.ps1
@@ -220,7 +220,7 @@ if ($is_web) {
         $max_requests = 1000
     }
 
-    iex "appcmd $appcmdargs set config /section:system.webServer/fastCGI ""/-[fullPath='$interp',arguments='\""""$wfastcgi\""""',instanceMaxRequests='$max_requests',signalBeforeTerminateSeconds='30']"""
+    iex "appcmd $appcmdargs set config /section:system.webServer/fastCGI ""/-[fullPath='$interp',arguments='\""""$wfastcgi\""""',instanceMaxRequests='$max_requests',signalBeforeTerminateSeconds='30']"" | Out-Null"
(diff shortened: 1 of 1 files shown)


zooba:
@YoungjaeKim What error do you see if it tries to remove an entry but there isn't one there?

YoungjaeKim:
Hi @zooba 
I tested upon your request; Added remove entry command multiple times.

It gives error message in each command as below.
> ERROR ( message:Cannot find requested collection element. )

Though remove command gives an error, I think my PR is safer as it gurantees to add new entry.

zooba:
What about using `appcmd.exe clear config /section:system.webServer/fastCGI`?

My biggest concern is having random error codes or messages floating around in the script, as they may trigger false failures (e.g. I've seen processes fail because of _any_ message in standard error, even when they aren't fatal).

YoungjaeKim:
Totally agree with your opinion.
I applied as your suggestion and got below log message.

> Cleared section "system.webServer/fastCgi" at configuration path "MACHINE/WEBROOT/APPHOST"

Seems more general solution than my PR. Would I close this PR and propose as your suggestion again?

zooba:
Yeah, you're right, we don't actually want to clear everything.

Let's just add `| Out-Null` at the end of the new line, so that at least we will hide the error message.

YoungjaeKim:
@zooba Applied Out-Null and confirmed that the `ERROR ( message:Cannot find requested collection element. )` message was not logged.

YoungjaeKim:
Thanks. my pleasure 👍

PR #1600, opened 2016-09-12
Add cookiecutter feature

huguesv:
Haven't touched the dev15 installer stuff yet, will do that now.
Edit: Setup updated but not tested

All under 'Shared' has been copied from various PTVS projects, in order to keep the cookiecutter extension independent from PTVS.  It's only using the build system files (.settings, .targets, etc). The test project isn't completely independent from PTVS yet, it reuses the TestUtilities for the TestData deployment.

image

image

image

context

Styling is mostly ok ie. dark theme:

search3

I picked my icons browsing this page:
http://glyphlist.azurewebsites.net/knownmonikers/


Python/Product/Cookiecutter/Commands/GitHubHomeCommand.cs
```diff
@@ -0,0 +1,47 @@
...
+            _commandId = commandId;
+        }
+
+        public override void DoCommand(object sender, EventArgs args) {
+            _window.NavigateToGitHub(_commandId);
+        }
+
+        public string Description {
+            get {
+                return "Cookiecutter";
```
zooba: >"Cookiecutter" [](start = 23, length = 14)

Is this the tooltip? Seems to be the same in a few commands

Python/Product/Cookiecutter/cookiecutter_run.py
```diff
@@ -0,0 +1,49 @@
...
+        cookiecutter(
+            template_path,
+            no_input=True,
+            extra_context=context,
+            output_dir=output_path,
+            config_file=user_config_path,
+        )
+    except Exception as error:
+        print(error)
+
```
zooba: All this really achieves is that the exit code will always be 0 and that errors are written to stdout rather than stderr. I'd just leave out the try/except block completely and let exceptions escape.

Python/Product/Cookiecutter/Shared/Interpreters/CPythonInterpreterFactoryConstants.cs
```diff
@@ -0,0 +1,64 @@
...
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.CookiecutterTools.Interpreters {
```
zooba: >Interpreters [](start = 38, length = 12)

Let's plan to aggressively simplify this namespace so we can avoid duplicating too much code (e.g. I'd rather leave out PythonLanguageVersion.cs, as it requires modification to avoid throwing errors when new Python versions come out)

Python/Product/Cookiecutter/Commands/GitHubHomeCommand.cs
huguesv: No, it's not used.  It's the one in the vsct that is used.

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -0,0 +1,726 @@
...
+                await AddFromSource(_githubSource, null, KnownMonikers.GitNoColor, GitHub, continuationToken);
+            }
+        }
+
+        private async Task AddFromSource(ITemplateSource source, string searchTerm, VisualStudio.Imaging.Interop.ImageMoniker image, CategorizedViewModel parent, string continuationToken = null) {
+            var loading = new LoadingViewModel();
+            parent.Templates.Add(loading);
+
+            try {
+                var result = await source.GetTemplatesAsync(searchTerm, continuationToken, CancellationToken.None);
```
zooba: >GetTemplatesAsync [](start = 42, length = 17)

I think we need to check the current search term after completing this method and abort if it has changed. Otherwise we'll get the race condition you were worried about.

zooba: Alternatively, you can keep a CancellationTokenSource in the VM and cancel it each time a new search begins to abort the previous calls. That's probably the nicer way to do it.

---
In reply to: [78426680](https://github.com/Microsoft/PTVS/pull/1600#discussion_r78426680) [](ancestors = 78426680)

Python/Product/Cookiecutter/ViewModel/TemplateViewModel.cs
```diff
@@ -0,0 +1,179 @@
...
+
+            set {
+                if (value.Guid != _image.Guid || value.Id != _image.Id) {
+                    _image = value;
+                    PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(Image)));
+                }
+            }
+        }
+
+        private static bool ParseRepoOwnerAndName(string repoUrl, out string owner, out string name) {
```
zooba: >ParseRepoOwnerAndName [](start = 28, length = 21)

I'd rather see this done with a regex:

var m = Regex.Match(@"http(s)?://github\.com/(?P<owner>.+?)/(?P<name>.+?)(/|$)", repoUrl);
owner = m.Groups["owner"].Value;
name = m.Groups["name"].Value;
return m.Groups["owner"].Success && m.Group["name"].Success;

zooba:
For the installer, if it's already correctly building into a `layouts\CookieCutter` folder then it should be trivial to clone one of the existing .swixproj files.

zooba:
:clock1:

zooba:
Ah, that's where the :shipit: guy comes from :)

I'm concerned about the potential race and I think we'd rather tidy that up, especially since it's a reaction to the user typing. The regex is a nice-to-have. Those are the only things holding this up IMO, though we'll certainly have work to do after the next release

huguesv:
done

---
In reply to: [246443905](https://github.com/Microsoft/PTVS/pull/1600#issuecomment-246443905) [](ancestors = 246443905)

Python/Product/Cookiecutter/cookiecutter_run.py
huguesv: done

---
In reply to: [78425152](https://github.com/Microsoft/PTVS/pull/1600#discussion_r78425152) [](ancestors = 78425152)

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
huguesv: done.

---
In reply to: [78426820](https://github.com/Microsoft/PTVS/pull/1600#discussion_r78426820) [](ancestors = 78426820,78426680)

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -0,0 +1,749 @@
...
+        }
+
+        public async Task SearchAsync() {
+            _templateRefreshCancelTokenSource?.Cancel();
+            _templateRefreshCancelTokenSource = new CancellationTokenSource();
+            try {
+                await RefreshTemplatesAsync(SearchTerm, _templateRefreshCancelTokenSource.Token);
+            } catch (OperationCanceledException) {
+            }
+            _templateRefreshCancelTokenSource = null;
```
zooba: >_templateRefreshCancelTokenSource [](start = 12, length = 33)

This is a new race condition here because it's after the await. I'd just leave the value there (or use Interlocked.CompareExchange, but since this function is always called from the UI thread that isn't necessary, and CTS is a cheap class)

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -0,0 +1,749 @@
...
+                    _installedSource.InvalidateCache();
+
+                    _templateRefreshCancelTokenSource?.Cancel();
+                    _templateRefreshCancelTokenSource = new CancellationTokenSource();
+                    try {
+                        Installed.Templates.Clear();
+                        await AddFromSource(_installedSource, SearchTerm, KnownMonikers.TestSuite, Installed, CancellationToken.None);
+                    } catch (OperationCanceledException) {
+                    }
+                    _templateRefreshCancelTokenSource = null;
```
zooba: >_templateRefreshCancelTokenSource [](start = 20, length = 33)

Same as above

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -0,0 +1,749 @@
...
+            var last = GitHub.Templates.LastOrDefault();
+            if (last is ContinuationViewModel) {
+                _templateRefreshCancelTokenSource?.Cancel();
+                _templateRefreshCancelTokenSource = new CancellationTokenSource();
+                try {
+                    GitHub.Templates.Remove(last);
+                    await AddFromSource(_githubSource, null, KnownMonikers.GitNoColor, GitHub, _templateRefreshCancelTokenSource.Token, continuationToken);
+                } catch (OperationCanceledException) {
+                }
+                _templateRefreshCancelTokenSource = null;
```
zooba: >_templateRefreshCancelTokenSource [](start = 16, length = 33)

Same as above

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -0,0 +1,749 @@
...
+                }
+            } catch (TemplateEnumerationException ex) {
+                var template = new ErrorViewModel() {
+                    ErrorDescription = ex.Message,
+                    ErrorDetails = ex.InnerException?.Message,
+                };
+                parent.Templates.Add(template);
+            }
+
+            parent.Templates.Remove(loading);
```
zooba: > parent.Templates.Remove(loading) [](start = 11, length = 33)

This probably needs to be in a finally, in case we cancel the operation.

zooba:
Feel free to merge once those comments are addressed.

PR #1605, opened 2016-09-13
Workspace

zooba:
This is not finished yet, but I'd like to start the review process and hopefully there are only incremental improvements to come.


Python/Product/Workspace/PythonLaunchDebugTargetProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.VisualStudio.Workspace.Debug;
+
+namespace Microsoft.PythonTools.Workspace {
+    [ExportLaunchConfigurationProvider(
+        ProviderType,
+        new[] { PythonConstants.FileExtension, PythonConstants.WindowsFileExtension },
+        PythonDebugLaunchProvider.LaunchTypeName,
+        PythonDebugLaunchProvider.JsonSchema
+    )]
+    class PythonLaunchDebugTargetProvider : ILaunchConfigurationProvider {
```
huguesv: >PythonLaunchDebugTargetProvider [](start = 10, length = 31)

You might want to base the name of the class on the interface it is implementing.  Right now it's confusing because there's a class that implements ILaunchDebugTargetProvider, and it's not this one.  So maybe "PythonLaunchDebugTargetProvider" for the other class, and "PythonLaunchConfigurationProvider" for this class.  Same comment for the Pyproj ones.

huguesv:
I'm not sure where the split into Core is going, file moves aren't part of the changes yet. 

Can you summarize what you intend on having in PythonTools, PythonTools.Core and Workspace and which way the dependencies will be?

huguesv:
Also, I presume that one goal is to prevent loading the more expensive PythonTools package when only workspace is used?

---
In reply to: [246819886](https://github.com/Microsoft/PTVS/pull/1605#issuecomment-246819886) [](ancestors = 246819886)

Python/Product/PythonTools/Properties/AssemblyInfo.cs
```diff
@@ -29,20 +29,9 @@
 [assembly: CLSCompliant(false)]
 [assembly: NeutralResourcesLanguage("en-US")]
 
-[assembly: ProvideCodeBase(AssemblyName = "Microsoft.PythonTools", CodeBase = "Microsoft.PythonTools.dll", Version = AssemblyVersionInfo.StableVersion)]
```
huguesv: Are the ProvideCodeBase attributes being moved somewhere else or did you determine we no longer need them at all?  It's making me a bit nervous, don't know what this is going to break if they go away.

zooba:
Ah sure. PythonTools.Core is the project that generates the VSIX (or the extension directory), while PythonTools contains the main package.

This lets us include assemblies in the VSIX that depend on PythonTools, such as Workspace. Currently we have a hacky MSBuild solution that doesn't work well in VS, as it creates circular dependencies.

Core is able to depend on everything that needs to be in the VSIX without creating circular dependencies - nothing can or should depend on Core, and the DLL it creates does not ship. There is no code, only dependencies and a pkgdef to register all the assembly bindings. Everything else gets pulled from other projects.

I had originally hoped to support Workspace without loading PythonTools, but it requires refactoring basically everything, so I didn't bother for now.

Python/Product/PythonTools/Properties/AssemblyInfo.cs
zooba: They're all in Core. We need an assembly reference to generate it, but PythonTools cannot reference assemblies that reference PythonTools because it's circular.

zooba:
Other assemblies that reference PythonTools include the project wizards and the test adapter. All of these now reliably build in VS and in MSBuild, and are included in the VSIX generated by Core.

huguesv:
Understood. Make sense. The only part that is a bit confusing is the name "core". Once you know that it refers to vsix (core vsix, django vsix, profiling vsix), then it's ok.

---
In reply to: [246822384](https://github.com/Microsoft/PTVS/pull/1605#issuecomment-246822384) [](ancestors = 246822384)

zooba:
A few more changes, but this now supports the min-bar for the next release (open, F5, Ctrl+F5, run with web browser).

I need to write tests still.

Python/Product/PythonTools/PythonTools/WebProject/PythonWebLauncher.cs
```diff
@@ -108,16 +93,10 @@ LaunchConfiguration defaultConfig
             if (debug) {
                 _pyService.Logger.LogEvent(Logging.PythonLogEvent.Launch, 1);
 
+                config.LaunchOptions[PythonConstants.WebBrowserUrlSetting] = url.AbsoluteUri;
                 using (var dsi = DebugLaunchHelper.CreateDebugTargetInfo(_serviceProvider, config)) {
                     dsi.Launch();
                 }
-
-                var debugger = (IVsDebugger)_serviceProvider.GetService(typeof(SVsShellDebugger));
```
huguesv: why is this being removed? we don't start browser anymore when debugging, or is this being taken care of elsewhere?

zooba: It's handled by the debugger when the web browser property is passed through, which it is in this case.

zooba:
I'll put it in and we'll sort out any issues asap, but I really want to get it running through the CI system.

PR #1607, opened 2016-09-14
Add more cookiecutter commands.

huguesv:
The one that was under Other Windows is moved up:
image

New:

image

In PTVS menu:

image


huguesv:
image

zooba:
I'd drop "Add" and "Create" from the start of the items and add "..." to both of those (but not the View or Python Tools ones). We should probably also aim for a later group in the Add context menu.

huguesv:
I fixed the text, still trying to decide on the location for the context menu.

It appears that the ordering of the items in that menu go from very generic to very specific, and cookiecutter is as generic as New Item.

Though if we are being realistic with regards to how popular the command will be, maybe putting it in its own group below `Analyzer...` would be more appropriate.

zooba:
Yeah, let's be realistic about the placement. Below Analyzer is fine by me.

eclectir:
I think it is fine where it is now. From a functional perspective, you are just adding files/items to your project. The difference is that you are using a template.The section the analyzer in has more references to "references".  I'd say we stick with this position, and change the verbage to say "Add from Cookiecutter Template".

PR #1622, opened 2016-09-15
Fix cookiecutter signing

zooba:
Fix cookiecutter signing
Deletes annoying unnecessary files.


huguesv:
core and workspace dlls too?

zooba:
I thought I'd already done workspace, but apparently not. Core's DLL never ships, so we don't need to sign it.

PR #1624, opened 2016-09-15
Fix 1620: cookiecutter templates failing to clone

huguesv:
Resolve symbolic link in the git folder path.

In case of a process error, always print out the executable and exit code.

Check strictly for exit code 0 for our python scripts. Cookiecutter will return positive value for error. We now have much better error reporting for iknite/cookiecutter-ansible-role template which fails during post hook, and other templates now report errors when they were ignored before.


zooba:
👍

zooba:
Wanted to merge my logging fix first so I can see why the test run was failing so bad.

PR #1631, opened 2016-09-15
Fixes #1601: async for/with not formatted correctly

### LINKED ISSUE #1601: Formatting document (ctrl-k,d) removes async keyword in front of "async for"
Hi,

we are observing some very peculiar behavior while using the 'Format Document' function of Visual Studio in a Python project. The function removes the `async` keyword and any single line comment before it from in front of a `async for` statement, even though it recogizes async as a keyword. This can be summarized by the code snippets below:

Before CTRL-K,D
```python
logger.debug('some code here')
#Some comment
async for event in device.async_read_loop():
    logger.debug('dispatch event')
```

After CTRL-K,D
```python
logger.debug('some code here') for event in device.async_read_loop():
    logger.debug('dispatch event')
```

using :
```
Visual Studio 2015 Update 3 (v14.0.25425.01)
Python Tools 2.2.4
```

Best regards,

zooba:
Fixes #1601: async for/with not formatted correctly
Enables round-tripping of async for/with statements
Adds tests

### DIFF
--- a/Python/Product/Analysis/Parsing/Ast/ForStatement.cs
+++ b/Python/Product/Analysis/Parsing/Ast/ForStatement.cs
@@ -89,6 +89,10 @@ namespace Microsoft.PythonTools.Parsing.Ast {
 
         internal override void AppendCodeStringStmt(StringBuilder res, PythonAst ast, CodeFormattingOptions format) {
             format.ReflowComment(res, this.GetProceedingWhiteSpace(ast));
+            if (IsAsync) {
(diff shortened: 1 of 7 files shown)


huguesv:
👍 

Why do you validate the results 50 times?  Did you sometimes have false positive?

zooba:
My new test doesn't actually expect to see a change, but if it's broken then it changes. However, on the first time through we see the correct (unchanged) code and the test passes even without the fix.

PR #1632, opened 2016-09-15
Fix description not being initialized.

huguesv:
Fixes https://github.com/Microsoft/PTVS/issues/1630

Test to be added in a later PR.  It requires having git clones as part of test data, still trying to decide on how to arrange it (each template needs its .git folder to be checked in, which is over 20 files).


zooba:
Can you mock out the git part of it completely?

PR #1640, opened 2016-09-16
Issue 1616

### LINKED ISSUE #1616: PEP 526 syntax support
[PEP 526](https://www.python.org/dev/peps/pep-0526/) adds syntax support for variable annotations.

We need to handle this syntax to avoid the parser breaking on user's code.

Supporting type annotations is still covered by #82 - this issue is just for the broken parser.

zooba:
Fixes #1616
Adds support for PEP 526 and PEP 498 syntax parsing and round-trip. Analysis is not yet implemented.

### DIFF
--- a/Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
+++ b/Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
@@ -201,6 +201,7 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
             { typeof(TupleExpression), ExpressionEvaluator.EvaluateSequence },
             { typeof(ListExpression), ExpressionEvaluator.EvaluateSequence },
             { typeof(SliceExpression), ExpressionEvaluator.EvaluateSlice },
+            { typeof(ExpressionWithAnnotation), ExpressionEvaluator.EvaluateAnnotatedExpression },
         };
 
         private static IAnalysisSet EvaluateSequence(ExpressionEvaluator ee, Node node) {
@@ -208,6 +209,14 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
             // TODO: We need to update the sequence on each re-evaluation, not just
             // evaluate it once.
             return ee.MakeSequence(ee, node);
+        }
+
+        private static IAnalysisSet EvaluateAnnotatedExpression(ExpressionEvaluator ee, Node node) {
+            var n = (ExpressionWithAnnotation)node;
+            if (n.Annotation != null) {
+                ee.Evaluate(n.Annotation);
+            }
+            return ee.Evaluate(n.Expression);
         }
 
         private static IAnalysisSet EvaluateParenthesis(ExpressionEvaluator ee, Node node) {
--- a/Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
+++ b/Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
@@ -446,6 +446,10 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
         }
 
         internal void AssignTo(Node assignStmt, Expression left, IAnalysisSet values) {
+            if (left is ExpressionWithAnnotation) {
+                left = ((ExpressionWithAnnotation)left).Expression;
+            }
+
             if (left is NameExpression) {
                 var l = (NameExpression)left;
                 if (!string.IsNullOrEmpty(l.Name)) {
--- a/Python/Product/Analysis/Parsing/Ast/ExpressionWithAnnotation.cs
+++ b/Python/Product/Analysis/Parsing/Ast/ExpressionWithAnnotation.cs
@@ -17,30 +17,38 @@
 using System.Text;
 
 namespace Microsoft.PythonTools.Parsing.Ast {
-    public class NameExpressionWithAnnotation : NameExpression {
+    public class ExpressionWithAnnotation : Expression {
+        private readonly Expression _expression;
         private readonly Expression _annotation;
 
-        public NameExpressionWithAnnotation(string name, Expression annotation) : base(name) {
+        public ExpressionWithAnnotation(Expression expression, Expression annotation) {
+            _expression = expression;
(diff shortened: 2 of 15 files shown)


Python/Product/Analysis/Parsing/Parser.cs
```diff
@@ -864,6 +954,15 @@ struct Name {
         private Statement ParseExprStmt() {
             Expression ret = ParseTestListAsExpr();
 
+            if (PeekToken(TokenKind.Colon) && _langVersion >= PythonLanguageVersion.V36) {
+                ret = ParseNameAnnotation(ret);
+                if (!PeekToken(TokenKind.Assign)) {
```
DinoV: The else should be handled here as well, otherwise you can do "foo: int = 2 = 3" which doesn't appear to be legal

Python/Tests/TestData/Grammar/VarAnnotation.py
```diff
@@ -0,0 +1,10 @@
+fob : oar
+fob : oar = 1
+
+class C:
+    fob : oar
+    fob : oar = 1
```
DinoV: Member expressions: C.fob : int = 1
Indicies: x[0] : int

Python/Tests/Analysis/AnalysisTest.cs
```diff
@@ -4301,6 +4317,34 @@ class C:
         }
 
         [TestMethod, Priority(0)]
+        public void AnnotatedAssign() {
```
DinoV: "Annotating a local variable will cause the interpreter to treat it as a local, even if it was never assigned to. Annotations for local variables will not be evaluated"
x = 1
def f():
    x: int

Type of inner x should be unknown (I suspect we fail this)

Python/Product/Analysis/Parsing/Ast/ExpressionWithAnnotation.cs
```diff
@@ -0,0 +1,64 @@
...
+        public Expression Annotation => _annotation;
+
+        public override string NodeName => "annotated expression";
+
+        internal override string CheckAssign() => null;
+        internal override string CheckAugmentedAssign() => "cannot assign to " + NodeName;
+        internal override string CheckDelete() => "cannot delete " + NodeName;
+
+        public override void Walk(PythonWalker walker) {
+            _expression.Walk(walker);
```
DinoV: if(walker.Walk(this)) {
}
walker.PostWalk(this);

PythonWalker also needs to be updated with an overload for this

Python/Tests/TestData/Grammar/VarAnnotation.py
DinoV: Member expressions: C.fob : int = 1
Indicies: x[0] : int

Python/Tests/Analysis/AnalysisTest.cs
```diff
@@ -4301,6 +4317,34 @@ class C:
...
+
+    def func(self):
+        self.abc : int = 42
+
+a = C()
+a.func()
+fob1 = a.abc
+fob2 = a.y
+fob3 = x
+";
```
DinoV: def f(val):
    print(val)

x: f(42)

val should be known to be an int
----
def f(val):
    print(val)

class C:
    y : f(42)

val should be known to be an int
---

def f(val):
    print(val)

def g():
    x : f(42)

val should have no type

Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
```diff
@@ -446,6 +446,10 @@ internal class ExpressionEvaluator {
         }
 
         internal void AssignTo(Node assignStmt, Expression left, IAnalysisSet values) {
+            if (left is ExpressionWithAnnotation) {
+                left = ((ExpressionWithAnnotation)left).Expression;
+            }
+
             if (left is NameExpression) {
                 var l = (NameExpression)left;
```
DinoV: You probably need to handle ExpressionWithAnnotation for walking somewhere in here, and evaluate the annotation if we're in a class or module scope (but not a local scope).

Python/Tests/Analysis/AnalysisTest.cs
```diff
@@ -2128,6 +2128,22 @@ public VariableLocation(int startLine, int startCol, VariableType type, string f
             AssertUtil.ContainsExactly(entry.GetTypeIdsByIndex("oar2", text.IndexOf("oar2 =")), BuiltinTypeId.Unicode);
         }
 
+        [TestMethod, Priority(0)]
+        public void StringFormattingV36() {
+            var text = @"
+y = f'abc {42}'
+ry = rf'abc {42}'
+yr = fr'abc {42}'
+fadd = f'abc{42}' + f'{42}'
```
DinoV:
```
def f(val):
    print(val)
f'abc {f(42)}'
```

val should be 42 (if you're not going to actually implement this you could leave it commented out)

Python/Tests/Analysis/ParserTests.cs
```diff
@@ -2653,6 +2669,40 @@ public class ParserTests {
             }
         }
 
+        [TestMethod, Priority(0)]
```
DinoV: Formatted string test, plus errors:
f'x={x'
f'x={!x}'

Python/Product/Analysis/Parsing/Parser.cs
zooba: Actually needs to be further below, but point taken.

Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
zooba: We actually never evaluate the LHS of an assignment statement... I wonder how much will break if I enable that...

Python/Tests/Analysis/ParserTests.cs
zooba: Currently we don't parse the content at all - we simply ignore the prefix and treat it like a regular string. (This is actually how Python handles it as well, which is suboptimal, but not trivial to solve right now...)

PR #1650, opened 2016-09-19
Fixes #1648 Fix debug-attach and profiling for Python 3.6

### LINKED ISSUE #1648: Fix debug-attach and profiling for Python 3.6
Because CPython 3.6 has different layouts of internal structures, we need to update our native code to handle it.

Otherwise, Debug Attach and profiling will not work.

zooba:
Fixes #1648 Fix debug-attach and profiling for Python 3.6
Also enables profiling and debug tests for 3.6.


huguesv:
Don't really know about the gil changes, but if tests pass then 👍

zooba:
The GIL changes were just deduplicating two variables that had identical contents.

PR #1652, opened 2016-09-21
Json version

zooba:
Updates Newtonsoft.Json version for Dev15 build and removes hard dependency on the older version.


zooba:
I've manually tested this already and it works as expected.

huguesv:
I checked the latest Preview build's PrivateAssemblies folder and the one that's there is 8.0.3, so that must be right :)

PR #1654, opened 2016-09-22
Install cookiecutter on demand

huguesv:
Our only external requirement is Python 3.3 or later.
Don't bother notifying for missing git, we know it's installed. In rare error case (like dev14), message in output window will be sufficient.
Virtual environment + cookiecutter package go in  %localappdata%/Microsoft/CookiecutterTools/env

Please check the strings/messages in missing dependencies xaml, and in strings.resx, I'm open to more detailed or better strings.

I've updated the feed so we don't recommend third party templates we haven't really tested.

### DIFF
--- a/Python/Product/Cookiecutter/Model/CookiecutterClientProvider.cs
+++ b/Python/Product/Cookiecutter/Model/CookiecutterClientProvider.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Microsoft.CookiecutterTools.Interpreters;
 
@@ -36,7 +37,10 @@ namespace Microsoft.CookiecutterTools.Model {
 
         private static CookiecutterPythonInterpreter FindCompatibleInterpreter() {
             var interpreters = PythonRegistrySearch.PerformDefaultSearch();
-            var compatible = interpreters.Where(interp => interp.Configuration.Version >= new Version(3, 3)).FirstOrDefault();
+            var compatible = interpreters
+                .Where(x => File.Exists(x.Configuration.InterpreterPath))
+                .OrderByDescending(x => x.Configuration.Version)
+                .FirstOrDefault(x => x.Configuration.Version >= new Version(3, 3));
             return compatible != null ? new CookiecutterPythonInterpreter(compatible.Configuration.InterpreterPath) : null;
         }
     }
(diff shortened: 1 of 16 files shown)


Python/Product/Cookiecutter/Model/CookiecutterClientProvider.cs
```diff
@@ -15,80 +15,29 @@
...
-            psi.CreateNoWindow = true;
-
-            var proc = Process.Start(psi);
-            var output = proc.StandardOutput.ReadToEnd().Trim();
-            proc.WaitForExit();
-
-            return output;
+        private static CookiecutterPythonInterpreter FindCompatibleInterpreter() {
+            var interpreters = PythonRegistrySearch.PerformDefaultSearch();
+            var compatible = interpreters.Where(interp => interp.Configuration.Version >= new Version(3, 3)).FirstOrDefault();
```
zooba: Ideally we'd get the latest version after 3.3 and also ensure that the interpreter path actually exists (I just hit a problem with registry corruption that was causing cookiecutter to fail to open).

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -165,6 +168,54 @@ class CookiecutterViewModel : INotifyPropertyChanged {
...
+
+            set
+            {
+                if (value != _isInstallingError) {
+                    _isInstallingError = value;
+                    PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(IsInstallingError)));
+                }
+            }
+        }
+
```
zooba: Inconsistent formatting

Python/Tests/CookiecutterTests/MockCookiecutterClient.cs
```diff
@@ -24,10 +24,30 @@
 
 namespace CookiecutterTests {
     class MockCookiecutterClient : ICookiecutterClient {
+        public bool CookiecutterInstalled
+        {
+            get
+            {
+                throw new NotImplementedException();
+            }
+        }
```
zooba: Inconsistent formatting

Python/Product/Cookiecutter/Model/CookiecutterClientProvider.cs
zooba: Something like:
```csharp
interpreters
    .Where(x => File.Exists(x.Configuration.InterpreterPath))
    .OrderByDescending(x => x.Configuration.Version)
    .FirstOrDefault(x => x.Configuration.Version >= new Version(3, 3)
```

PR #1661, opened 2016-09-23
Targets and references

### LINKED ISSUE #1584: Dev15 test targets do not work
We need to add a test build task that can get the right path from VS to find the extension. Our usual assumptions are no longer valid.

### LINKED ISSUE #1657: Build fails when Lightweight Project Load is enabled
Builds are always out-of-proc when this option is enabled, even if the project is loaded. Currently we depend on some VS DLLs that are not available building out-of-proc.

Currently we force the project to load on debug. Build still fails, which blocks the user with a dialog, but if they continue then debugging is fine.

zooba:
Fixes #1657 Build fails when Lightweight Project Load is enabled
Fixes #1584 Dev15 Exp Hive targets do not work

Also migrates us to build with VS 15 post-Preview 4, which is going to require a rebuild of the interactive window assemblies before we can build on the new machine. But the machine will be ready shortly so we can switch the builds when convenient.


huguesv:
👍

PR #1685, opened 2016-09-30
Fixes #1672 Double backslash in Python environments prevent from open…

### LINKED ISSUE #1672: Double backslash in Python environments prevent from opening interpreter folder
Dev 15 Preview 5, Default Data Science workload

If you click on the `C:\Program Files\\Anaconda3` in the image below, it opens the user's Documents folder.  I assume that's because it can't deal with the double backslash in the path.

FYI Anaconda was auto-detected like this.

image

zooba:
Fixes #1672 Double backslash in Python environments prevent from opening interpreter folder

Normalise paths when reading from registry.

### DIFF
--- a/Python/Product/Common/Infrastructure/PathUtils.cs
+++ b/Python/Product/Common/Infrastructure/PathUtils.cs
@@ -73,7 +73,6 @@ namespace Microsoft.PythonTools.Infrastructure {
 
             var root = EnsureEndSeparator(Path.GetPathRoot(path));
             var parts = path.Substring(root.Length).Split(DirectorySeparators);
-            bool isAbs = parts[0].EndsWith(":");
             bool isDir = string.IsNullOrWhiteSpace(parts[parts.Length - 1]);
 
             for (int i = 0; i < parts.Length; ++i) {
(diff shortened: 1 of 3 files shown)


Python/Product/Common/Infrastructure/PathUtils.cs
```diff
@@ -62,40 +62,66 @@ public static class PathUtils {
...
-                    return uri.AbsoluteUri.Replace(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+            Uri uri;
+            if (TryMakeUri(path, false, UriKind.Absolute, out uri) &&
+                uri.Scheme != Uri.UriSchemeFile) {
+                return uri.AbsoluteUri;
+            }
+
+            var root = EnsureEndSeparator(Path.GetPathRoot(path));
+            var parts = path.Substring(root.Length).Split(DirectorySeparators);
+            bool isAbs = parts[0].EndsWith(":");
```
huguesv: `isAbs` appears unused

huguesv:
Could you make sure that NormalizePath also works with network paths ie `\\myserver\myfolder\myfile.txt` ?  It doesn't look like there is a test for that case, so I'm worried about regressions.

zooba:
There are tests for that - it's in the `root` iterable. But there aren't any tests currently for `..`ing past the start, so I'll add those.

PR #1686, opened 2016-09-30
Fixes #1677 Package uninstall from solution explorer doesn't prompt t…

### LINKED ISSUE #1677: Package uninstall from solution explorer doesn't prompt to elevate
Packages tab of Python Environments prompts, but not solution explorer.

It's a problem with the default install of Anaconda from Data Science workload, which requires elevation.

zooba:
Fixes #1677 Package uninstall from solution explorer doesn't prompt to elevate

Generalises the prompt to be used from the UI hierarchy.


huguesv:
👍

PR #1688, opened 2016-09-30
Profiler issues

zooba:
Fixes search for profiler tools and improves the error message when they aren't found.

We still need to properly handle the case where it is not found - the call logic is crazy and handling the exception in the right place for all entry points is not trivial.

### DIFF
--- a/Python/Product/Profiling/Profiling/ProfiledProcess.cs
+++ b/Python/Product/Profiling/Profiling/ProfiledProcess.cs
@@ -113,6 +113,10 @@ namespace Microsoft.PythonTools.Profiling {
 
             string perfMonPath = Path.Combine(perfToolsPath, "VSPerfMon.exe");
 
+            if (!File.Exists(perfMonPath)) {
+                throw new InvalidOperationException("Cannot locate performance tools.");
+            }
+
             var psi = new ProcessStartInfo(perfMonPath, "/trace /output:" + ProcessOutput.QuoteSingleArgument(filename));
             psi.CreateNoWindow = true;
             psi.UseShellExecute = false;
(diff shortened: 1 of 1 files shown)


Python/Product/Profiling/Profiling/ProfiledProcess.cs
```diff
@@ -150,6 +155,38 @@ sealed class ProfiledProcess : IDisposable {
...
+                        if (_arch == ProcessorArchitecture.Amd64) {
+                            path = PathUtils.GetAbsoluteDirectoryPath(path, "x64");
+                        }
+                        if (Directory.Exists(path)) {
+                            return path;
+                        }
+                    }
+                }
+
+                using (var key = baseKey.OpenSubKey(@"SOFTWARE\Microsoft\VisualStudio\RemoteTools\{0}\DiagnosticsHub".FormatUI(AssemblyVersionInfo.VSVersion))) {
```
huguesv: It probably won't make much of a difference because the formatted parameter is a string, but shouldn't registry keys use invariant culture?

Python/Product/Profiling/Profiling/ProfiledProcess.cs
```diff
@@ -112,6 +113,10 @@ sealed class ProfiledProcess : IDisposable {
 
             string perfMonPath = Path.Combine(perfToolsPath, "VSPerfMon.exe");
 
+            if (!File.Exists(perfMonPath)) {
+                throw new InvalidOperationException("Cannot locate performance tools.");
```
huguesv: Will the callers catch InvalidOperationException or will this crash VS?

zooba: There are already InvalidOperationExceptions raised from this function, and clearly something is catching the Win32Exception that currently gets thrown below. I don't see any explicit catches in the call chain, so guessing it's a protected function via the commanding system.

Python/Product/Profiling/Profiling/ProfiledProcess.cs
zooba: Yes, you're right.

huguesv:
👍

PR #1697, opened 2016-10-05
Fixes extraction of files.

zooba:


zooba:
Please merge this one before #1696 as it will help our test deployment.

PR #1701, opened 2016-10-07
Fixes #1700 IronPython does not have a DB or package support

### LINKED ISSUE #1700: IronPython does not have a DB or package support
The instantiation for interpreters changed and it wasn't updated. Need to do this.

zooba:
Fixes #1700 IronPython does not have a DB or package support
Adds creation options to enable the database and pip support.


huguesv:
👍

PR #1709, opened 2016-10-10
Add event logging to cookiecutter

huguesv:


Python/Product/Cookiecutter/Telemetry/DictionaryExtension.cs
```diff
@@ -0,0 +1,42 @@
...
+                if (dict == null) {
+                    var attr = BindingFlags.Public | BindingFlags.Instance;
+                    dict = new Dictionary<string, object>();
+                    foreach (var property in data.GetType().GetProperties(attr)) {
+                        if (property.CanRead) {
+                            dict.Add(property.Name, property.GetValue(data, null));
+                        }
+                    }
+                }
+            }
```
zooba: Formatting

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -569,11 +584,14 @@ class CookiecutterViewModel : INotifyPropertyChanged {
             IsCreatingError = false;
             IsCreatingSuccess = false;
 
+            var selection = SelectedTemplate;
+            Debug.Assert(selection != null);
```
zooba: We should do something here if `selection` is null - either raise a more useful exception or abort the operation with an error for the user.

PR #1721, opened 2016-10-14
Bugfixes

alanch-ms:
- Update PythonUwpSdkNotFound dialog error message. 
- Add extra null check 
- Revert packages guids to match the guids in Dev15.


alanch-ms:
@zooba  - I tested this fix in the vspreview branch.  There's something wrong with the bcb2 branch.  I can't deploy to the PI for some reason.  I'll look into that later.

PR #1726, opened 2016-10-14
Update README.md

HHest:
Readme is out of date. I downloaded the wrong version because of this. Kept thinking 2.2 was the latest because Readme said so. :(


huguesv:
Thanks!

PR #1731, opened 2016-10-18
Cookiecutter template updating

huguesv:
Automatic update option
 - Delayed after initialization (may need to adjust those delays)
Check for update command
Update template command

Reduce code duplication in CookiecutterViewModel
 - Enum variables instead of multiple bool variables for state
 - Pass down the Redirector so that writing to output is immediate (especially good for when generating files), reduces client code needed to print to output, and simplifies the APIs that don't need to return captured output

Some non integration tests enabled, will add more over time.

### DIFF
--- a/Python/Product/Cookiecutter/Cookiecutter.cs
+++ b/Python/Product/Cookiecutter/Cookiecutter.cs
@@ -32,6 +32,8 @@ namespace Microsoft.CookiecutterTools
         public const int cmdidLinkGitHubWiki = 0x1006;
         public const int cmdidCreateFromCookiecutter = 0x1007;
         public const int cmdidAddFromCookiecutter = 0x1008;
+        public const int cmdidUpdateTemplate = 0x1009;
+        public const int cmdidCheckForUpdates = 0x1010;
         public const int ContextMenu = 0x0065;
         public const int WindowToolBarId = 0x2000;
         public const int ToolbarNavGroup = 0x3000;
--- a/Python/Product/Cookiecutter/Cookiecutter.cs
+++ b/Python/Product/Cookiecutter/Cookiecutter.cs
@@ -33,7 +33,7 @@ namespace Microsoft.CookiecutterTools
         public const int cmdidCreateFromCookiecutter = 0x1007;
         public const int cmdidAddFromCookiecutter = 0x1008;
         public const int cmdidUpdateTemplate = 0x1009;
-        public const int cmdidCheckForUpdates = 0x1010;
+        public const int cmdidCheckForUpdates = 0x100A;
         public const int ContextMenu = 0x0065;
         public const int WindowToolBarId = 0x2000;
         public const int ToolbarNavGroup = 0x3000;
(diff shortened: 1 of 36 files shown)


Python/Product/Cookiecutter/Cookiecutter.cs
```diff
@@ -32,6 +32,8 @@ internal sealed partial class PackageIds
         public const int cmdidLinkGitHubWiki = 0x1006;
         public const int cmdidCreateFromCookiecutter = 0x1007;
         public const int cmdidAddFromCookiecutter = 0x1008;
+        public const int cmdidUpdateTemplate = 0x1009;
+        public const int cmdidCheckForUpdates = 0x1010;
```
zooba: >10 [](start = 52, length = 2)

0x100A comes after 0x1009

Python/Product/Cookiecutter/CookiecutterToolWindow.cs
```diff
@@ -86,13 +84,16 @@ sealed class CookiecutterToolWindow : ToolWindowPane, IVsInfoBarUIEvents {
                 _cookiecutterControl = new CookiecutterControl(_outputWindow, CookiecutterTelemetry.Current, new Uri(feedUrl), OpenGeneratedFolder, UpdateCommandUI);
                 _cookiecutterControl.ContextMenuRequested += OnContextMenuRequested;
                 control = _cookiecutterControl;
+                _cookiecutterControl.InitializeAsync(CookiecutterPackage.Instance.CheckForTemplateUpdate).DoNotWait();
```
zooba: >.DoNotWait(); [](start = 105, length = 13)

Ideally I'd like to see a HandleAllExceptions() in here too.

Python/Product/Cookiecutter/Model/GitClient.cs
```diff
@@ -125,6 +123,59 @@ private static string TeamExplorerGitFilePath
...
         }
 
+        public async Task<DateTime?> GetLastCommitDateAsync(string repoFolderPath, string branch) {
+            var arguments = new List<string>() { "log", "-1", "--date=iso" };
+            if (!string.IsNullOrEmpty(branch)) {
+                arguments.Add(branch);
+            }
+
+            var output = ProcessOutput.Run(_gitExeFilePath, arguments, repoFolderPath, null, false, null);
+            using (output) {
```
zooba: You can combine these lines together - using(var output = ...). It helps contain the scope of `output`

Python/Product/Cookiecutter/Model/GitClient.cs
```diff
@@ -125,6 +123,59 @@ private static string TeamExplorerGitFilePath
...
+            }
+            return null;
+        }
+
+        public async Task FetchAsync(string repoFolderPath) {
+            var arguments = new string[] { "fetch" };
+            var output = ProcessOutput.Run(_gitExeFilePath, arguments, repoFolderPath, null, false, _redirector);
+            using (output) {
+                await output;
+                if (output.ExitCode < 0) {
```
zooba: This can be written as `if ((await output) < 0) {`

zooba:
:shipit:

PR #1733, opened 2016-10-18
Fixes #1732: Infinite DB refresh loop

### LINKED ISSUE #1732: Infinite DB refresh loop
Some state became inconsistent and caused VS to continually try and refresh the DB. This was because the `_generating` field was set, even though we weren't actually generating.

We should add an extra check when `IsGenerating` is set to ensure we actually are, and just reset the state if we are not.

zooba:
Fixes #1732: Infinite DB refresh loop
Checkes whether DB is actually generating when the flag is set and resets the state.


huguesv:
👍

PR #1735, opened 2016-10-19
Cookiecutter extra metadata

huguesv:
Add ability of reading description and data types from a known private variable defined in cookiecutter.json.
Display description as tooltip, fall back to default value when description is not provided.
Add a y/n combo box, may change to a checkbox later.
Fix sending of template event data for templates from local folder with no remote.


Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -131,6 +141,102 @@ class CookiecutterClient : ICookiecutterClient {
...
 
+        private static void LoadVisualStudioSpecificContext(List<ContextItem> items, JProperty vsExtrasProp) {
+            // This section reads additional metadata for Visual Studio.
+            // All fields are optional, but if they are specified, they are validated.
+            //
+            // Example:
+            // "_visual_studio" : {
+            //   "var1" : {
+            //     "description" : "description for var 1",
+            //     "type" : "yesno"
```
zooba: I prefer the term "selector" here, so that it won't be confused with the Python type of the result.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -88,24 +88,34 @@ class CookiecutterClient : ICookiecutterClient {
...
                         } else {
                             throw new InvalidOperationException(string.Format("Unsupported json element type in context file for property '{0}'.", prop.Name));
                         }
+                    } else if (prop.Name == "_visual_studio") {
+                        vsExtrasProp = prop;
                     }
                 }
+
+                if (vsExtrasProp != null) {
+                    LoadVisualStudioSpecificContext(items, vsExtrasProp);
```
zooba: Is there some way we can report errors in this section without aborting the entire template? Just fall back to regular behaviour

Python/Product/Cookiecutter/View/CookiecutterOptionsPage.xaml.cs
```diff
@@ -76,7 +77,28 @@ class TemplateContextItemTemplateSelector : DataTemplateSelector {
             var element = container as FrameworkElement;
             var p = item as ContextItemViewModel;
             if (element != null && p != null) {
-                var templateName = p.Items.Count == 0 ? "textTemplate" : "listTemplate";
+                string templateName;
+                switch (p.ValueType) {
+                    case Model.ContextItemValueType.String:
+                        templateName = "textTemplate";
```
zooba: What if we skip the enum and just pass through the selector name + "Template" as the template name? I'm just trying to minimize the changes necessary to add new ones later.

zooba: Cap the length of the selector name (30 chars seems like enough), make it lowercase and then use `Regex.IsMatch(selector, "^[a-z]+$")` to check that it's safe first.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
huguesv: Yeah, I can easily change it to WriteErrorLine to the output window.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
huguesv: Agreed

Python/Product/Cookiecutter/View/CookiecutterOptionsPage.xaml.cs
huguesv: Yeah, after I checked it in I thought I could use the enum value name directly to match the template, but passing through the original string without converting to enum is even less work ;)

PR #1738, opened 2016-10-20
Cookiecutter odbc + labels

huguesv:


Python/Product/Cookiecutter/Model/ConnectionStringConverter.cs
```diff
@@ -0,0 +1,98 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Data.Odbc;
+using System.Data.SqlClient;
+
+namespace Microsoft.CookiecutterTools.Model {
+    public static class ConnectionStringConverter {
```
zooba: Does this need to be public? Best to start private and then increase as needed.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -148,27 +148,30 @@ class CookiecutterClient : ICookiecutterClient {
...
-            // - connection
+            // - odbcConnection
             if (vsExtrasProp.Value.Type == JTokenType.Object) {
                 var vsExtrasObj = (JObject)vsExtrasProp.Value;
                 foreach (JProperty prop in vsExtrasObj.Properties()) {
                     var item = items.SingleOrDefault(ctx => ctx.Name == prop.Name);
                     if (item != null) {
                         if (prop.Value.Type == JTokenType.Object) {
                             var itemObj = (JObject)prop.Value;
+                            ReadLabel(item, itemObj);
```
zooba: If we don't specify a label, where does the value get set?

huguesv: ContextItemViewModel ctor sets it to label if not null/empty, otherwise param name.

Python/Product/Cookiecutter/Model/ConnectionStringConverter.cs
huguesv: Right, it wasn't meant to be public.

zooba:
Oops, sorry. I just created a merge conflict with my other change

PR #1749, opened 2016-10-20
Fixes #1748: Debug unit test confused about which debugger to use

### LINKED ISSUE #1748: Debug unit test confused about which debugger to use
VS 2015 Update 3 + PTVS 3.0

This happens with the azure-sdk-for-python solution. I will try to repro on another project.

It's trying to use the mixed-mode debugger, but that's turned off in my project settings.

image

image

zooba:
Fixes #1748: Debug unit test confused about which debugger to use
Uses correct boolean when parsing boolean values.


huguesv:
👍

PR #1752, opened 2016-10-20
Fixes #1751 Smart indent does not work in REPL

### LINKED ISSUE #1751: Smart indent does not work in REPL
We aren't correctly handling the difference between the line number in the projection buffer and the subject buffer in `AutoIndent.GetLineIndentation`, so the line number comparison fails and we never attempt to indent.

zooba:
Fixes #1751 Smart indent does not work in REPL
Removes old compatibility code
Removes duplicate REPLs opening during tests and tries harder to close the one we just finished using.


zooba:
Just hit merge if there are no concerns - I'll be out until tomorrow.

Issue #1727, opened 2016-10-14
Tests are not launched in VS15 preview 5


lmazuel:
Using VS15 preview 5, I can see my tests in the test explorer:
![image](https://cloud.githubusercontent.com/assets/1050156/19405208/f748be76-9229-11e6-8945-fac6c6231e18.png)

But if try to launch them, it does nothing. "Run selected tests" does nothing. "Run all" seems to think (button switch to Cancel), but nothing happens.

Got a message on TEst output

```
Cannot discover test extensions installed by NuGet. The NuGet service is not available. Tests may not be found due to missing test adapter extensions.
```

Microsoft Visual Studio Enterprise 15 Preview 5
Version 15.0.25802.0 PREVIEW
Microsoft .NET Framework
Version 4.6.01586

Installed Version: Enterprise

Visual Basic 15 Preview 5   00369-50000-00000-AA578
Microsoft Visual Basic 15 Preview 5

Visual C# 15 Preview 5   00369-50000-00000-AA578
Microsoft Visual C# 15 Preview 5

ASP.NET and Web Tools 2016   15.0.20908.0
ASP.NET and Web Tools 2016

Common Azure Tools   1.8
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

JavaScript Language Service   2.0
JavaScript Language Service

Microsoft .NET Core Tools   15.0.20908.0
Microsoft .NET Core Tools

Python Tools for Visual Studio   3.0.16266.1
Python Tools for Visual Studio provides IntelliSense, projects, templates, Interactive windows, and other support for Python developers.

Python Tools for Visual Studio - Cookiecutter Integration   3.0.16266.1
Provides cookiecutter integration.

Python Tools for Visual Studio - Django Integration   3.0.16266.1
Provides templates and integration for the Django web framework.

Python Tools for Visual Studio - Profiling Support   3.0.16266.1
Profiling support for Python projects.

TypeScript   2.0.5.0
TypeScript tools for Visual Studio

lmazuel:
In "Run all" mode, when I click "Cancel" I got:
```
========== Run test finished: 0 run (0:02:38.880) ==========
```

lmazuel:
Launching my test directly using "Start without debugging" is ok:
![image](https://cloud.githubusercontent.com/assets/1050156/19405569/bdd391e4-922d-11e6-98e9-ea31112b7854.png)

huguesv:
[...]

PR #1756, opened 2016-10-21
Fix 1727 test not executing

huguesv:
Fixes https://github.com/Microsoft/PTVS/issues/1727

It fixes some of the existing test adapter tests (unrelated to 1727).  Inheritance related tests are still failing, so I've moved them to a separate test so that the main tests `Run` and `RunAll` now pass.


zooba:
👍

PR #1761, opened 2016-10-24
Prevents RemoveItem from crashing VS when racing with another removal.

zooba:
Reported internally via crash reporting.


huguesv:
👍

Issue #1271, opened 2016-05-08
visualstudio_py_debugger on Jython 2.5.3


abbgrade:
Hello,
I have to debug a Python function, that need to run on Jython 2.5.3, because its part of Microsoft Azure HD Insight 3.2 and i cannot change it.

On CPython 2.5.3 with PTVS 2.2.3 everything is fine but on Jython 2.5.3 it failes with:

> Traceback (most recent call last):
>   File "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\Microsoft\Python Tools for Visual Studio\2.2\visualstudio_py_launcher.py", line 25, in <module>
>     import visualstudio_py_debugger as vspd
>   File "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\Microsoft\Python Tools for Visual Studio\2.2\visualstudio_py_debugger.py", line 562, in <module>
>     attach_lock = thread.allocate()
> AttributeError: type object 'org.python.modules.thread.thread' has no attribute 'allocate'
> 
> Internal error detected. Please copy the above traceback and report at
> http://go.microsoft.com/fwlink/?LinkId=293415

Thanks
Steffen

PR #1763, opened 2016-10-24
Fixing #1271 regarding debugging with Jython

erl987:
This fixes two issues that let debugging with Jython fail. Needs an review, because I am not deep in the code. Tested with Jython 2.7.0 and Python 3.5.

### DIFF
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -564,7 +564,7 @@ def should_debug_code(code):
 
     return True
 
-attach_lock = thread.allocate()
+attach_lock = thread.allocate_lock()
 attach_sent_break = False
 
 local_path_to_vs_path = {}
@@ -2318,7 +2318,7 @@ def detach_threads():
 
 def new_thread(tid = None, set_break = False, frame = None):
     # called during attach w/ a thread ID provided.
-    if tid == debugger_thread_id:
+    if tid and tid == debugger_thread_id:
         return None
 
     cur_thread = Thread(tid)    
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -1459,7 +1459,7 @@ class Thread(object):
                 # collect globals used locally, skipping undefined found in builtins
                 f_globals = cur_frame.f_globals
                 if f_globals: # ensure globals to work with (IPy may have None for cur_frame.f_globals for frames within stdlib)
-                    self.collect_variables(vars, f_globals, cur_frame.f_code.co_names, treated, skip_unknown = True)
+                    self.collect_variables(vars, f_globals, cur_frame.f_code.co_name, treated, skip_unknown = True)
             
             frame_info = None
 
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -374,9 +374,11 @@ def is_file_in_zip(filename):
     elif path.isfile(parent):
         KNOWN_ZIPS.add(parent) 
         return True     
-    else:    
+    elif path.isdir(parent):    
         KNOWN_DIRECTORIES.add(parent)
         return False
+    else:
+        return is_file_in_zip(parent)
 
 def lookup_builtin(name, frame):
     try:
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -1459,7 +1459,13 @@ class Thread(object):
                 # collect globals used locally, skipping undefined found in builtins
                 f_globals = cur_frame.f_globals
                 if f_globals: # ensure globals to work with (IPy may have None for cur_frame.f_globals for frames within stdlib)
-                    self.collect_variables(vars, f_globals, cur_frame.f_code.co_name, treated, skip_unknown = True)
+                    self.collect_variables(
+                        vars,
+                        f_globals,
+                        getattr(cur_frame.f_code, 'co_names', ()),
+                        treated,
+                        skip_unknown = True
+                    )
             
             frame_info = None
 
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -2324,7 +2324,7 @@ def detach_threads():
 
 def new_thread(tid = None, set_break = False, frame = None):
     # called during attach w/ a thread ID provided.
-    if tid and tid == debugger_thread_id:
+    if tid == debugger_thread_id:
         return None
 
     cur_thread = Thread(tid)    
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -110,7 +110,21 @@ def thread_creator(func, args, kwargs = {}, *extra_args):
 
     return _start_new_thread(new_thread_wrapper, (func, args, kwargs))
 
-_start_new_thread = thread.start_new_thread
+_thread_start_new_thread = thread.start_new_thread
+def _start_new_thread(func, args, kwargs = {}):
+    t_lock = thread.allocate_lock()
+    t_lock.acquire()
+    
+    tid = []
+    def thread_starter(a, kw):
+        tid.append(thread.get_ident())
+        t_lock.release()
+        return func(*a, **kw)
+    
+    _thread_start_new_thread(thread_starter, (args, kwargs))
+    with t_lock:
+        return tid[0]
+
 THREADS = {}
 THREADS_LOCK = thread.allocate_lock()
 MODULES = []
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -371,12 +371,12 @@ def is_file_in_zip(filename):
         return False
     elif parent in KNOWN_ZIPS:
         return True
-    elif path.isdir(parent):
+    elif path.isfile(parent):
+        KNOWN_ZIPS.add(parent) 
+        return True     
+    else:    
         KNOWN_DIRECTORIES.add(parent)
         return False
-    else:
-        KNOWN_ZIPS.add(parent)
-        return True
 
 def lookup_builtin(name, frame):
     try:
--- a/Python/Product/PythonTools/visualstudio_py_util.py
+++ b/Python/Product/PythonTools/visualstudio_py_util.py
@@ -134,7 +134,9 @@ NONE_PREFIX = to_bytes('N')
 def read_bytes(conn, count):
     b = to_bytes('')
     while len(b) < count:
-        b += conn.recv(count - len(b))
+        received_data = conn.recv(count - len(b))
+        if received_data:
+            b += received_data
     return b
 
 
--- a/Python/Product/PythonTools/visualstudio_py_util.py
+++ b/Python/Product/PythonTools/visualstudio_py_util.py
@@ -135,8 +135,9 @@ def read_bytes(conn, count):
     b = to_bytes('')
     while len(b) < count:
         received_data = conn.recv(count - len(b))
-        if received_data:
-            b += received_data
+        if received_data is None:
+            break
+        b += received_data
     return b
 
 
--- a/Python/Product/PythonTools/ptvsd/attach_server.py
+++ b/Python/Product/PythonTools/ptvsd/attach_server.py
@@ -162,7 +162,7 @@ def enable_attach(secret, address = ('0.0.0.0', DEFAULT_PORT), certfile = None, 
 
     atexit.register(vspd.detach_process_and_notify_debugger)
 
-    server = socket.socket()
+    server = socket.socket(proto=socket.IPPROTO_TCP)
     server.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     server.bind(address)
     server.listen(1)


Python/Product/PythonTools/visualstudio_py_debugger.py
```diff
@@ -2318,7 +2318,7 @@ def detach_threads():
 
 def new_thread(tid = None, set_break = False, frame = None):
     # called during attach w/ a thread ID provided.
-    if tid == debugger_thread_id:
+    if tid and tid == debugger_thread_id:
```
zooba: Just wondering what issue this resolves? Does Jython not allow you to compare None with an int for equality?

Python/Product/PythonTools/visualstudio_py_debugger.py
```diff
@@ -1459,7 +1459,7 @@ def get_frame_list(self):
                 # collect globals used locally, skipping undefined found in builtins
                 f_globals = cur_frame.f_globals
                 if f_globals: # ensure globals to work with (IPy may have None for cur_frame.f_globals for frames within stdlib)
-                    self.collect_variables(vars, f_globals, cur_frame.f_code.co_names, treated, skip_unknown = True)
+                    self.collect_variables(vars, f_globals, cur_frame.f_code.co_name, treated, skip_unknown = True)
```
zooba: This is actually a significant change - `co_name` is the name of the code object, while `co_names` is a tuple containing globals referenced by the code object. If Jython does not support `co_names`, this will need to become a `getattr(cur_frame.f_code, 'co_names', ())` (ideally on a separate line, to keep this one from getting too long)

Python/Product/PythonTools/visualstudio_py_debugger.py
erl987: When new_thread() is called, tid is None and then the comparison tid==debugger_thread_id evaluates to true and new_thread returns with a None. I assume that this is not intended.

Python/Product/PythonTools/visualstudio_py_debugger.py
erl987: Ok, actually I did not recognize that these changes regarding remote debugging go directly into a pull request. I will look on that more carefully later.

erl987:
I have corrected the mistake and I think the pull request is now ready. Debugging and remote debugging is now working with Jython 2.7.0, including single line stepping.

One final issue left: Finally I need to remote debug Jython code that is actually called by Java. It is working fine to attach to that process and to stop at breakpoints and read variable values. However, it is not possible to step by single lines - the debugger just jumps to the next breakpoint. It should be running with the same Jython library and the same Java. Do you have any clue on the reason and how to debug that?

Python/Product/PythonTools/visualstudio_py_debugger.py
zooba: `None == <anything except None>` should always return `False`, so there shouldn't be a need to check that `tid` is not None first. I'm wondering whether Jython gets this wrong somehow, or if its implementation of `thread.start_new_thread()` does not return an identifier?

Python/Product/PythonTools/visualstudio_py_util.py
```diff
@@ -134,7 +134,9 @@ def exec_module(module, global_variables):
 def read_bytes(conn, count):
     b = to_bytes('')
     while len(b) < count:
-        b += conn.recv(count - len(b))
+        received_data = conn.recv(count - len(b))
+        if received_data:
+            b += received_data
```
zooba: Receiving no data typically indicates the end of the stream - is looping the right behavior here or should we break? @int19h may have some idea.

zooba:
Have a look at `should_block_on_frame` for that behavior. What's probably happening is that we are finding a file in `DONT_DEBUG` before finding one in `DEBUG_ENTRYPOINTS`, which is basically how we differentiate between user code and non-user code.

You can temporarily modify `should_send_frame` to always return `True`, which will let you see within VS all the frames that we know about. I'm not entirely sure what you'll see, but there may be something that can be easily inserted into `DEBUG_ENTRYPOINTS` that will fix this.

Python/Product/PythonTools/visualstudio_py_util.py
int19h: For a TCP connection, AFAIK, you can get a zero-length read if the other side did a zero-length write in the middle of a stream, so it's not necessarily a reliable EOF marker. But it should still be an empty byte array in that case, and so it'd be safe to `+=` it. If this rather checks for `None`, I don't think that can happen in any case other than connection closing.

zooba: That was my assumption as well, and in that case we probably want:

```python
if received_data is None:
    break
```

Otherwise we could get stuck in an infinite loop here. If we're actually getting `b''` back then we'll still be stuck in the loop, but that was the prior behavior anyway and it hasn't been a problem.

Python/Product/PythonTools/visualstudio_py_debugger.py
erl987: I can't reproduce that issue anymore, therefore I have reverted it. Maybe some other changes had an influence.

erl987:
The mentioned problem that the debugger does not stop between breakpoints is caused by the fact, that with Jython the filenames of the models appear as: __pyclasspath__/Building.py for file in the source root directory. The causes the test for ZIP file to fail in the test for isdir(parent). One obvious solution would be:

`JYTHON_CLASSPATH = '__pyclasspath__'

def is_file_in_zip(filename):
    parent, name = path.split(path.abspath(filename))
    next_level_parent, file_directory_name = path.split(parent)
    if file_directory_name == JYTHON_CLASSPATH:
	    return False
    if parent in KNOWN_DIRECTORIES:
        return False
    elif parent in KNOWN_ZIPS:
        return True
    elif path.isdir(parent):
        KNOWN_DIRECTORIES.add(parent)
        return False
    else:
        KNOWN_ZIPS.add(parent)
        return True`

With this change, stepping line by line works normally. However, it may not be the best solution. Especially when the source tree has serveral layers. One issue to explore here, is that the debug server exits with code -1 (Das Programm "[10236] <python> @ tcp://localhost/" wurde mit Code -1 (0xffffffff) beendet.)

zooba:
I'd rather have an extra condition in `should_debug_code` - something like:

```python
if sys.platform.startswith('java') and filename.startswith('__pyclasspath__/'):
    return True
```

Looking at the zip code, we should probably also be checking `path.isfile(parent)` to detect when a file is inside a ZIP file instead of just detecting when it isn't in a directory. That would also fix this case without the extra check above, and might help others as well.

```python
def is_file_in_zip(filename):
    parent, name = path.split(path.abspath(filename))
    if parent in KNOWN_DIRECTORIES:
        return False
    elif parent in KNOWN_ZIPS:
        return True
    elif path.isfile(parent):
        KNOWN_ZIPS.add(parent)
        return True
    else:
        KNOWN_DIRECTORIES.add(parent)
        return False
```

However, part of the point of that function is to avoid sending non-filesystem paths over to VS. If `__pyclasspath__/Building.py` can't actually be opened from the filesystem, it would also suffice to add it to `KNOWN_ZIPS` at startup. Otherwise, it could be added to `KNOWN_DIRECTORIES` at startup.

```python
KNOWN_DIRECTORIES = set((None, ''))
KNOWN_ZIPS = set()
if sys.platform.startswith('java'):
    KNOWN_ZIPS.add('__pyclasspath__')  # if we cannot read files in there
    # or
    KNOWN_DIRECTORIES.add('__pyclasspath__') # if we _can_ read files in there
```

I suspect that making both of the last two changes is the best approach, but would need to test it. Think you can try these out?

erl987:
I will give them a try.
But when trying to debug in our production code, where also the Jython code is called from Java, even after fixing the problem above (in the previous raw way), another issue occurs: The debugger seems to step initially before any other file through the lines of its own file "visualstudio_py_debugger.py" and ends in a deadlock in the following acquire-call:

```
def get_thread_from_id(id):
    THREADS_LOCK.acquire()
    (...)
```

In handle_lines() I can log the following order of excecution (shown are method / file):

wait C:\Users\ralph\.m2\repository\org\python\jython-standalone\2.7.0\jython-standalone-2.7.0.jar\Lib\threading.py
__init__ <string>
write __pyclasspath__/ptvsd/visualstudio_py_debugger.py
(...)
probe_stack __pyclasspath__/ptvsd/visualstudio_py_debugger.py
probe_stack __pyclasspath__/ptvsd/visualstudio_py_debugger.py
probe_stack __pyclasspath__/ptvsd/visualstudio_py_debugger.py
probe_stack __pyclasspath__/ptvsd/visualstudio_py_debugger.py
__enter__ __pyclasspath__/ptvsd/visualstudio_py_debugger.py
get_thread_from_id __pyclasspath__/ptvsd/visualstudio_py_debugger.py

I am not sure, but possibly there is a nested call to get_thread_from_id() when debugging the method itself?

erl987:
I justed recognized that the deadlock was probably caused by other code changes during the testing. A clean current source file with your suggested changes seems to work (at least for a test case). I will try tommorrow with the production code, but I suspect it should work now.

erl987:
The deadlock occured while the debugger was debugging its own source file (in handle_line()). However, it was actually caused by importing problems of the Jython module in our own codebase. It is not great that the real reason is so obfuscated, but that is probably hard to change and no bug in that sense.

erl987:
I have refactored is_file_in_zip() as suggested and it works fine. I would not add:

```Python
KNOWN_DIRECTORIES = set((None, ''))
KNOWN_ZIPS = set()
if sys.platform.startswith('java'):
    KNOWN_ZIPS.add('__pyclasspath__')  # if we cannot read files in there
    # or
    KNOWN_DIRECTORIES.add('__pyclasspath__') # if we _can_ read files in there
```

Depending on the level of the source file in the directory tree, the file path exposed to the debugger may also look like:
C:\Users\User\Documents\code\JavaJython2\__pyclasspath__\level1\module.py

So there is no general way to determine the critical directory name in advance, and there is only a slight overhead if the method is once run completely, then it is anyway added to KNOWN_DIRECTORIES.

Python/Product/PythonTools/visualstudio_py_debugger.py
erl987: When testing now local debugging of Jython again, I detected the issue again. In fact, the expression ` tid == debugger_thread_id` evaluates in that case to `True` because also `debugger_thread_id` is `None`. This seems only to happen for local debugging. The previous change makes it running, but is not the right solution.

Python/Product/PythonTools/visualstudio_py_debugger.py
```diff
@@ -371,12 +371,12 @@ def is_file_in_zip(filename):
         return False
     elif parent in KNOWN_ZIPS:
         return True
-    elif path.isdir(parent):
+    elif path.isfile(parent):
```
zooba: I was wrong about this change - we need to keep splitting the path until we definitively have a path or a file (otherwise we may be in a subdirectory within a ZIP file, which will be false for both isdir and isfile):

```python
def is_file_in_zip(filename):
    parent, name = path.split(path.abspath(filename))
    if parent in KNOWN_DIRECTORIES:
        return False
    elif parent in KNOWN_ZIPS:
        return True
    elif path.isfile(parent):
        KNOWN_ZIPS.add(parent)
        return True
    elif path.isdir(parent):
        KNOWN_DIRECTORIES.add(parent)
        return False
    else:
        return is_file_in_zip(parent)
```

With this change everything looks to be fine against Python 2.6 and 3.5.

erl987: Seems to work also with Jython 2.7.0.

Python/Product/PythonTools/visualstudio_py_debugger.py
erl987: Did you see my comment on visualstudio_py_debugger.py? I wonder how to proceed there.

erl987:
1. Did you see my comment for visualstudio_py_debugger.py? I wonder how to proceed there.

2. For my use case, things work fine now. But I found a serious issue when performing local debugging having imports - less relevant for me but probably for others. When the debugger steps to the import, it fires several error messages. When you continue finally the debugging goes on. This does not happen in the Release mode.

When tracking that down, it happens in def exec_code(code, file, global_variables) in visualstudio_py_utils.py when compiling / executing. 

message_1
message_2
message_3
message_4

I have no clue about the reason. Based on the error messages attached (in the order of occurence until the debugger proceeds) it is somehow failing to parse through the source directories?

zooba:
> Did you see my comment for visualstudio_py_debugger.py?

Lots of comments. Which one? :)

For the exceptions during import behavior, we have a `BUILT_IN_HANDLERS` dictionary in `ExceptionBreakInfo` that lists frames we assume handle all exceptions. For example, `<frozen importlib._bootstrap>` is listed there to handle exceptions raised within Python 3's importer.

If you break at one of those exceptions you're seeing, I'd expect to see a frame on the stack that looks like import machinery. Adding it to the dictionary (with value `[(None, None, '*')]`, which means it handles all (`'*'`) exceptions on all lines) will prevent those dialogs from being shown to users.

erl987:
That one: When testing now local debugging of Jython again, I detected the issue again. In fact, the expression tid == debugger_thread_id evaluates in that case to True because also debugger_thread_id is None. This seems only to happen for local debugging. The previous change (to check for tid not None) makes it running, but is not the right solution.

zooba:
Ah, okay. That means that `_thread.start_new_thread` is returning `None` instead of a thread identifier. We might have to use `getident()` from inside the thread somehow, and block the _start_new_thread caller until we get it. Something like this:

```python
_thread_start_new_thread = thread.start_new_thread
def _start_new_thread(func, args, kwargs = {}):
    t_lock = thread.allocate_lock()
    t_lock.acquire()

    tid = []
    def thread_starter(*a, **kw):
        tid.append(thread.getident())
        t_lock.release()
        return func(*a, **kw)

    _thread_start_new_thread(thread_starter, args, kwargs)
    with t_lock:
        return tid[0]
```

I'm not entirely sure that'll work (1. because I typed it into a browser, and 2. because I don't know whether `getident()` works correctly either), but it would seem to be the only available option here.

erl987:
I have commited something following your idea. I had to do some changes to get it running. I am not sure if I made the variadic parameters correct. It is now running for Jython 2.7.0 and CPython 3.5 in local debugging as well as remote debugging. 

The issue regarding importing ist still under investigation. For some reason the line `exec(code_obj, global_variables)` in` def exec_code(code, file, global_variables)` fails with with an error without throwing an exception, therefore there is no call stack giving a hint. The same function call works in the Jython command line with the code having been exported in a text file.

erl987:
The stack frame where the import problem occurs, is in fact the Python source file itself. Excluding exceptions.ImportError and zipimport.zipimport.ZipImportError helps preventing the occurence of the unhandled errors.  I was only able to get it work by adding those exceptions to ExceptionBreakInfo and with commenting its clear-method out. So that's only suitable for tracking it down. Any suggestions on how to proceed with this last issue?

zooba:
You may need to patch `should_send_frame` again to see everything that's actually there. If the top one looks like the Python file, we're hiding some frames below that.

zooba:
I'm keen to get these changes in soon, and I'm willing to take them without the import error handling in order to get the other fixes.

Do you think you'll get a chance to figure out the rest? Or should I merge this now and we can file another issue for the unhandled import errors?

erl987:
Actually even patching `should_send_frame` does not help to see any more information on the import error traceback. So I don't know how to proceed there. I would also finish that pull request so far and separate the remaining issue. I have used the changes in the meanwhile and they seem to work well with Jython as well as CPython.

zooba:
Awesome! I've opened #1838 to track the import error issue.

PR #1769, opened 2016-10-25
Cloud Service fixes

### LINKED ISSUE #1764: Cannot create Cloud project
Our Cloud project wizard is outdated and needs to be updated for VS "15".

We shouldn't need to prompt to download and install Azure Tools, but I'm not entirely sure what we should be doing instead.

### LINKED ISSUE #1767: Cannot add new web roles
The `roletemplatedata.xml` files still specify `.zip` on the template name. Since we are no longer shipping zip files, these fail.

We should just have to remove the `.zip`.

zooba:
Fixes #1764 Cannot create Cloud project
Fixes #1767 Cannot add new web roles


zooba:
Still doing some validation, so there may be another commit or two, but this is very close to what we need.

huguesv:
👍

zooba:
Seems to be one lingering issue that probably isn't our fault, so I'll check this in and we'll keep validating.

Issue #1777, opened 2016-10-27
Build error for AnalysisBrowser on Dev 15 build 15.0.25826.0


huguesv:
```
        private static string GetTypeName(object item) {
            if (item == null) {
                return "null";
            } else if (item is string) {
                return "string";
            } else if (item is Dictionary<string, object>) {
                return "dict(str -> object)";
            } else if (item is List<Dictionary<string, object>>) {
                return "list(dict)";
            } else if (item is List<object>) {
                return "list(object)";
            } else if (item is Dictionary<string, object>[]) { // this line doesn't compile
                return "tuple(dict(str -> object))";
            } else if (item is object[]) {
                return "tuple(object)";
            } else {
                return item.GetType().Name;
            }
        }
```

I did some test in a separate console app, and it looks like we could replace that problematic condition with this:

```
typeof(Dictionary<string, object>[]).IsAssignableFrom(item.GetType())
```

It's rare to see a breaking change like this. I could file it in roslyn github, unless you happen to have talked to them already.

zooba:
I haven't spoken to them - go ahead and file it. There's no reason for that to break.

huguesv:
Filed https://github.com/dotnet/roslyn/issues/14785

int19h:
Does it work if you use an alias?
```cs
using Dict = Dictionary<string, object>[];
...
if (item is Dict) ...
```

huguesv:
No, but this compiles:
```
using Dict = System.Collections.Generic.Dictionary<string, object>;
...
if (item is Dict[]) ...
```

zooba:
This has been fixed in the compiler.

PR #1781, opened 2016-10-28
Workaround for dev15 compiler bug.

huguesv:
For https://github.com/Microsoft/PTVS/issues/1777


zooba:
Let's hold this until they acknowledge they won't be fixing the bug (but I suspect they'll fix it). For now, just unload the project so it doesn't try and build.

zooba:
This has been fixed in the compiler, so we don't need to merge this fix.

Issue #1791, opened 2016-10-31
Azure WebApp Publish failed due to env not found error


YoungjaeKim:
## Symptom

I tried to publish Flask app to Azure WebApp, but failed due to the error message below. Sadly, default Azure Flask template also failed to publish.

> A virtual environment is required to deploy dependencies to Microsoft Azure Web Sites. Either add a virtual environment or add <SuppressWebSitesVirtualEnvError>true</SuppressWebSitesVirtualEnvError> to your project if you have no dependencies other than the Python standard library.	

## Reproduce Steps

1. git pull https://github.com/azureappserviceoss/FlaskAzure.git (azure default template)
2. In Visual Studio, New > Project > From Existing Python Code > select pulled folder.
3. In setup wizard, set as Flask app and Detect virtual environment.
4. It seems failed to detenct `env` directory. Project has `Python27 (global default)` as environment.
5. Right-click `Python Environment` and click `Add Existing Virtual Environment`.
6. Select `env` directory and it shows "We could not find a base interpreter at D:\Python27" error.
7. When right-click `Python Environment` and click `Add/Remove Python Environments`, the dialog shows `env (unavailable)` in environment list. 

## Questions

1. Why does VisualStudio point `D:\Python27` in **Step 6** and how can I fix it? I do not have `D:` drive and cannot find hard-coded line for that.
1. Where can I put `<SuppressWebSitesVirtualEnvError>true</SuppressWebSitesVirtualEnvError>` in web.config? I cannot find any documentation and example.

huguesv:
Virtual environments have hard coded paths to the base python path, which can cause problems if you try to reuse them across machines like you are doing here.

I would recommend that you delete it and recreate it using PTVS.  Since there's a requirements.txt at the root, PTVS will check the "install packages" in the new virtual env dialog by default.

FYI SuppressWebSitesVirtualEnvError element is meant for the .pyproj file, not the web.config file.

zooba:
Also so you're aware, we're probably going to remove the virtualenv publishing functionality in the next version (with Visual Studio "15"). We'll recommend the process outlined at https://aka.ms/PythonOnAppService for getting started (and hopefully streamline it a bit by then, but even now we know customers are happier with that approach).

YoungjaeKim:
Hi @zooba 
[...]

PR #1794, opened 2016-10-31
Simplify Azure publishing

zooba:
This goes a long way to fixing issues like #1791, but it does it by removing most of the old functionality.

* no more auto-generated `web.config`
* no more deploying virtualenvs to App Service/Cloud Service
* no more installing from WebPI feeds
* no more `AzureSetup.exe` or `AzureSetup.cfg` (`WebRole.dll` still exists for remote debugging)
* simplified `ConfigureCloudService.ps1` script (still to be completed)
* safer `pip_downloader.py` script
* new `web.config` templates
* property changes when going from "this project is not a web/worker role" to "this project is a web/worker role"
 * previously we'd try (and fail) to detect this at build time

Still need to do a lot of testing and probably fixing stuff, but the bulk of the change is here for comment/discussion. Combined with the https://aka.ms/PythonOnAppService changes, I think this is worthwhile bringing in for 3.0.


zooba:
I need to run some testing on the cloud service scripts still, but since I'm about to change work machine I figured I'd push.

zooba:
I believe this is ready to merge now. @DinoV @huguesv if you feel like taking a break from vacation and having a review, I'd appreciate it.

Python/Product/BuildTasks/Microsoft.PythonTools.Web.targets
```diff
@@ -94,30 +93,23 @@ permissions and limitations under the License.
     </ItemGroup>
   </Target>
 
+  <Target Name="CollectFilesFromIntermediateAssembly" />
```
huguesv: Just checking that this is needed, since it's new and doesn't appear to be doing anything.

zooba: It's actually suppressing the "standard" target by the same name, which tries to get the compiled DLL and its debugging symbols.

Python/Product/BuildTasks/Microsoft.PythonTools.Web.targets
```diff
@@ -129,544 +121,18 @@ permissions and limitations under the License.
...
-
-    <ItemGroup>
-      <FileWrites Include="@(_PythonWebSitesFiles)" />
-      <FilesForPackagingFromProject Include="@(_PythonWebSitesFiles->'%(FullPath)')">
-        <DestinationRelativePath>%(_PythonWebSitesFiles.DestinationRelativeDir)%(FileName)%(Extension)</DestinationRelativePath>
-        <FromTarget>CollectPythonWebSitesFiles</FromTarget>
-        <Category>Run</Category>
-      </FilesForPackagingFromProject>
-    </ItemGroup>
+  <Target Name="_WarnAboutLegacySettings">
```
huguesv: Looks like this isn't used yet?

Python/Product/BuildTasks/Microsoft.PythonTools.targets
```diff
@@ -200,36 +200,20 @@ permissions and limitations under the License.
...
-          Returns="@(BuiltProjectOutputGroupFastOutput)"
-          DependsOnTargets="$(BuiltProjectOutputGroupFastDependsOn)">
+  <Target Name="BuiltProjectOutputGroup"
+          Returns="@(BuiltProjectOutputGroupOutput)"
+          DependsOnTargets="$(BuiltProjectOutputGroupDependsOn)">
     <ItemGroup>
-      <BuiltProjectOutputGroupFastOutput Remove="@(BuiltProjectOutputGroupFastOutput)" />
-      <_BuiltProjectOutputGroupFastOutput Remove="@(_BuiltProjectOutputGroupFastOutput)" />
+      <BuiltProjectOutputGroupFastOutput Remove="@(BuiltProjectOutputGroupOutput)" />
+      <_BuiltProjectOutputGroupFastOutput Remove="@(_BuiltProjectOutputGroupOutput)" />
```
huguesv: These 2 'FastOutput' are the only references left to Fast in the file. Is that correct?

zooba: No, they need renaming to remove the 'Fast'.

Python/Templates/Templates.csproj
```diff
@@ -74,9 +74,253 @@
     <None Include="ReadmeFiles\**\*">
       <IncludeInVsix>false</IncludeInVsix>
     </None>
-    <Content Include="Core\ItemTemplates\**\*;Django\ItemTemplates\**\*;IronPython\ItemTemplates\**\*;Web\ItemTemplates\**\*">
+    <Content Include="Core\ItemTemplates\Python\EmptyPackage\package.vstemplate">
```
huguesv: You accidentally modified the project here

huguesv: Or maybe not.

huguesv:
Let's do it 👍

zooba:
It is done!

PR #1796, opened 2016-10-31
Fix test launcher handling of import errors on Python 3.5

huguesv:
Fixes https://github.com/Microsoft/PTVS/issues/1727 for Python 3.5

Implementation of unittest changed for 3.5, they now handle import errors and create fake test cases for them.  Problem is that the implementation of id() on the fake test cases doesn't match the id that VS has for the test.


zooba:
👍

PR #1813, opened 2016-11-03
Cookiecutter git.exe location

huguesv:
Use the VS running instance location in order to locate git.exe, instead of using registry.
That's not always possible (tests, dev 14), so we still fallback to registry and global git.exe.

Don't merge yet, I still need to test on Dev 15.


huguesv:
Ok, I checked that it still works on Dev 15.

zooba:
Arguably the best way to do this is to load the Team Explorer extension and query its install location, but I doubt it's going to move anywhere.

huguesv:
I'm not worried about it moving. If they made a symbolic link, it's indication that they'll keep the path to it valid.  If not, we'll catch it.
Loading the extension may not cost anything if it's already loaded, not sure if they are auto load or not. There will still be a risk that their git subdir changes.

zooba:
👍

PR #1816, opened 2016-11-04
Fixes #1634 Fix mixed-mode debugging for 3.6

### LINKED ISSUE #1634: Fix mixed-mode debugging for 3.6
Python 3.6 almost certainly doesn't work for MMD anymore because the internals have changed so much.

zooba:
Fixes #1634 Fix mixed-mode debugging for 3.6
Enables MMD for 3.6 via eval_frame hook
Fixes PyDictObject implementation
Removes PGO block to evaluate whether it is still necessary

I want to leave the PGO checks commented out for now, in case we have to bring them back before final release. In my testing, I encountered no problems with Python 3.6 x64, which was compiled with PGO, so either the PDBs are now more valid or the code that was relying on valid FuncDebugEnd values is totally gone (I definitely removed some in the past).

I ran this change through the internal VS test infrastructure for debuggers and it passed 100%.

### DIFF
--- a/Python/Product/Common/Infrastructure/EnumerableExtensions.cs
+++ b/Python/Product/Common/Infrastructure/EnumerableExtensions.cs
@@ -63,5 +63,25 @@ namespace Microsoft.PythonTools.Infrastructure {
                 yield return new KeyValuePair<TKey, TValue>((TKey)entry.Key, (TValue)entry.Value);
             }
         }
+
+        private class TakeWhileCounter<T> {
+            private ulong _remaining;
+
+            public TakeWhileCounter(ulong count) {
+                _remaining = count;
+            }
+
+            public bool ShouldTake(T value) {
+                if (_remaining == 0) {
+                    return false;
+                }
+                _remaining -= 1;
+                return true;
+            }
+        }
+
+        public static IEnumerable<T> Take<T>(this IEnumerable<T> source, ulong count) {
+            return source.TakeWhile(new TakeWhileCounter<T>(count).ShouldTake);
+        }
     }
 }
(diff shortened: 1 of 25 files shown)


Python/Product/Debugger/DkmDebugger/Proxies/Structs/PyFrameObject.cs
```diff
@@ -74,24 +112,29 @@ public PyFrameObject(DkmProcess process, ulong address)
             return new PyFrameObject(frame.Process, framePtr);
         }
 
-        public PointerProxy<PyCodeObject> f_code {
-            get { return GetFieldProxy(_fields.f_code); }
+        public virtual PointerProxy<PyFrameObject> f_back {
```
huguesv: It doesn't look like you added any overrides for these properties you are making virtual. Do they need to be virtual?

Python/Product/Common/Infrastructure/EnumerableExtensions.cs
```diff
@@ -63,5 +63,25 @@ public static class EnumerableExtensions {
...
+            public bool ShouldTake(T value) {
+                if (_remaining == 0) {
+                    return false;
+                }
+                _remaining -= 1;
+                return true;
+            }
+        }
+
+        public static IEnumerable<T> Take<T>(this IEnumerable<T> source, ulong count) {
```
huguesv: You didn't add any calls to Take that passes a ulong.  Did you mean to change any? Otherwise this code is not used.

zooba: Ah, I probably removed them. I think it's okay - these are internal functions and we are likely to legitimately need it somewhere.

zooba:
I found some places where we needed to use the new Take override, and also one for `long`. Otherwise we were incorrectly casting values.

We're likely to hit an OutOfMemory error if someone does send back >2 billion elements of anything, but less likely than we were to fail when we incorrectly cast it.

huguesv:
👍

Issue #1787, opened 2016-10-28
New Project From Existing Code creates upgrade files


huguesv:
Dev 15 RC

Create python files using cookiecutter
New Project, From Existing Code

When done, you'll see UpgradeLog and backup folder, as if an upgrade step occurred.

PR #1819, opened 2016-11-04
Fix new project from existing code wizard

huguesv:
Fixes https://github.com/Microsoft/PTVS/issues/1787 New Project From Existing Code creates upgrade files

All Import Wizard tests pass.


Python/Product/PythonTools/PythonTools/Project/ImportWizard/ImportSettings.cs
```diff
@@ -454,7 +454,6 @@ InterpreterConfiguration config
                 new Dictionary<string, string> {
                     { MSBuildConstants.IdKey, Path.GetFileName(sourcePath) },
                     { MSBuildConstants.DescriptionKey, config.Description },
-                    { MSBuildConstants.BaseInterpreterKey, config.Id },
```
huguesv: Please confirm this is correct. This is based on what I see in .pyproj when I add a new virtual environment to an existing project.

zooba: Yep, that's correct. We don't require a base interpreter for virtual environments anymore.

PR #1821, opened 2016-11-04
Prevent UI delay caused by not returning to UI thread after downloading package data

zooba:
Prevent UI delay caused by not returning to UI thread after downloading package data
Switch to using a fwlink as the base URL for PyPI


huguesv:
👍

Issue #1817, opened 2016-11-04
[Cookiecutter Explorer] Clicking "Home" button should load default view 


eclectir:
If I search something, and then want to go to that default you, I have no way to do that other than hitting search on a blank search box or "esc". Clicking the home button is my foremost intuition, but it does not take me to the default view – I recommend that it does. Hitting “Esc” does feel very obvious.

huguesv:
I had been waiting for this feedback :)  I agree.

PR #1823, opened 2016-11-04
Reset search when navigating to home page

huguesv:
These operations navigate to the search page, clear the search term and refreshes the search results.
- Click Home on toolbar
- Click Cancel on options page
- Successful file creation

Fixes https://github.com/Microsoft/PTVS/issues/1817


zooba:
👍

PR #1835, opened 2016-11-09
Fixes #1834 Environment package window oddity

### LINKED ISSUE #1834: Environment package window oddity
Search results all show up as `/simple/<name>/`, I don't remember seeing that before.

image

Clicking the link fails with error about not finding simple distribution.

Tested on a second machine, seeing it there too.

zooba:
Fixes #1834 Environment package window oddity
Needed to change the regex to handle slightly different /simple format on pypi.org vs. pypi.python.org.


huguesv:
👍

PR #1837, opened 2016-11-09
Fixes #1836 Isolate code coverage

### LINKED ISSUE #1836: Isolate code coverage
Currently, our test adapter can do code coverage (disabled in recent builds because of an issue with the code coverage extension). However, when code coverage is not available, this causes all unit testing to be unavailable.

Since code coverage integration is an Enterprise-only feature, we need to support unit tests independently of code coverage. In practice, this means separating out the handling so that MEF can compose the test adapter assembly even when the code coverage dependency is not available.

zooba:
Fixes #1836 Isolate code coverage
Removes direct dependency on code coverage assembly
Removes IncludeCodeCoverage property

### DIFF
--- a/Python/Product/TestAdapter/TestAdapter.csproj
+++ b/Python/Product/TestAdapter/TestAdapter.csproj
@@ -34,7 +34,6 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.PythonTools.TestAdapter</RootNamespace>
     <AssemblyName>Microsoft.PythonTools.TestAdapter</AssemblyName>
-    <DefineConstants Condition="$(IncludeCodeCoverage)">$(DefineConstants);INCLUDE_CODE_COVERAGE</DefineConstants>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


Python/Product/TestAdapter/TestAdapter.csproj
```diff
@@ -103,13 +102,7 @@
...
   </Choose>
   <Choose>
-    <When Condition="$(IncludeCodeCoverage)">
-      <ItemGroup>
-        <Reference Include="Microsoft.VisualStudio.TestWindow.CodeCoverage">
-          <HintPath>$(DevEnvDir)\CommonExtensions\Microsoft\AnalyzeCodeCoverage\Microsoft.VisualStudio.TestWindow.CodeCoverage.dll</HintPath>
-        </Reference>
-      </ItemGroup>
-    </When>
+    <When Condition="$(IncludeCodeCoverage)" />
```
huguesv: Delete empty Choose + When

huguesv:
👍  with above change

PR #1839, opened 2016-11-10
Adds support for VS's telemetry reporting system.

zooba:
Adds support for VS's telemetry reporting system.
Count numbers of environments, v2 vs v3, completion DB analysis time, and installed packages.
Ensures private data is protected before reporting.

All data that is collected at any point appears in Tools, Python Tools, Diagnostic Info. Users can opt-out from sending usage information under Help, Send Feedback, Settings.

When reviewing, please check that I've protected _all_ telemetry-only code with exception handlers. It is totally unacceptable to crash the product at any point because of telemetry issues.


Python/Product/PythonTools/PythonTools/Logging/VsTelemetryLogger.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            var evt = new TelemetryEvent(logEvent.ToString());
+            var props = PythonToolsLoggerData.AsDictionary(argument);
+            if (props != null) {
+                foreach (var kv in props) {
+                    evt.Properties[PropertyPrefix + kv.Key] = kv.Value;
+                }
+            } else if (argument != null) {
+                evt.Properties[PropertyPrefix + "Value"] = argument;
```
huguesv: RTVS and cookiecutter don't prefix "Value" property name, but I'm fine with it either way.  VS will add its own set of properties which don't have a prefix, so presumably we prefix everything to avoid conflicting with those.

huguesv:
The rest LGTM.

Python/Product/PythonTools/PythonTools/Logging/VsTelemetryLogger.cs
zooba: Yeah, I'd rather prefix everything that isn't explicitly available to us.

PR #1851, opened 2016-11-14
Fixes #1850 Use pip's JSON format for list

### LINKED ISSUE #1850: Use pip's JSON format for list
The `pip list` format will be changing.

Currently we use this command to display the installed packages (it differs slightly from `pip freeze`) and rely on parsing the output with a regex. As this will someday break, we should switch to passing `--format=json` and parsing that output now.

zooba:
Fixes #1850 Use pip's JSON format for list
Uses JSON format by default, falling back on the regex if passing the JSON option fails.

### DIFF
--- a/Python/Product/VSInterpreters/PipPackageManager.cs
+++ b/Python/Product/VSInterpreters/PipPackageManager.cs
@@ -24,6 +24,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Infrastructure;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.PythonTools.Interpreter {
     class PipPackageManager : IPackageManager, IDisposable {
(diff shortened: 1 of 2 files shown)


Python/Product/VSInterpreters/PipPackageManager.cs
```diff
@@ -436,12 +447,38 @@ class PipPackageManager : IPackageManager, IDisposable {
...
                         try {
-                            if (await proc == 0) {
-                                packages = proc.StandardOutputLines
-                                    .Select(i => PackageSpec.FromPipList(i))
-                                    .Where(p => p.IsValid)
-                                    .OrderBy(p => p.Name)
-                                    .ToList();
+                            var exitCode = await proc;
+                            Debug.WriteLine(string.Join("\r\n", proc.StandardOutputLines));
+                            Debug.WriteLine(string.Join("\r\n", proc.StandardErrorLines));
```
huguesv: You could use Environment.NewLine instead of `"\r\n"` like a few lines below

huguesv:
👍

Python/Product/VSInterpreters/PipPackageManager.cs
zooba: Those were meant to be deleted...

PR #1855, opened 2016-11-14
Add ability to specify a URL on cookiecutter parameters

### LINKED ISSUE #1827: cookiecutter parameter help link
Add the ability to specify a URL for each parameter (in addition to the existing Label and Description).

When a URL is specified, we can make the Label a clickable link to that URL.

huguesv:
Fix #1827


huguesv:
image

huguesv:
Should we put a tooltip that shows the URL?  If yes, I can do it for GitHub links in the search page too.

Python/Product/Cookiecutter/View/CookiecutterOptionsPage.xaml
```diff
@@ -30,19 +30,38 @@
...
+                    <wpf:Lambda x:Key="OperationStatusFailedToVisibleOrCollapsed">(OperationStatus s) => s == OperationStatus.Failed ? Visibility.Visible : Visibility.Collapsed</wpf:Lambda>
+                    <wpf:Lambda x:Key="OperationStatusSucceededToVisibleOrCollapsed">(OperationStatus s) => s == OperationStatus.Succeeded ? Visibility.Visible : Visibility.Collapsed</wpf:Lambda>
+                    <wpf:Lambda x:Key="BoolToVisibleOrCollapsed">(bool b) => b ? Visibility.Visible : Visibility.Collapsed</wpf:Lambda>
+                    <wpf:Lambda x:Key="BoolToVisible">(bool b) => b ? Visibility.Visible : Visibility.Hidden</wpf:Lambda>
+                    <wpf:Lambda x:Key="BoolToNotVisible">(bool b) => b ? Visibility.Hidden : Visibility.Visible</wpf:Lambda>
+                    <wpf:Lambda x:Key="EmptyStringToVisibleOrCollapsed">(string s) => !string.IsNullOrEmpty(s) ? Visibility.Collapsed: Visibility.Visible</wpf:Lambda>
+                    <wpf:Lambda x:Key="EmptyStringToNotVisibleOrCollapsed">(string s) => string.IsNullOrEmpty(s) ? Visibility.Collapsed : Visibility.Visible</wpf:Lambda>
```
zooba: I'd like to switch as many of these as possible to be DataTriggers at some point, but this is fine for now.

Python/Product/Cookiecutter/View/CookiecutterOptionsPage.xaml.cs
```diff
@@ -80,6 +80,17 @@ internal partial class CookiecutterOptionsPage : Page {
...
+
+        private void OpenInBrowser_CanExecute(object sender, CanExecuteRoutedEventArgs e) {
+            var url = (string)e.Parameter;
+            e.CanExecute = !string.IsNullOrEmpty(url);
+            e.Handled = true;
+        }
+
+        private void OpenInBrowser_Executed(object sender, ExecutedRoutedEventArgs e) {
+            var url = (string)e.Parameter;
+            Process.Start(url)?.Dispose();
```
zooba: This is a security risk, given it's an arbitrary string from an untrusted source. We should at least `Uri.Parse` it and make sure the scheme is http or https, and probably use of the internal browser rather than passing the string to the shell.

zooba:
Yes, we should definitely show the URL. If we have a description, we probably want to show both (URL first).

huguesv:
FYI we can't use the internal browser because of this (taken from Dev 15 RC):
image

I'm adding the http/https validation, it will warn at template load time in the output window and skip any url that's not valid.

zooba:
👍

PR #1856, opened 2016-11-15
Detect fatal errors when cloning a cookiecutter template.

### LINKED ISSUE #1673: cookiecutter clone 'unable to create file' should be treated as error, not warning
In the case of the template below, a directory will be missing from the template.

We shouldn't allow loading it, and we probably should delete it immediately from the `.cookiecutters` folder, so they don't attempt to use the installed version later on.

This happens in the case of a folder name that isn't Windows compatible. Aside from asking the author from fixing the template, we can't make it work. But we should make that clear to the user. Right now if they don't pay attention to the output window, they may not notice the problem until later on.

```
----- Cloning template 'robinandeer/cookiecutter-pyvanguard' -----

Cloning into 'cookiecutter-pyvanguard'...
error: unable to create file {{cookiecutter.repo_name}}/tests/test_{{ cookiecutter.repo_name|replace("-", "_") }}.py: Invalid argument
fatal: cannot create directory at '{{cookiecutter.repo_name}}/{{cookiecutter.repo_name|replace("-", "_")}}': Invalid argument
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'

----- Successfully cloned template 'robinandeer/cookiecutter-pyvanguard' to 'C:\Users\huvalo\.cookiecutters\cookiecutter-pyvanguard' -----

----- Successfully loaded template 'robinandeer/cookiecutter-pyvanguard' -----
```

huguesv:
Fix #1673

### DIFF
--- a/Python/Tests/CookiecutterTests/GitClientTests.cs
+++ b/Python/Tests/CookiecutterTests/GitClientTests.cs
@@ -1,0 +1,46 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.IO;
+using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


Python/Tests/CookiecutterTests/GitClientTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.CookiecutterTools.Model;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TestUtilities;
+
+namespace CookiecutterTests {
+    [TestClass]
+    public class GitClientTests {
+        private const string GitHubWindowsIncompatibleRepoUrl = "https://github.com/huguesv/cookiecutter-pyvanguard";
```
zooba: Can we avoid this dependency? What if we zipped up the repo and checked it in, then extracted it locally for this test?

zooba: Or even just recreate the repo automatically for the test

zooba:
👍 unless you update the test.

Python/Tests/CookiecutterTests/GitClientTests.cs
huguesv: Manipulating the repo is difficult on windows machine.  I filed an item for it here: https://github.com/Microsoft/PTVS/issues/1862

PR #1869, opened 2016-11-15
Fixes #1611 cookiecutter should default to user's projects path

### LINKED ISSUE #1611: cookiecutter should default to user's projects path
The one you can set on the Tools, Options, Projects and Solutions, General page. IIRC, the IVsSolution service can provide this directory.

zooba:
Fixes #1611 cookiecutter should default to user's projects path
Sets a default project path when moving to the template.

We still need to add support for using the current project's path when "Add new item" entry point is used. I'll file a new issue for that.

### DIFF
--- a/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
+++ b/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
@@ -21,14 +21,19 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.CookiecutterTools.Infrastructure;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell.Interop;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.CookiecutterTools.Model {
     class CookiecutterClient : ICookiecutterClient {
+        private readonly IServiceProvider _provider;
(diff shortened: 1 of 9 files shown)


Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -141,6 +147,26 @@ class CookiecutterClient : ICookiecutterClient {
             MoveToDesiredFolder(outputFolderPath, tempFolder);
         }
 
+        public Task<string> GetDefaultOutputFolderAsync(string shortName) {
```
huguesv: This doesn't quite fit in with the rest of CookiecutterClient's responsibilities, which is to load/execute templates (interfacing with the python package).  As long as we don't start dealing with VS projects and the like, it can stay here, and move it out when necessary.

huguesv:
👍

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
zooba: It'll have to become some sort of query event then, or a new object that gets passed into the view model, as it doesn't make sense to have this function anywhere else in the current object model. (The problem being that users can change their project path while VS is running - if it were static then it wouldn't matter.)

PR #1874, opened 2016-11-17
Cookiecutter default list selection

### LINKED ISSUE #1799: cookiecutter should have default list selection
Either we default to the first item in the list, or we add an item to the `_visual_studio` element to specify the default selection.

huguesv:
Fix #1799

Also updates git clone implementation to fix tests that were passing a mock redirector, ending up with null StandardErrorLines.

By the way, cookiecutter always uses the first entry in a list as the default.

I considered having an extra combo box item (first in list) for selecting default value (basically treat it as 'unset'), but since that would always be the same as the second item (the true first item in the list), there didn't seem to be much benefit.


zooba:
👍

PR #1876, opened 2016-11-17
Send PII values properly using TelemetryPiiProperty.

### LINKED ISSUE #1861: Switch cookiecutter event data to use TelemetryPiiProperty
For consistency with PTVS.

huguesv:
Fix #1861

I double-checked the data types (using the debugger) in the properties dictionary at the point where TelemetrySession.PostEvent is called, to make sure it's sending the class instance, not a string representation.


zooba:
👍

PR #1879, opened 2016-11-18
Fixes #1739 Analysis does not restart if Python is installed while VS is running

### LINKED ISSUE #1739: Analysis does not restart if Python is installed while VS is running
The analyzer process fails if it has no copies of Python available, and then does not start when you add one.

This also leads to unit tests not displaying, colorization not working correctly, and obviously no completions.

zooba:
Fixes #1739 Analysis does not restart if Python is installed while VS is running
Fixes notifications when going from "no interpreters" to "global default" interpreter
Fixes analyzer not starting when no interpreters are available
Fixes errors being swallowed by analyzer
Fixes thread safety issues when loading the default interpreter
Excludes some logged events from telemetry collection
Excludes (slow) Event Log data from diagnostics when Shift is held

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
@@ -175,12 +175,14 @@ namespace Microsoft.PythonTools.Project {
         }
 
         private void OnInterpreterRegistryChanged(object sender, EventArgs e) {
-            // Check whether the active interpreter factory has changed.
-            var fact = InterpreterRegistry.FindInterpreter(ActiveInterpreter.Configuration.Id);
-            if (fact != null && fact != ActiveInterpreter) {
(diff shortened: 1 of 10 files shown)


Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -276,11 +277,10 @@ internal class PythonProjectNode :
         }
 
         private void GlobalDefaultInterpreterChanged(object sender, EventArgs e) {
-            // This event is only raised when our active interpreter is the
-            // global default.
-            var evt = ActiveInterpreterChanged;
-            if (evt != null) {
-                evt(this, EventArgs.Empty);
+            if (_active == null) {
```
huguesv: `_active` could use a better name at some point (I realize you didn't introduce it in this PR).

huguesv:
👍

PR #1895, opened 2016-11-30
Fixes #1394 Test adapter should not list all tests on command line

### LINKED ISSUE #1394: Test adapter should not list all tests on command line
With the new test adapter we can run many tests in a single process. Currently, they are each listed individually on the command line.

This is very likely to run out of command line space, as we only get 32K chars. We should dump tests into a file and parse them instead (or reuse an existing file that they're already listed in?)

zooba:
Fixes #1394 Test adapter should not list all tests on command line
Adds support for passing tests in a plain-text file.


Python/Product/TestAdapter.Executor/TestExecutor.cs
```diff
@@ -573,6 +605,12 @@ out testMethod
                             }
                         }
                     }
+                    if (File.Exists(testList)) {
+                        try {
```
huguesv: Probably should do this in a finally block

huguesv:
👍

Python/Product/TestAdapter.Executor/TestExecutor.cs
zooba: It's deliberate. If tests fail, I want to be able to find the file later on. It's in TEMP, so it'll be cleaned up eventually anyway.

PR #1899, opened 2016-12-01
Django template update

### LINKED ISSUE #1890: Django template needs updating
Looks like we still have some old Django in our templates.

These must be reviewed and updated before release.

huguesv:
Fix #1890 Django template needs updating

This updates the polls sample based on the version in PTVS-samples repo.  I took care of not reverting any updates we already had (.pyproj changes for 3.0).

I looked at all the other django templates, and could not see any using old django code.

By the way, I compared django-manage createproject/createapp between django 1.9 and 1.10, and the only difference is is comments/urls where 1.9 is replaced with 1.10.


zooba:
👍

PR #1900, opened 2016-12-01
Fix reference to django and empty web role templates.

### LINKED ISSUE #1898: Empty Web Role and Django Web Role templates don't work
New Project, Azure Cloud Service
If you pick either Empty or Django Web roles, it will fail:
```
---------------------------
Microsoft Visual Studio
---------------------------
The project template referenced by the roletemplatedata.xml file cannot be found.
---------------------------
OK   
---------------------------
```

Bottle and Flask Web roles are ok, Worker role is ok.

I have a fix in progress.

huguesv:
Fix #1898 Empty Web Role and Django Web Role templates don't work


zooba:
Make sure we don't have any other references to these by name (thinking specifically in unit tests, but a full text search is going to be the easiest way to check anyway).

huguesv:
The only hits are in the old msi. Fix it or leave it?  I don't even know if that still builds.

zooba:
It should still build, but it's not of huge value. Let's fix them so that we don't leave it too far from working.

huguesv:
Updated installer.

zooba:
LGTM

PR #1902, opened 2016-12-02
Fixes #1897 Python Environments only shows one venv

### LINKED ISSUE #1897: "Python Environments" view only show one venv in case of same venv name in two projects
Being one solution with two projects, each one having its own venv named the same:
image

Changing to "Python Environments" view shows only one (in this case, PythonApplication2)
image

Microsoft Visual Studio Enterprise 2015
Version 14.0.25431.01 Update 3
Microsoft .NET Framework
Version 4.6.01586

Installed Version: Enterprise

Python Tools for Visual Studio   2.2.40825.00

zooba:
Fixes #1897 Python Environments only shows one venv in case of same venv name in two projects
Updates ordering function to sort by ID if descriptions are equal.
Displays project name for the company for virtual environments.
Fixes oddly named function.

### DIFF
--- a/Python/Product/EnvironmentsList/ToolWindow.xaml.cs
+++ b/Python/Product/EnvironmentsList/ToolWindow.xaml.cs
@@ -549,10 +549,28 @@ namespace Microsoft.PythonTools.EnvironmentsList {
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


huguesv:
👍

PR #1946, opened 2016-12-05
Custom environment fixes

### LINKED ISSUE #1906: New custom environment does not land user in configure
image

### LINKED ISSUE #1943: Creating Python environment forcibly prepends it with "VisualStudio".  Editing re-prepends.
Repro steps:
1. Create a new Python environment
2. Rename it to something "Foo"
3. Configure it properly, click 'Apply'

Note the environment is called 'VisualStudio Foo'
4. Edit the environment, Click Apply
5. Note a new environment called 'VisualStudio VisualStudio Foo', along with the old one.

Expected:
Environment should be called 'Foo'.  Editing it should not create a new one with a new name.

### LINKED ISSUE #1944: Changing environment library path does not toggle Apply button/link state
Repro steps:
1. Create a new environment
2. Configure it, including with a library path.
3. Click 'Apply'
4. Re-configure the environment editing the library path to a different directory.
5. Note 'Apply' is disabled.
6. Change the Python version
7. Note 'Apply' is now enabled.

Expected:
Changing the library path should allow the configuration change to be saved.

zooba:
Fixes #1906 New custom environment does not land user in configure
Fixes #1943 Creating Python environment forcibly prepends it with "VisualStudio"
Fixes #1944 Changing environment library path does not toggle Apply button

Reworks how environments are created so that we don't eagerly save any details. Allows much easier cancellation.
Removes the UI for Library Path
Fixes the registry key where we save the display name.
Auto-detect will now fill in the description
Adds a button to the Overview tab to jump to the configure panel.

### DIFF
--- a/Python/Product/Analysis/Interpreter/InterpreterArchitecture.cs
+++ b/Python/Product/Analysis/Interpreter/InterpreterArchitecture.cs
@@ -15,6 +15,9 @@
 // permissions and limitations under the License.
 
 using System;
+using System.Diagnostics;
+using System.Reflection;
+using Microsoft.PythonTools.Infrastructure;
 
 namespace Microsoft.PythonTools.Interpreter {
     public abstract class InterpreterArchitecture : IFormattable, IComparable<InterpreterArchitecture> {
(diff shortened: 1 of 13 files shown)


huguesv:
:shipit:

zooba:
I have a few test fixes still to come - it relied on the command that I removed.

Issue #1926, opened 2016-12-02
DS sample breaks because UCI has moved their dataset


smortaz:
the new location is:

http://mlr.cs.umass.edu/ml/machine-learning-databases/spambase/spambase.data

---

============
Example Data
============
The example is from http://archive.ics.uci.edu/ml/datasets/Spambase
It contains pre-processed metrics, such as the frequency of certain
words and letters, from a collection of emails. A classification for
each one indicating 'spam' or 'not spam' is in the final column.
See the linked page for full details of the data set.

This script uses three classifiers to predict the class of an email
based on the metrics. These are not representative of modern spam
detection systems.
'''

# Remember to update the script for the new data when you change this URL
URL = "http://archive.ics.uci.edu/ml/machine-learning-databases/spambase/spambase.data"

zooba:
As a quick fix we can switch the URL in the template. Long term we need to find a dataset that we can mirror somewhere so it won't disappear on us.

zooba:
The PR above is the quick fix. Leaving this open so that we can move to our own data sets.

huguesv:
We're closing old issues with no recent activity. If this is still important, please open a new issue.

PR #1950, opened 2016-12-06
Fixes the classifier and clustering templates with the new URL.

zooba:
This is the quick fix for #1926.

We still need to figure out why the regression sample is broken (seems like a temporary server error rather than a broken link), and really ought to switch these samples over to our own datasets so that we know they won't disappear.


huguesv:
👍

PR #1955, opened 2016-12-08
Fixes #1951 Cookiecutter treeviewitem Names are not accessible to Screen readers

### LINKED ISSUE #1951: Cookiecutter treeviewitem Names are not accessible to Screen Readers
The cookiecutter treeview does not properly define AutomationProperties for its items, so Windows UIAutomation api exposes the type name of the viewmodels backing the items. This prevents users relying on assistive technology (such as screen readers) from knowing what those items are.

zooba:
Fixes #1951 Cookiecutter treeviewitem Names are not accessible to Screen Readers
Fixes UI automation properties in cookiecutter and environments list.
Also make it so pressing Enter in the search box will start installing based on the current text.


huguesv:
👍

PR #1956, opened 2016-12-08
Fixes #1916 Exception popup doesn't show exception name

### LINKED ISSUE #1916: Exception popup doesn't show exception name
The new VS exception popup doesn't show exception type name as part of the message:

image

As many exceptions (e.g. from failed asserts) don't include the message text, either, the dialog becomes rather useless, since neither the type nor the descriptive message is visible.

image

For comparison, here's how this looks for F#:

image

zooba:
Fixes #1916 Exception popup doesn't show exception name
Provides detailed information to the new exception dialog in VS 2017
Adds a new (optional) exception message format to ptvsd.

This is still in validation, but feel free to weigh in with comments.


zooba:
@int19h may have some interest.

int19h:
:+1:

Issue #1870, opened 2016-11-15
cookiecutter should default to current project path when entering from Add New Item


zooba:
We have multiple entry points into the cookiecutter Explorer window. If you come in from the Add New Item entry point (rather than New Project), we should default to installing the template into the current project directory. We can keep installing into that directory until another entry point is used.

We probably also need to display a little note underneath the path ("We chose your current project to add new items" or something), otherwise it may not be immediately obvious where we're going to put the files.

huguesv:
Also, the generated files should be added to the project. I guess do it automatically when Create succeeds. Don't know how that affects Open in Solution Explorer link when we return to the home page. Doesn't make a lot of sense to have it in that scenario.

huguesv:
Since this isn't a modal dialog, the user can leave the tool open for a long period and forget that it was opened with a "Add From".  They may have deleted the parent folder that was there when invoked, and the project may have been closed.

zooba:
We can detect that the directory is gone and reset. We can probably also handle the case where the project is closed or the workspace is changed too. But I think having an info notice next to the path is the best way to deal with this.

Issue #1768, opened 2016-10-25
Add from cookiecutter appears on web/worker role elements


zooba:
Add a Cloud project and add a role to it. When you right-click `CloudService1\Roles\WebRole1` in Solution Explorer you get the option to "Add from cookiecutter".

The menu works fine, so it's not a major issue, but we should not be showing up here if possible. No other Add items appear.

PR #1958, opened 2016-12-08
Add From Cookiecutter

huguesv:
Fixes https://github.com/Microsoft/PTVS/issues/1870 cookiecutter should default to current project path when entering from Add New Item

Fixes https://github.com/Microsoft/PTVS/issues/1768 Add from cookiecutter appears on web/worker role elements
 - the command is disabled for cloud service project itself as well
 - when the command is disabled, it is also invisible to match behavior of the other Add commands

Also gets rid of the warning when output folder is not empty.  When we're about to overwrite a file that already exists, we check the contents first and make a backup if the new file is different.

Finally, this also makes Add From Cookiecutter actually add the item to your project, since that's where the command shows up.  This works to varying degrees depending on the target project type. 
- C# and Python appear to be working well.
- C++ doesn't support real folders, so all items appear at the root.
- F# project system appears to be a bit buggy regarding folders, so I've put in some extra code that seems ok.
- R -- the command doesn't appear in context menu there, will have to look into it.
- Cloud Service - command is now disabled for that project type

With this functionality, someone that wants a project (instead of relying on folder view) can:
- Create an empty python project (Python Application), delete any unused .py file
- Add From Cookiecutter, pick anything you like
- You now have a project and didn't have to go through Project From Existing Code.

Edit: Tested on VS 2017.
Edit2: Tests added for backup, and for add from template (though not for the project system interacting parts).


Python/Product/Cookiecutter/Commands/CreateFromCookiecutterCommand.cs
```diff
@@ -26,7 +26,7 @@ class CreateFromCookiecutterCommand : Command {
         }
 
         public override void DoCommand(object sender, EventArgs args) {
-            CookiecutterPackage.Instance.ShowWindowPane(typeof(CookiecutterToolWindow), true);
+            CookiecutterPackage.Instance.NewCookiecutterSession();
```
zooba: Again, passing in an IServiceProvider or package instance is better than using the global.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -19,6 +19,7 @@
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
```
zooba: Is this being used? I don't see where

Python/Product/Cookiecutter/Model/ProjectSystemClient.cs
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+            return null;
+        }
+
+        public void AddToProject(ProjectLocation location, CreateFilesOperationResult creationResult) {
+            string folderPath = location.FolderPath;
+            string targetProjectUniqueName = location.ProjectUniqueName;
+
+            var project = FindProject(targetProjectUniqueName);
```
zooba: Telemetry for project kind?

PR #1959, opened 2016-12-08
Updates branding to remove "Python Tools for Visual Studio" name.

zooba:
Updates branding to remove "Python Tools for Visual Studio" name.
Moves more strings into common resource file.
Removes ML project
Removes IPyTools detection.


huguesv:
:shipit:

PR #1962, opened 2016-12-09
Issue 1911

### LINKED ISSUE #1911: Default environment does not stick.
image

zooba:
Fixes #1911 Default environment does not stick
Fixes issue where we were recreating environment views too often.
Simplifies default interpreter handling to avoid cyclic and repeated event triggers.

### DIFF
--- a/Python/Product/Common/Infrastructure/ObservableCollectionExtensions.cs
+++ b/Python/Product/Common/Infrastructure/ObservableCollectionExtensions.cs
@@ -60,5 +60,50 @@ namespace Microsoft.PythonTools.Infrastructure {
                 left.Insert(index, item);
             }
         }
+
+        public static void Merge<TLeft, TRight, TKey>(
+            this ObservableCollection<TLeft> left,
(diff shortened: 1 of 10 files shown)


Python/Product/Common/Infrastructure/ObservableCollectionExtensions.cs
```diff
@@ -60,5 +60,50 @@ IComparer<T> compareSortKey
                 left.Insert(index, item);
             }
         }
+
+        public static void Merge<TLeft, TRight, TKey>(
```
huguesv: there's a couple of copies of this file, don't know if you want to keep them in sync or not.

huguesv:
👍

Python/Product/Common/Infrastructure/ObservableCollectionExtensions.cs
zooba: Yeah, I do. I'll copy into the other one and then push.

PR #1964, opened 2016-12-09
Fixes #1939 MMD breakpoint in Python file gives weird call stack

### LINKED ISSUE #1939: MMD: breakpoint in Python file gives weird call stack
Python 3.5.2 32-bit.
```py
import sys
import module1

print('before')
module1.example(module1, 123)
print('after')
```
Tried this with breakpoint on either the first print, or on the function call after. Either way, when the breakpoint is hit, the stack looks like this:
image
and in the editor, you get:
image

However, Python debugger is actually active, and in particular, you may Step In (F11). If you do, everything is suddenly back to normal - the stack now looks like this:
image

and the editor shows the correct line:
image

The Python frame on the stack is also correct, and double-clicking it shows the appropriate line in .py file.

Furthermore, if you now Step Out of native code, you get the expected behavior also:
image

image

zooba:
Fixes #1939 MMD breakpoint in Python file gives weird call stack
Ensures the symbols for Microsoft.PythonTools.Debugger.Helper are included.

### DIFF
--- a/Python/Setup/swix/Packages/Microsoft.PythonTools.Core.Vsix.swixproj
+++ b/Python/Setup/swix/Packages/Microsoft.PythonTools.Core.Vsix.swixproj
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <Package Include="file_associations.swr" />
+    <Package Include="debug_symbols.swr" />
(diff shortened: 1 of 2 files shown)


huguesv:
👍

PR #1968, opened 2016-12-10
Fixes #1864 Missing environments are no longer show in solution explorer

### LINKED ISSUE #1864: Missing environments are no longer shown in solution explorer
If you open a project that includes some environments that are not installed on the current machine, they do not appear in the list.

We should be displaying "(unavailable)" entries so they can be activated/removed still.

zooba:
Fixes #1864 Missing environments are no longer show in solution explorer
Switches to using IDs in a number of places in the UI so that we can support missing environments
Massively simplifies the view objects for environments in certain dialogs
Allows removing missing environments from a project

### DIFF
--- a/Python/Product/Common/Strings.Designer.cs
+++ b/Python/Product/Common/Strings.Designer.cs
@@ -670,6 +670,15 @@ namespace Microsoft.PythonTools {
         }
         
         /// <summary>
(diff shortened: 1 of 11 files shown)


huguesv:
👍

PR #1977, opened 2016-12-13
Fix the Open in Solution Explorer link disappearing

### LINKED ISSUE #1974: Cookiecutter automatic check for updates may be clearing the open in solution explorer section
On first time use, I left the machine idle after generating my first cookiecutter project, and when I came back it was on the home page but the Open in Solution Explorer link wasn't there.

Shahrokh saw this during the bug bash as well.  This is on the public RC update.

huguesv:
Fix #1974 Cookiecutter automatic check for updates may be clearing the open in solution explorer section


zooba:
👍

PR #1980, opened 2016-12-14
Fixes #1903 code snippet 'main' not working via shortcut

### LINKED ISSUE #1903: code snippet 'main' not working via shortcut
I used to be able to type `main` and hit TAB, and it would trigger the code snippet.

That still works for some other snippets, such as `while`, `except`, but for some reason it no longer works for `main`.  It doesn't show up in the completion list either like the other snippets.

Note that I can still add it via the Insert Snippet command.

zooba:
Fixes #1903 code snippet 'main' not working via shortcut
Adds code snippet items into completion list
Removes unused options from PythonToolsService
Makes option initialization lazier to reduce startup time


huguesv:
👍

PR #1997, opened 2016-12-27
Fixes #1948 Crash on Go To Definition

### LINKED ISSUE #1466: Question: F12 doesn't take me to definition of function but rather proposes a choice
The choice is not on different implementations but usage of the module and definition of function. Would be nice if it simply could jump to definition straight away.

Thanks,
--rustam

### LINKED ISSUE #1948: Crash on Go To Definition (F12)
Using PTVS 3 preview, on VS2017 RC, clicking F12 or Right-click + Go To Definition, results in a crash of visual studio whenever the definition is not in the same file.

zooba:
Fixes #1948 Crash on Go To Definition
Checks for nulls and quietly ignores them (with an assert in debug builds)

Fixes #1466 F12 doesn't take me to definition of function but rather proposes a choice
Stops creating Definitions for from...import names.


huguesv:
👍

PR #2001, opened 2016-12-28
Fixes #1909 Cancelling elevation during uninstall throws error

### LINKED ISSUE #1909: Cancelling elevation during uninstall throws error
System.ComponentModel.Win32Exception (0x80004005): The operation was canceled by the user
   at System.Diagnostics.Process.StartWithShellExecuteEx(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at Microsoft.PythonTools.Infrastructure.ProcessOutput..ctor(Process process, Redirector redirector)
----- Error while uninstalling 'azure' -----

zooba:
Fixes #1909 Cancelling elevation during uninstall throws error
Adds an elevation manifest to RunElevated and improves logging of Win32Exception messages.
Fixes "installing" string displayed when uninstalling packages.

### DIFF
--- a/Python/Product/RunElevated/app.manifest
+++ b/Python/Product/RunElevated/app.manifest
@@ -1,6 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
-  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
   <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
     <security>
       <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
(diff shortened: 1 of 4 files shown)


Python/Product/RunElevated/app.manifest
```diff
@@ -0,0 +1,40 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
```
huguesv: From what I understand of the doc, this is required but only really matters when other assemblies need to specify a dependency on this assembly.
Given that, maybe we can just fix the name and major/minor version without worrying about build number?

huguesv:
👍  up to you if you want to fix assembly identity.

Python/Product/RunElevated/app.manifest
zooba: I'm just going to remove it. It is left over from the app.manifest template.

PR #2002, opened 2016-12-29
Fixes #1991 .vsp files generated by profile give sharing violation

### LINKED ISSUE #1991: .vsp files generated by profiler give "VSP1237: sharing violation" error when Anaconda env selected
I was trying to profile some code with Anaconda, but got the following error. It worked when I switched to a Python 3.5 environment. What profiler are we using? What work would we need to do to enable profiling in the Anaconda environment?
 
Given Anaconda is an important part of the VS data science workload, we should resolve this for VS2017 if possible. 

image

zooba:
Fixes #1991 .vsp files generated by profile give sharing violation
Waits up to 1 second for file to become accessible before opening.

I can't reproduce the actual issue, so I'm assuming it's a race with the profiling process closing. This change defers the open until we've successfully opened the file without sharing.

As far as I can tell, nothing to do with Anaconda.

### DIFF
--- a/Python/Product/Profiling/PythonProfilingPackage.cs
+++ b/Python/Product/Profiling/PythonProfilingPackage.cs
@@ -325,6 +325,7 @@ namespace Microsoft.PythonTools.Profiling {
                     for (int retries = 10; retries > 0; --retries) {
                         try {
                             using (new FileStream(outPath, FileMode.Open, FileAccess.Read, FileShare.None)) { }
+                            break;
                         } catch (IOException) {
                             Thread.Sleep(100);
                         }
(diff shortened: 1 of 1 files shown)


Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -321,6 +322,13 @@ out guid
                 _stopCommand.Enabled = false;
                 _startCommand.Enabled = true;
                 if (openReport && File.Exists(outPath)) {
+                    for (int retries = 10; retries > 0; --retries) {
+                        try {
+                            using (new FileStream(outPath, FileMode.Open, FileAccess.Read, FileShare.None)) { }
+                        } catch (IOException) {
```
huguesv: break to avoid doing this more than once?

PR #2011, opened 2017-01-03
Fixes #2010 Interpreter link crashes when path variable not set

### LINKED ISSUE #2010: Interpreter link crashes when path variable not set
If you configure a custom environment and do not set the "Path environment field", clicking the link in the overview tab to run the console will crash.

zooba:
Fixes #2010 Interpreter link crashes when path variable not set
Checks path variable before adding a variable with no name into the environment.

This doesn't affect VS 2017 because we automatically change empty variable names to `PYTHONPATH`.


huguesv:
👍

PR #2014, opened 2017-01-04
Allow cookiecutter template to run VS commands 

huguesv:
This covers the following scenarios:
 - Open a web url in internal or external web browser
 - Open a generated file in internal or external web browser
 - Open a generated file in the editor
 - Processing of a requirements text file, which proposes to the user:
   - install requirements into the active environment
   - create a virtual environment and install requirements into it (if active environment isn't already a virtual env)
   - do nothing

If adding to an existing project, the commands run immediately after the files are added to the project.

If creating a new 'project', the commands run after the 'project' is opened in solution explorer, triggered by the user clicking on the Open in Solution Explorer link.  I've made that link disappear once it's been clicked.

Command parameters are rendered by jinja, so they can contain cookiecutter variables.

There are examples of how you specify commands in my comments in CookiecutterClient.cs, and in the templates I am adding under TestData folder.

Building under VS 2017 brought up some warnings for async methods not having Async suffix, so I fixed those.

I got rid of OleMenuCommandService in cookiecutter package, since commands with arguments can't be implemented using it and it wasn't worth it for the other 2-3 commands.

Change to PTVS are for the new "process" requirements txt command, which takes the path to the reqs.txt file as parameter.  In order to make it work with existing code, I had to remove the hard coding of requirements.txt in a few places.

I'm thinking we need to add a warning before running the commands, to confirm that the user wants to enable running the commands.  With a task dialog, we could show the list of commands, so they can review before accepting.


Python/Product/Cookiecutter/CookiecutterPackage.cs
```diff
@@ -148,22 +221,31 @@ public sealed class CookiecutterPackage : Package {
...
+                return VSConstants.S_OK;
+            }
+
+            var name = GetStringArgument(variantIn) ?? "";
+            var uri = name.Trim('"');
+
+            if (File.Exists(uri)) {
+                var ext = Path.GetExtension(uri);
+                if (string.Compare(ext, ".htm", StringComparison.CurrentCultureIgnoreCase) == 0 ||
+                    string.Compare(ext, ".htm", StringComparison.CurrentCultureIgnoreCase) == 0) {
```
zooba: Should these be different string literals?

Python/Product/Cookiecutter/CookiecutterToolWindow.cs
```diff
@@ -209,21 +209,19 @@ sealed class CookiecutterToolWindow : ToolWindowPane, IVsInfoBarUIEvents {
...
         }
 
-        private void OpenGeneratedFolder(string folderPath) {
-#if DEV15_OR_LATER
-            OpenInSolutionExplorer(folderPath);
-#else
-            OpenInWindowsExplorer(folderPath);
+        private void ExecuteCommand(string name, string args) {
+            try {
+                _dte.ExecuteCommand(name, args ?? string.Empty);
```
zooba: We should be using `IVsUIShell.PostExecCommand` here with a GUID and ID. Using strings is going to break on non-English VS, as the names are localised. See ImportWizard.Wizard.RunStarted for an example (including passing parameters).

Python/Product/Cookiecutter/View/CookiecutterSearchPage.xaml.cs
```diff
@@ -100,7 +100,7 @@ internal partial class CookiecutterSearchPage : Page {
 
         private void TreeView_SelectedItemChanged(object sender, RoutedPropertyChangedEventArgs<object> e) {
             var val = e.NewValue as TemplateViewModel;
-            ViewModel.SelectTemplate(val).DoNotWait();
+            ViewModel.SelectTemplateAsync(val).DoNotWait();
 
             SelectedTemplateChanged?.Invoke(this, EventArgs.Empty);
```
zooba: Should we really be raising the event after the call above has completed? It's actually fairly easy - just make this event handler `async void` and await the SelectTemplateAsync call (event handlers are basically why `async void` was added).

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -1725,6 +1814,8 @@ private static T FindImmediateChild<T>(HierarchyNode parent)
                         return "e,env,environment: p,package: a,admin";
                     case PythonConstants.GenerateRequirementsTxt:
                         return "e,env,environment:";
+                    case PythonConstants.ProcessRequirementsTxt:
+                        return "path";
```
zooba: IIRC, you need the trailing colon to indicate that the parameter takes a value. You can test this from the Command Window.

Python/Product/PythonTools/PythonTools.vsct
```diff
@@ -624,8 +624,17 @@ permissions and limitations under the License.
         </Strings>
       </Button>
 
+      <Button guid="guidPythonToolsCmdSet" id="cmdidProcessRequirementsTxt" priority="0x0610" type="Button">
+        <CommandFlag>AllowParams</CommandFlag>
+        <CommandFlag>CommandWellOnly</CommandFlag>
+        <Strings>
+          <ButtonText>Pro&amp;cess Requirements</ButtonText>
```
zooba: Is there a more descriptive name than "process"? "InstallProjectRequirements" maybe? Can/should we switch the existing wizard to use this command as well?

Python/Tests/TestData/Cookiecutter/template_opentxt/cookiecutter.json
```diff
@@ -0,0 +1,10 @@
+{
+  "project_name": "Default Project Name",
+  "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '_') }}",
+  "_visual_studio_post_cmds": [
+    {
+      "name": "File.OpenFile",
+      "args": "\"{{cookiecutter._output_folder_path}}\\readme.txt\""
```
zooba: I'd rather `args` be a list rather than having to quote each argument. I expect most to be a single argument though, so perhaps treat a string as one argument and a list as multiple? `ProcessOutput.QuoteSingleArgument` is your friend when going to a single string (though passing `string[]` as a variant should be fine too).

Python/Product/Cookiecutter/CookiecutterToolWindow.cs
huguesv: PTVS english commands work on non-english.  I don't want to use guids, that makes the template unreadable.  If for some reason some of our commands don't work on non-english, we should make them work.  VS commands do work.

zooba:
> I'm thinking we need to add a warning before running the commands, to confirm that the user wants to enable running the commands

Since we know that there are commands at the start, maybe add a checkbox to the wizard that says "Run additional tasks on completion"?

Python/Product/Cookiecutter/CookiecutterToolWindow.cs
zooba: We don't localize, that's why. If/when we do localize command names, then it'll stop working.

Python/Product/Cookiecutter/CookiecutterPackage.cs
huguesv: oops

Python/Product/Cookiecutter/CookiecutterToolWindow.cs
```diff
@@ -209,21 +209,19 @@ sealed class CookiecutterToolWindow : ToolWindowPane, IVsInfoBarUIEvents {
...
+            } catch (Exception ex) when (!ex.IsCriticalException()) {
+#if !DEV15_OR_LATER
+                if (name == "File.OpenFolder") {
```
huguesv: It's a special case for VS 2015, which we don't officially support. I don't know the state of the open folder support extension on VS 2015.  I figured let it try, in case it's there.

Python/Product/PythonTools/PythonTools.vsct
huguesv: I didn't like the name either, we can use your suggestion.  It may require changing some strings to be a bit more generic, but reusing the command from the wizard may be possible.  Let me see.

Python/Product/Cookiecutter/CookiecutterToolWindow.cs
huguesv: Well VS localizes and the english names, ie File.Open still work.  I verified it on French VS.

zooba: We've definitely hit issues here in the past. I'd rather not risk it. (Also, localization may not be 100% complete for VS 2017 yet...)

Python/Product/Cookiecutter/CookiecutterToolWindow.cs
zooba: Ah, good point, I misread the ifdef. It's fine there.

The extension is totally unsupported, so I wouldn't worry about it.

Python/Product/Cookiecutter/CookiecutterToolWindow.cs
huguesv: I remember we had an issue with invoking new project from existing code a while back.  That was probably when our command names were a mess, before you cleaned them all up.
I'll test it again and I'll ask on the extensibility list.  I'd like proof of an actual issue before forcing template authors into using guids.

zooba: Knowing that list, they'll tell you to avoid DTE just for the sake of avoiding DTE :)

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
huguesv: How do you ask for command parameter help from the command window?  I've looked for this before, and again today, and the docs I'm looking at say nothing about it. https://msdn.microsoft.com/en-us/library/c785s0kz.aspx

zooba: Just try invoking the command and passing the arguments. Though it may be irrelevant unless you're using [IVsParseCommandLine](https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.shell.interop.ivsparsecommandline.aspx) (like we do for the other commands, via the `ParseCommandArgs` helper)

huguesv: Yeah that works fine, as I'm not using that parser. For some reason I always imagined there was a way to get parameter help from command window, I guess there isn't and that explains why I can't find it in the docs!

huguesv:
Yeah I think that's good. Less disruptive than a dialog that pops up.

Cases similar to this are VS warning you when you open a project downloaded from the web, and MS Word telling you it disabled macros when you load a document.

AFAIK, in both of these cases, they don't really help you decide if you should accept or not. It's pretty much, do you trust that source or not.  As opposed to "let me take a look at it before I decide".

Though with a checkbox, we have to decide if it's checked or unchecked by default, if it needs to be remembered per template, and/or globally...

If unchecked by default, most users will likely not touch that option and completely miss out on the feature.

But if checked by default, they will also likely not touch it, and it's a security risk.

Is it then better to have a simple message box with "don't ask me again" checkbox?  This would guarantee they pay attention to it at least once.

Python/Product/Cookiecutter/View/CookiecutterSearchPage.xaml.cs
huguesv: fixed

Python/Product/PythonTools/PythonTools.vsct
huguesv: Renamed & now shared by the wizard.

zooba:
> But if checked by default, they will also likely not touch it, and it's a security risk.

It's not a huge security risk, given it can only execute code that's already on their machine. If it's not checked by default, all the partner teams who have asked for this will be upset, and most users will end up with sample code that doesn't work.

We should just have it always checked by default. If feedback indicates that this isn't a good idea, we can look at alternatives, but right now it needs to be checked by default to work properly.

zooba:
:shipit:

PR #2018, opened 2017-01-05
Test improvements

zooba:
Predominantly test improvements, but also updates IronPython reference assemblies to 2.7.7. This _may_ prevent the use of earlier versions of IronPython, but the current situation is that we _may_ be preventing the use of versions later than 2.7.5. I simply can't reliably evaluate it without a proper build.

At worst, we may have to ask users to upgrade IronPython, but that shouldn't be difficult to justify. My big concern is that _sometimes_ it works with a mismatched assembly version in the GAC and sometimes it fails, and I simply can't tell when or why.


zooba:
On the IronPython side, I just finished a test run for this branch that fixed all the load errors against 2.7.7, so that confirms the "at worst" scenario.

huguesv:
👍

PR #2019, opened 2017-01-06
Fix 2015 Add From Cookiecutter sometimes reverts to global mode

huguesv:
Add From Cookiecutter was basically ignored when the command was ending up creating the tool window for the first time. Due to the tool window performing its actual initialization on idle, the NewSession() which was passing in the target project location was a no-op due to the control not being ready.


zooba:
+1

PR #2038, opened 2017-01-08
Improve django url intellisense parameters completion

David-Wobrock:
Hey hey,

Hesse commits:
- Fix a small typo
- Remove a redundant cast
- add the PTVS Apache License in a file where it should be

but more importantly, it improves the way django url parameters are completed.
It will propose the context variables as completion, as long as not enough parameters have been provided to the url template tag. And it will propose the unused named parameters also as completion.

Some changes are a "preparation"  to the django url namespace support I'm working on.

Cheers,
David


zooba:
Just so you're aware, I'm happy to take this now as is, but we are in the final stages of locking down for VS 2017's release. I haven't yet figured out what our branching strategy will be - will probably start a "3.0" branch for that release and keep using "master" for our first update - but the future namespace support will probably not be able to make it into the initial release at this stage.

PR #2042, opened 2017-01-09
Fixes #2041 Constantly creating processes while checking packages

### LINKED ISSUE #2041: Constantly creating processes while checking packages
Our file system watchers are detecting any change and retriggering scans. These scans import pip, which creates .pyc files. These files trigger another scan.

We should filter the file changes that may trigger a scan.

zooba:
Fixes #2041 Constantly creating processes while checking packages
Checks whether the file change is a Python source or compiled file before retriggering package scan.

### DIFF
--- a/Python/Product/VSInterpreters/PipPackageManager.cs
+++ b/Python/Product/VSInterpreters/PipPackageManager.cs
@@ -711,22 +711,8 @@ namespace Microsoft.PythonTools.Interpreter {
             }
         }
 
-        private static bool IsDirectory(string path) {
-            if (!Directory.Exists(path)) {
-                return false;
-            }
-
-            // These are the most likely directories to be created, and we know
-            // that we do not care about the contents, so ignore it.
-            if ("__pycache__".Equals(PathUtils.GetFileOrDirectoryName(path), StringComparison.OrdinalIgnoreCase)) {
-                return false;
-            }
-
-            return true;
-        }
-
         private void OnRenamed(object sender, RenamedEventArgs e) {
-            if (IsDirectory(e.FullPath) ||
+            if (Directory.Exists(e.FullPath) ||
                 ModulePath.IsPythonFile(e.FullPath, false, true, false) ||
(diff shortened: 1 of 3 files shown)


Python/Product/VSInterpreters/PipPackageManager.cs
```diff
@@ -675,37 +675,50 @@ private sealed class Suppressed : IDisposable {
...
+            }
         }
 
         public void NotifyPackagesChanged() {
             _refreshIsCurrentTrigger.Change(100, Timeout.Infinite);
         }
 
         private void OnRenamed(object sender, RenamedEventArgs e) {
-            _refreshIsCurrentTrigger.Change(1000, Timeout.Infinite);
+            if (ModulePath.IsPythonFile(e.FullPath, false, true, false) ||
```
huguesv: What if it's a folder that's renamed?  We need to trigger unconditionally?

Python/Product/VSInterpreters/PipPackageManager.cs
```diff
@@ -675,37 +675,50 @@ private sealed class Suppressed : IDisposable {
...
+                ModulePath.IsPythonFile(e.OldFullPath, false, true, false)) {
+                _refreshIsCurrentTrigger.Change(1000, Timeout.Infinite);
```
huguesv: I'm wondering if you need to protect against ObjectDisposedException when changing the timer.

zooba: Probably for the best.

Python/Product/VSInterpreters/PipPackageManager.cs
zooba: I believe we'll get a notification for each file contained in the folder, but I'll check.

zooba: Nope, just one for the directory. I'll add that check.

huguesv:
👍

PR #2050, opened 2017-01-10
Range of important fixes

zooba:
Prevents a few crashes at shutdown
Improves encapsulation/creation of BufferParser instances (doesn't quite fix all the races here, but it's a necessary start point - more work coming here)
Avoids sending too many AnalysisComplete events
Avoids creating filesystem watchers and package managers in analyzer process
Fixes analyzer not resetting after database update
Fixes retriggering package scan because of a package scan
Prevents OOM when analyzing delphixpy package

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntry.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntry.cs
@@ -16,7 +16,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.PythonTools.Infrastructure;
 using Microsoft.VisualStudio.Text;
 
 namespace Microsoft.PythonTools.Intellisense {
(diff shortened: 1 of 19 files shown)


zooba:
AnalysisEntry.cs has the biggest changes, though I don't expect them to last to the final release since they don't fix all the race conditions here.

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntry.cs
```diff
@@ -83,32 +180,38 @@ internal sealed class AnalysisEntry {
...
+                var events = buffer.GetNewAnalysisEntryRegistrations();
+                foreach (var notify in events) {
+                    notify(this);
                 }
             }
         }
 
         public int GetBufferId(ITextBuffer buffer) {
-            var bufferParser = BufferParser;
+            var bufferParser = _bufferParser;
```
huguesv: >_bufferParser [](start = 31, length = 13)

this here and in the function below are the only places accessing _bufferParser without lock (_bufferParserLock)

zooba: They're reads where we check the value afterwards and are "okay" if it fails. Partial writes are not a problem here. Plus, this code is hopefully not going to survive too long, so I'm not concerned.

huguesv:
:shipit:

PR #2054, opened 2017-01-11
Fixes #2045 Code snippets don't expand in REPL

### LINKED ISSUE #2045: Code snippets don't expand in REPL
They show up in completion list, but pressing TAB doesn't expand them.

FYI we've had several issues in RTVS regarding snippets in REPL (they now work).

If it's expensive to fix, I'm tempted to say remove them from REPL completions.  Simply to avoid this frustrating scenario:
image

zooba:
Fixes #2045 Code snippets don't expand in REPL
Exclude snippets from completion list in the repl.


huguesv:
👍

PR #2060, opened 2017-01-12
Fixes #2057 Multiple project IntelliSense does not work

### LINKED ISSUE #2057: Multiple project IntelliSense does not work
Loading multiple projects appears to reset analysis as each one is loaded.

We should either load them all in separate analyzer processes (and rely on search paths to cross-reference them), or fix reload.

zooba:
Fixes #2057 Multiple project IntelliSense does not work
Stops trying to share analyzer processes between projects.
Also fixes incorrect hash code when strongly merging object values.


huguesv:
So intellisense and refactor will work, but user needs to set search path. Correct?  If so ok with me, search path needed to be set anyway if you wanted F5 to work.

huguesv:
👍

zooba:
Cross-project refactoring won't work anymore, but Go to Definition should.

PR #2061, opened 2017-01-12
Fixes #2033 Keyboard scrolling in completions list broken

### LINKED ISSUE #2033: Keyboard scrolling in completions list broken
Using this week's build of VS 2017, when you hit arrow down on the completion list in Python, for example in the list that appears when you type `import <space>`, you see the selection change until you go below the last visible item, where it looks like selection keeps moving down, but you don't see it because the list isn't scrolling.

I checked C# and F#, and they don't have the issue.

zooba:
Fixes #2033 Keyboard scrolling in completions list broken
Fixes definition of scroll viewer.


huguesv:
👍

PR #2067, opened 2017-01-12
Partially fixes #2602 No navigate to results of object browser items

zooba:
Partially fixes #2602 No navigate to results of object browser items
Switches to an analyzer command for Navigate To results. Object browser is still busted.


huguesv:
👍

PR #2076, opened 2017-01-13
Test fixes

zooba:
Fixes a couple of tests.
Only product impact is defaulting to PYTHONPATH when no variable name is set. This should only affect tests, as we already default to PYTHONPATH elsewhere in the actual product.


huguesv:
👍

PR #2079, opened 2017-01-14
Fixes #2030 First repl command assumes multiple statements are supported

### LINKED ISSUE #2030: First repl command assumes multiple statements are supported
Open a new VS instance and Ctrl+Enter to send a code block to a normal Python repl. It will error out saying "multiple statements found ...".

If you do it again, we've learned that it only supports single statements and the code will be split. We should assume that single statements are required initially, and switch to multiple statements after the first one if possible.

zooba:
Fixes #2030 First repl command assumes multiple statements are supported
Switches to get the language version from a better API.


huguesv:
👍

PR #2090, opened 2017-01-23
Vtune integration

rgesteve:
This is the initial VTune integration work, adding the capability to profile Python projects with VTune (if VTune -- aka "Amplifier XE" 2017+ -- is already present in the system).  Any comments welcome.


zooba:
Just so you're aware, we won't be merging this into `master` until after VS 2017 releases - we're in stabilization mode right now. If you want to get this stored in the main repository, target it against the  `vtune` branch.

rgesteve:
Understood, I'm closing this and submitting #2095 against the vtune branch.  Thanks!

PR #2095, opened 2017-01-25
Vtune basic python profiling

rgesteve:


zooba:
Thanks! Unfortunately, with the merge of all the rest of our changes here it's basically impossible to do a review, so I'm just accepting the merge and we can do the review when we merge into master.

PR #2102, opened 2017-02-01
Fix build and update build tools.

huguesv:
I'm on build 26130.1 and one of the debug interfaces has been changed. I updated the build tools while I was thinking of it, but it wasn't strictly necessary for it to build.

I suspect you'll want to update the build machine before merging this.


zooba:
Thanks, I have a fix for this in one of my working copies, but it needs to go in. I'll update the build machine and then merge.

zooba:
Main build machine is updated, and secondary build VM and the test VMs are updating now. Once they're all up and running again I'll merge.

PR #2105, opened 2017-02-02
Fix cookiecutter git client

### LINKED ISSUE #2104: Cookiecutter clone broken in latest VS 2017 builds
Error code 128 when calling git.exe

It's not finding the following dlls:

```
---------------------------
git-remote-https.exe - System Error
---------------------------
The program can't start because libcurl-4.dll is missing from your computer. Try reinstalling the program to fix this problem. 
---------------------------
OK   
---------------------------


---------------------------
git-remote-https.exe - System Error
---------------------------
The program can't start because libintl-8.dll is missing from your computer. Try reinstalling the program to fix this problem. 
---------------------------
OK   
---------------------------



---------------------------
git-remote-https.exe - System Error
---------------------------
The program can't start because libiconv-2.dll is missing from your computer. Try reinstalling the program to fix this problem. 
---------------------------
OK   
---------------------------

```

Instead of using:
`C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TeamFoundation\Team Explorer\Git\cmd\git.exe`

we need to use:
`C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TeamFoundation\Team Explorer\Git\mingw32\bin\git.exe`

Instead of using the target parent folder for the working directory, we have to set it to the git.exe folder.

That also means we need to pass in another argument to git clone so it doesn't try to clone into a subfolder of git.exe, ex:
```
`"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TeamFoundation\Team Explorer\Git\mingw32\bin\git.exe" clone  "https://github.com/brettcannon/python-azure-web-app-cookiecutter" "C:\users\huvalo\.cookiecutters"`
```

huguesv:
Fix #2104 Cookiecutter clone broken in latest VS 2017 builds

Make sure that the dependent dlls can be found by adding the folder to the PATH. That seems safer than changing the working dir, which would also work but requires adding more params to the command line to tell git which folder to really work in.

### DIFF
--- a/Python/Product/Cookiecutter/Model/GitClient.cs
+++ b/Python/Product/Cookiecutter/Model/GitClient.cs
@@ -148,8 +148,12 @@ namespace Microsoft.CookiecutterTools.Model {
         }
 
         private Dictionary<string, string> GetEnvironment() {
+            var path =
+                Path.GetDirectoryName(_gitExeFilePath) + ";" +
+                Environment.GetEnvironmentVariable("PATH");
+
             return new Dictionary<string, string>() {
-                { "PATH", Path.GetDirectoryName(_gitExeFilePath) },
+                { "PATH", path },
             };
         }
 
--- a/Python/Product/Cookiecutter/Model/GitClient.cs
+++ b/Python/Product/Cookiecutter/Model/GitClient.cs
@@ -150,7 +150,7 @@ namespace Microsoft.CookiecutterTools.Model {
         private Dictionary<string, string> GetEnvironment() {
             var path =
                 Path.GetDirectoryName(_gitExeFilePath) + ";" +
-                Environment.GetEnvironmentVariable("PATH");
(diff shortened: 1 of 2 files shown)


Python/Product/Cookiecutter/Model/GitClient.cs
```diff
@@ -147,6 +147,12 @@ class GitClient : IGitClient {
             }
         }
 
+        private Dictionary<string, string> GetEnvironment() {
+            return new Dictionary<string, string>() {
+                { "PATH", Path.GetDirectoryName(_gitExeFilePath) },
```
zooba: We should add this path to the current value of PATH, probably at the start, or else we may break other expectations about what it can find.

PR #2106, opened 2017-02-02
Extract cookiecutter string resources

huguesv:
Part 1 of localization work.

Also deleted some unused code, and turned some fields into readonly.

There's a few more strings in the Shared code that is copy/pasted between PTVS and Cookiecutter, I will take care of that in a subsequent PR.

### DIFF
--- a/Python/Product/Cookiecutter/ProjectResources.cs
+++ b/Python/Product/Cookiecutter/ProjectResources.cs
@@ -14,10 +14,8 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-namespace Microsoft.CookiecutterTools
-{
-    internal class SR
-    {
+namespace Microsoft.CookiecutterTools {
+    internal class SR {
         internal const string SettingsGeneralCategory = "Settings_General_Category";
         internal const string SettingsShowHelpName = "Settings_ShowHelp_Name";
         internal const string SettingsShowHelpDescription = "Settings_ShowHelp_Description";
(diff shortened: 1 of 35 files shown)


Python/Product/Cookiecutter/ProjectResources.cs
```diff
@@ -0,0 +1,29 @@
...
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+namespace Microsoft.CookiecutterTools
+{
+    internal class SR
+    {
```
zooba: Newline formatting

Python/Product/Cookiecutter/ViewModel/TemplateViewModel.cs
```diff
@@ -173,14 +176,28 @@ class TemplateViewModel : INotifyPropertyChanged {
                 return _ownerUrl;
             }
 
-            set {
+            set
+            {
```
zooba: Newline formatting

zooba:
At some point we ought to go through and fix up the canonical names in the vsct to avoid the leading dots. Probably we should do that before submitting for localization. (IIRC, the way to fix it was to create a menu called "Python" or "Cookiecutter", set it as the parent for all the commands, and then use CommandPlacements to also put them in the right places.)

Issue #1234, opened 2016-04-22
comments in Django templates


eijoac:
when i type {#, i got {##}}. Is this an auto-completion bug? Thanks.

PR #2108, opened 2017-02-03
Move profiling strings into .resx

huguesv:


Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -44,7 +44,7 @@ namespace Microsoft.PythonTools.Profiling {
     // This attribute tells the PkgDef creation utility (CreatePkgDef.exe) that this class is
     // a package.
     [PackageRegistration(UseManagedResourcesOnly = true)]
-    [Description("Python Tools Profiling Package")]
+    [Description("Python Tools Profiling Package")] // TODO: Localization: does this do anything?
```
huguesv: Do you know what this is for?  RTVS and Cookiecutter package class don't have it.  I'm hoping we can just remove it.  Most of PTVS packages have the attribute.

zooba: No idea. Let's just leave it unlocalized for now - it'll either have no impact or we'll get a loc bug on it.

Python/Product/Profiling/Profiling/CompareReportsWindow.xaml
```diff
@@ -3,8 +3,9 @@
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:ui="clr-namespace:Microsoft.PythonTools.Profiling"
+    xmlns:pt="clr-namespace:Microsoft.PythonTools"
```
zooba: Can we get the strings resources in the Profiling namespace? Having the same class fullname may cause issues if both are referenced (e.g. in test assemblies)

Python/Product/Profiling/Profiling/ProfiledProcess.cs
```diff
@@ -128,7 +128,7 @@ sealed class ProfiledProcess : IDisposable {
             using (var p = ProcessOutput.RunHiddenAndCapture(perfCmdPath, "/waitstart")) {
                 p.Wait();
                 if (p.ExitCode != 0) {
-                    throw new InvalidOperationException("Starting perf cmd failed{0}{0}Output:{0}{1}{0}{0}Error:{0}{2}".FormatUI(
+                    throw new InvalidOperationException(Strings.StartPerfCmdError.FormatUI(
```
zooba: Newlines can be explicitly entered into the resource, so we don't need to use {0} to insert them.

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -63,7 +63,7 @@ sealed class PythonProfilingPackage : Package {
         internal static PythonProfilingPackage Instance;
         private static ProfiledProcess _profilingProcess;   // process currently being profiled
         internal static string PythonProjectGuid = "{888888a0-9f3d-457c-b088-3a5042f75d52}";
-        internal const string PerformanceFileFilter = "Performance Report Files|*.vsp;*.vsps";
+        internal static string PerformanceFileFilter = Strings.PerformanceReportFilesFilter;
```
zooba: `static readonly`

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -53,7 +53,7 @@ namespace Microsoft.PythonTools.Profiling {
     [Guid(GuidList.guidPythonProfilingPkgString)]
     // set the window to dock where Toolbox/Performance Explorer dock by default
     [ProvideToolWindow(typeof(PerfToolWindow), Orientation = ToolWindowOrientation.Left, Style = VsDockStyle.Tabbed, Window = EnvDTE.Constants.vsWindowKindToolbox)]
-    [ProvideFileFilterAttribute("{81da0100-e6db-4783-91ea-c38c3fa1b81e}", "/1", "Python Performance Session (*.pyperf);*.pyperf", 100)]
+    [ProvideFileFilterAttribute("{81da0100-e6db-4783-91ea-c38c3fa1b81e}", "/1", "Python Performance Session (*.pyperf);*.pyperf", 100)] // TODO: Localization
```
zooba: Should be able to localize the filter string here by putting it into `VSPackage.resx` and then putting `"#1234"` as the string here (where 1234 is the resource ID).

huguesv: Makes sense, but it didn't work :(

Python/Product/Profiling/PythonProfilingPackage.cs
huguesv: done

Python/Product/Profiling/Profiling/ProfiledProcess.cs
huguesv: done

Python/Product/Profiling/Profiling/CompareReportsWindow.xaml
huguesv: done

Python/Product/Profiling/PythonProfilingPackage.cs
huguesv: Will come back to it later. There's a TODO so I won't forget.

zooba:
👍

PR #2113, opened 2017-02-03
Update reference assemblies

zooba:
Updates our reference assemblies and makes it easier to have VS 2015 and VS 2017 actively working on the same repository.


huguesv:
👍

PR #2114, opened 2017-02-03
Move django strings into .resx file.

huguesv:
For DjangoPropertyPageControl (a windows forms control), the only change I made was set Localizable property to true (the tool did the rest).

### DIFF
--- a/Python/Product/Django/Intellisense/DjangoCompletionSourceBase.cs
+++ b/Python/Product/Django/Intellisense/DjangoCompletionSourceBase.cs
@@ -94,7 +94,7 @@ namespace Microsoft.PythonTools.Django.Intellisense {
                     "tag"));
             return new FuzzyCompletionSet(
                 "PythonDjangoTags",
-                "Django Tags",
+                Resources.DjangoTagsCompletionSetDisplayName,
                 applicableSpan,
                 completions,
                 options,
--- a/Python/Product/Django/TemplateParsing/DjangoFormatDefinitions.cs
+++ b/Python/Product/Django/TemplateParsing/DjangoFormatDefinitions.cs
@@ -24,8 +24,8 @@ using Microsoft.VisualStudio.Utilities;
 namespace Microsoft.PythonTools.Django.TemplateParsing {
     [Export(typeof(EditorFormatDefinition))]
     [ClassificationType(ClassificationTypeNames = DjangoPredefinedClassificationTypeNames.TemplateTag)]
-    [Name(DjangoPredefinedClassificationTypeNames.TemplateTag)]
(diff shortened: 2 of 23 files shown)


Python/Product/Django/Intellisense/DjangoCompletionSourceBase.cs
```diff
@@ -94,7 +94,7 @@ internal abstract class DjangoCompletionSourceBase : ICompletionSource {
                     "tag"));
             return new FuzzyCompletionSet(
                 "PythonDjangoTags",
-                "Django Tags",
+                Resources.DjangoTagsCompletionSetDisplayName,
```
huguesv: Parameter name says it's a display name, but I'm not sure where it gets displayed, ie. I don't know how to test it.

Python/Product/Django/TemplateParsing/DjangoFormatDefinitions.cs
```diff
@@ -24,8 +24,8 @@
 namespace Microsoft.PythonTools.Django.TemplateParsing {
     [Export(typeof(EditorFormatDefinition))]
     [ClassificationType(ClassificationTypeNames = DjangoPredefinedClassificationTypeNames.TemplateTag)]
-    [Name(DjangoPredefinedClassificationTypeNames.TemplateTag)]
-    [DisplayName(DjangoPredefinedClassificationTypeNames.TemplateTag)]
+    [Name(DjangoPredefinedClassificationTypeNames.TemplateTag)] // TODO: Localization - this string appears in fonts page in tools options
+    [DisplayName(DjangoPredefinedClassificationTypeNames.TemplateTag)] // TODO: Localization - not sure if this is used, does not affect fonts page
```
huguesv: I'm setting up a machine to see if the strings in fonts page get translated or not. Will know later today.

Looking at our code index, it doesn't look like DisplayName is being used by others that deriving from ClassificiationFormatDefinition.  If I make Name and DisplayName different, I can see that it's Name that is shown in fonts page.

huguesv: On French VS, names in the font page are translated. Will have to investigate how it's done later (I'll review to all TODOs).

zooba: You shouldn't need to set up a machine - just add a language pack through the installer and change to it in Tools, Options, International Settings.

Python/Product/Django/Intellisense/DjangoCompletionSourceBase.cs
zooba: You need two completion sets in the same completion popup. We don't do this because it's ugly and difficult to use.

zooba:
👍

PR #2116, opened 2017-02-04
Adds build support for localization.

zooba:
This gets us as far as a build when using our internal systems (or if you manually install the loc package you can get it locally as well). I haven't tested signed builds yet, but it should all be in the right order.

From a quick check, approx. 10% of strings are currently being localized :) We have some work to do!


huguesv:
👍

PR #2122, opened 2017-02-07
Converts MHT files into HTML with embedded images and stylesheets.

zooba:
There's probably a better project layout for this, but since it all stays checked in I doubt it'll matter much. Once we are localizing these files it's going to be a pain to modify them anyway.


zooba:
I'm waiting for a test build before merging this

huguesv:
👍

PR #2123, opened 2017-02-07
Move REPL strings to resx.

huguesv:


Python/Product/PythonTools/PythonTools/Repl/PythonInteractiveEvaluator.CommandProcessorThread.cs
```diff
@@ -730,6 +746,8 @@ public DisconnectedException(string message)
                     _stream.WriteInt64(thread);
                     _stream.WriteInt32(frame);
                     _stream.WriteInt32((int)frameKind);
+                    // TODO: Localization: we may need to do something with <CurrentFrame> string. Is it displayed?
+                    // It also appears visualstudio_py_repl.py so it probably needs to be in sync with it.
                     _currentScope = "<CurrentFrame>";
```
zooba: It appears in a dropdown alongside user-provided values. We should probably localize it, but as `"<Current frame>"` rather than the spaceless one.

zooba: Actually, let's leave it as is in case there are cross-dependencies in the .py file. But we might be able to hard-code a check and map to the localized string. I'll merge with the TODO comment.

PR #2125, opened 2017-02-08
Move more strings to resx

huguesv:
Web project related windows forms controls.
Commands subfolder code, including diagnostics window.
Ensure we have both CanonicalName and LocCanonicalName on commands.


Python/Product/Common/Strings.resx
```diff
@@ -1184,6 +1184,162 @@ Please refer to &lt;a href="https://aka.ms/pythononappservice"&gt;aka.ms/PythonO
...
+    <value>_Copy all...</value>
+  </data>
+  <data name="DiagnosticsWindow_Close" xml:space="preserve">
+    <value>C_lose</value>
+  </data>
+  <data name="DiagnosticsWindow_Title" xml:space="preserve">
+    <value>Visual Studio - Python Diagnostic Info</value>
+  </data>
+  <data name="DiagnosticsWindow_MaskedValue" xml:space="preserve">
+    <value> (masked)</value>
```
zooba: The strings we add into the log file should not be localized. These are intended to be sent to us, and we want them in English.

Let's take these back to literal strings. I'll make the case that we should not localize them if it becomes an issue.

huguesv:
Rolled back the localization of the diagnostics log.

PR #2128, opened 2017-02-08
Move strings for Navigation and Debugger folders to resx

huguesv:


Python/Product/PythonTools/PythonTools/Navigation/PythonLibraryNode.cs
```diff
@@ -116,7 +113,7 @@ public PythonLibraryNode(LibraryNode parent, CompletionResult value, IVsHierarch
 
                 if (isAlias) {
                     res.Append(Name);
-                    res.Append(" (alias of ");
+                    res.Append(" (alias of "); // TODO: Localization ?
```
zooba: Yes. The whole thing should become a format string like `{0} (alias of {1})`

PR #2130, opened 2017-02-08
Move strings to resx for code under Project.

huguesv:


Python/Product/Common/Strings.resx
```diff
@@ -495,6 +495,221 @@ The web.config file will be configured to resolve URIs matching this pattern to
...
+  </data>
+  <data name="AddVirtualEnvironmentWindow_BrowsePathOrigPrefixError" xml:space="preserve">
+    <value>We could not find a base interpreter at </value>
+    <comment>A folder path is concatenated to the end of this string</comment>
+  </data>
+  <data name="AddVirtualEnvironmentWindow_Actions" xml:space="preserve">
+    <value>Actions we will perform:</value>
+  </data>
+  <data name="AddVirtualEnvironmentWindow_InstallPipAndVirtualEnv" xml:space="preserve">
+    <value>Install pip and virtualenv packages</value>
```
zooba: Just a note that we need to remember to specify that "pip" is a product name and should not be localized. I'm guessing they won't try to localize virtualenv.

Python/Product/Common/Strings.resx
```diff
@@ -495,6 +495,221 @@ The web.config file will be configured to resolve URIs matching this pattern to
...
+    <value>Failed to upload file: {0}</value>
+    <comment>{0} is error details</comment>
+  </data>
+  <data name="PublishPropertyControl_LocationLabel" xml:space="preserve">
+    <value>Publishing folder location ({0}): </value>
+    <comment>{0} is a dynamically built list of location types
+Location types examples:
+ftp or file
+web, ftp or file
+azure, web, ftp or file</comment>
```
zooba: We may need to precompute these lists if we can, but we'll see whether the localizers complain.

Python/Product/PythonTools/PythonTools/Project/AddVirtualEnvironment.xaml
```diff
@@ -53,13 +52,13 @@
...
-                            </Label>
+                                   Name="BrowsePathError"
+                                   Content="{x:Static common:Strings.AddVirtualEnvironmentWindow_BrowsePathError}"/>
 
                             <Label Name="BrowsePathOrigPrefixError">
                                 <TextBlock TextWrapping="Wrap">
-                                    <Run>We could not find a base interpreter at </Run>
+                                    <!-- TODO: Localization make this a single string -->
+                                    <Run Text="{x:Static common:Strings.AddVirtualEnvironmentWindow_BrowsePathOrigPrefixError}"/>
                                     <Run Text="{Binding BrowseOrigPrefix}" />
```
zooba: You should be able to specify the StringFormat parameter on the BrowseOrigPrefix binding to do string formatting. If it is harder then that, leave the TODO here.

zooba:
Okay, I hit Merge and didn't mean to, so the BrowseOrigPrefix function change will have to be a separate PR.

PR #2131, opened 2017-02-08
Move Uwp strings to resx.

huguesv:


zooba:
@alanch-ms FYI, but the change looks safe so I'm merging it.

PR #2132, opened 2017-02-09
Moved debugger strings to resx.

huguesv:
I've added some TODOs for some that I still have remaining, and others that I'm not sure about.


Python/Product/Common/Strings.resx
```diff
@@ -1162,6 +1162,127 @@ Please refer to &lt;a href="https://aka.ms/pythononappservice"&gt;aka.ms/PythonO
...
+  </data>
+  <data name="DebugThreadUnknownLocation" xml:space="preserve">
+    <value>&lt;unknown location, not in Python code&gt;</value>
+  </data>
+  <data name="DebugTcpsTransportConnectionError" xml:space="preserve">
+    <value>Could not establish secure connection to {0} because of the following SSL issues:
+
+</value>
+    <comment>{0} is a URI
+A list of issues is concatenated to the message (one per line)</comment>
```
zooba: I _think_ we should be adding {1} for the list of issues, or we may not handle RTL text properly (thinking specifically of the marker characters, if any, that may not behave consistently when blindly concatenating strings).

Python/Product/Debugger/Debugger/DebugEngine/AD7Property.cs
```diff
@@ -60,6 +58,7 @@ class AD7Property : IDebugProperty2, IDebugProperty3 {
 
             if ((dwFields & enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_TYPE) != 0) {
                 if (_evalResult.ExceptionText != null) {
+                    // TODO: Localization
```
zooba: I vote no.

Python/Product/Debugger/Debugger/DebugEngine/AD7Property.cs
```diff
@@ -69,6 +68,7 @@ class AD7Property : IDebugProperty2, IDebugProperty3 {
 
             if ((dwFields & enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_VALUE) != 0) {
                 if (_evalResult.ExceptionText != null) {
+                    // TODO: Localization
```
zooba: No, because the evaluator result will not be localized.

Python/Product/Debugger/Debugger/DebugEngine/AD7Thread.cs
```diff
@@ -126,6 +124,7 @@ class AD7Thread : IDisposable, IDebugThread2, IDebugThread100 {
                 props.dwFields |= enum_THREADPROPERTY_FIELDS.TPF_STATE;
             }
             if ((dwFields & enum_THREADPROPERTY_FIELDS.TPF_PRIORITY) != 0) {
+                // TODO: Localization
                 props.bstrPriority = "Normal";
```
zooba: I believe so, but we should be able to check whether other languages localize these values in the Threads window.

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteDebugPortSupplier.cs
```diff
@@ -97,6 +103,7 @@ public class PythonRemoteDebugPortSupplier : IDebugPortSupplier2, IDebugPortSupp
         }
 
         public int GetPortSupplierName(out string pbstrName) {
+            // TODO: Localization - translate or not?
             pbstrName = "Python remote (ptvsd)";
```
zooba: I think so, but then again, 2/3 of the string won't be localized anyway. Leave it for now

Python/Product/Debugger/DkmDebugger/ExceptionManager.cs
```diff
@@ -151,6 +151,7 @@ internal class ExceptionManagerLocalHelper : DkmDataItem {
 
             var reprOptions = new ReprOptions(process);
 
+            // TODO: localization
             string typeName = "<unknown exception type>";
```
zooba: Yes, we should

Python/Product/Debugger/DkmDebugger/PyObjectNativeVisualizer.cs
```diff
@@ -73,6 +73,7 @@ internal class PyObjectNativeVisualizer : DkmDataItem {
                 return null;
             }
 
+            // TODO: Localization ?
             var pyEvalResult = new PythonEvaluationResult(objRef, "[Python view]")
```
zooba: I think so. Leave it for now and we'll see whether the loc testers call it out (though there's a real good chance they'll never see it - this is fairly advanced stuff).

Python/Product/Debugger/Debugger/DebugEngine/AD7Property.cs
huguesv: TODO removed

Python/Product/Debugger/Debugger/DebugEngine/AD7Property.cs
huguesv: TODO removed

Python/Product/Debugger/DkmDebugger/ExceptionManager.cs
huguesv: Done, already had a resx string for it

Python/Product/Debugger/DkmDebugger/PyObjectNativeVisualizer.cs
huguesv: "[Python view]" appears in `\Python\Product\Debugger\DkmDebugger\PythonDkm.natvis`, so I think it's safer to not change it for now.

PR #2161, opened 2017-02-14
Fixes #2142: NullReferenceException when using lightbulb extract method

### LINKED ISSUE #2142: NullReferenceException when using lightbulb extract method
I had the following code:
```
def testing1():
    a = 1
    b = 2
    yield a
    yield 1

```

I selected the contents of the method, clicked on lightbulb, selected extract method and got this error:
image

zooba:
Fixes #2142: NullReferenceException when using lightbulb extract method
Adds null checks.


huguesv:
👍

PR #2164, opened 2017-02-14
Updates reference assemblies, since the nuget packages are not up to date

zooba:
Updates reference assemblies, since the nuget packages are not up to date
We'll switch back to the nuget packages once they have the correct bits in them.


huguesv:
👍  if you've verified they are signed

zooba:
I have

Issue #2150, opened 2017-02-13
Move to shared interactive window implementation


zooba:
Rather than independently resolving issues in our clone of the interactive window, we should switch back to using the common one. This will save us from having to manually fix problems that have likely been fixed already.

PR #2168, opened 2017-02-16
Use shared interactive window component

huguesv:
Fix https://github.com/Microsoft/PTVS/issues/2150

Python & R toolbars don't conflict. Note that I made groups, previous implementation was a single group for all toolbar items.
image

VS 2015, which we don't ship, has slightly different order, due to them defining their group priority as 0, so we can't have the New Window button as the first item. Well we could, if we integrated in their group, but that would also look different, and have more conditional code in the vsct, which I want to minimize.

image

Need to test Jupyter stuff, there's a TODO in the code. Don't merge yet.


Python/Product/PythonTools/PythonTools/Repl/PythonInteractiveEvaluator.cs
```diff
@@ -661,7 +661,9 @@ partial class PythonInteractiveEvaluator :
         }
 
         internal Task InvokeAsync(Action action) {
-            return ((System.Windows.UIElement)_window.GetTextViewHost()).Dispatcher.InvokeAsync(action).Task;
+            // TOOD: Verify this
```
huguesv: This is the thing I need to test. Does it look like it would work? There's no GetTextViewHost() method on the shared implementation.

huguesv: It's called by DisplayImage and DisplayXaml, and I can see it working for image:
image

huguesv:
This build includes the changes:
\\pytools\rolldata\PTVS\Build\PTVS-Github-Huguesv-Dev15\17046.2\release

It's working fine so far.

Python/Product/PythonTools/PythonTools/Repl/PythonInteractiveEvaluator.cs
zooba: I think this is fine. Might be worth ensuring that the action is actually run on the UI thread, but then again it should raise an exception if we create new controls on the wrong thread.

If the repl tests run okay, then tidy up the comments here and we'll call it fine.

zooba:
Please also go into Python/setup/swix/install.ps1 and add `Microsoft.VisualStudio.InteractiveWindow` to the list of components to install in VS. The shared interactive window is not installed by default.

zooba:
:+1

PR #2174, opened 2017-02-16
VSCT localization

### LINKED ISSUE #2107: Avoid leading dot hack in vsct canonical names
At some point we ought to go through and fix up the canonical names in the vsct to avoid the leading dots. Probably we should do that before submitting for localization. (IIRC, the way to fix it was to create a menu called "Python" or "Cookiecutter", set it as the parent for all the commands, and then use CommandPlacements to also put them in the right places.)

### LINKED ISSUE #2148: PythonTools .vsct localization
It's working for cookiecutter, django and profiling, but not for python tools.

image

Edit: Confirmed django vsct working.

zooba:
Fixes #2148 PythonTools .vsct localization
Fixes #2107 Avoid leading dot hack in .vsct files

Still more changes coming, but wanted to get eyes on this. I merged in the interactive window change too, so we can test builds from my drop for all the changes together.


huguesv:
Is there a benefit to putting spaces in the canonical names?

zooba:
I'm hoping it causes less confusion for the localizers. Spaces get stripped before they become command names.

zooba:
I think this is ready to merge now.

huguesv:
I'm not able to run any of the cookiecutter commands using the command window.
```
>Cookiecutter.Home
Command "Cookiecutter.Home" is not valid.
>Cookiecutter.Explorer
Command "Cookiecutter.Explorer" is not valid.
>
```

zooba:
:( The Cookiecutter menus are parenting off the context menu first - it's `OtherContextMenus.CookiecutterContext.Home` for home, and I can't find Explorer anywhere.

This is painful enough and not really a high priority feature - I'll file a bug on it and we'll deal with it later.

zooba:
Preemptively filed #2177

huguesv:
`Cookiecutter.ExternalWebBrowser` is meant to be used from templates, so they can launch an external browser as a task, and it doesn't work.
`Python.LaunchProfiling` works, but `Python.PerformanceExplorer` doesn't.

I would rather keep the dot hack than remove the hack and break many commands. Especially if the point of getting rid of the hack is to make commands work better.

zooba:
Ah you're right. I'll bring it back for the ones we need.

zooba:
I've spent altogether too long on this now - commanding in VS is just busted and there seems to be no nice way to get it all to work.

The hacks are back in for the commands I know we need. They don't appear in the command window intellisense, but they do execute.

We should reach out to the team that understands this and get their advice, but I have other things I need to work on and don't want to block it on this. If you spot any other issues, let me know, but I need to do my string fixups before we can send off the first batch for translation.

huguesv:
Alright.

PR #2180, opened 2017-02-18
String fixes

### LINKED ISSUE #2179: Some cookiecutter strings not localized
When Python isn't installed, this is displayed
image

zooba:
Mostly updating resource strings for readability.

Also include fix #2179 

Still have 3-4 assemblies to go through.


Python/Product/Cookiecutter/Strings.resx
```diff
@@ -447,15 +447,21 @@ Please delete the installed template and try again.</value>
     <value>Show the help information bar.</value>
   </data>
   <data name="Settings_FeedUrl_Name" xml:space="preserve">
-    <value>Recommended Feed URL</value>
+    <value>Recommended feed URL</value>
   </data>
   <data name="Settings_FeedUrl_Description" xml:space="preserve">
-    <value>Location of the feed. The contents of the feed consists of line separated URLs of template locations.</value>
+    <value>Location of the feed file. Feed files have one template URLs on each line.</value>
```
huguesv: singular URL

huguesv:
I wonder if we they'll know how to deal with the file filter strings (the special | separators)?

Also on pseudo-loc builds they don't work that great because the junk chars are added at the beginning and end of the whole string, not each section.

I'm starting to wonder if we need to split them in loc and non-loc strings.

huguesv:
👍  except given URLs -> URL change.

PR #2182, opened 2017-02-21
correct spelling mistake

EdwardBetts:


zooba:
Thanks for the contribution. This isn't actually part of our code - it's a virtual environment that we include in our test data - and the spelling mistake has already been fixed in the original project (CPython).

Issue #2183, opened 2017-02-21
Localize project template content


zooba:
Ensure all comments, descriptive strings, and essentially anything other than pure code is localized.

huguesv:
We're closing old issues with no recent activity. If this is still important, please open a new issue.

PR #2184, opened 2017-02-21
Extracts template names into resources.

### LINKED ISSUE #2173: Project template localization
Names and descriptions in .vstemplate files

May not need to localize the generated code. I'll check ASP.NET to see if they do.

zooba:
Fixes #2173


zooba:
I filed #2183 to track localizing the content of templates - I don't think it's important enough to do it right away. (Right now we need to get the critical stuff done so we can do the first handoff)

huguesv:
👍

PR #2190, opened 2017-02-23
Fixes #2160 Custom commands cannot detect installed packages

### LINKED ISSUE #2160: Custom commands cannot detect installed packages
(From [developer community](https://developercommunity.visualstudio.com/content/problem/7246/run-pylint-does-not-recognize-that-pylint-is-alrea.html))

If you run pylint, it prompts to install even if pylint is already installed.

I suspect we are including the version specifier in a call to check whether it is present and hence failing. This is a regression since VS 2015.

zooba:
Fixes #2160 Custom commands cannot detect installed packages
Switches check to treat package specs as requirements rather than names.


huguesv:
👍

PR #2191, opened 2017-02-23
Fixes #2111 RemoveNonDocument ordering issue

### LINKED ISSUE #2111: RemoveNonDocument ordering issue
In `HierarchyNode.RemoveNonDocument` we have this code:

```
            // Remove child if any before removing from the hierarchy
            for (HierarchyNode child = this.FirstChild; child != null; child = child.NextSibling) {
                child.Remove(removeFromStorage);
            }
```

We should get `NextSibling` before removing the child, since removing it may take it out of the linked list.

zooba:
Fixes #2111 RemoveNonDocument ordering issue
Correctly traverses the linked list while removing elements.

This was reported by NTVS team when they discovered a crash in their version of this file.


huguesv:
👍

PR #2206, opened 2017-03-01
Fixes #2043 Intellisense performance issues

### LINKED ISSUE #2043: Intellisense performance issues
We seem to have performance issues with the amount of time taken to respond to queries for completions and signature help.

Any bottlenecks or synchronization issues in the out-of-proc analyzer are worth investigating and removing.

zooba:
Fixes #2043 Intellisense performance issues

Broadly, this reduces the timeouts based on telemetry information, but the more significant change is consistency in how we wait for out-of-proc analysis messages and also adding telemetry points for these waits.

To do this, we need direct access to the analyzer earlier in the call, so this resulted in a few functions changing from static to instance methods.

As usual, all telemetry-only code is (or should be) protected by catch-all handlers that assert in debug builds. Telemetry failures should _never_ cause the product to break.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
@@ -17,23 +17,22 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using System.Windows.Input;
 using Microsoft.PythonTools.Editor;
 using Microsoft.PythonTools.Editor.Core;
-using Microsoft.VisualStudio.InteractiveWindow;
 using Microsoft.PythonTools.Parsing;
 using Microsoft.PythonTools.Parsing.Ast;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Editor;
+using Microsoft.VisualStudio.InteractiveWindow;
 using Microsoft.VisualStudio.Language.Intellisense;
 using Microsoft.VisualStudio.Language.StandardClassification;
 using Microsoft.VisualStudio.OLE.Interop;
 using Microsoft.VisualStudio.Text;
-using Microsoft.VisualStudio.Text.Classification;
 using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Text.IncrementalSearch;
 using Microsoft.VisualStudio.Text.Operations;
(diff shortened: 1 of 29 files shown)


Python/Product/PythonTools/PythonTools/Logging/InMemoryLogger.cs
```diff
@@ -119,6 +131,18 @@ class InMemoryLogger : IPythonToolsLogger {
                 }
             }
 
+            lock (_analysisTiming) {
+                if (_analysisTiming.Any()) {
+                    res.AppendLine("Analysis timing:");
+                    foreach (var kv in _analysisTiming.OrderBy(kv => kv.Key)) {
+                        res.AppendFormat("    {0} (count {1}, {2} timeouts, max {3:N0}ms, mean {4:N2}ms)",
```
huguesv: average, not mean?

Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
```diff
@@ -114,28 +113,27 @@ internal sealed class IntellisenseController : IIntellisenseController, IOleComm
...
-                    _serviceProvider,
-                    _textView,
-                    pt.Value,
-                    TimeSpan.FromSeconds(1.0)
-                );
+                var entry = _textView.GetAnalysisEntry(pt.Value.Snapshot.TextBuffer, _serviceProvider);
+                var t = entry.Analyzer.GetQuickInfoAsync(entry, _textView, pt.Value);
+                var quickInfo = await Task.Run(() => entry.Analyzer.WaitForRequest(t, "GetQuickInfo", null, 2));
 
                 QuickInfoSource.AddQuickInfo(_textView, quickInfo);
```
huguesv: Is this ok if quickInfo is null?

Python/Product/PythonTools/PythonTools/Refactoring/ExtractMethodRequestView.cs
```diff
@@ -256,9 +257,14 @@ public ExtractMethodRequestView(IServiceProvider serviceProvider, ExtractedMetho
         private void UpdatePreview() {
             var info = GetRequest();
             if (info != null) {
-                _previewer.GetExtractionResult(info).ContinueWith(
-                    x => PreviewText = x?.WaitOrDefault(1000)?.methodBody ?? Strings.ExtractMethod_FailedToGetPreview
-                ).DoNotWait();
+                _previewer.GetExtractionResult(info).ContinueWith(t => {
+                    try {
+                        PreviewText = t.Result.methodBody;
```
huguesv: is this the thing that used to return null when user selected code that isn't valid to be extracted and then crashed?  If so you may want to keep the r.Result?.methodBody.  I know you catch Exception, but failing to get preview due to a normal scenario doesn't need to Debug.Fail.

huguesv: I found what I was thinking of, and it wasn't this.
https://github.com/Microsoft/PTVS/pull/2161/files

Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
zooba: Yes

Python/Product/PythonTools/PythonTools/Logging/InMemoryLogger.cs
zooba: Arguably the average isn't all that interesting here anyway (mean or otherwise), but I'm not going to write code for bucketing just for this display.

huguesv:
👍

PR #2207, opened 2017-03-01
Fixes expected filename for localisation files.

zooba:


zooba:
Depending on the resolution for #2208 regarding snippet LCL filenames, we may want to remove the `%(RecursiveDir)` values from the metadata for `FilesToLocalize Include="@(CodeSnippets)"` so that we get everything in a flat layout.

huguesv:
No opinion on this.  👍  if this works.

PR #2210, opened 2017-03-01
Fix #1289: PTVS Concord components should avoid using NoFilter in their .vsdconfigxml

### LINKED ISSUE #1289: PTVS Concord components should avoid using NoFilter in their .vsdconfigxml
PTVS uses NoFilter for three of its the interfaces -- [IDkmExceptionManager](https://github.com/Microsoft/PTVS/blob/819bd32ea0f4a812300bf9ff9392d4dd79ca4264/Python/Product/Debugger/DkmDebugger/RemoteComponent.vsdconfigxml#L78), [IDkmAsyncBreakCompleteReceived](https://github.com/Microsoft/PTVS/blob/819bd32ea0f4a812300bf9ff9392d4dd79ca4264/Python/Product/Debugger/DkmDebugger/RemoteComponent.vsdconfigxml#L78) and [IDkmCallStackFilter](https://github.com/Microsoft/PTVS/blob/819bd32ea0f4a812300bf9ff9392d4dd79ca4264/Python/Product/Debugger/DkmDebugger/LocalComponent.vsdconfigxml#L57).

For IDkmExceptionManager, I think you could use an EngineId filter. For the others you may be able to use either a CallOnlyWhenLoaded constraint.

Having the exception manager loaded causes an issue with XBox game development. This is somewhat a bug in the core debugger. But it seems like a good fix here anyway.

Using CallOnlyWhenLoaded, I believe you can cut down on the number of scenarios that the code loads in, which is good for perf for scenarios where the user isn't using Python. You could use your custom message listener, or an IDkmRuntimeLoadNotification to get your code to load when Python does load in the target process.

int19h:
Filter by EngineId, or use CallOnlyWhenLoaded, as appropriate.

### DIFF
--- a/Python/Product/Debugger/DkmDebugger/LocalComponent.vsdconfigxml
+++ b/Python/Product/Debugger/DkmDebugger/LocalComponent.vsdconfigxml
@@ -53,7 +53,7 @@ permissions and limitations under the License.
           <Interface Name="IDkmModuleSymbolsLoadedNotification"/>
         </InterfaceGroup>
 
-        <InterfaceGroup>
+        <InterfaceGroup CallOnlyWhenLoaded="true">
           <NoFilter/>
           <Interface Name="IDkmCallStackFilter"/>
         </InterfaceGroup>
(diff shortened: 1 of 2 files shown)


int19h:
All MMD tests pass with the change, and manual testing didn't uncover any issues. Not sure what the problem was last time I tried it, but looks like we are good with the cheap fix after all.

zooba:
👍

PR #2211, opened 2017-03-01
Updates wfastcgi in preparation for release.

zooba:
Note that the rich diff view for README.rst (near the "View" button) works for ReST files.


huguesv:
👍

PR #2217, opened 2017-03-02
Fixes #1542 Create virtual env dialog doesn't always come up via new project

### LINKED ISSUE #1542: Create virtual env dialog doesn't always come up via new project
VS 15 Preview 4 (daily)

Start VS
New Project->Bottle Web Project
Install into a virtual environment

Result: Virtual env dialog doesn't come up, virtual env not created

If you repeat this in the same instance of VS, it will work the second time.

You'll need to restart VS to repro again.

zooba:
Fixes #1542 Create virtual env dialog doesn't always come up via new project
Exposes IOleCommandTarget via the DTE Project so we can invoke the command without going through DTE.
Adds test ensuring project has the right types exposed.
Fixes implicit circular dependency between PythonToolsTests and PythonToolsUITests.

Validated with manual testing:
* I found that creating Flask project, Django project, Flask project would somehow get VS stuck not showing the dialog
* After this change, I can no longer reproduce this
* Over 100 projects were created and deleted in the process of fixing this bug :)

### DIFF
--- a/Common/Product/SharedProject/Automation/OAProject.cs
+++ b/Common/Product/SharedProject/Automation/OAProject.cs
@@ -19,11 +19,12 @@ using System.Globalization;
 using System.Runtime.InteropServices;
 using EnvDTE;
 using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.OLE.Interop;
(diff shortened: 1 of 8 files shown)


int19h:
👍

Issue #1953, opened 2016-12-07
Tests are not shown on VS 2017


lmazuel:
Opening the Azure SDK for Python project, after migration, analysis, several open/close of VS, still does not see my tests.

Microsoft Visual Studio Enterprise 2017 RC
Version 15.0.26006.2 D15REL
Microsoft .NET Framework
Version 4.6.01586

Installed Version: Enterprise

Visual Basic 2017 RC   00369-50000-00000-AA399
Microsoft Visual Basic 2017 RC

Visual C# 2017 RC   00369-50000-00000-AA399
Microsoft Visual C# 2017 RC

Application Insights Tools for Visual Studio Package   8.4.01118.2
Application Insights Tools for Visual Studio

ASP.NET and Web Tools 2016   15.0.21206.0
ASP.NET and Web Tools 2016

ASP.NET Web Frameworks and Tools 2013   5.2.41123.0
For additional information, visit https://www.asp.net/

Azure App Service Tools v2.9.6   15.0.21130.0
Azure App Service Tools v2.9.6

Common Azure Tools   1.8
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

JavaScript Language Service   2.0
JavaScript Language Service

Microsoft Azure Tools   2.9
Microsoft Azure Tools for Microsoft Visual Studio vNext - v2.9.41121.2

NuGet Package Manager   4.0.0
NuGet Package Manager in Visual Studio. For more information about NuGet, visit http://docs.nuget.org/.

Python Tools for Visual Studio   3.0.16334.1
Python Tools for Visual Studio provides IntelliSense, projects, templates, Interactive windows, and other support for Python developers.

Python Tools for Visual Studio - Cookiecutter Integration   3.0.16334.1
Provides cookiecutter integration.

Python Tools for Visual Studio - Django Integration   3.0.16334.1
Provides templates and integration for the Django web framework.

Python Tools for Visual Studio - Profiling Support   3.0.16334.1
Profiling support for Python projects.

SQL Server Data Tools   15.1.61124.0
Microsoft SQL Server Data Tools

TypeScript   2.1.3.0
TypeScript tools for Visual Studio

huguesv:
I'm on the same build.  I had the project opened with the tests not discovered, I left it for about 5 mins and when I came back they were there and I was able to run one.  That was with a Python 3.5 64-bit venv, with requirements.txt installed into it.

[...]

PR #2218, opened 2017-03-02
Adds support for logging all messages sent over IPC.

zooba:
Adds support for logging all messages sent over IPC.
This will be helpful for debugging Intellisense, test discovery and soon debugging issues.


zooba:
@huguesv This will potentially affect your current work, but may also help debug any issues that arise.

I'm going to need it for solving #1953, but I won't get through that today and wanted to at least get the extra logging in.

int19h:
👍

huguesv:
Yeah I found the logging to output a bit too simplistic.  I had made a small improvement so that at least the Debug.WriteLine calls print out the name of the connection so I can distinguish the debugger ones from the analyzer ones.

👍

PR #2249, opened 2017-03-08
Fix ptvsd stackless schedule_callback

maxrossi:
Stackless-only issue:
`stackless.set_schedule_callback` was missing and `_stackless_schedule_cb` never called

### DIFF
--- a/Python/Product/PythonTools/visualstudio_py_debugger.py
+++ b/Python/Product/PythonTools/visualstudio_py_debugger.py
@@ -790,6 +790,7 @@ class Thread(object):
     def _stackless_attach(self):
         try:
             stackless.tasklet.trace_function
+            stackless.set_schedule_callback(self._stackless_schedule_cb)
         except AttributeError:
             # the tasklets need to be traced on a case by case basis
             # sys.trace needs to be called within their calling context


zooba:
Thanks. I'm not sure that's the right place for the change - the `try`/`except AttributeError` block is a very narrow test for whether `trace_function` is available. I think in this case it ought to be in an `else` block after the `except`:

```python
try:
    stackless.tasklet.trace_function 
except AttributeError:
    # handle when trace_function is not available
else:
    # handle when trace_function is available
```

I'll admit I'm not familiar enough with stackless's debugging infrastructure to tell for certain whether the set_schedule_callback is exactly right, but it seems it will be once we aren't risking swallowing AttributeErrors from the call. (We rely entirely on external contributions for stackless support, so thanks for sending this in!)

maxrossi:
Yes, moving `set_schedule_callback` in the `else` block surely prevents swallowing AttributeErrors, even if `set_schedule_callback` is a `stackless` function present since early versions of Stackless Python, so it should never raise such errors.

Anyway, `_stackless_schedule_cb` was defined but actually never set as stackless callback. Here's the official example about debugging tasklets, that uses `set_schedule_callback` http://stackless.readthedocs.io/en/2.7-slp/library/stackless/debugging.html

zooba:
@int19h This looks safe enough to take, and it's my fault the PR has languished for so long. It needs to be updated for the refactor so far - want to make a new PR and close this one?

int19h:
I've started the new PR.

PR #2253, opened 2017-03-09
Fixes #2239 Update build dependencies.

### LINKED ISSUE #2239: Fix reference to the DB connection dialog for Dev15
https://github.com/Microsoft/RTVS/pull/3267

zooba:
Fixes #2239 Update build dependencies.
Also simplifies nuget package management so we shouldn't have to clean up so often.


huguesv:
Will I still need to delete the packages folder when switching between 2015 and 2017?  I was confused when I was getting newtonsoft json dll load errors in unit tests, then realized that the last time I had fetched packages was using the "other" version of VS, so I was building against the wrong newtonsoft.  Not a problem when running the product as VS redirects to the correct newtonsoft, but the unit tests don't do that.

zooba:
You can install into `packages.14.0` and `packages.15.0` and those directories will be preferred by the build files.

huguesv:
👍

PR #2269, opened 2017-03-10
Action provider telemetry

zooba:
This should help identify telemetry being sent back that relates to our commands.

(Internal request from the VS team)


huguesv:
👍

PR #2270, opened 2017-03-11
Fixes some performance issues

zooba:
Fixes some performance issues:
* response to AddBulkFiles was delayed until after all parsing had completed
* reduce logging in TaskProvider by default
Also fixes a null reference crash when hovering over text.


zooba:
Feel free to hit merge if no concerns.

PR #2297, opened 2017-03-16
Slightly improves cookiecutter accessibility.

zooba:


Python/Product/Cookiecutter/ViewModel/CategorizedViewModel.cs
```diff
@@ -32,7 +32,7 @@ class CategorizedViewModel : INotifyPropertyChanged {
             _displayName = displayName;
         }
 
-        public bool Selectable => false;
+        public bool Selectable => true;
```
huguesv: If you haven't already, check that things work properly when the category is selected. For example, menu/toolbar items disabled as appropriate, template description section at the bottom is correct.

Python/Product/Cookiecutter/View/CookiecutterSearchPage.xaml
```diff
@@ -146,7 +145,8 @@
                 <HierarchicalDataTemplate DataType="{x:Type vm:CategorizedViewModel}" ItemsSource="{Binding Path=Templates}">
                     <TextBlock Margin="0 4 0 4"
                                FontWeight="Bold"
-                               Text="{Binding Path=DisplayName}"/>
+                               AutomationProperties.ItemType="Category"
```
huguesv: Looks like "Category" should be localized?  It gives examples of "Document File" and "Folder".
https://msdn.microsoft.com/en-us/library/system.windows.automation.automationproperties.itemtype(v=vs.110).aspx

huguesv:
Looks good otherwise

Python/Product/Cookiecutter/View/CookiecutterSearchPage.xaml
zooba: It doesn't work anyway - I'll remove it.

Python/Product/Cookiecutter/ViewModel/CategorizedViewModel.cs
zooba: Thought I checked this, but I'll check again.

zooba: Yep, it works fine.

PR #2309, opened 2017-03-21
Fixes #2306 No way to get PythonProject from IWpfTextView

### LINKED ISSUE #2306: No way to get PythonProject from IWpfTextView
Extensions that add editor adornments are notified on text view creation, but do not have any way to get the associated PythonProject and hence, the ProjectAnalyzer they need to query analysis info.

We should provide some way of getting the project - probably via an `IPythonProjectProvider` object in the text view property bag.

(We might also want global notifications for project creation, since there is no easy way to search all possible places where a project may exist.)

zooba:
Fixes #2306 No way to get PythonProject from IWpfTextView
Moves analyzer getters into shared service

This is still in progress, but interested in feedback on the approach.


zooba:
@DinoV This is the PR I wouldn't mind you having a look at. I think it's basically ready now in its current form, but not 100% convinced it's the right approach.

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntryService.cs
```diff
@@ -0,0 +1,198 @@
...
+            }
+            return null;
+        }
+
+        #region IAnalysisEntryService members
+
+        public ProjectAnalyzer DefaultAnalyzer => _site.GetPythonToolsService().DefaultAnalyzer;
+
+        public Task WaitForAnalyzerAsync(ITextBuffer textBuffer, CancellationToken cancellationToken) {
+            if (textBuffer.Properties.ContainsProperty(typeof(VsProjectAnalyzer))) {
```
huguesv: >textBuffer [](start = 16, length = 10)

Given this is a public API we are exposing, you should throw ArgumentNullException instead of causing a NRE

huguesv: Same for all 3 functions below.

---
In reply to: [108487412](https://github.com/Microsoft/PTVS/pull/2309#discussion_r108487412) [](ancestors = 108487412)

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntryService.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// For files on disk this is pretty easy - we analyze each file on it's own in a buffer parser.  
+        /// Therefore we map filename -> analyzer and then get the analysis from teh analyzer.  If we
+        /// determine an analyzer but the file isn't loaded into it for some reason this would return null.
+        /// We can also apply some policy to buffers depending upon the view that they're hosted in.  For
+        /// example if a buffer is outside of any projects, but hosted in a difference view with a buffer
+        /// that is in a project, then we'll look in the view that has the project.
+        /// 
+        /// For interactive windows we will use the analyzer that's configured for the window.
+        /// </summary>
+        public bool TryGetAnalysisEntry(ITextView textView, ITextBuffer textBuffer, out AnalysisEntry entry) {
```
huguesv: >TryGetAnalysisEntry [](start = 20, length = 19)

You use these a lot, but don't expose them in IAnalysisEntryService, and I'm not sure why.

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntryService.cs
```diff
@@ -0,0 +1,198 @@
...
+                tcs.TrySetResult(null);
+                textBuffer.Properties.RemoveProperty(_waitForAnalyzerKey);
+            }
+        }
+
+        /// <summary>
+        /// Gets the analysis entry for the given view and buffer.
+        /// 
+        /// For files on disk this is pretty easy - we analyze each file on it's own in a buffer parser.  
+        /// Therefore we map filename -> analyzer and then get the analysis from teh analyzer.  If we
```
huguesv: >teh [](start = 81, length = 3)

typo

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntryService.cs
zooba: AnalysisEntry is part of the internal API.

huguesv: What's beneficial about the IAnalysisEntryService public interface then?  They can get the ProjectAnalyzer, what will they do with it?

---
In reply to: [108496312](https://github.com/Microsoft/PTVS/pull/2309#discussion_r108496312) [](ancestors = 108496312)

huguesv:
👍 if @DinoV doesn't have time to review I would say you can get this merged.

PR #2312, opened 2017-03-21
Fixes #2279 ExecuteInRepl command may let exception escape

### LINKED ISSUE #2279: ExecuteInRepl command may let exception escape
Where it calls `GetLaunchConfigurationOrThrow` an exception may be raised for misconfigured environments. Because it's in an `async void` method, this will cause VS to crash.

We shouldn't crash here.

zooba:
Fixes #2279 ExecuteInRepl command may let exception escape
Handles exceptions and displays appropriate user feedback.


huguesv:
👍

PR #2321, opened 2017-03-24
Fixes #2319 Unexpected error dialog needs version number

### LINKED ISSUE #2319: Unexpected error dialog needs version number
When people copy-paste the error message into a bug (such as #2318 and #2317), it isn't obvious what version of PTVS they're using.

We should include the build version in that message.

zooba:
Fixes #2319 Unexpected error dialog needs version number
Adds version number to details.
Moves strings to resource files so they will be localized.

Note that the details string is deliberately non-localized, since the intent is to send it back to the development team for review (and we all speak English right now).


huguesv:
👍  this should be very useful

PR #2322, opened 2017-03-24
Fixes #2320 Unknown interpreter ID keeps causing upgrades

### LINKED ISSUE #2320: Unknown interpreter ID keeps causing upgrades
(From https://developercommunity.visualstudio.com/content/problem/31074/python-tools-pyproj-endless-migration.html)

When we run a project upgrade, if we don't recognize an interpreter ID then we leave it alone. This means that next time the project is opened, we will prompt to upgrade again.

We should remove the unrecognized ID and display a clear message explaining what was changed.

zooba:
Fixes #2320 Unknown interpreter ID keeps causing upgrades
Removes unknown IDs rather than ignoring them

### DIFF
--- a/Python/Product/Common/Strings.Designer.cs
+++ b/Python/Product/Common/Strings.Designer.cs
@@ -4993,6 +4993,15 @@ namespace Microsoft.PythonTools {
         }
         
         /// <summary>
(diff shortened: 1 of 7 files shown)


huguesv:
👍

Issue #2063, opened 2017-01-12
Asserts when loading project


huguesv:
Using latest from master

I created a new ironpython wpf app
I realized I was missing IronPython so I closed VS, I installed IronPython 2.7.7
I start VS, open my previously created project.

This first assert happens every time:
```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
Analyzer initialization failed with no result



   at Microsoft.PythonTools.Intellisense.VsProjectAnalyzer.<.ctor>b__27_3(Task`1 task) in C:\Projects\GitHub\huguesv\ptvs\Python\Product\PythonTools\PythonTools\Intellisense\ProjectAnalyzer.cs:line 184

   at System.Threading.Tasks.ContinuationTaskFromResultTask`1.InnerInvoke()

   at System.Threading.Tasks.Task.Execute()

   at System.Threading.Tasks.Task.ExecutionContextCallback(Object obj)

   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)

   at System.Threading.Tasks.Task.ExecuteEntry(Boolean bPreventDoubleExecution)

   at System.Threading.Tasks.Task.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()

   at System.Threading.ThreadPoolWorkQueue.Dispatch()

   at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback()


---------------------------
Abort   Retry   Ignore   
---------------------------
```

This one I only saw once:
```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
An exception in a task was not observed. See ActivityLog.xml for more details.

System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. ---> System.ArgumentException: Stream was not readable.

   at System.IO.StreamReader..ctor(Stream stream, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize, Boolean leaveOpen)

[...]

PR #2323, opened 2017-03-24
Change debugger protocol from binary to json.

huguesv:
Use our existing ipc json infrastructure to replace the binary protocol used by debugger/ptvsd.
Add minimal ipc json infrastructure python implementation (client only, socket only).
Add a couple of tests at the ipc json infrastructure level for both c# and python.
Fix django debugger test data so they pass on latest django version.
Various other test fixes, such as handling the new VS UI for reporting exceptions as editor adornment.
Removed support for Python 2.5 debugging, as it doesn't have the json module. You'll get a message box saying it's not supported when you start debugging or attach.

While the debugger tests mostly pass, some of them only are when you run them individually a second time.

The code on both sides of the debugger (but mostly Python side) needs to better handle malformed/invalid data, and there are no tests for that yet.

I discovered inconsistencies in the debug script working folder behavior, but haven't changed anything regarding this, so this DebugProject tests still fail.
- StartWithDebuggingSubfolderInProject

These other DebugProject tests sometimes fail due to https://github.com/Microsoft/PTVS/issues/2063:
- DebugPythonProjectSubFolderStartupFileSysPath
- StartWithoutDebuggingInProject
- StartWithoutDebuggingNotInProject

These DebugProject tests pass, but are pretty flaky despite my attempts to improve them:
- TestLaunchWithErrorsDontRun
- TestShowCallStackOnCodeMap

I need to investigate this DebugProject test failure:
- ExpressionEvaluation

The following AttachTests tests always fail, and they were passing before my changes.  I haven't spent significant time debugging them yet.
- AttachTimeout
- AttachTimeoutThreadsInitialized

This test failed before my changes and still fails:
- AttachTests.AttachNewThread_PyGILState_Ensure

There's a build with these changes here:
`\rolldata\PTVS\Build\PTVS-Github-Huguesv-Dev15\17083.2\release`


Python/Product/Debugger/Debugger/DebugEngine/AD7BoundBreakpoint.cs
```diff
@@ -51,7 +50,7 @@ class AD7BoundBreakpoint : IDebugBoundBreakpoint2 {
 
             if (!_deleted) {
                 _deleted = true;
-                _breakpoint.Remove();
+                TaskExtensions.RunSynchronouslyOnUIThread(ct => _breakpoint.RemoveAsync(ct));
```
zooba: This isn't really an extension method anymore if we need to use the class name each time... do we have a Utilities or Helpers class? `*Extensions` should be almost entirely extension methods.

zooba: If not, making it a method of some object that we always have here (`_engine` maybe?) would work (and might also allow us to detect timeouts more reliably and send telemetry info if we detect problems)

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteProcess.cs
```diff
@@ -91,41 +88,62 @@ private PythonRemoteProcess(int pid, Uri uri, PythonLanguageVersion langVer)
                 }
             } while (stream == null);
 
+            var debugConn = new DebugConnection(stream);
             bool connected = false;
             try {
-                string sig = stream.ReadAsciiString(DebuggerSignature.Length);
-                if (sig != DebuggerSignature) {
-                    throw new ConnectionException(ConnErrorMessages.RemoteUnsupportedServer);
+                var connectedEvent = new AutoResetEvent(false);
```
zooba: Needs dispose

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteProcess.cs
```diff
@@ -91,41 +88,62 @@ private PythonRemoteProcess(int pid, Uri uri, PythonLanguageVersion langVer)
...
+
+                string serverDebuggerName = string.Empty;
+                int serverDebuggerProtocolVersion = 0;
+
+                EventHandler<LDP.RemoteConnectedEvent> eventReceived = delegate (object sender, LDP.RemoteConnectedEvent ea) {
+                    serverDebuggerName = ea.debuggerName;
+                    serverDebuggerProtocolVersion = ea.debuggerProtocolVersion;
+                    connectedEvent.Set();
```
zooba: Handle disposed

Python/Product/Debugger/Debugger/LegacyDebuggerProtocol.cs
```diff
@@ -0,0 +1,593 @@
...
+        //////////////////////////////////////////////////////////////////////
+        /// Requests
+        //////////////////////////////////////////////////////////////////////
+
+        public sealed class StepIntoRequest : GenericRequest {
+            public const string Command = "legacyStepInto";
+
+            public override string command => Command;
+
+            public long threadId;
```
zooba: Does JSON handle longs okay? Would we be better to send these as strings (in case someone takes the JavaScript part seriously and casts them to doubles)?

Python/Product/Debugger/Debugger/LegacyDebuggerProtocol.cs
```diff
@@ -0,0 +1,593 @@
...
+
+        public enum LanguageKind {
+            Python = 0,
+            Django = 1,
+        }
+
+        public enum FrameKind {
+            None,
+            Python,
+            Django,
```
zooba: Let's start with explicit values on all enums, just to make it clear that they shouldn't be reordered.

Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -197,14 +176,31 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
...
+                        _connection.LegacyLast -= OnLegacyLast;
+                        _connection.LegacyModuleLoad -= OnLegacyModuleLoad;
+                        _connection.LegacyProcessLoad -= OnLegacyProcessLoad;
+                        _connection.LegacyRequestHandlers -= OnLegacyRequestHandlers;
+                        _connection.LegacyStepDone -= OnLegacyStepDone;
+                        _connection.LegacyThreadCreate -= OnLegacyThreadCreate;
+                        _connection.LegacyThreadExit -= OnLegacyThreadExit;
+                        _connection.LegacyThreadFrameList -= OnLegacyThreadFrameList;
+                        _connection.LegacyInternalError -= OnLegacyInternalError;
+                        _connection.Dispose();
```
zooba: >_connection.Dispose(); [](start = 24, length = 22)

_connection = null;

Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -620,476 +535,337 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
...
-                _stream.Write(StepOutCommandBytes);
-                _stream.WriteInt64(threadId);
-            }
+        private void OnLegacyInternalError(object sender, LDP.InternalErrorEvent e) {
+            // This event is sent when an internal error occurs
+            // on python side while processing a request.
+            // Request handlers should probably catch errors and send them back in the response
+            // instead of letting it propagate up.
+            Debug.Fail(e.output);
         }
```
zooba: Can we write this to an output window or somewhere the user will see in a release build? Maybe create an exception object and use ReportUnhandledException?

Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -620,476 +535,337 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
...
+
+            TaskCompletionSource<int>[] completions;
+            lock (_pendingGetThreadFramesRequestsLock) {
+                completions = _pendingGetThreadFramesRequests.ToArray();
+                _pendingGetThreadFramesRequests.Clear();
             }
 
-            return tcs.Task;
+            foreach (TaskCompletionSource<int> tcs in completions) {
+                tcs.SetResult(0);
```
zooba: >SetResult [](start = 20, length = 9)

TrySetResult - or else we'll crash if we race with a cancellation.

Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -1131,201 +907,176 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
...
 
-            DebugWriteCommand("Set Line Number");
-            lock (_streamLock) {
-                _setLineResult = false;
-                _stream.Write(SetLineNumberCommand);
-                _stream.WriteInt64(pythonStackFrame.Thread.Id);
-                _stream.WriteInt32(pythonStackFrame.FrameId);
-                _stream.WriteInt32(lineNo);
-            }
+            var timeoutSource = new CancellationTokenSource(2000);
```
zooba: >CancellationTokenSource(2000) [](start = 36, length = 29)

Timeouts via CTSs make me nervous - can you run a test pass with timeouts set to 0 to make sure that we don't have unhandled exceptions leaking out? (You should also use the CancellationTokens static class here, which will prevent cancellations while the debugger is attached so we can step through this code. It should also make it easier for you to return an "already cancelled" token for testing purposes.)

Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -1131,201 +907,176 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
...
-                }
-            }
+            var requestTask = SendDebugRequestAsync(new LDP.SetLineNumberRequest() {
+                threadId = pythonStackFrame.Thread.Id,
+                frameId = pythonStackFrame.FrameId,
+                lineNo = lineNo,
+            }, linkedSource.Token);
+
+            try {
+                return await requestTask.ContinueWith(t => {
```
zooba: >await requestTask.ContinueWith [](start = 23, length = 30)

Why do we need to await/ContinueWith? If you want to avoid trying to marshal back to the original thread, use await requestTask.ConfigureAwait()

Python/Product/Debugger/Debugger/PythonThread.cs
```diff
@@ -31,34 +33,30 @@ class PythonThread {
             _name = "";
         }
 
-        public void StepInto() {
-            _process.SendStepInto(_identity);
+        public Task StepIntoAsync(CancellationToken ct) {
+            return _process.SendStepIntoAsync(_identity, ct);
```
zooba: The => syntax also works for methods (as well as read-only properties), if you wanted to simplify these down. I don't have any preference either way (though I do prefer not using async and await for trivial functions like these)

Python/Product/Debugger/Debugger/TaskExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using static System.FormattableString;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.PythonTools.Debugger {
+    class TaskExtensions {
```
zooba: >TaskExtensions [](start = 10, length = 14)

`static class TaskHelpers`

Python/Product/Debugger/Debugger/TaskExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+        private static readonly Lazy<IVsThreadedWaitDialogFactory> _twdf = new Lazy<IVsThreadedWaitDialogFactory>(() => {
+            var twdf = (IVsThreadedWaitDialogFactory)Package.GetGlobalService(typeof(SVsThreadedWaitDialogFactory));
```
zooba: >Package [](start = 53, length = 7)

Do we have anywhere suitable to get an `IServiceProvider` from? Relying on `GetGlobalService` is a great way to break tests...

zooba: Then again, we're also reliant on `ThreadHelper` and we can't fix that one. Carry on...

---
In reply to: [107987804](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107987804) [](ancestors = 107987804)

Python/Product/Debugger/Debugger/TaskExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (twdf == null) {
+                var err = Invariant($"{nameof(SVsThreadedWaitDialogFactory)} is not available");
+                Trace.Fail(err);
+                throw new InvalidOperationException(err);
```
zooba: >InvalidOperationException [](start = 26, length = 25)

This will likely crash the entire process if it occurs. Better to just not show the dialog.

Python/Product/Ipc.Json/Connection.cs
```diff
@@ -47,15 +48,21 @@ public sealed class Connection : IDisposable {
...
         /// <param name="types">A set of registered types for receiving requests and events.  The key is "request.command_name" or "event.event_name"
         /// where command_name and event_name correspond to the fields in the Request and Event objects.  The type is the type of object
         /// which will be deserialized and instantiated.  If a type is not registered a GenericRequest or GenericEvent object will be created
-        /// which will include the complete body of the request as a dictionary.</param>
+        /// which will include the complete body of the request as a dictionary.
+        /// </param>
+        /// <param name="connectionLogKey">Name of registry key used to determine if we should log messages and exceptions to disk.</param>
         public Connection(
             Stream writer,
+            bool ownWriter,
```
zooba: >ownWriter [](start = 17, length = 9)

Make this `disposeWriter` and `disposeReader`

Python/Product/Ipc.Json/Connection.cs
```diff
@@ -253,6 +270,8 @@ public sealed class Connection : IDisposable {
                     var type = packet["type"].ToObject<string>();
                     var seq = packet["seq"].ToObject<int?>();
 
+                    Debug.WriteLine(string.Format("Received packet seq={0}", seq));
```
zooba: >Debug.WriteLine(string.Format("Received packet seq={0}", seq)); [](start = 20, length = 63)

Do we need this? I suspect it'll be noisy

zooba:
>                         packet = JsonConvert.DeserializeObject<JObject>(line);

This is the point that fuzz testing will break stuff. I'm also a little concerned about the WriteError() further below, especially for when we're using this within VS (why would we want to write the error back to the remote process? Maybe it doesn't?)

Just a point to keep in mind.

---
Refers to: Python/Product/Ipc.Json/Connection.cs:255 in 808e07d. [](commit_id = 808e07dd945600508a916753f49edd1739b7a015, deletion_comment = False)

Python/Product/PythonTools/visualstudio_py_debugger.py
```diff
@@ -2071,22 +2101,11 @@ def report_exception(frame, exc_info, tid, break_type):
...
-            # debug options, we'll send this format.
-            write_bytes(conn, EXCP)
-            write_string(conn, str(data['typename']))
-            write_int(conn, tid)
-            write_int(conn, 1 if 'breaktype' in data else 0)
-            write_string(conn, str(data['message']))
+    send_debug_event(
+        name='legacyException',
+        threadId=tid,
+        data=[{'key':k, 'val':str(v)} for k, v in data.items()],
```
zooba: >[{'key':k, 'val':str(v)} for k, v in data.items()] [](start = 13, length = 50)

Does this need to be a list of pairs? Can it just be "dict((k, str(v)) for k, v in data.items())"?

zooba: Guessing it's minimal change right now. Probably not worth doing anything about it

---
In reply to: [107989912](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107989912) [](ancestors = 107989912)

Python/Product/PythonTools/visualstudio_py_ipcjson.py
```diff
@@ -0,0 +1,268 @@
...
+# 
+# THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+# OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+# IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+# MERCHANTABLITY OR NON-INFRINGEMENT.
+# 
+# See the Apache Version 2.0 License for specific language governing
+# permissions and limitations under the License.
+
+from __future__ import division, with_statement, absolute_import
```
zooba: >division [](start = 23, length = 8)

I see no uses of division here - we can remove this

Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
```diff
@@ -16,25 +16,29 @@
...
 using Microsoft.PythonTools.Parsing;
+using Microsoft.VisualStudio.InteractiveWindow;
+using Microsoft.VisualStudio.InteractiveWindow.Commands;
 using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Utilities;
 using Microsoft.VisualStudioTools;
 
 namespace Microsoft.PythonTools.Repl {
+    using LDP = Microsoft.PythonTools.Debugger.LegacyDebuggerProtocol;
+
```
zooba: Move outside the namespace

Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
```diff
@@ -101,8 +109,9 @@ internal class PythonDebugReplEvaluator :
             if (engine != null) {
                 long? activeThreadId = ((IThreadIdMapper)engine).GetPythonThreadId((uint)_serviceProvider.GetDTE().Debugger.CurrentThread.ID);
                 if (activeThreadId != null) {
-                    AttachProcess(engine.Process, engine);
-                    ChangeActiveThread(activeThreadId.Value, false);
+                    AttachProcessAsync(engine.Process, engine).ContinueWith(t => {
+                        ChangeActiveThread(activeThreadId.Value, false);
+                    });
```
zooba: >; [](start = 22, length = 1)

.HandleAllExceptions(GetType()).DoNotWait()

Python/Tests/Core/DebugReplEvaluatorTests.cs
```diff
@@ -334,14 +363,19 @@ public class DebugReplEvaluatorTests {
...
+            await _evaluator.AttachProcessAsync(process, new MockThreadIdMapper());
         }
 
         private class MockThreadIdMapper : IThreadIdMapper {
             public long? GetPythonThreadId(uint vsThreadId) {
                 return vsThreadId;
             }
         }
+
+        protected static CancellationToken TimeoutToken() {
```
zooba: >TimeoutToken [](start = 43, length = 12)

Use CancellationTokens class

Python/Tests/DebuggerTests/IpcJsonTests.cs
```diff
@@ -0,0 +1,312 @@
...
+using System.Threading.Tasks;
+using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Ipc.Json;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TestUtilities;
+using TestUtilities.Python;
+
+namespace DebuggerTests {
+    [TestClass]
+    public class IpcJsonTests {
```
zooba: >IpcJsonTests [](start = 17, length = 12)

If these are not specific to debugging, can we move them to AnalysisTests? And make them Priority(0).

Python/Tests/DebuggerUITests/DebugProject.cs
```diff
@@ -33,6 +33,7 @@
 using TestUtilities.UI.Python;
 using Assert = Microsoft.VisualStudio.TestTools.UnitTesting.Assert;
 using Thread = System.Threading.Thread;
+using Microsoft.VisualStudioTools;
```
zooba: Before the `using Assert ...`

Python/Tests/TestData/DjangoDebugProject/DjangoDebugProject.pyproj
```diff
@@ -11,7 +11,7 @@
     </SearchPath>
     <WorkingDirectory>.</WorkingDirectory>
     <LaunchProvider>Django launcher</LaunchProvider>
-    <WebBrowserUrl></WebBrowserUrl>
+    <WebBrowserUrl>http://localhost</WebBrowserUrl>
```
zooba: >http://localhost [](start = 19, length = 16)

Is this change likely to cause the test machines to end up with 1000s of open browser windows?

Python/Tests/TestData/SysPath/SysPath.pyproj
```diff
@@ -32,5 +31,9 @@
   <ItemGroup>
     <Folder Include="Sub\" />
   </ItemGroup>
+  <ItemGroup>
+    <InterpreterReference Include="Global|PythonCore|2.7-32" />
+  </ItemGroup>
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\Python Tools\Microsoft.PythonTools.targets" />
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\Python Tools\Microsoft.PythonTools.targets" />
-</Project>
+</Project>
```
zooba: Don't want this twice (I've already fixed the bug that caused this)

Python/Tests/TestData/DjangoDebugProject/DjangoDebugProject.pyproj
huguesv: Probably. Do we still have hope that test machines will be able to handle the UI tests? ;)
At least one test using this project is setting a breakpoint and launch debugging, and waits for the breakpoint to hit.

---
In reply to: [107991880](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107991880) [](ancestors = 107991880)

Python/Tests/DebuggerTests/IpcJsonTests.cs
huguesv: I meant to be move them somewhere with fewer dependencies than debugger tests, and forgot.  I'll try moving them to AnalysisTests.

---
In reply to: [107991525](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107991525) [](ancestors = 107991525)

huguesv:
> # Must be in sync with enum PythonBreakpointConditionKind in PythonBreakpoint.cs

I'll update these 2 comments, they are supposed to match the values in LegacyDebuggerProtocol.cs now.

---
Refers to: Python/Product/PythonTools/visualstudio_py_debugger.py:184 in 808e07d. [](commit_id = 808e07dd945600508a916753f49edd1739b7a015, deletion_comment = False)

Python/Product/Ipc.Json/Connection.cs
huguesv: I'll remove it. The output printed by ProcessResponse should be sufficient. I added it when I was being paranoid when I was suffering from the TCS issue where my process messages thread was off executing client request continuation code instead of processing messages.

---
In reply to: [107988499](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107988499) [](ancestors = 107988499)

Python/Tests/DebuggerTests/IpcJsonTests.cs
huguesv: AnalysisTests doesn't depend on Ipc.Json project (Analyzer projects depends on Ipc.Json, but not Analysis).  So I don't think that's the right place either.  It would probably be best to be in its own test project.

---
In reply to: [108000250](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108000250) [](ancestors = 108000250,107991525)

Python/Product/PythonTools/visualstudio_py_debugger.py
huguesv: I'll change it.

---
In reply to: [107990008](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107990008) [](ancestors = 107990008,107989912)

Python/Product/Debugger/Debugger/LegacyDebuggerProtocol.cs
```diff
@@ -0,0 +1,593 @@
...
+            public override string command => Command;
+        }
+
+        public sealed class SetExceptionInfoRequest : GenericRequest {
+            public const string Command = "legacySetExceptionInfo";
+
+            public override string command => Command;
+
+            public int defaultBreakOnMode;
+            public ExceptionInfo[] breakOn;
```
huguesv: >breakOn [](start = 35, length = 7)

Although the intermediate representation in PythonProcess is in a Enumerable<KeyValuePair<string,int>, the original data from AD7Engine is stored in a dictionary<string,int>, and the final destination on Python side is also a dictionary.  It feels like it could be transported in json as a dictionary too.

Python/Product/Debugger/Debugger/PythonThread.cs
huguesv: I changed it. It would have been better if it fit on line, but that made some of the lines too long.  So in the end we're saving one line per method.

---
In reply to: [107987565](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107987565) [](ancestors = 107987565)

Python/Product/Debugger/Debugger/LegacyDebuggerProtocol.cs
huguesv: Sigh. JavaScript ruining your fun at every opportunity :(

---
In reply to: [107983458](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107983458) [](ancestors = 107983458)

Python/Product/Debugger/Debugger/PythonProcess.cs
huguesv: Will do.

---
In reply to: [107985719](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107985719) [](ancestors = 107985719)

Python/Product/Debugger/Debugger/PythonThread.cs
```diff
@@ -33,31 +33,23 @@ class PythonThread {
             _name = "";
         }
 
-        public Task StepIntoAsync(CancellationToken ct) {
-            return _process.SendStepIntoAsync(_identity, ct);
-        }
+        public Task StepIntoAsync(CancellationToken ct) =>
+            _process.SendStepIntoAsync(_identity, ct);
```
zooba: Now I see it, I actually prefer the version with braces.

I'm not so worried about long lines for trivial implementations like these. It's obviously not worth reading, so it doesn't really matter if it goes off the screen IMHO.

Python/Product/Debugger/Debugger/LegacyDebuggerProtocol.cs
huguesv: VSC protocol is defining thread id as a number
[https://github.com/Microsoft/vscode-debugadapter-node/blob/master/debugProtocol.json#L2102](https://github.com/Microsoft/vscode-debugadapter-node/blob/master/debugProtocol.json#L2102)
[https://github.com/Microsoft/vscode-debugadapter-node/blob/master/protocol/src/debugProtocol.ts#L1078](https://github.com/Microsoft/vscode-debugadapter-node/blob/master/protocol/src/debugProtocol.ts#L1078)
I would have been convinced if they used such a workaround themselves.

---
In reply to: [108021582](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108021582) [](ancestors = 108021582,107983458)

Python/Product/Debugger/Debugger/PythonProcess.cs
huguesv: No need, that was an oversight. Thanks!

---
In reply to: [107987006](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107987006) [](ancestors = 107987006)

Python/Product/Debugger/Debugger/PythonProcess.cs
huguesv: FYI ReportUnhandledException requires adding a reference on VSCommon.

---
In reply to: [108023087](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108023087) [](ancestors = 108023087,107985719)

huguesv: Looking at it more closely, my comment above is wrong.
on_request calls the specific request handler ie. on_legacyStepInto, and catches any exception and does a send_response(success=False, ...) with exception info.
The only way that an internal error would be sent is if that send_response reporting the error fails. If that fails, it's probably because there's a problem with the socket, and sending the internal error event would fail too.
So I'm thinking we just remove this whole internal error event and just do a traceback.print_exc() on the python side.

---
In reply to: [108259436](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108259436) [](ancestors = 108259436,108023087,107985719)

huguesv: It's removed now.

---
In reply to: [108267972](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108267972) [](ancestors = 108267972,108259436,108023087,107985719)

Python/Product/Debugger/Debugger/PythonThread.cs
huguesv: I reverted the change.

---
In reply to: [108208788](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108208788) [](ancestors = 108208788)

Python/Product/Debugger/Debugger/PythonProcess.cs
huguesv: I changed them all to use CancellationTokens. I've tested the 4 places where we use it in the product by forcing it to 0 ms. Error messages may not be ideal, but it doesn't break.

AD7Property.SetValueAsString: The value could not be set.  Please check your entry.

PythonProcess.SetLineNumberAsync: Unable to set the next statement. Operation not supported. Unknown error: 0x80004005.

AD7BoundBreakpoint.GetHitCount: No error from debugger, we let exception propagate out

AD7Property.EnumChildren: No error from debugger, we catch and return S_FALSE


---
In reply to: [107986767](https://github.com/Microsoft/PTVS/pull/2323#discussion_r107986767) [](ancestors = 107986767)

Python/Tests/DebuggerTests/IpcJsonTests.cs
huguesv: I've moved them to their own test project.

---
In reply to: [108011092](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108011092) [](ancestors = 108011092,108000250,107991525)

Python/Tests/IpcJsonTests/IpcJsonTests.cs
```diff
@@ -256,6 +256,8 @@ public class IpcJsonTests {
         }
     }
 
+#pragma warning disable 0649
```
zooba: What warning is this? Can you add a comment with the warning message as well, otherwise every time I see it I'm going to go look it up :)

Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
```diff
@@ -84,7 +84,9 @@ sealed class OutOfProcProjectAnalyzer : IDisposable {
             _projectFiles = new ProjectEntryMap();
             _connection = new Connection(
                 writer,
+                true,
```
int19h: Might want to use named arguments here to make it clear what these mean.

Python/Product/Common/Infrastructure/CancellationTokens.cs
```diff
@@ -20,7 +20,7 @@
 
 namespace Microsoft.PythonTools.Infrastructure {
     public static class CancellationTokens {
```
int19h: I'm confused about how this entire class is supposed to work. Aren't all these tokens essentially single-use for the lifetime of the process they're created in? And won't they start counting from the moment they're created?

Python/Product/Debugger/Debugger/DebugConnection.cs
```diff
@@ -0,0 +1,302 @@
...
+            _debuggerThread = new Thread(MessageProcessingThread);
+            _debuggerThread.Name = "Python Debugger Message Processing " + _processGuid;
+            _debuggerThread.Start();
+
+            while (true) {
+                lock (_isListeningLock) {
+                    if (_isListening) {
+                        break;
+                    }
+                    Thread.Sleep(10);
```
int19h: This looks like it should be outside of the lock?

int19h: Also, it looks like you could just wait on `_listeningReadyEvent`.

Python/Product/Debugger/Debugger/DebugConnection.cs
```diff
@@ -0,0 +1,302 @@
...
+            }
+
+            Debug.WriteLine("EventHandlingThread Ended");
+        }
+
+        private void HandleEvent(EventReceivedEventArgs e) {
+            Debug.WriteLine(string.Format("PythonProcess handling event: {0}", e.Event.name));
+            lock (_eventHandlingLock) {
+                Debug.Assert(e.Event.name == LDP.LocalConnectedEvent.Name || e.Event.name == LDP.RemoteConnectedEvent.Name || _isAuthenticated);
+                switch (e.Event.name) {
```
int19h: This will be shorter and typesafer if you use the new typeswitch on the event class:
```cs
switch (e.Event) {
case LDP.AsyncBreakEvent evt:
  LegacyAsyncBreak?.Invoke(this, evt);
  break;
case LDP.BreakpointFailedEvent evt:
  ...
}

```

Python/Product/Debugger/Debugger/DebugConnectionListener.cs
```diff
@@ -95,39 +98,64 @@ static class DebugConnectionListener {
...
-                        if (result == ConnErrorMessages.None) {
-                            targetProcess.Connect(stream);
-                            stream = null;
-                            socket = null;
+                var debugConn = new DebugConnection(stream);
+                try {
+                    string debugId = string.Empty;
+                    var result = ConnErrorMessages.None;
+                    using (var connectedEvent = new AutoResetEvent(false)) {
+                        EventHandler<LDP.LocalConnectedEvent> eventReceived = (object sender, LDP.LocalConnectedEvent ea) => {
```
int19h: You can declare named local functions in C# now! :)
```cs
void eventReceived(object sender, LDP.LocalConnectedEvent ea) { ... }
```

Python/Product/Debugger/Debugger/DebugConnectionListener.cs
```diff
@@ -95,39 +98,64 @@ static class DebugConnectionListener {
...
+                            result = (ConnErrorMessages)ea.result;
+                            debugId = ea.processGuid;
+                            try {
+                                connectedEvent.Set();
+                            } catch (ObjectDisposedException) {
+                            }
+                            debugConn.Authenticated();
+                        };
```
int19h: Should this also unsubscribe itself? Even if it doesn't get invoked twice, it's a closure that'll have a bunch of references...

int19h: Ah, never mind, I see the unsubscribe is handled elsewhere.

Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -561,7 +562,7 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
             _breakOnException.Clear();
             _defaultBreakOnExceptionMode = (int)enum_EXCEPTION_STATE.EXCEPTION_STOP_USER_UNCAUGHT;
 
-            SetExceptionInfo(_defaultBreakOnExceptionMode, _breakOnException);
+            TaskHelpers.RunSynchronouslyOnUIThread(ct => SetExceptionInfoAsync(_defaultBreakOnExceptionMode, _breakOnException, ct));
```
int19h: I notice that there are a lot of these peppered over the various AD7 classes. Why do we need them? These methods are all implementations of a COM interface invoked by the VS debugger, which is supposed to always invoke them on its UI thread (well, technically, in the STA apartment). So these should be no-ops. Do we actually see cases where the calls come on the background thread?

huguesv: It's about using JoinableTaskFactory to block until the asynchronous call is done. The TaskHelpers require you to be on the UI thread already.  The name can be misleading, thinking that it will do the switch to UI thread. It's just what they were named in RTVS.

---
In reply to: [108532812](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108532812) [](ancestors = 108532812)

zooba: The main thing it fixes is the continuations continuing as part of the COM message pump, rather than blocking other events. The first attempt at this regularly deadlocked between COM calls attempting to reach the UI thread while we were dealing with another event.

That said, I believe we are assuming that this call is coming from a non-UI thread, so having an assertion (even temporarily) to confirm that we're not unnecessarily using the helper. But I don't think it hurts.

Python/Product/Debugger/Debugger/DebugConnectionListener.cs
huguesv: While it would be nice to use latest C# features, I think I prefer keeping the ability to compile with VS 2015, at least until we're past our stable release.

---
In reply to: [108531583](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108531583) [](ancestors = 108531583)

Python/Product/Common/Infrastructure/CancellationTokens.cs
huguesv: I thought the same thing initially :) Each call to the getter will get you a new token (they're not singletons), so it's fine.

---
In reply to: [108528637](https://github.com/Microsoft/PTVS/pull/2323#discussion_r108528637) [](ancestors = 108528637)

Python/Product/Debugger/Debugger/DebugConnectionListener.cs
zooba: Agreed, mainly for the benefit of those people who are building PTVS themselves, to save them having to do too much work to get it to build.

zooba:
+1 on the new changes

huguesv:
I pushed a new change that fixes a regression during the updates to this PR, which was causing an assert when debug process ended and debug repl was being detached.

With this change the product appears to be working fine.  Debug Repl tests are still unstable (but pass when you re-run one by one).  Other tests pass, but there are frequent exceptions on background threads that cause events to be logged (seen in Event Viewer).  All are due to ObjectDisposedException, on the lock objects in Connection.cs.

zooba:
Those event log items are fine - we aggressively log NullReference and ObjectDisposed exceptions during tests (see `AssertListener.CurrentDomain_FirstChanceException`) because it helps with debugging crashes, but if they aren't causing crashes then it's fine to ignore them.

Most REPL tests are unstable due to shared state. It's tough to separate these out well, but we need to spend the time on it soon.

int19h:
👍

PR #2325, opened 2017-03-24
Fixes #2324 Should not have to reanalyze completion DB on minor updates

### LINKED ISSUE #2324: Should not have to reanalyze completion DB on minor updates
Currently we store all DBs in a version-specific folder. However, most of our updates do not change the format, and so we could easily migrate them forward.

Alternatively, we could go back to versioning the DB independently from the entire product, but this time using separate folders rather than an embedded value (which should deal with the issues that we used to suffer).

zooba:
Fixes #2324 Should not have to reanalyze completion DB on minor updates
Switches DB directory to include explicit format version

Removes test for deprecated/removed RequestContext
Fixes Templates project rebuilding so often


zooba:
@DinoV Went with a split version design - one version for compatibility (in the path to allow SxS); one version indicating the quality of the contents (to trigger reanalysis)

huguesv:
How does this affect the database stored in a virtual env's .ptvs folder?

zooba:
Ah, good point. We should write both values into the `database.ver` file to ensure that incompatible DBs aren't carried over.

zooba:
I made a simpler fix, which is to not reset `CurrentVersion` ever. That way we allow side-by-side VS versions to use different formats, but on a per-project level we will retrigger analysis (which is typically much faster) when opening it with a different supported format. (Essentially, this is status quo for venvs, and less frequent refreshes for global environments.)

huguesv:
👍

PR #2327, opened 2017-03-24
Fixes #2316 Mixed-mode attach fails

### LINKED ISSUE #2316: Mixed-mode attach fails
Running Python 3.5.3 32-bit, I open the Attach dialog in VS, select Python+Native, and attach to the process.

I get an invalid operation exception attaching the Python part. Digging in a little bit more, it looks like an invalid parameter error being returned from the `AD7LoadCompleteEvent.Send(this)` in [AD7Engine.cs](https://github.com/Microsoft/PTVS/blob/master/Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs#L326). This is a stopping event, which IIUC requires a thread object to be passed - but we aren't passing it.

Commenting out this event doesn't appear to break anything - do we need to include it? We don't have a thread object handy at the point where we are currently calling it.

zooba:
Fixes #2316 Mixed-mode attach fails
Corrects the AD7LoadCompleteEvent parameters


int19h:
👍

PR #2338, opened 2017-03-28
Fixes #2280 EnvironmentList ToolWindow crashes on invalid environment

### LINKED ISSUE #2280: EnvironmentList ToolWindow crashes on invalid environment
When constructing `EnvironmentView`, the tool window does not handle invalid argument exceptions. If an environment view cannot be created (presumably because of a race condition with the environment being removed), VS will crash.

We should not crash here. Ideally, we shouldn't add an invalid environment either.

zooba:
Fixes #2280 EnvironmentList ToolWindow crashes on invalid environment
Handles errors creating a view for an environment, including retrying the entire update to exclude "bad" entries rather than leaving them visible and unusable
Improves handling on the online help item, to avoid static construction

Also fixes layout of "make default" label.

### DIFF
--- a/Python/Product/EnvironmentsList/ToolWindow.xaml.cs
+++ b/Python/Product/EnvironmentsList/ToolWindow.xaml.cs
@@ -364,14 +364,19 @@ namespace Microsoft.PythonTools.EnvironmentsList {
                     select = selectView?.Configuration?.Id;
                 }
 
+                // We allow up to three retries at this process to handle race
+                // conditions where an interpreter disappears between the
+                // point where we enumerate known configuration IDs and convert
+                // them into a view. The first time that happens, we insert a
+                // stub entry, and on the subsequent pass it will be removed.
                 bool anyMissing = true;
                 for (int retries = 3; retries > 0 && anyMissing; --retries) {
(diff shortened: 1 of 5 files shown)


Python/Product/EnvironmentsList/ToolWindow.xaml.cs
```diff
@@ -363,29 +364,46 @@ await StartRefreshDBAsync(view)
...
+                                return _onlineHelpView;
+                            }
+                            var fact = _interpreters.FindInterpreter(c.Id);
+                            EnvironmentView view = null;
+                            try {
+                                if (fact != null) {
+                                    view = new EnvironmentView(_service, _interpreters, fact, null);
+                                }
+                            } catch (ArgumentException) {
+                            }
```
huguesv: What changes in the 2nd or 3rd retry that would make it work when it failed the first time?

zooba: Typically it'll fail when it races with an environment being removed, so we got the ID initially, but by the time we go to create the view it's gone. We'll add a stub item the first time through, and the second time we won't get the ID, so we'll remove the stub item.

huguesv:
👍 it could be worth adding that explanation as a comment.

Issue #2271, opened 2017-03-11
Go to definition or Find references not working


andyp05:
Using latest preview Release Date: March 7, 2017 (build 26304.00).

Hitting F12 will work to find definition of functions in imported or inherited modules, but won't find variable definitions. If in same module, it works fine. Get a 'Cannot go to definition' error.

The same issue for Ctrl-K,R Find all references.

zooba:
So I can F12 to a variable definition in another module just fine, though trying to rename it crashed VS (trace below). FAR also found all the references I expected, though some of the lines showed as blank in the results window.

I was working in our Flask web project template. Perhaps there's something about your project layout that's causing us trouble? Can you share your folder/file structure and show the ones where variables should be resolved properly?


```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ArgumentOutOfRangeException
   at System.Globalization.CompareInfo.Compare(System.String, Int32, Int32, System.String, Int32, Int32, System.Globalization.CompareOptions)
   at System.String.Compare(System.String, Int32, System.String, Int32, Int32)
   at Microsoft.PythonTools.Refactoring.LocationPreviewItem..ctor(Microsoft.PythonTools.Intellisense.VsProjectAnalyzer, Microsoft.PythonTools.Refactoring.FilePreviewItem, Microsoft.PythonTools.Intellisense.AnalysisLocation, Microsoft.PythonTools.Analysis.VariableType)
   at Microsoft.PythonTools.Refactoring.PreviewChangesEngine.CreatePreviewItems()
   at Microsoft.PythonTools.Refactoring.PreviewChangesEngine..ctor(System.IServiceProvider, Microsoft.PythonTools.Refactoring.IRenameVariableInput, System.String, Microsoft.PythonTools.Refactoring.RenameVariableRequest, System.String, System.String, Microsoft.PythonTools.Intellisense.VsProjectAnalyzer, System.Collections.Generic.IEnumerable`1<Microsoft.PythonTools.Intellisense.AnalysisVariable>)
   at Microsoft.PythonTools.Refactoring.VariableRenamer+<RenameVariable>d__4.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
[...]

PR #2341, opened 2017-03-29
Fixes refactor/rename crash from #2271

zooba:
Fixes refactor/rename crash from #2271
Rewrites name analysis for ease of reading

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Refactoring/LocationPreviewItem.cs
+++ b/Python/Product/PythonTools/PythonTools/Refactoring/LocationPreviewItem.cs
@@ -16,7 +16,9 @@
 
 using System;
 using System.Diagnostics;
+using System.Globalization;
 using Microsoft.PythonTools.Analysis;
+using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.Intellisense;
 using Microsoft.VisualStudio.Language.Intellisense;
 using Microsoft.VisualStudio.Shell.Interop;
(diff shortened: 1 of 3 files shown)


Python/Product/PythonTools/PythonTools/Refactoring/LocationPreviewItem.cs
```diff
@@ -30,66 +32,128 @@ namespace Microsoft.PythonTools.Refactoring {
...
         private static readonly char[] _whitespace = new[] { ' ', '\t', '\f' };
 
         public LocationPreviewItem(VsProjectAnalyzer analyzer, FilePreviewItem parent, AnalysisLocation locationInfo, VariableType type) {
-            _lineNo = locationInfo.Line;
-            _columnNo = locationInfo.Column;            
+            Debug.Assert(locationInfo.Column >= 1, "Invalid location info (Column)");
+            Debug.Assert(locationInfo.Line >= 1, "Invalid location info (Line)");
+            Line = locationInfo.Line;
+            //Column = locationInfo.Column;
             _parent = parent;
```
huguesv: If we return before Column is assigned, will that cause trouble?  It will be 0, and it's supposed to be 1-based.  For example, DisplayPreview below will end up setting a start index of -1.

zooba: If we return before Column is assigned, there's nothing to display. But we aren't checking that properly and we should be filtering out invalid items.

Python/Product/PythonTools/PythonTools/Refactoring/LocationPreviewItem.cs
```diff
@@ -30,66 +32,127 @@ namespace Microsoft.PythonTools.Refactoring {
...
+            }
+
+            return true;
         }
 
+        public int Line { get; }
+        public int Column { get; }
+        public VariableType Type { get; }
+
+        public int Length => _span.Length;
```
huguesv: >_span [](start = 29, length = 5)

_span may be null if you've returned early from the constructor.

huguesv: nevermind, I just noticed it's a struct.

---
In reply to: [109052129](https://github.com/Microsoft/PTVS/pull/2341#discussion_r109052129) [](ancestors = 109052129)

huguesv:
:shipit:

Python/Product/PythonTools/PythonTools/Refactoring/LocationPreviewItem.cs
zooba: Span is a struct - it'll be zero'd out.

PR #2350, opened 2017-03-30
Telemetry

zooba:
Reduces the level of information being collected for things we don't care about/get any value from.
Adds debug repl point.


huguesv:
👍

Issue #2357, opened 2017-03-31
Assert while remote debugging


huguesv:
This happened while debugging azure website.
```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
Invalid cross-thread call



   at Microsoft.VisualStudioTools.UIThreadExtensions.MustBeCalledFromUIThread(UIThreadBase self, String message) in C:\Projects\huguesv\PTVS\Python\Product\VSCommon\Infrastructure\UIThreadBase.cs:line 169

   at Microsoft.PythonTools.Extensions.GetPythonToolsService(IServiceProvider serviceProvider) in C:\Projects\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Extensions.cs:line 495

   at Microsoft.PythonTools.Repl.PythonDebugProcessReplEvaluator.<ConnectAsync>b__17_0(RemoteReplAttachResponse resp) in C:\Projects\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Repl\PythonDebugReplEvaluator.cs:line 579

   at Microsoft.PythonTools.Ipc.Json.Connection.RequestInfo`1.SetResponse(JToken obj) in C:\Projects\huguesv\PTVS\Python\Product\Ipc.Json\Connection.cs:line 613

   at Microsoft.PythonTools.Ipc.Json.Connection.ProcessResponse(JObject packet) in C:\Projects\huguesv\PTVS\Python\Product\Ipc.Json\Connection.cs:line 335

   at Microsoft.PythonTools.Ipc.Json.Connection.<ProcessMessages>d__26.MoveNext() in C:\Projects\huguesv\PTVS\Python\Product\Ipc.Json\Connection.cs:line 279

   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.InvokeMoveNext(Object stateMachine)

   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.Run()

   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c__DisplayClass4_0.<OutputAsyncCausalityEvents>b__0()

[...]

PR #2358, opened 2017-03-31
Add better error handling around attach to debug repl

huguesv:
Fix https://github.com/Microsoft/PTVS/issues/2357

Add error handling so that VS doesn't crash when attach to debug repl is refused or there's any other error around that request.

This will allow us to turn off the debug repl attach feature in attach_server.py, if necessary.


int19h:
👍

Issue #2369, opened 2017-04-05
Debugger Set Next Statement times out, causes trouble after


huguesv:
Set breakpoint in django template. I used django polls sample, details.html line 17.
When you hit the breakpoint, right-click on a different line and Set Next Statement (I tried to set it to details.html line 18).
It will time out after 2 secs and put an error message.
Then try to evaluate one of the variables (such as choice).  This will hang the debugger (you may be able to cancel though).

Python side of the debugger must be getting an exception that makes it exit the message processing loop, so we never get a response and it never listens for any more requests.

Edit: This is also a problem for regular Python debugging, not just Django templates. I verified that this is a regression with the debugger protocol rewrite.

huguesv:
I must have been testing it wrong, it appears to be working fine for regular Python debugging.  For Django templates, it was never supported.

The debugger hanging after attempting it in a Django template was because it wasn't properly releasing the THREADS lock when an exception was thrown, and so any future request handler that needed to acquire the lock was blocked.

PR #2371, opened 2017-04-06
Ensure that threads list lock is always released.

huguesv:
Fixes https://github.com/Microsoft/PTVS/issues/2369
Removes unused internal error function.


int19h:
👍

PR #2373, opened 2017-04-06
Accessibility fixes for environment window and profiling dialog

### LINKED ISSUE #2234: No screen reader output when tabbing through New Environment Config Form UI
Tool tips available, but when you hover over the textbox, not field name. This tip is not read by the screen reader.

huguesv:
Fixes #2234 No screen reader output when tabbing through New Environment Config Form UI
Fixes similar bug in the profiling dialog

This uses the AutomationProperties.LabeledBy attribute to associate a control with its label.  When tabbing to the control, the label is now read by narrator.

This also sets AutomationProperties.Name on the browse buttons so that for example narrator reads "Browse for interpreter path" followed by "button" instead of just reading "button" (because the button text is `...` which is not read).

Prevents options pane from receiving focus, as it's a container and non-interactive.

Prevent focus from going to images in labels and buttons in environments window.

### DIFF
--- a/Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
+++ b/Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
@@ -39,7 +39,8 @@
                                     </Grid.ColumnDefinitions>
                                     <Control Grid.Column="0"
                                                     Margin="3 0"
-                                                    Style="{StaticResource PythonEnvironmentImage}" />
+                                                    Style="{StaticResource PythonEnvironmentImage}"
+                                             IsTabStop="False"/>
                                     <TextBlock Grid.Column="1"
                                                TextWrapping="WrapWithOverflow"
                                                Text="{x:Static l:Resources.EnvironmentPathsExtensionMakeDefaultLabel}"/>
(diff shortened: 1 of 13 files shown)


Python/Product/EnvironmentsList/Resources.resx
```diff
@@ -223,6 +223,9 @@
   <data name="ConfigurationExtensionPrefixPathLabel" xml:space="preserve">
     <value>Prefix path</value>
   </data>
+  <data name="ConfigurationExtensionPrefixPathBrowseButton" xml:space="preserve">
+    <value>Browse folder dialog for prefix path</value>
```
zooba: Don't need "folder dialog" in these

Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
```diff
@@ -39,7 +39,8 @@
                                     </Grid.ColumnDefinitions>
                                     <Control Grid.Column="0"
                                                     Margin="3 0"
-                                                    Style="{StaticResource PythonEnvironmentImage}" />
+                                                    Style="{StaticResource PythonEnvironmentImage}"
+                                             IsTabStop="False"/>
```
zooba: Indentation

PR #2380, opened 2017-04-11
Fixes #2368 add new empty python file makes a ANSI encoding py file

### LINKED ISSUE #2368: add new empty python file makes a ANSI encoding py file
which should better be utf-8

zooba:
Fixes #2368 add new empty python file makes a ANSI encoding py file
Makes empty files non-empty so that our default encoding is selected.


huguesv:
👍

PR #2383, opened 2017-04-11
Fixes #2379 Import coverage file error

### LINKED ISSUE #2379: Import coverage file error
Create new python application project
Paste the following code in a `program.py` file:
```
import time

def f():
        for i in range(10000):
                time.sleep(0)

f()
```

From command line, run the profiler:
```
py -3.6 -m coverage run program.py
py -3.6 -m coverage xml
```

Close `program.py` if it's open in  the editor.

Tools, Python Tools, Import coverage file
Select coverage.xml from the project folder

Result:
```
---------------------------
Microsoft Visual Studio
---------------------------
Value does not fall within the expected range.
---------------------------
OK   
---------------------------
```

It's thrown by this piece of code, which is trying to get to the project, with the goal of getting the python environment.  When the filename is not open in the editor, `IVsRunningDocumentTable4.GetDocumentCookie` throws instead of returning `VSCOOKIE_NIL`.

```
        internal static PythonProjectNode GetProjectFromFile(this IServiceProvider serviceProvider, string filename) {
            var docTable = serviceProvider.GetService(typeof(SVsRunningDocumentTable)) as IVsRunningDocumentTable4;
            var cookie = docTable.GetDocumentCookie(filename);
            if (cookie != VSConstants.VSCOOKIE_NIL) {
```

zooba:
Fixes #2379 Import coverage file error
Switches AnalysisEntryService to use existing helper function
Minor refactoring

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Extensions.cs
+++ b/Python/Product/PythonTools/PythonTools/Extensions.cs
@@ -375,22 +375,27 @@ namespace Microsoft.PythonTools {
 
         internal static PythonProjectNode GetProjectFromFile(this IServiceProvider serviceProvider, string filename) {
             var docTable = serviceProvider.GetService(typeof(SVsRunningDocumentTable)) as IVsRunningDocumentTable4;
-            var cookie = docTable.GetDocumentCookie(filename);
-
-            if (cookie != VSConstants.VSCOOKIE_NIL) {
-                IVsHierarchy hierarchy;
-                uint itemid;
-                docTable.GetDocumentHierarchyItem(cookie, out hierarchy, out itemid);
-                var project = hierarchy.GetProject();
-                if (project != null) {
-                    return project.GetPythonProject();
-                }
-
-                object projectObj;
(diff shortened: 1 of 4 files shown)


Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -375,22 +375,27 @@ public static class Extensions {
 
         internal static PythonProjectNode GetProjectFromFile(this IServiceProvider serviceProvider, string filename) {
```
huguesv: I'd like to make it clear that this is not going to return the project if the file isn't opened in an editor.  Maybe add a summary comment or rename the method to make it more obvious.

huguesv:
The version used for the parser may end up being wrong just because none of the files were opened in the editor and the global default was picked, but I suppose that's not so terrible.

👍

zooba:
You're right - we should be searching all open projects here as well before falling back. Let me revise the API a bit.

zooba:
@huguesv New code added - can you take another look?

huguesv:
Looks ok 👍

PR #2388, opened 2017-04-13
Fixes analysis logging in tests.

zooba:
Fixes analysis logging in tests.
Adds test for #2271 (currently expected to fail)


huguesv:
👍

Issue #2129, opened 2017-02-08
Suggestsion: Consider not iterating over OLECMD[] arrays


JoshVarty:
There are a [few places](https://github.com/Microsoft/PTVS/blob/f71df86801f4bb1b75deaff2e874b73442be9c4d/Python/Product/PythonTools/PythonTools/Navigation/EditFilter.cs#L839-L847) in the `IOleCommandTargets` where we're iterating over `OLECMD[]` arrays. I've spoken to other Visual Studio team members and confirmed that these arrays should never have more than one element in them.

You can see this assumption has been made in Roslyn [here](http://source.roslyn.io/#Microsoft.VisualStudio.LanguageServices/Implementation/AbstractOleCommandTarget.Query.cs,ca2d73c9587da87b) where they always index `OLECMD[]` arrays with 0.

Why is it an array at all? It's my understanding that this is legacy as the `IOleCommandTarget` interface is borrowed from Windows: https://msdn.microsoft.com/en-us/library/windows/desktop/ms683797(v=vs.85).aspx

If you guys are comfortable with this change I'd be happy to go through and make it for you.

zooba:
I'm very neutral on this. I don't really see any reason to change code without any assumptions into code that makes assumptions, and it seems quite reasonable to query for multiple commands.

What I'd rather see fixed is the blatant bug where we return immediately and never get past element zero. It would seem from this that there are no cases where VS will query multiple commands at once, but since this is an easily invoked method I find it hard to rely on that.

@DinoV may have a relevant opinion here. For mine, if we did remove the loops, I'd want an assertion that nCmds is always 1 so we find out very quickly if that is ever invalidated.

JoshVarty:
>It would seem from this that there are no cases where VS will query multiple commands at once, but since this is an easily invoked method I find it hard to rely on that.

In part I'm thinking it might future-proof PTVS against changes in the commanding system. (eg. Not requiring that more than one command will ever be queried simultaneously). 

I'm prototyping a redesign of commanding within the editor and was just going through PTVS to ensure that we could help move you over with little pain.

zooba:
> In part I'm thinking it might future-proof PTVS against changes in the commanding system

Again, I don't see how making an assumption like this helps anything. The API is documented as passing a parameter specifying the number of elements in the array, and using that properly (which we currently don't, but that's a bug) makes us more robust than using it incorrectly.

[...]

PR #2393, opened 2017-04-13
Fixes #2260 Start with/without debugging launches the wrong file

### LINKED ISSUE #2260: Start with/without debugging launches the wrong file
Open a python project
Drag & drop a python file that doesn't belong in that project into the editor
Right-click on editor and select Start with debugging, or Start without Debugging.

Result: The project's startup file is launched, instead of the file that you right-clicked on.

`DebugProject.StartWithoutDebuggingNotInProject` fails while doing these exact steps.

zooba:
Fixes #2260 Start with/without debugging launches the wrong file
Removes global command handlers and handles them in the edit filter and project nodes.
Corrects the issue identified in #2129 for EditFilter
Fixes potential infinite loop when abandoning out-of-proc analyzers


int19h:
👍

huguesv:
👍  I'll retest remote debugging to Azure VM to ensure the stream reading change doesn't break it.

zooba:
It should be an okay change because of the `await` - if the stream is still alive but there's nothing to read it will wait, and it's only when the stream is actually complete (but not closed) that it exits. But I didn't test any of the cases where we provide a stream to it (rather than stdio), so I'll do a quick check before merging.

zooba:
It has no impact on local-remote debugging, and I don't see any reason why remote-remote debugging would be different. But please retest that once the build comes out.

PR #2396, opened 2017-04-13
Fixes #2395 VSHPROPID_Name should not return caption

### LINKED ISSUE #2395: VSHPROPID_Name should not return caption
Currently we return the caption, which is incorrect in cases where we've modified the caption.

We should return the name of the file/folder instead of whatever caption we are displaying.

zooba:
Fixes #2395 VSHPROPID_Name should not return caption
Adds Name property to allow overriding node names.
Override name by default for files and folders.


zooba:
Feel free to merge when happy - I'll be afk for a few days probably.

PR #2406, opened 2017-04-17
Fixes #2374 IPython does not work in specific project

### LINKED ISSUE #2374: VS2017 Ipython does not work in specific project
Dear all, 

I am running VS2017 preview and using Anaconda to build my virtual environments. I am able to add my custom virtual environments to "Python Environments" and use Ipython by calling the REPL from the option "Open Interactive Window". Unfortunately, I cannot use Ipython  when I execute the REPL from my python projects. In this case I have access only to the standard repl with no Ipython.

zooba:
Fixes #2374 IPython does not work in specific project
Changes script locator to look in interpreter's directory for projects.


huguesv:
👍

Issue #2402, opened 2017-04-15
Connection class not handling unicode properly


huguesv:
I'm currently writing low level message reading/parsing tests, and I'm finding out we have an issue with unicode text when encodedBytes.Length > text.Length, meaning there are multi-byte characters.

The message format is like this:
```
Content-Length: <content byte length>

<content>
```

When writing a message, we use the byte length for Content-Header value:
```
var str = JsonConvert.SerializeObject(packet);
var bytes = Encoding.UTF8.GetBytes(str);
var contentLengthStr = "Content-Length: " + bytes.Length + "\n\n";
var contentLength = Encoding.UTF8.GetBytes(contentLengthStr);

await _writer.WriteAsync(contentLength, 0, contentLength.Length).ConfigureAwait(false);
await _writer.WriteAsync(bytes, 0, bytes.Length).ConfigureAwait(false);
await _writer.FlushAsync(cancel).ConfigureAwait(false);
```

When reading, StreamReader.ReadAsync takes the number of chars to read, and but we give it a number bytes instead.  There are no methods on StreamReader that lets you do binary reads.

So in cases of multi-bytes characters, we read too much. Result will vary depending on whether there are messages queued up in the stream, as we will read into the next message.

As I'm writing tests I've been making the reading code more strict, and now it's obvious when there's a mismatch like this.

StreamReader is buffered, so it's not like we could temporarily read the bytes directly from the Stream, as the Stream position will likely be further ahead than we expect from the prior ReadLineAsync calls.

One solution is to stop using StreamReader altogether, just write our own reader class that has the ability to read both text and binary.  That's pretty much what I did on the Python side.

I'd like to confirm whether VSCode debugger protocol, which is similar that what we are using, considers Content-Length to be character or byte count.  If it's character count, then the solution is to make that change.

huguesv:
I've replaced StreamReader with my own class that can read a specific number of bytes, and can also do ReadLine, so far all my low level tests are passing but the debugger tests are all failing.

zooba:
@huguesv This is a fairly critical bug that we need to fix asap.

I'd suggest we use a binary reader - the separators are always `b"\r\n"`, so we can do a basic readline (perhaps emulated with a small buffer, if we need to) to get the headers.

zooba:
Also, it's byte count, not character count: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md

huguesv:
[...]

PR #2409, opened 2017-04-17
Fixes and tests for IPC JSON

huguesv:
Add lower-level tests for the reading of packets on C# side, fix issues discovered and refactor as necessary for tests.

Change the protocol to be closer to the VSC base protocol (ascii headers, `\r\n` separators).

The new reading code no longer allocates a char array based on the value of Content-Length, which could have caused out of memory errors if that value was incorrect.  In the new code an incorrect large value won't cause as much damage.

Fixes https://github.com/Microsoft/PTVS/issues/2402

### DIFF
--- a/Python/Product/Ipc.Json/ProtocolReader.cs
+++ b/Python/Product/Ipc.Json/ProtocolReader.cs
@@ -1,0 +1,121 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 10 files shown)


Python/Product/TestAdapter/visualstudio_py_testlauncher.py
```diff
@@ -101,23 +102,9 @@ def tell(self):
...
-
-    def send_event(self, name, **args):
-        with self.lock:
-            body = {'type': 'event', 'seq': self.seq, 'event':name, 'body':args}
-            self.seq += 1
-            content = json.dumps(body).encode('utf-8')
-            headers = ('Content-Length: %d\r\n\r\n' % (len(content), )).encode('ascii')
-            self.socket.send(headers)
-            self.socket.send(content)
+        super(_IpcChannel, self).__init__(socket=socket)
```
zooba: You can omit the `__init__` completely here if the arguments match.

Python/Product/Ipc.Json/ProtocolReader.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        /// <summary>
+        /// Reads bytes from the stream into the buffer, in chunks.
+        /// </summary>
+        /// <returns>Number of bytes that were added to the buffer.</returns>
+        private async Task<int> ReadIntoBuffer() {
+            var temp = new byte[1024];
+            var read = await _stream.ReadAsync(temp, 0, temp.Length);
+            _buffer.AddRange(temp.Take(read).ToArray());
```
zooba: Omit `ToArray` here, but since we know the size you can use `_buffer.Capacity += read;` to ensure we only reallocate once.

zooba: Better yet - `_buffer.Capacity = _buffer.Count + read;`

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
huguesv: I was debating this. Base class requires it to be pass by name.  I could change the caller to pass it by name instead of delete the init.  I think I'll do that.

Python/Product/Ipc.Json/ProtocolReader.cs
huguesv: Performance impact should be minimal.  I will do this when I have enough time to rerun tests.

PR #2410, opened 2017-04-17
Fixes #973 Text Editor Advanced options layout expands

### LINKED ISSUE #973: Text Editor Advanced options layout expands in Russian VS on Japanese OS
(From DD131362, for internal people)

OS: Windows 10 Enterprise JPN (winmain 10131.0.150526-1541) 
VS Enterprise 23009.00 RUS

image

The easy workaround is resizing the dialog, but we should enable autoscroll so that for weird situations where the layout changes we don't have to rely on resizing.

zooba:
Fixes #973 Text Editor Advanced options layout expands
Enables auto-scroll on options page
Also reduces control margins to match other pages


huguesv:
👍

Issue #2400, opened 2017-04-14
Publish ptvsd with next update


zooba:
Because we've changed the format of the debugger, existing `ptvsd` will not work. We need to publish an update asap.

We should just rev the minor version number on this one - people on VS 2015 should be installing (at worst) `ptvsd<3`, and everyone in VS 2017 is in preview so we can make the change. Also, with #750 planned I think we'll probably switch to proper semver after the refactor.

zooba:
Ping @huguesv and @int19h for awareness.

huguesv:
We currently have all our script version number set to `3.0.0.0`.  That includes files such as PythonScraper.py, pip_downloader.py, and all the files that go under ptvsd.

Do we want to have all the files under ptvsd be 3.1.0.0, and all of the others stay at 3.0.0.0?  Or bump them all to 3.1.0.0?

zooba:
Let's go to `3.1.0.0`. I have some reasons to accelerate parts of the refactor and publishing, so we'll likely bounce to `4.0` fairly quickly anyway.

zooba:
We need to do one more update and publish to get rid of the RC tag on `ptvsd`.

PR #2418, opened 2017-04-18
Update version from 3.0.0.0 to 3.1.0.0.

huguesv:
https://github.com/Microsoft/PTVS/issues/2400


zooba:
Only the files in `ptvsd` should have to change - everything else can remain 3.0 because we haven't changed their format.

huguesv:
Please double check, it should limited to the ptvsd files now.

PR #2427, opened 2017-04-20
Fixes #2424 Padding on "This environment cannot be made the default"

### LINKED ISSUE #2424: Padding on "This environment cannot be made the default"
image

Should align with the buttons below.

zooba:
Fixes #2424 Padding on "This environment cannot be made the default"
Removes unnecessary padding.
Also adjusts margin on support URL so spacing is correct when hidden.


huguesv:
👍

Issue #2420, opened 2017-04-19
REPL analyzer processes do not exit


zooba:
If you open 100 REPL windows, you will have 100 analyzer processes until you exit VS.

We should have them time out and close, or close them when the window is closed.

zooba:
Also, interactive analyzers should have a different comment on their command line.

PR #2428, opened 2017-04-20
Improves command line comment for interactive window analyzers

zooba:
Improves command line comment for interactive window analyzers
See #2420


huguesv:
👍

PR #2429, opened 2017-04-20
Fixes #2426 Cannot disambiguate Anaconda installs

### LINKED ISSUE #2426: Cannot disambiguate Anaconda installs
The display name for Anaconda is the same for a given version, regardless of language version or architecture.

We should detect the case where a display name is repeated (perhaps just within a given company?) and add disambiguators.

zooba:
Fixes #2426 Cannot disambiguate Anaconda installs
Adds a DisambiguateDescriptions function to InterpreterConfiguration and makes the current description part of the configuration identity.

### DIFF
--- a/Python/Product/Analysis/Interpreter/InterpreterConfiguration.cs
+++ b/Python/Product/Analysis/Interpreter/InterpreterConfiguration.cs
@@ -16,13 +16,13 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using Microsoft.PythonTools.Infrastructure;
 
 namespace Microsoft.PythonTools.Interpreter {
     public sealed class InterpreterConfiguration : IEquatable<InterpreterConfiguration> {
-        private string _description, _fullDescription;
+        private readonly string _description;
+        private string _fullDescription;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Python/Product/Analysis/Interpreter/InterpreterConfiguration.cs
```diff
@@ -55,7 +61,28 @@ public sealed class InterpreterConfiguration : IEquatable<InterpreterConfigurati
...
+
+            if (hasVersion && hasArch) {
+                // Regular description is sufficient
+                _fullDescription = null;
+            } else if (hasVersion) {
+                _fullDescription = "{0} ({1})".FormatUI(_description, Architecture);
+            } else if (hasArch) {
+                _fullDescription = "{0} ({1})".FormatUI(_description, Version);
+            } else {
+                _fullDescription = "{0} ({1}, {2})".FormatUI(_description, Version, Architecture);
```
huguesv: These will look strange if `_description` is empty, but looking at the code PythonRegistrySearch, it doesn't look like that happens in practice.

huguesv:
👍

zooba:
Everything will look strange if description is empty - please don't do that :)

PR #2433, opened 2017-04-21
Some accessibility fixes for pip package view

huguesv:
404497 Accessibility : MAS40B : INSPECT : Python+ Python Environments : Name property for the Search edit in Packages window is null
404519 Accessibility : MAS36 : Keyboard : 'Delete(X)' and 'Upgrade To set up tools(Up Arrow mark) buttons are not accessible in Packages window of Python Environment


zooba:
👍

PR #2438, opened 2017-04-24
Fixes #2352 Intellisense hangs VS when interacting with filepath that contains many files

### LINKED ISSUE #2352: Intellisense hangs VS when interacting with filepath that contains many files
I've got a folder with 90,000 text files in it. (I'm doing some machine learning and this contains my training set).

When I manipulate a string with this filepath, VS hangs momentarily. (I assume when populating the Intellisense object).

Perhaps we should consider setting an upper boundary on the number of elements we're willing to put in this dialog. 

Sample:

image

zooba:
Fixes #2352 Intellisense hangs VS when interacting with filepath that contains many files
Filters by partial name once items exceed 10000
Since the completion is synchronous and the disk IO causes the hang, we also disable automatic triggering for paths. You now need to Ctrl+Space/J to trigger completions in a string literal.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
@@ -258,9 +258,10 @@ namespace Microsoft.PythonTools.Intellisense {
                 var sigHelpSession = Volatile.Read(ref _sigHelpSession);
                 var literalSpan = GetStringLiteralSpan();
                 if (literalSpan.HasValue &&
-                    ShouldTriggerStringCompletionSession(prefs, literalSpan.Value) &&
+                    // UNDONE: Do not automatically trigger file path completions github#2352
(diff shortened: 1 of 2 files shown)


huguesv:
Code changes look okay, but I'm wondering about the max of 10000.  How does it perform with 9999 items?

zooba:
That's a slippery slope :P

The slow part is reading all the names from disk. That always reads all files, regardless of how many we filter to. A fix for that part is far more significant and I don't want to rush it in - having a manually triggered completion list is the best way to avoid it being an issue for now.

huguesv:
I was worried that 10,000 items is a lot to return, and if the IntelliSense UI was able to handle it without performance issues.  If that works fine then 👍

zooba:
Well, it's fine on my machine with 10K files :) I'd hope that anyone else who has 10K files on their machine also has a machine that can handle it, but it is entirely possible that some machines won't be able to (for which we should see telemetry data or hear feedback proving it)

PR #2441, opened 2017-04-25
More tests for ipc json protocol

huguesv:
Add tests for Python ipc json protocol socket reading code, for Python 2.6+ and IronPython.

Since that code reads using the socket functions, it was easiest to make the tests write the test data to a socket, unlike C# where I was able to use a MemoryStream and bypass the socket.

I refactored the C# tests so that the same test data is used for both Python and C# implementations.

There are changes to the product code, so I'll wait until 15.2 is done before merging.


zooba:
👍

Issue #2447, opened 2017-04-26
Opening the Python Environments tab from Tools\Python\Python Environments causes VS2017 to crash & restart.


JeffHanna:
Every time I open the Python Environments tab it opens, changes my cursor to a wait cursor, and then crashes VS 2017. I've uninstalled/reinstalled Python 3.6, the Python workflow in VS 2017 Community Preview, and VS 2017 Preview entirely. Nothing fixes this issue. 

[PTVS_Diagnostics_utf8.txt](https://github.com/Microsoft/PTVS/files/958835/PTVS_Diagnostics_utf8.txt)

zooba:
For convenience, I've copied the exception below. 

Typically this would occur because of an odd or unsupported path format, though I don't see anything in the log that would be the cause of that.

```
Description: The process was terminated due to an unhandled exception.
Exception Info: System.UriFormatException
   at System.Uri.CreateThis(System.String, Boolean, System.UriKind)
   at System.Uri..ctor(System.String, System.UriKind)
   at Microsoft.PythonTools.Infrastructure.PathUtils.MakeUri(System.String, Boolean, System.UriKind, System.String)

Exception Info: System.ArgumentException
   at Microsoft.PythonTools.Infrastructure.PathUtils.MakeUri(System.String, Boolean, System.UriKind, System.String)
   at Microsoft.PythonTools.Infrastructure.PathUtils.GetAbsoluteDirectoryPath(System.String, System.String)
   at Microsoft.PythonTools.Repl.PythonInteractiveEvaluator.GetScriptsPath(System.IServiceProvider, System.String, Microsoft.PythonTools.Interpreter.InterpreterConfiguration, Boolean)
   at Microsoft.PythonTools.InterpreterList.InterpreterListToolWindow.GetScriptPath(Microsoft.PythonTools.EnvironmentsList.EnvironmentView)
   at Microsoft.PythonTools.InterpreterList.InterpreterListToolWindow.OpenInteractiveScripts_CanExecute(System.Object, System.Windows.Input.CanExecuteRoutedEventArgs)
   at System.Windows.Input.CommandBinding.OnCanExecute(System.Object, System.Windows.Input.CanExecuteRoutedEventArgs)
   at System.Windows.Input.CommandManager.FindCommandBinding(System.Windows.Input.CommandBindingCollection, System.Object, System.Windows.RoutedEventArgs, System.Windows.Input.ICommand, Boolean)
[...]

PR #2448, opened 2017-04-26
Fixes how the default scripts location is determined.

zooba:
We shouldn't be assuming that it's in My Documents.

According to #2447 there may still be an actual crash involved here, and I don't think this will necessarily resolve it.


huguesv:
👍

PR #2468, opened 2017-04-26
Analyzer logging improvements

zooba:
Reduce the amount of "good" log messages and increase the number of "bad" messages to help with debugging issues.


huguesv:
👍

PR #2477, opened 2017-04-28
Cookiecutter accessibility fixes

huguesv:
More reasonable TAB and SHIFT-TAB behavior
- Cycles through all controls forward and backward, no longer takes you to the toolbar
- Load more is no longer a tab stop, just go to the tree node by navigating the tree with arrow keys

Focus now remains in the tree view after clicking load more, instead of going to the output window.

Toolbar is accessed with Shift-Alt (standard in VS)

405000 Accessibility : MAS05 : Keyboard : Python + Cookiecutter : Focus order improper for tool-bar in Cookie cutter window
403507 Accessibility : MAS05 : Keyboard : Python + Cookiecutter : Focus order improper in CookieCutter window after refreshing or loading the page
404511 Accessibility : MAS07 : Keyboard : Python+Cookiecutter : Keyboard trap found over 'Search' edit box when perfoming reverse tab (Shift +Tab) in CookieCutter window
403505 Accessibility : MAS36 : Keyboard : Python + Cookiecutter : 'Load more' is not accessible through reverse tab (Shift + TAB) in CookieCutter window

### DIFF
--- a/Python/Product/Cookiecutter/Cookiecutter.csproj
+++ b/Python/Product/Cookiecutter/Cookiecutter.csproj
@@ -161,6 +161,7 @@
     <Compile Include="Model\ILocalTemplateSource.cs" />
     <Compile Include="Model\TemplateContext.cs" />
     <Compile Include="Model\DteCommand.cs" />
+    <Compile Include="ViewModel\TreeItemViewModel.cs" />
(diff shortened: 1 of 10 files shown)


huguesv:
Oh, and in the tree view, I made ENTER load the template, just like ENTER loads more results.

zooba:
👍

PR #2483, opened 2017-05-01
Fixes #2481 No AnalysisEntryService in tests

### LINKED ISSUE #2481: No AnalysisEntryService in tests
Causes many test failures.

We should provide a mock service of some form.

zooba:
Fixes #2481 No AnalysisEntryService in tests
Ensures the entry service is added to the mock component model.


huguesv:
👍

PR #2486, opened 2017-05-02
More cookiecutter accessibility fixes

huguesv:
Bug 410725: Usability : Unable to invoke the context menu in Cookie cutter window using Shift key + F10
Bug 404076: Accessibility : MAS40B : INSPECT : Python + Cookiecutter : Name property is inappropriate for the interactive elements in Templates present in Cookiecutter window

Fixes alignment of search text box / button.
Prevents changing the selection if there's already a node that is selected.

### DIFF
--- a/Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
+++ b/Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
@@ -476,16 +476,20 @@ namespace Microsoft.CookiecutterTools.ViewModel {
             SearchResults.Add(Recommended);
             SearchResults.Add(GitHub);
 
+            // Ensure that there's a selection, to avoid focus issues
+            // when tabbing to the search results.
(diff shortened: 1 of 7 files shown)


Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -476,16 +476,20 @@ class CookiecutterViewModel : INotifyPropertyChanged {
             SearchResults.Add(Recommended);
             SearchResults.Add(GitHub);
 
+            // Ensure that there's a selection, to avoid focus issues
+            // when tabbing to the search results.
+            if (SearchResults.Where(cat => cat.IsSelected).Count() == 0) {
```
zooba: Prefer `SearchResults.Any(cat => cat.IsSelected)` to save enumerating every single item.

zooba:
👍

PR #2500, opened 2017-05-04
Adds a prototype WSL extension to enable debugging.

zooba:
Adds a prototype WSL extension to enable debugging.
Enables building VSIX packages as part of release.


zooba:
I have a test build running of this to make sure the build/signing changes work. But other than that, this is sufficient to check in as a prototype

zooba:
Signed build seems to be fine.

Python/Product/Wsl/Debugger/WslLauncher.cs
```diff
@@ -0,0 +1,327 @@
...
+
+            return (
+                "if hash python{0}.{1} 2>/dev/null; then export py=python{0}.{1};" +
+                "elif hash python{0} 2>/dev/null; then export py=python{0}; echo 'WARNING: Using python{0} instead of python{0}.{1}';" +
+                "else echo 'Cannot locate Python {0}.{1}'; read -p 'Press Enter to exit . . .'; exit 1;" +
+                "fi"
+                ).FormatInvariant(version.Major, version.Minor);
+        }
+
+        private string CreateScriptFile(params string[] lines) {
```
int19h: You can tell bash.exe to run arbitrary code directly via `-c`, e.g. `bash.exe -c "echo foo"`.

Python/Product/Wsl/Debugger/WslLauncher.cs
```diff
@@ -0,0 +1,327 @@
...
+                var root = Path.GetPathRoot(path);
+                var newRoot = "/mnt/" + root.ToLowerInvariant().Replace(":", "");
+                path = newRoot + path.Substring(root.Length);
+            }
+
+            return path.Replace('\\', '/');
+        }
+
+        private static string QuoteSingleArgument(string arg) {
+            return "'{0}'".FormatInvariant(arg.Replace("'", "'\\''"));
```
int19h: Bash single quotes [don't allow any escaping whatsoever](https://www.gnu.org/software/bash/manual/html_node/Single-Quotes.html, not even of nested single quotes.

To allow for arbitrary strings here, you probably need to use [this](https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html).

zooba: You mean for arbitrary strings that allow escaping? This function is only used for paths, which shouldn't allow escaping, and single quotes are already handled by the replace string.

Python/Product/Wsl/Debugger/WslLauncher.cs
zooba: The escaping rules got way too complicated for me to figure out. If you can give me the exact rules to convert this script into a single correctly quoted command line, I'll consider it :)

```
if hash python3.5 2>/dev/null; then export py=python3.5;elif hash python3 2>/dev/null; then export py=python3; echo 'WARNING: Using python3 instead of python3.5';else echo 'Cannot locate Python 3.5'; read -p 'Press Enter to exit . . .'; exit 1;fi
export PYTHONPATH='/mnt/c/Program Files (x86)/Microsoft Visual Studio/Preview/Enterprise/Common7/IDE/Extensions/Microsoft/Python/Core'; $py  -m ptvsd -s mwmxoy0zpjc -i 127.0.0.1 -p 50046 --wait '/mnt/c/Users/Steve/Documents/Visual Studio 2017/Projects/PythonApplication1/PythonApplication1/PythonApplication1.py' some arguments
```

Python/Product/Wsl/Debugger/WslLauncher.cs
int19h: Ah, I missed the second pair of single quotes used as the replacement. Never mind,

Python/Product/Wsl/Debugger/WslLauncher.cs
int19h: So far as I can tell, `bash.exe` simply passes the entire command line (sans command itself) to the real Bash as is, so the normal Bash quoting rules are in effect. So I think you can just use `QuoteSingleArgument`?

Python/Product/Wsl/Debugger/WslLauncher.cs
```diff
@@ -0,0 +1,327 @@
...
+                dte.SuppressUI = false;
+            }
+
+            // Another attempt, but display UI.
+            process.Attach();
+            return true;
+        }
+
+
+        private static string FixPath(string path) {
```
int19h: Some further thinking of this. I'm not sure to what extent we can rely on simple substitution to be the right thing - it appears that `/mnt` can be changed, for example (it's just a mount point), and then there are all the weird corner cases with UNC paths and whatnot. I wonder if it's better to just let WSL do the mapping itself? They don't have an API for this, but you can always do `bash.exe -c pwd` with the directory you want mapped as the current directory.

Python/Product/Wsl/Debugger/WslLauncher.cs
zooba: We're going to end up with double quoting then, which is also going to mess with user-provided script arguments. I'm not sure the benefit is actually worth it. Especially since right now I'm just trying to dump the code somewhere safe and get back to bugfixing ;)

Python/Product/Wsl/Debugger/WslLauncher.cs
```diff
@@ -0,0 +1,327 @@
...
+            int port = Enumerable.Range(new Random().Next(49152, 65536), 60000)
+                .Except(IPGlobalProperties.GetIPGlobalProperties().GetActiveTcpConnections().Select(c => c.LocalEndPoint.Port))
+                .First();
+
+            string secret = Path.GetRandomFileName().Replace(".", "");
+
+            var sh = CreateScriptFile(
+                GetPyVariable(config.Interpreter?.Version),
+                string.Join(" ",
+                    "export PYTHONPATH={0};".FormatInvariant(QuoteSingleArgument(FixPath(PtvsdSearchPath))),
```
zooba: Yes, though we still need to provide search paths here. Separating it into its own line in the script entirely will make life much easier for debugging.

Python/Product/Wsl/Debugger/WslLauncher.cs
zooba: Oh, that's a nice trick. Should be more reliable than what's currently there.

I'll add a TODO - none of this code is shipping yet

int19h:
👍

PR #2514, opened 2017-05-08
Fix profiling tests

huguesv:
You should look at the first 2 individual commits.  In the 3rd commit, I've moved a test to a different class and renamed the exising test class, so the full diff view isn't very useful.


huguesv:
Together with https://github.com/Microsoft/PTVS/pull/2515, all profiling tests pass now.

zooba:
👍

PR #2516, opened 2017-05-08
Fix add/remove import tests noticed on test machine

huguesv:
'Remove import' tests failures were easy to repro on my machine, and adding the wait for analyzer makes them succeed consistently.
'Add import' tests were not failing on my machine. On test machine they fail waiting for the lightbulb session, so adding a wait on the analyzer makes some sense.


zooba:
👍

PR #2520, opened 2017-05-09
Fix ipc json unicode tests failures

huguesv:
Don't print the message, as it may contain unicode chars that cannot be printed to the console.

I was testing an unrelated code page issue on my desktop machine at the time I wrote the test, and ended up with a false positive because of it.

The intention with printing the message was that the test could validate the message read. It only checks if the message was read and deserialized successfully. If we ever want to validate more than that, we could print in base64, or write to a file.


zooba:
Alternatively, you could print with `file=sys.stderr`, which will escape unprintable characters, or you could recreate `sys.stdout` with a different error mode at the start.

I'm okay with this change though if we don't desperately need the message.

PR #2521, opened 2017-05-09
Fix cookiecutter integration test

huguesv:
Extend the range of valid count for GitHub results so that we don't have to keep adjusting the value as GitHub returns more or less results over time.


zooba:
👍 

Didn't we have another issue to make these tests not rely on access to GitHub?

huguesv:
I've made another small adjustment, and with that, 20/23 cookiecutter tests pass while offline.  The other 3 are focused on online scenarios (Search, InstallFromOnlineMultipleTimes and CreateFromOnlineTemplate). I would be more worried if they didn't test online, since that's the primary scenario for cookiecutter explorer.

PR #2522, opened 2017-05-09
Fixes #2493 Execute File in Interactive executes wrong file

### LINKED ISSUE #2493: Execute File in Interactive executes wrong file
Load a Python project.
Close all document windows.
Check under Tools->Python Tools that the command caption is "Execute Project in Interactive".
Open one of the project Python files, any will do as long as it's not the startup file.
Check under Tools->Python Tools that the command caption is "Execute File in Interactive".
If you execute the command, it will execute the project startup file, not the one you opened. That's unexpected / misleading.

It will execute the opened file only if you don't have a project loaded.

zooba:
Now ignores the project if the file is not in it (or is not visible), and selects the right project if it's not in the startup project.


huguesv:
👍

PR #2524, opened 2017-05-09
Fix build

huguesv:
I made the mistake of being the first to upgrade to 15.3 ;)

I've updated the ThreadHelper deprecated code according to the guidelines in Andrew's email.

Microsoft.VisualStudio.Threading assembly is now version 15.3.  I don't know how careful we want to be, my changes removes the version check only on Dev15, but we could simplify it by making it ignore the version unconditionally ie. no need for choose element.

I fixed the Dev14 build while I was there.

### DIFF
--- a/Common/Tests/SharedProjectUITests/BasicProjectTests.cs
+++ b/Common/Tests/SharedProjectUITests/BasicProjectTests.cs
@@ -1743,11 +1743,17 @@ namespace Microsoft.VisualStudioTools.SharedProjectTests {
                             uint itemid;
 
                             Console.WriteLine(relativeName);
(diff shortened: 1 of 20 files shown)


huguesv:
I'm getting "Invalid cross-thread call" when running some tests that were working just fine yesterday, so something ain't right.

huguesv:
That invalid cross-thread call was unrelated, I went back prior to these changes and it was still there. No idea why this didn't show up yesterday when I made the change that ended up introducing the issue.  Anyway, I've included the fix in the latest commit.

zooba:
I'd rather use a non-specific version unconditionally (really, I'd rather assembly versions weren't changed for "stable" DLLs, but I guess the name should be promise enough).

huguesv:
Updated references.

zooba:
👍

PR #2526, opened 2017-05-10
Fix DjangoProjectTests

huguesv:
8/9 of DjangoProjectTests now pass

The remaining failure is due to UI automation being broken right now on Dev15 for the app designer tabs. Filed a bug against VS.


zooba:
👍

PR #2532, opened 2017-05-11
Use external web browser

huguesv:
Remaining calls to OpenVsWebBrowser are:
- CommonNonCodeFileNode.DoDefaultAction
- OpenNoInterpretersHelpPage

### DIFF
--- a/Common/Product/SharedProject/CommonPackage.cs
+++ b/Common/Product/SharedProject/CommonPackage.cs
@@ -214,9 +214,11 @@ namespace Microsoft.VisualStudioTools {
         }
 
         internal static void OpenWebBrowser(string url) {
-            var uri = new Uri(url);
-            Process.Start(new ProcessStartInfo(uri.AbsoluteUri));
-            return;
+            try {
+                var uri = new Uri(url);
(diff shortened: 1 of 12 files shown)


Common/Product/SharedProject/CommonPackage.cs
```diff
@@ -214,9 +214,11 @@ public abstract class CommonPackage : Package, IOleComponent {
...
 
         internal static void OpenWebBrowser(string url) {
-            var uri = new Uri(url);
-            Process.Start(new ProcessStartInfo(uri.AbsoluteUri));
-            return;
+            try {
+                var uri = new Uri(url);
+                Process.Start(new ProcessStartInfo(uri.AbsoluteUri));
+            } catch (UriFormatException) {
+            }
```
huguesv: Not ideal. Won't be very common. Some URLs are coming from registry key values, which is where this could happen.

zooba: Can we report the error? Since this should only occur due to corruption on the user's machine, it seems like a reasonable thing to pop up in a dialog box.

huguesv:
Updated code to report any errors.

zooba:
👍

PR #2533, opened 2017-05-11
Some fixes for virtual environment tests

huguesv:
UnavailableEnvironments still fails.  All others can now pass but they are still flaky.


zooba:
👍

PR #2537, opened 2017-05-11
Accessibility fixes for environment window

huguesv:
Make the 2 buttons for enable IPython mode into a real checkbox.

404495 Accessibility : MAS40B : INSPECT : Python+ Python Environments : Control type is incorrect for the checkbox in Python Environments window
403510 Accessibility : MAS05 : Keyboard : Python + Python Environment : Focus order improper in Python Environment Window


zooba:
👍

PR #2538, opened 2017-05-11
Fixes #2365 Mixed mode debugging on 3.6 gives exe_common.inl not found

### LINKED ISSUE #2365: Mixed mode debugging on 3.6 gives exe_common.inl not found
Tried with both CPython 3.6 and Anaconda 3.6, which both have pdb files, and my script is this with a breakpoint on the print line:
```
import sys
print(sys.version)
```

It happens once per breakpoint.  F5 continue execution.  At no point did it show me my Python code where my breakpoint(s) are located.

image

zooba:
Fixes #2365 Mixed mode debugging on 3.6 gives exe_common.inl not found
Fixes build options so that debug helpers are usable.
Fixes source of PythonDeveloper registry value.

Might also be nice to filter all the stack below the `invoke_main` call, since it's really not relevant. I'm not sure how to do that, but maybe it's easy for @int19h ?

### DIFF
--- a/Python/Product/DebuggerHelper/DebuggerHelper.vcxproj
+++ b/Python/Product/DebuggerHelper/DebuggerHelper.vcxproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" TreatAsLocalProperty="Platform">
   <PropertyGroup>
     <Platform>x64</Platform>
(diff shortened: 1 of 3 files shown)


int19h:
Not filtering out frames in the .exe is originally by design - basically, it doesn't touch any frames outside of python*.dll, to avoid mucking around with the host app, since that can be anything.

But we can certainly check for python.exe specifically, and exclude things from it. In fact, I think we can exclude all frames from that module entirely, I don't think there's anything useful there except if you're working on the interpreter itself (and then you should use the Python Developer mode, and we can check for that).

If filtering by module, this is really easy, because we already do such filtering for our own debugger helper, so that would be more of the same:
https://github.com/Microsoft/PTVS/blob/2fc9d456c14d109c670bdfa1d5b93bf84ea49275/Python/Product/Debugger/DkmDebugger/CallStackFilter.cs#L46

zooba:
Yeah, there's nothing useful in [python.exe](https://github.com/python/cpython/blob/master/Programs/python.c) (and note that the `#ifdef MS_WINDOWS` excludes essentially the whole file). Unless someone writes their own executable with the same name...

I filed #2544 for that enhancement - I'm not going to work it into this fix.

PR #2539, opened 2017-05-12
Move diagnostics log functionality to tools/options dialog.

huguesv:
401930 Accessibility : MAS40B : INSPECT : Python + Diagnostic info + Name property for the edit in Diagnostic info window is null

image


zooba:
👍  Yay!

PR #2551, opened 2017-05-14
Use constant instead of magic number

fniephaus:


zooba:
Thanks, @fniephaus! Nice catch

PR #2568, opened 2017-05-17
Polls project templates fixes

huguesv:
Fix django settings modules so they work when project file contains spaces.
Flatten folder structure to match the other project templates and avoid the Folder element in vstemplate, which doesn't work when the project name includes spaces (files under the safeprojectname folder were not being generated).
Update to latest pyjade.
Update dependency on azure to install only the latest azure-storage, and fix the calling code as necessary.


zooba:
👍

PR #2590, opened 2017-05-26
Issue 2586

### LINKED ISSUE #2586: REPL default colours are not accessible
We need to ensure minimum levels of contrast in all our themes

We also need to rename the colour definitions to include "Python", since they are our colours and are not shared with other extensions.

zooba:
Fixes #2586 REPL default colours are not accessible
Changes default colours for themes to have sufficient contrast (4.5:1) - except high contrast theme (need 14:1)
Updates strings to clearly identify colors as being for Python interactive only
Updates high contrast theme to use only white text


huguesv:
👍

PR #2591, opened 2017-05-26
Fixes some high contrast color issues

zooba:
Fixes some high contrast color issues
Synchronises Controls.xaml
Fixes some resource strings


zooba:
Another accessibility fix incoming - I'll add it to this PR instead of switching my current branch and dealing with all that mess

huguesv:
👍

PR #2593, opened 2017-05-26
Cookiecutter accessibility fixes

huguesv:
Bug 404069: Accessibility : MAS40B : INSPECT : Python + Cookiecutter : Inspect properties are incorrect for the tree items in Cookie cutter window

Fix automation peers for the cookiecutter tree view so that there's only one peer per tree view item. This makes the parent-children relationship more clear, and narrator now correctly reports item count.

Make the peers for templates and continuation support the invoke pattern so that hitting 'enter' is accessible through automation.

This is the result:
image


zooba:
👍  Looks great!

PR #2605, opened 2017-05-30
Fixes #2595 Use interactive mode checkbox is invisible on high contrast

### LINKED ISSUE #2595: Use interactive mode checkbox is invisible on high contrast
Using the very latest from master.

Regular VS themes are ok.
OS Light and Dark High contrast are not.
When you move the mouse over the checkbox, it becomes invisible.

image

zooba:
Fixes #2595 Use interactive mode checkbox is invisible on high contrast
Removes checkbox hover color.
Fixes button hover colors.


huguesv:
👍

PR #2607, opened 2017-05-30
Build and test fix

huguesv:
I upgraded a machine from 15.3 build 26507 to 26526, and it required adding a referenced assembly.  It still builds on the older build even with these changes (the 'new' assembly was there before).

Also fixes a REPL test (I forgot which one) which was failing with NRE.


huguesv:
FYI the build was failing because it couldn't find the referenced type `Microsoft.VisualStudio.Shell.Interop.IVsToolboxItemProvider2`.

Issue #2610, opened 2017-05-31
Tests failing to load Microsoft.VisualStudio.Validation assembly


huguesv:
Example:
```
Test Name:	ImportClassifications
Test Outcome:	Failed
Result Message:	
Test method PythonToolsMockTests.ClassifierTests.ImportClassifications threw exception: 
System.IO.FileLoadException: Could not load file or assembly 'Microsoft.VisualStudio.Validation, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```

In 15.3, they've changed the version of that assembly from 15.0.0.0 to 15.3.0.0.

devenv.exe.config has a bindingRedirect for it.

PR #2617, opened 2017-05-31
Add binding redirect for Microsoft.VisualStudio.Validation.

huguesv:
Fixes https://github.com/Microsoft/PTVS/issues/2610

On 15.3, all tests from PythonToolsMockTests were failing trying to load an older version of Microsoft.VisualStudio.Validation. Now a lot of the tests are passing again, and those who don't are failing for a different reason.

Note that we do not directly reference that assembly from our projects.

### DIFF
--- a/Python/Tests/PythonToolsMockTests/App.config
+++ b/Python/Tests/PythonToolsMockTests/App.config
@@ -1,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.VisualStudio.Validation" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>
+        <bindingRedirect oldVersion="15.0.0.0-15.3.0.0" newVersion="15.3.0.0"/>
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
+</configuration>
--- a/Python/Tests/PythonToolsMockTests/PythonToolsMockTests.csproj
+++ b/Python/Tests/PythonToolsMockTests/PythonToolsMockTests.csproj
@@ -177,5 +177,8 @@
       <Name>VSCommon</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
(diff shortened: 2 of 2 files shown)


zooba:
I think we actually just want to update this in our machine setup scripts, so that the changes end up in QTAgent.exe.config and similar.

zooba:
Which is actually now the install script in the VSTestHost repo. Looks like it'll also have to be updated for the Newtonsoft.Json version too.

zooba:
Actually, do you know whether this works reliably for all tests? Maybe we can do this instead of modifying the installed config files.

huguesv:
FYI this is where I got the tip of using app.config: https://github.com/Microsoft/vsts-tasks/issues/546

The app.config that I'm adding here is for PythonToolsMockTests, which don't use VSTestHost.

It would be nice to stop updating VS installed .config files, because when you update VS it overwrites them.

I was able to make this work:
- don't update the .config files
- put a second copy of Microsoft.VisualStudioTools.VSTestHost.15.0.dll in PrivateAssemblies

The script that updates the .config files is adding ";CommonExtensions\Platform;CommonExtensions\Microsoft\Editor" to probing private path. First part is for locating the test host dll, but I'm not sure what the second part is for.  I ran some editor tests that passed fine.

By the way, I found this in my PrivateAssemblies:
Microsoft.VSSDK.TestHostAdapter.dll
Microsoft.VSSDK.TestHostAdapter.pdb

zooba:
The second one is to ensure the VisualStudio.Text DLLs can be loaded for non-UI tests. Possibly that issue has been fixed generally if they're being resolved now (and not just copied into your output directory on build).

The VSSDK test host adapter is forked from the same source code that ours was originally forked from, but unless they've improved it significantly since then it likely still has the same reliability issues that plagued us before (and may not support Debug Test either).

zooba:
@huguesv I'm going to be getting some new test machines soon - want to look at making this work for all test assemblies? Or shall I take it over?

huguesv:
Feel free to take over if you like.

huguesv:
Can we merge this so that I can at least run the PythonToolsMockTests?

I need to add tests for https://github.com/Microsoft/PTVS/pull/2860

I don't mind if you improve it later on and end up undoing this, I just want something that works in the meantime.

zooba:
Sure.

PR #2623, opened 2017-05-31
Adds workaround for building against stable VS.

zooba:
This will keep us building with the current version of VS while we figure out how to deal with the brokenness in the next update.

We'll just have to ignore the failed tests for now until we can get all the build machines upgraded properly (which they probably won't do for us, so we're back to maintaining our own build infrastructure...)

### DIFF
--- a/Common/Tests/Utilities/Mocks/TextVersionWorkaround.cs
+++ b/Common/Tests/Utilities/Mocks/TextVersionWorkaround.cs
@@ -1,0 +1,24 @@
+﻿// Visual Studio Shared Project
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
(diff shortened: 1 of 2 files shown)


zooba:
I've already updated the build definitions to use a working build pool, so feel free to merge this whenever and we'll get tests going again.

huguesv:
What do you mean by working build pool?  Is it machines that we manage ourselves that are currently on 15.2 and that we'll update to 15.3 shortly?

zooba:
No, we switched to the one that is managed by the build team a while back, but they won't update to 15.3 until after it releases. I upgraded our own one and switched back to that briefly, but it won't build for a variety of other reasons and I don't want to spend the time debugging that when we could (and should) just be getting the editor team to fix their API breakage.

PR #2626, opened 2017-06-01
Debugger test run improvement

huguesv:
When an exception occurred in ProcessLoaded event handler (on background thread), it wasn't caught and that was causing the test agent to terminate. Not a big deal when using Test Explorer, as it only affects that one test.  On the test machine, that is causing a whole bunch of tests to not be executed.


zooba:
👍

Issue #2628, opened 2017-06-01
DebuggerTests that log warnings to event viewer


huguesv:
Lots of ObjectDisposedException events are logged while running DebuggerTests, and I've narrowed it down to the following tests. All others run 'clean'.

LocalReprRestrictionsTest
StepToEntryPoint
TestBreakAll
TestBreakAllThreads
TestBreakpointHitOtherThreadStackTrace

zooba:
Are these unhandled exceptions? We log them immediately during tests to help find where they are raised and not handled.

Shouldn't be an issue to have too many, provided they're all being raised during the test and not after we think we've finished.

huguesv:
They are handled.

What these all have in common is that they call `TerminateProcess` (which calls `Process.Kill`), because that's the only way to stop end the script being debugged (infinite loop).

It's normal for `TerminateProcess` to cause `ObjectDisposedException`, we're reading from a stream that was closed when the process was killed.

I just found the logging code in AssertListener.cs:
```
        static void CurrentDomain_FirstChanceException(object sender, FirstChanceExceptionEventArgs e) {
            if (e.Exception is NullReferenceException || e.Exception is ObjectDisposedException) {
                // Exclude safe handle messages because they are noisy
                if (!e.Exception.Message.Contains("Safe handle has been closed")) {
                    var log = new EventLog("Application");
                    log.Source = "Application Error";
                    log.WriteEntry(
                        "First-chance exception: " + e.Exception.ToString(),
                        EventLogEntryType.Warning
                    );
                }
            }
        }
```

I don't like the noise that this is generating in this scenario, it can cause us to miss events that matter. Looking more closely at all the events, I just found an event for a NullReferenceException amongst all the ObjectDisposedException ones.

14 interpreters * 5 tests causing an average of 6 events each = 420 events in event viewer.

What if I was to temporarily suspend then restore logging of ObjectDisposedException (only) in BaseDebuggerTests.TerminateProcess?

zooba:
Sure, just make sure that it's definitely going to be restored. Missing log entries are much much worse than too many log entries.

PR #2631, opened 2017-06-02
Temporarily disable event logging of ObjectDisposedException

huguesv:
Fix for #2628 DebuggerTests that log warnings to event viewer.


zooba:
Feels overengineered for this bit (which is mostly a hack anyway, so a simple `bool` would have sufficed in my opinion). If you feel okay with this though then go ahead and merge, but I really don't want this diagnostic to be too flexible - it's really meant to be as hardcoded and foolproof as possible.

huguesv:
Sure, I can simplify it.

Issue #2579, opened 2017-05-22
Paste remove REPL prompt doesn't work for IPython style prompts


huguesv:
Copy-pasting from the Pandas documentation into the editor is painful because the prompts are not getting removed.

I'm trying to copy from code snippets that look like this:
```
In [8]: df = pd.DataFrame({'A': [1, 2, 3],
   ...:                    'B': [1., 2., 3.],
   ...:                    'C': ['foo', 'bar', 'baz'],
   ...:                    'D': pd.date_range('20130101', periods=3)})
   ...: 

In [9]: df.dtypes
Out[9]: 
A             int64
B           float64
C            object
D    datetime64[ns]
dtype: object
```

http://pandas.pydata.org/pandas-docs/stable/whatsnew.html

zooba:
Fixed in the above PR

PR #2637, opened 2017-06-02
Fixes #2398 Paste adds extra whitespaces at the end

### LINKED ISSUE #2398: Paste adds extra whitespaces at the end
In the attached image, notice how the cursor is a few characters away right after paste.
capture

zooba:
Fixes #2398 Paste adds extra whitespaces at the end
Replaces prompt removal with a two-step process so that Paste works properly.
Expands prompts to include IPython-style prompts


huguesv:
👍 thanks for the IPython prompts :) https://github.com/Microsoft/PTVS/issues/2579

PR #2638, opened 2017-06-02
Adds a user agent string for requests to pypi.org.

zooba:
Looks like we also need to reduce the number of calls overall, but I'll do that as a separate issue.


huguesv:
👍

PR #2649, opened 2017-06-06
Improve debugger tests reliability.

huguesv:
I ran all the debugger tests twice and did not see a single failure (compared to 3-5% failure before).

Changes the callback passed to CreateProcess to be `Func<Task>` instead of `Action`, and ensure that it is awaited on. This ensures proper order of execution during process load event handling.

The callers that were already passing an async delegate didn't have to change, just the ones that were passing non-async delegate.


zooba:
👍

PR #2660, opened 2017-06-09
Fixes extra tab stop on list item around link to help page.

zooba:


Python/Product/EnvironmentsList/ToolWindow.xaml.cs
```diff
@@ -556,6 +557,25 @@ await StartRefreshDBAsync(view)
             }
         }
 
+        private static void SelectListBoxItem(DependencyObject child, bool isSelected) {
+            while (child != null) {
+                var lbi = child as ListBoxItem;
+                if (lbi != null) {
+                    lbi.IsSelected = true;
```
zooba: Will fix this - should be `isSelected` (or perhaps removed completely... since it works despite this...)

zooba:
Approved by @huguesv

PR #2676, opened 2017-06-15
Enables live region support using internal VS API.

zooba:
Enables live region support using internal VS API. This should be replaced with proper WPF support when that is added.

Simplifies status display in cookiecutter.
Fixes some missing help text properties.


Python/Product/EnvironmentsList/DBExtension.xaml
```diff
@@ -125,10 +74,47 @@
             <CollectionViewSource x:Key="SortedPackages" Source="{Binding Packages}" />
         </Grid.Resources>
 
-        <ContentControl Content="{Binding}"
-                        ContentTemplate="{StaticResource ProgressView}"
-                        Focusable="False"/>
-        
+        <Grid DataContext="{Binding EnvironmentView}">
```
zooba: This change may be unnecessary now that I've improved the SetLiveProperty() call - previously you couldn't use it in a DataTemplate because the control was not correctly parented on creation, but that shouldn't matter now.

In any case, I kind of like the new code better, at least now it's written.

huguesv:
👍

PR #2677, opened 2017-06-15
Debug interactive window protocol changes

huguesv:
Makes the DebugReplEvaluatorTests* pass more reliably, and add some more coverage.
Change the protocol to be the debugger's, which is JSON, instead of the REPL's, which is binary.

Note that IntelliSense is broken in debug repl prior to these changes (the 2 IntelliSense related methods that are marked TODO are never called).

Need to decide how to increment the ptvsd version (this is breaking change, so 4.0?)

I've added an intermediate base class `PythonCommonInteractiveEvaluator` that the regular repl and debug repl now derive from, so that debug repl doesn't inherit any of the binary repl protocol specific code. A lot of that code is the same, but it's now split between the new base class and the original `PythonInteractiveEvaluator`.

### DIFF
--- a/Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
+++ b/Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
@@ -1436,9 +1436,13 @@ namespace Microsoft.PythonTools.Debugger.DebugEngine {
         }
 
         private void OnDebuggerOutput(object sender, OutputEventArgs e) {
-            AD7Thread thread;
-            if (!_threads.TryGetValue(e.Thread, out thread)) {
-                _threads[e.Thread] = thread = new AD7Thread(this, e.Thread);
+            // Output from debug REPL code execution may be run on debugger
+            // event handling thread, and e.Thread will be null.
+            AD7Thread thread = null;
+            if (e.Thread != null) {
+                if (!_threads.TryGetValue(e.Thread, out thread)) {
+                    _threads[e.Thread] = thread = new AD7Thread(this, e.Thread);
+                }
             }
 
             Send(new AD7DebugOutputStringEvent2(e.Output), AD7DebugOutputStringEvent2.IID, thread);
--- a/Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
+++ b/Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
@@ -16,27 +16,18 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
-using System.Net;
-using System.Net.Sockets;
-using System.Runtime.ExceptionServices;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PythonTools.Debugger;
 using Microsoft.PythonTools.Debugger.DebugEngine;
-using Microsoft.PythonTools.Debugger.Remote;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.Intellisense;
-using Microsoft.PythonTools.Ipc.Json;
 using Microsoft.PythonTools.Options;
-using Microsoft.PythonTools.Parsing;
 using Microsoft.VisualStudio.InteractiveWindow;
 using Microsoft.VisualStudio.InteractiveWindow.Commands;
 using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Utilities;
 using Microsoft.VisualStudioTools;
-using LDP = Microsoft.PythonTools.Debugger.LegacyDebuggerProtocol;
 
 namespace Microsoft.PythonTools.Repl {
     [InteractiveWindowRole("Debug")]
@@ -50,6 +41,7 @@ namespace Microsoft.PythonTools.Repl {
     {
         private PythonDebugProcessReplEvaluator _activeEvaluator;
         private readonly Dictionary<int, PythonDebugProcessReplEvaluator> _evaluators = new Dictionary<int, PythonDebugProcessReplEvaluator>(); // process id to evaluator
+        private readonly Dictionary<int, Task> _attachingTasks = new Dictionary<int, Task>();
         private EnvDTE.DebuggerEvents _debuggerEvents;
         private readonly PythonToolsService _pyService;
         private readonly IServiceProvider _serviceProvider;
@@ -156,7 +148,6 @@ namespace Microsoft.PythonTools.Repl {
             if (t != null) {
                 return await t;
             }
-            // No evaluator, so say we succeeded
             return true;
         }
 
@@ -180,9 +171,7 @@ namespace Microsoft.PythonTools.Repl {
         }
 
         public IEnumerable<string> GetAvailableScopes() {
-            // TODO: Localization: we may need to do something with <CurrentFrame> string. Is it displayed?
-            // It also appears visualstudio_py_repl.py so it probably needs to be in sync with it.
-            string[] fixedScopes = new string[] { "<CurrentFrame>" };
+            string[] fixedScopes = new string[] { Strings.DebugReplCurrentFrameScope };
             if (_activeEvaluator != null) {
                 return fixedScopes.Concat(_activeEvaluator.GetAvailableScopes());
             } else {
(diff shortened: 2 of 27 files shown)


int19h:
So far we have versioned ptvsd according to PTVS version numbers, and every release was assumed to be incompatible with every other. Until we fully switch to JSON everywhere, I think we should stick to that scheme. Once we do switch, we can start versioning the package separately and use semver.

zooba:
Agreed. Let's get to our "1.0" (which will be 4.0) before switching versioning scheme.

But check the behaviour if you have the newer `ptvsd` in the remote process with the current VS, and the newer VS with the older `ptvsd`. At the very least, we should add a known issue to the release notes, and for that we'll have to describe the behaviour.

Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -1436,9 +1436,13 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
...
 
         private void OnDebuggerOutput(object sender, OutputEventArgs e) {
-            AD7Thread thread;
-            if (!_threads.TryGetValue(e.Thread, out thread)) {
-                _threads[e.Thread] = thread = new AD7Thread(this, e.Thread);
+            // Output from debug REPL code execution may be run on debugger
+            // event handling thread, and e.Thread will be null.
+            AD7Thread thread = null;
+            if (e.Thread != null) {
+                if (!_threads.TryGetValue(e.Thread, out thread)) {
```
zooba: Do we need to lock `_threads` here?

Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
```diff
@@ -421,15 +410,38 @@ internal class PythonDebugReplEvaluator :
...
 
+            // Multiple callers can execute in this method simultaneously (due to async/await)
+            // so we need to protect against 2 callers trying to create evaulators for the same process.
+
+            // Keep track of evaluators that are in progress of attaching, and if
+            // we are getting called to attach for one that is already in progress,
+            // just wait for it to finish before returning.
+            Task attachingTask;
+            TaskCompletionSource<object> attachingTcs = null;
+            if (_attachingTasks.TryGetValue(process.Id, out attachingTask)) {
```
zooba: Need a lock here to avoid races (two threads check, then both add the task).

Concurrent dict doesn't solve this - it doesn't have a function for "get it if it's there, create it if not, **and tell me whether you just created it**".

huguesv: I'll try to clarify the comment above. Locking isn't needed because the simultaneous callers are both on the UI thread. The second UI thread caller won't enter this function until the first caller uses await down below and yields execution. So as long as we check the dictionary before we call await, it's ok because they aren't really running in parallel.

Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
huguesv: It's not protected anywhere else, but that doesn't mean that's correct ;) That could be the cause of the multithreading debugging issues we have trouble reproducing.

huguesv:
In release builds, behavior is the same whether version is incremented or not: after clicking attach, debugging terminates without any error message. I'll look into that.

Python/Product/PythonTools/visualstudio_py_debugger.py
```diff
@@ -94,6 +94,25 @@
     from System.Runtime.CompilerServices import ConditionalWeakTable
     IPY_SEEN_MODULES = ConditionalWeakTable[object, object]()
 
+    clr.AddReference('Microsoft.Dynamic')
```
huguesv: I copied all of these from the repl implementation, and I didn't try to minimize them to see which ones were really needed.  Without any of it, executing code in modules was failing.

Python/Tests/Core/DebugReplEvaluatorTests.cs
```diff
@@ -63,6 +65,7 @@ public class DebugReplEvaluatorTests {
             _window = new MockReplWindow(_evaluator);
             _evaluator._Initialize(_window);
             _processes = new List<PythonProcess>();
+            _moduleFileExtension = Version.Version < Microsoft.PythonTools.Parsing.PythonLanguageVersion.V30 && Version.IsCPython ? ".pyc" : ".py";
```
zooba: As this is class-specific state, it would be better as a property than an instance variable.

At the very least, the instance variable should be readonly. Arguably we should also fix the module file extension to be corrected to `.py` wherever possible as well, which would presumably also make this redundant.

zooba: Another alternative - when comparing paths use `Path.ChangeExtension(path, null)` on both to remove the extension first.

huguesv:
Regarding the versioning, that was my mistake, I forgot to increment the protocol version. That's what is being compared. This is what you see if you have the wrong version:
```
---------------------------
---------------------------
The remote server at tcp://my_secret@127.0.0.1:5713/ is not running a remote debugging server, or may be running an unsupported version.
---------------------------
Retry   Cancel   
---------------------------
```

So I'll increment the protocol version to 8 and version to 3.2.0.0

zooba:
Unless @int19h wants to look more closely, I don't have any more comments.

Python/Tests/Core/DebugReplEvaluatorTests.cs
huguesv: After looking at the code that makes use of the module file (called from command handler for `PkgCmdIDList.cmdidOpenInteractiveScopeInEditor`), I think I'll go with the ChangeExtension(null).
```
        private static readonly IEnumerable<string> SourceExtensions = new[] {
            PythonConstants.FileExtension,
            PythonConstants.WindowsFileExtension,
        };

        private string GetCurrentScopeSourcePath() {
            var path = (_interactive.Evaluator as IMultipleScopeEvaluator)?.CurrentScopePath;
            if (string.IsNullOrEmpty(path)) {
                return null;
            }

            foreach (var ext in SourceExtensions) {
                var source = Path.ChangeExtension(path, ext);
                if (File.Exists(source)) {
                    return source;
                }
            }

            return null;
        }
```

zooba: Yeah, that should be fixing the extension already... guess we have cases where it can't.

huguesv:
About 50% of the time, I get this in the output window when my python app exits. I'm pretty sure that's a regression.
```
The thread 'MainThread' (0x3ac8) has exited with code 0 (0x0).
IOError: [Errno 9] Bad file descriptor
Error in sys.excepthook:
The program '[11568] pythonw.exe' has exited with code 1 (0x1).
```
or this:
```
The thread 'MainThread' (0xac4) has exited with code 0 (0x0).
The thread 0x349c has exited with code 0 (0x0).
The thread 0x156c has exited with code 0 (0x0).
IOError: [Errno 9] Bad file descriptor
Error in sys.excepthook:
  File "C:\USERS\HUVALO\APPDATA\LOCAL\MICROSOFT\VISUALSTUDIO\15.0_909856D1EXP\EXTENSIONS\MICROSOFT CORPORATION\PYTHON\3.0\visualstudio_py_debugger.py", line 2370, in _excepthook
    do_wait()
  File "C:\USERS\HUVALO\APPDATA\LOCAL\MICROSOFT\VISUALSTUDIO\15.0_909856D1EXP\EXTENSIONS\MICROSOFT CORPORATION\PYTHON\3.0\visualstudio_py_debugger.py", line 2503, in do_wait
The program '[14276] pythonw.exe' has exited with code 1 (0x1).
```

zooba:
@huguesv We probably just want to handle it in that case.

huguesv:
The whole "Press any key to continue" feature (which I have ON for both expected and unexpected exit) is no longer working (even when I don't see that error, it doesn't prompt). It's broken even when the Debug REPL is never instantiated.  There's nothing obvious in my changes to debugger.py that would cause this.

zooba:
What if an exception is occurring inside the Python code in one of the bare try/except blocks?

huguesv:
False alarm. I forgot that I had checked "Windows Application" in project settings a few days ago for testing and never put it back to console. I was comparing with the pre-PR code on my other machine whose project was console application.

zooba:
We've already locked in the build for our next Preview release, so feel free to check this in whenever you're happy. Anything we need to fix urgently can use a branch from before this change, so we have time to test things ourselves before inflicting it on real users.

huguesv:
Okay I'll merge to avoid delaying @int19h 's work any further. There are no known regression from my manual testing, but I'll look at the REPL test failures to make sure.

PR #2697, opened 2017-06-20
REPL fixes

### LINKED ISSUE #2688: CSV paste in REPL functionality missing
The test `ReplWindowTests.CsvPaste` checks that if the clipboard contains data in regular text and `DataFormats.CommaSeparatedValue` formats, that the CSV format is used when `Edit.Paste` command is issued.

The test fails because the text format is pasted.

I looked at the interactive window implementation:
https://github.com/dotnet/interactive-window/blob/3114957d329e725982d19dc04db6bfb6152f88b0/src/Microsoft.VisualStudio.InteractiveWindow/InteractiveWindow.UIThreadOnly.cs#L2660

There is no check for CSV format.

We probably should just delete this test.

huguesv:
Fix #2688 CSV paste in REPL functionality missing 
Also fixes 2 NullReferenceException found by tests.

More fixes for REPL tests will come in a later PR. The changes included here are the only fixes in product code I had accumulated so far.


zooba:
👍

PR #2699, opened 2017-06-20
Fixes #2301 Cannot customise interactive scripts path

### LINKED ISSUE #2301: Cannot customize interactive scripts path
We have a field in Tools, Options, Python [Tools!], Interactive Options to let you change the scripts directory for interactive scripts, but it appears to be ignored.

zooba:
Fixes #2301 Cannot customise interactive scripts path
Enables customization of the path
Also prevents calculating the path too eagerly
Prevents calling pip list too eagerly


zooba:
Anyone can merge after reviewing this if there are no questions

PR #2700, opened 2017-06-20
Fix localization issues.

zooba:
Fix localization issues.
Makes examples localizable
Enables auto-ellipsis on text that may be truncated in check boxes.

These were all reported on the internal tracker.


zooba:
Anyone can merge after reviewing this

PR #2702, opened 2017-06-21
Fix 2531

### LINKED ISSUE #2531: help() output is not displayed in interactive window
If I type `help(print)` in a Python 3 interactive window, the help text is not displayed.

However, if I then immediately reset the window, the text is displayed as an error message.

Testing with `py -3.6 -c "help(print)" >nul`, it's clearly sending the text to stdout and not stderr, so we are probably misdirecting it.

huguesv:
Fix #2531 help() output is not displayed in interactive window


zooba:
👍

PR #2715, opened 2017-06-22
Fixes #2389 Cannot go to definition of variables

### LINKED ISSUE #2389: Cannot go to definition of variables
If you define a top-level variable, you should be able to use `Ctrl+,` to find and navigate to it.

Currently, only lists/dicts/etc. appear in the list, and selecting them does not navigate to the definition.

zooba:
Fixes #2389 Cannot go to definition of variables
Ensures all definition locations are made available for variables.
Improves MemberResult implementation using Lazy.
Fixes MemberResult not ignoring Unknown type information properly.


huguesv:
👍

PR #2717, opened 2017-06-22
Fixes #1694 Intellisense packages are not sorted

### LINKED ISSUE #1694: Intellisense packages are not sorted
In addition to #1585, we don't sort the packages in this list. They should be sorted alphabetically.

zooba:
Fixes #1694 Intellisense packages are not sorted
Displays full package name in list so it is obvious that they are sorted.


huguesv:
👍

PR #2723, opened 2017-06-22
Fixes #2648 REPL code analysis does not include variables

### LINKED ISSUE #2648: REPL code analysis does not include variables
(From https://developercommunity.visualstudio.com/content/problem/65321/python-interactive-window-does-not-remember-variab.html)

VERY frequently, the autocomplete for the interactive IPython window in PTVS 2017 will stop remembering the names of my variables. It then autocorrects me when I try to type my variable name into a default function or method, which isn't what I want for quick, iterative prototyping. 

I do not know what triggers it, but it happens on most of my workflows (which include Pandas + Numpy). Creating a dataframe called "df" for example, will almost always cause the autocomplete to fail within a few code block executions.

zooba:
Fixes #2648 REPL code analysis does not include variables
Fixes handling of empty expression completion for IPython mode
Fixes error in heartbeat channel
Prevents invalid names being shown in completion list (since they do not filter properly)
Fixes debug mode failing due to incorrect use of ShellExecute.


huguesv:
👍

PR #2726, opened 2017-06-23
Fixes #2085 Cannot see imports from search paths

### LINKED ISSUE #2085: Cannot see imports from search paths
After adding a search path, completions for `import` do not include available names from that path.

Once you finish typing the name, the import is resolved. We should eagerly figure out names that may be imported and offer them as completions.

zooba:
Fixes #2085 Cannot see imports from search paths
Adds background task to enumerate packages/modules in search path directories (ZIP files not yet supported)
Makes unimported modules lazier
Fixes NullReferenceException in Navigate To.

### DIFF
--- a/Python/Product/Analysis/Interpreter/Default/CPythonInterpreter.cs
+++ b/Python/Product/Analysis/Interpreter/Default/CPythonInterpreter.cs
@@ -121,7 +121,28 @@ namespace Microsoft.PythonTools.Interpreter.Default {
             }
         }
 
-        private async Task UpdateSearchPathPackages(CancellationToken cancellationToken) {
+        private async void BeginUpdateSearchPathPackages() {
+            var cts = new CancellationTokenSource();
+            var oldCts = Interlocked.Exchange(ref _searchPathPackagesCancellation, cts);
+            try {
+                oldCts?.Cancel();
+                oldCts?.Dispose();
+            } catch (ObjectDisposedException) {
+            }
+
+            try {
+                await Task.Run(() => UpdateSearchPathPackagesAsync(cts.Token));
+            } catch (OperationCanceledException) {
+            } catch (ObjectDisposedException) {
+            } catch (Exception ex) {
+                // Cannot do anything more useful with the exception message here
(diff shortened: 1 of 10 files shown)


zooba:
Making an update for better async (I hope)

zooba:
Yep, it's better :)

Python/Product/Analysis/Interpreter/Default/CPythonInterpreter.cs
```diff
@@ -247,6 +307,9 @@ class GetModuleCallable {
             _searchPaths = _state.GetSearchPaths();
             _searchPathDb = null;
             _zipPackageCache = null;
+
+            BeginUpdateSearchPathPackages();
+
             ModuleNamesChanged?.Invoke(this, EventArgs.Empty);
```
huguesv: Do you still want to fire this event here, it will be done at the end of background task too.

Python/Product/Analysis/Values/SyntheticDefinitionInfo.cs
```diff
@@ -19,14 +19,18 @@
...
     class SyntheticDefinitionInfo : AnalysisValue {
-        public SyntheticDefinitionInfo(IEnumerable<LocationInfo> locations) {
+        public SyntheticDefinitionInfo(
+            string name,
+            string doc,
+            IEnumerable<LocationInfo> locations
+        ) {
+            Name = name;
+            Documentation = doc;
             Locations = locations.ToArray();
```
huguesv: Should throw `ArgumentNullException` for `locations`

Python/Product/Analysis/Interpreter/Default/CPythonInterpreter.cs
zooba: No, you're right.

Python/Product/Analysis/Values/SyntheticDefinitionInfo.cs
zooba: I'll handle null instead.

huguesv:
👍

PR #2733, opened 2017-06-26
Fixes #2706 $appcmdargs miscalculated in default web role template

### LINKED ISSUE #2706: $appcmdargs miscalculated in default web role template
In ConfigureCloudService.ps1, we have this snippet of code:

```
    $appcmdargs += @(
        "set",
        "config",
        ...
    )
```

At this point, $appcmdargs may be `@()` or one+ arguments extracted from an environment variable. In the case where it is exactly one argument, the `+` causes that argument and `"set"` to be concatenated, rather than kept as separate elements.

Correct code is (probably):

```
    $appcmdargs = @(
        $appcmdargs,
        "set",
        "config",
        ...
    )
```

zooba:
Fixes #2706 $appcmdargs miscalculated in default web role template
Uses correct syntax to concatenate lists in Powershell.
Avoids joining absolute paths together.


huguesv:
👍

PR #2747, opened 2017-06-29
Always load Python debugger code as ptvsd package in all scenarios

int19h:
Always load Python debugger code as ptvsd package in all scenarios:
- local launch
- local attach
- remote attach

Fixes of various unrelated product and test issues that came up while testing the change.

Diagnostic improvements for attach tests - if the test times out, and python.exe is still alive, it'll automatically create a stack minidump.

Note that this doesn't significantly change the internal structure of the package, or attempt to merge the modules yet.


Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -1098,7 +1073,19 @@ bool DoAttach(HMODULE module, ConnectionInfo& connInfo, bool isDebug) {
...
         auto debugOptions = PyObjectHolder(isDebug, strFromString(connInfo.Buffer->DebugOptions));
         DecRef(call(attach_process.ToPython(), pyPortNum.ToPython(), debugId.ToPython(), debugOptions.ToPython(), pyTrue, pyFalse, NULL), isDebug);
+        if (auto err = errOccurred()) {
+            PyObject *type, *value, *traceback;
+            pyErrFetch(&type, &value, &traceback);
+
+            auto repr = PyObjectHolder(isDebug, pyObjectRepr(value));
+            wchar_t reprText[0x1000] = {};
+            pyUnicodeAsWideChar(repr.ToPython(), reprText, _countof(reprText) - 1);
+            fputws(reprText, stderr);
```
int19h: Future plans are to propagate the entire text of this error to the caller - this will require switching to something like JSON to communicate attach results, instead of the current system with simple error codes.

Python/Product/PythonTools/ptvsd/__init__.py
```diff
@@ -19,4 +19,94 @@
...
+    optional timeout occurs.
+
+    Parameters
+    ----------
+    timeout : float, optional
+        The timeout for the operation in seconds (or fractions thereof).
+    """
+    return _attach_server().wait_for_attach(timeout)
+
+def break_into_debugger():
```
int19h: The following two functions now work for regular debugging, not just remote.

Python/Product/PythonTools/ptvsd/attach_server.py
```diff
@@ -156,12 +100,12 @@ def enable_attach(secret, address = ('0.0.0.0', DEFAULT_PORT), certfile = None,
...
     global _attach_enabled
     if _attach_enabled:
-        raise AttachAlreadyEnabledError('ptvsd.enable_attach() has already been called in this process.')
+        raise ptvsd.AttachAlreadyEnabledError('ptvsd.enable_attach() has already been called in this process.')
     _attach_enabled = True
 
     atexit.register(vspd.detach_process_and_notify_debugger)
 
-    server = socket.socket(proto=socket.IPPROTO_TCP)
+    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM, socket.IPPROTO_TCP)
```
int19h: Some older Python versions don't support keyword arguments here.

Python/Product/PythonTools/ptvsd/debugger.py
```diff
@@ -880,8 +861,9 @@ def _stackless_schedule_cb(self, prev, next):
             current.trace_function = current_tf
 
     def trace_func(self, frame, event, arg):
-        # If we're so far into process shutdown that sys is already gone, just stop tracing.
-        if sys is None:
+        # If we're so far into process shutdown that modules are being unloaded, stop tracing
+        # since we can't rely on any modules we've imported to still be working.
+        if sys is None or not sys.modules:
```
int19h: In 3.6, the shutdown process is subtly different, and `sys is None` doesn't cut it anymore. This fixes issues where you see a bunch of gunk about unhandled exceptions being ignored when deleting objects referenced from module's globals on shutdown (said exceptions coming from our tracefunc).

Python/Product/PythonTools/ptvsd/debugger.py
```diff
@@ -2318,15 +2300,18 @@ def start_debugger_loop(sock):
...
     for i in xrange(50):
+        failure_cause = None
         try:
             sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             sock.connect(('127.0.0.1', port_num))
             break
-        except:
+        except Exception:
+            import traceback
+            failure_cause = traceback.format_exc()
```
int19h: The trace is captured as string here to make it easy to extract it from PyDebugAttach.

Python/Tests/DebuggerTests/AttachTests.cs
```diff
@@ -667,36 +676,40 @@ void main()
...
+                        }
+                    });
 
-                        attachStarted.Set();
-                        await proc.StartListeningAsync(10000);
-                    } finally {
-                        await DetachProcessAsync(proc);
-                    }
-                });
+                    await Task.WhenAny(attachTask, attachStarted.Task).Unwrap().WithTimeout(10000, "Failed to start attaching within 10s");
```
int19h: This is the only place in this file where the switch from events to tasks makes a big difference - waiting on `attachTask` here propagates any exceptions that it might throw onto the test thread, so that they're captured in test results. I changed other tests for consistency, and as a safeguard to avoid mixing async and blocking sync code in general.

Python/Tests/DebuggerTests/AttachTests.cs
```diff
@@ -943,7 +968,7 @@ class TraceRedirector : Redirector {
 
             using (var p = ProcessOutput.Run(
                 Path.Combine(vc.BinPath, "cl.exe"),
-                new[] { "/MD", "test.cpp" },
+                new[] { "/Zi", "/MD", "test.cpp" },
```
int19h: Enable debug information - for attaching, and better minidumps.

Python/Product/Debugger/Debugger/DebugConnectionListener.cs
```diff
@@ -82,80 +82,83 @@ static class DebugConnectionListener {
...
-                            Debug.WriteLine("Unknown debug target: {0}", debugId);
                         }
+                    } finally {
+                        debugConn?.Dispose();
                     }
+                } catch (IOException) {
+                } catch (SocketException) {
+                } catch (TimeoutException) {
+                } catch (Exception ex) {
+                    Environment.FailFast("DebugConnectionListener.AcceptConnection", ex);
```
zooba: Is this better than letting the exception fall out?

Python/Product/PythonTools/ptvsd/__init__.py
zooba: 👍

Python/Product/PythonTools/ptvsd/debugger.py
```diff
@@ -43,41 +43,16 @@
...
-try:
-    # In the local attach scenario, visualstudio_py_repl is injected into globals()
-    # by PyDebugAttach before loading this module, and cannot be imported.
-    _vspr = visualstudio_py_repl
-except:
-    try:
-        import visualstudio_py_repl as _vspr
-    except ImportError:
-        import ptvsd.visualstudio_py_repl as _vspr
+import ptvsd.repl as _vspr
```
zooba: Any reason not to group this with the earlier imports?

Python/Product/PythonTools/ptvsd/debugger.py
```diff
@@ -122,6 +97,12 @@
 # save start_new_thread so we can call it later, we'll intercept others calls to it.
 
 debugger_dll_handle = None
+
+def set_debugger_dll_handle(handle):
```
zooba: Worth a comment that this will be called from C code, otherwise we'll regularly be trying to figure out why we have a function for this.

Python/Product/PythonTools/ptvsd/debugger.py
```diff
@@ -2318,15 +2300,18 @@ def start_debugger_loop(sock):
...
             break
-        except:
+        except Exception:
+            import traceback
+            failure_cause = traceback.format_exc()
             import time
             time.sleep(50./1000)
     else:
-        raise Exception('failed to attach')
+        raise Exception('Attach failed:\n\n' + failure_cause)
```
zooba: Initialize `failure_cause` to `""` instead of `None`?

Python/Product/PythonTools/ptvsd_loader.py
```diff
@@ -0,0 +1,40 @@
...
+    from ptvsd.debugger import attach_process, new_thread, new_external_thread, set_debugger_dll_handle
+finally:
+    sys.path.remove(ptvs_lib_path)
+
+# If threading was not loaded before ptvsd imports, it must still be unloaded afterwards.
+# It assumes that when it's loaded, that happens on the main thread, and there are no other
+# threads in the process - and saves the current thread ID as that of the main thread. In
+# local attach, however, this code runs on the injected debugger thread; and if threading
+# is loaded on that thread, it will misidentify it as a main thread, breaking things later.
+assert(was_threading_loaded or 'threading' not in sys.modules)
```
zooba: No parens on assert

Python/Tests/DebuggerTests/MiniDumpWriter.cs
```diff
@@ -0,0 +1,95 @@
...
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PythonTools.Debugger;
+using TestUtilities;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using Microsoft.Win32.SafeHandles;
```
zooba: Sort usings

Python/Tests/DebuggerTests/TaskExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public static Task CancelAfter(int milliseconds, string message = null) =>
+            CancelAfter<object>(milliseconds, message);
+
+        public static Task WithTimeout(this Task task, int milliseconds, string message = null) =>
+            Task.WhenAny(task, CancelAfter(milliseconds, message)).Unwrap();
+
+        public static Task<T> WithTimeout<T>(this Task<T> task, int milliseconds, string message = null) =>
+            Task.WhenAny(task, CancelAfter<T>(milliseconds, message)).Unwrap();
+    }
```
zooba: Can the CancellationTokens class help here? At the very least, turning timeouts to infinite when debugging is helpful.

Python/Tests/IronPython/ReplEvaluatorTests.cs
```diff
@@ -170,7 +170,7 @@ public class IronPythonReplEvaluatorTests {
 
                 await replEval.ExecuteText("print '[%s]' % sys.gettrace()");
                 AssertUtil.AreEqual(
-                    new Regex(@"\[\<bound method Thread.trace_func of \<Thread.+\>\>\]"),
+                    new Regex(@"\[None\]"),
```
zooba: Doesn't have to be a regex anymore

Python/Product/Debugger/Debugger/DebugConnectionListener.cs
int19h: If I remember correctly, as written originally, it wasn't bringing down the process for me, so it was hard to figure out what went wrong.

Python/Product/PythonTools/ptvsd/debugger.py
int19h: None that I can think of. I'll change this.

Python/Tests/DebuggerTests/TaskExtensions.cs
int19h: Cancellation tokens are for cooperative cancellation when the functions themselves support it - something has to check the token on the other side. Ideally this should be the case everywhere in the product, but for tests, this is a cheaper/easier option than writing the test code with tokens. 

With respect to timeouts, do you suggest ignoring the argument and using infinity `#if DEBUG`?

Python/Product/PythonTools/ptvsd_loader.py
```diff
@@ -0,0 +1,40 @@
...
+# 
+# See the Apache Version 2.0 License for specific language governing
+# permissions and limitations under the License.
+
+"""
+Bootstrap loader for ptvsd, used only for local attach scenarios. See PyDebugAttach
+project for usage details.
+"""
+
+import sys
```
huguesv: add `__author__` and `__version__`?

Python/Product/PythonTools/ptvsd_repl_launcher.py
```diff
@@ -0,0 +1,119 @@
...
+
+if __name__ == '__main__':
+    try:
+        _run_repl()
+    except:
+        if repl.DEBUG:
+            sys.__stdout__.write(traceback.format_exc())
+            sys.__stdout__.write('\n\nPress Enter to close...')
+            sys.__stdout__.flush()
+            input()
```
huguesv: try/except with `raw_input` like done near top of the file?

Python/Tests/DebuggerTests/MiniDumpWriter.cs
```diff
@@ -0,0 +1,92 @@
...
+            uint id;
+            try {
+                handle = _process.Handle;
+                id = (uint)_process.Id;
+            } catch (InvalidOperationException) {
+                return;
+            }
+
+            var dumpDir = TestData.GetTempPath(randomSubPath: true);
+            var dumpPath = Path.Combine(dumpDir, "dump.dmp");
```
huguesv: I don't know how frequently you had to review dump files, but would it be useful to have the ability to control the dump file name? Constructor could accept a name, and tests could pass TestContext.FullyQualifiedTestClassName.

huguesv:
That's all I have 👍

Python/Tests/DebuggerTests/MiniDumpWriter.cs
int19h: If the test fails (and triggers this), it will also print out the path to the file to test results, which is easily accessible from the run, or from the test logs if it's running on the server - so it wouldn't add much, I think.

Python/Product/PythonTools/ptvsd_loader.py
int19h: It is not a module - it's loaded from the file as a string, and executed as such. Not meant to be used by the users directly, either. Hence why no docstrings etc.

PR #2748, opened 2017-06-29
Fixes #2727 Editors opened by LSL are not associated with the project

### LINKED ISSUE #2727: Editors opened by LSL are not associated with the project
When you open a solution with lightweight solution load enabled, existing editor windows are opened before our project loads. As a result, we do not associate these windows with the project and they will have incorrect settings (language version, search paths, etc.)

We should search for any open windows on project load and reassociate them with our project automatically.

zooba:
Fixes #2727 Editors opened by LSL are not associated with the project
Updates open editors on project reanalysis to ensure they are associated with the correct project
Ensures squiggles are cleared.
Prevents overlapping asynchronous requests to out-of-proc analyzer.
Prevents unnecessary notifications after disconnecting a buffer parser
Some minor code cleaning.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
@@ -85,6 +85,8 @@ namespace Microsoft.PythonTools.Intellisense {
         private readonly HashSet<ProjectReference> _references = new HashSet<ProjectReference>();
         private bool _disposing;
 
+        private readonly ConcurrentDictionary<object, object> _activeRequests = new ConcurrentDictionary<object, object>();
+
         private readonly IPythonToolsLogger _logger;
 
         internal Task ReloadTask;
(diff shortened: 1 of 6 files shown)


Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -2264,9 +2341,19 @@ class ExprWalker : PythonWalker {
                 fileId = analysis.Entry.FileId,
             };
 
-            var resp = await analysis.Entry.Analyzer.SendRequestAsync(req, timeout: timeout).ConfigureAwait(false);
-
-            return resp?.expression;
+            return analysis.Entry.Analyzer.EnsureSingleRequest(
+                typeof(AP.ExpressionForDataTipRequest),
+                req.expr,
+                t => t == req.expr,
```
huguesv: Shouldn't it also compare the location? What if it's 2 different request for tooltips for the same expression, but in different locations which would produce different results?

zooba: Yeah, it probably should. That's very unlikely to happen in reality (VS only supports one data tip at a time), but I could see it coming up in the test suite.

zooba:
@huguesv Fixed the key you commented on. If there's nothing else that concerns you, please hit merge whenever you like.

PR #2751, opened 2017-06-29
Fixes #161 Rename directory error message

### LINKED ISSUE #161: Rename directory error message
PTVS 2.2 30515.00
VS 2015 RC

I was trying to repro this issue from codeplex and got a slightly different error instead.
https://pytools.codeplex.com/workitem/2418

I started with a new console app.
I added a bunch of folders and files, and packages using Add New Item.
I renamed some of the folders and that worked fine.
I close and reopened the solution and kept on renaming folders.
I eventually got this error:

```
---------------------------
Microsoft Visual Studio
---------------------------
Rename directory failed. Error HRESULT E_FAIL has been returned from a call to a COM component.
---------------------------
OK   
---------------------------
```

It may not be easy to repro but if you try long enough, you'll get an error.

zooba:
Fixes #161 Rename directory error message
Stops raising on non-fatal failure to refresh property browser.

As far as I can tell, this is the only place we can get an `E_FAIL` from this code path. It's certainly not a big deal, but no need to scare the user with a dialog.


huguesv:
👍

PR #2780, opened 2017-07-11
Support debugger reattach

### LINKED ISSUE #2612: VS2017 Reattach to process fails with "Unable to get the Transport / Qualifier"
Attach to process works fine, but using 'Reattach to Process' results in:

```
---------------------------
Microsoft Visual Studio
---------------------------
Unable to reattach to the process:

 python.exe @ tcp://127.0.0.1:3000/: Unable to get the Transport / Qualifier used in the previous debug session.

---------------------------
OK   
---------------------------
```

Code:
```
import ptvsd
ptvsd.enable_attach(None, address = ('0.0.0.0', 3000))
#Enable the below line of code only if you want the application to wait untill the debugger has attached to it
print("wait_for_attach")
ptvsd.wait_for_attach()
print("test")
```

huguesv:
Fix #2612 VS2017 Reattach to process fails with "Unable to get the Transport / Qualifier"

In addition to reattach, the other benefit is that the attach dialog now remembers previously attached URIs, so you can select from the dropdown instead of having to retype them.


Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
```diff
@@ -0,0 +1,70 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.VisualStudio;
+
+namespace Microsoft.PythonTools.Debugger.Remote {
+    internal class PythonRemoteEnumListDebug<T>
+        where T : class {
```
zooba: I thought we already had a base class for this?

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        private readonly IList<T> _elements;
+        private int _index;
+
+        public PythonRemoteEnumListDebug(IList<T> elem) {
+            _elements = elem;
```
zooba: Null check and exception

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
```diff
@@ -0,0 +1,70 @@
...
+        public int GetCount(out uint pcelt) {
+            pcelt = (uint)_elements.Count;
+            return VSConstants.S_OK;
+        }
+
+        public int Next(uint celt, T[] rgelt, ref uint pceltFetched) {
+            uint actual = (uint)Math.Max(0, Math.Min(celt, _elements.Count - _index));
+
+            for (var i = 0; i < actual; i++) {
+                rgelt[i] = _elements[_index];
```
zooba: Range check on `rgelt` and return HRESULT

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
```diff
@@ -0,0 +1,70 @@
...
+                _index++;
+            }
+
+            // If successful, returns S_OK. Returns S_FALSE if fewer than
+            // the requested number of elements could be returned;
```
zooba: Is this the right behaviour? Or is it `S_FALSE` if no elements could be returned?

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        public int Reset() {
+            _index = 0;
+            return VSConstants.S_OK;
+        }
+
+        public int Skip(uint celt) {
+            if (_index + celt <= _elements.Count) {
+                _index += (int)celt;
+                return VSConstants.S_OK;
```
zooba: Should we return `S_FALSE` here if `_index == _elements.Count`? (Same post-condition as the else block, just a different way of reaching it.)

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
int19h: We have one for arrays. But this needs a list so that items can be added and removed.

We could get rid of the array one, and switch all existing uses of that to lists with this code.

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
int19h: `S_FALSE` is supposed to be the indicator that less than `celt` were returned, at least for the standard COM enumeration protocol. This way the caller can spin in a loop until they get `S_FALSE`, and then return however many they've actually got from the last call (which may be 0).

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
int19h: MSDN says to return `S_FALSE` if "the end of the sequence was reached before skipping the requested number of elements".

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
zooba: Even while someone is enumerating it? That sounds like a bad idea.

Better to convert it to an array when requested.

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
zooba: OK

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
zooba: COM is weird... :)

huguesv: The comment I put there was copy/pasted from MSDN

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteEnumListDebug.cs
huguesv: Oh I see you're talking about AD7Enums.cs.  I can try to reuse that one.  There's PythonRemoteEnumDebug.cs that I knew about, which is only able to enumerate 0 or 1 element (null or non-null object).

huguesv:
I changed it to reuse the existing AD7Enum class.

PR #2814, opened 2017-07-20
Fix broken IntelliSense and high CPU

### LINKED ISSUE #2802: IntelliSense stops working and devenv.exe CPU usage goes up
I created a 3.5 64-bit venv and installed tensorflow in it.
I waited for all analysis to be done (hit ignore on about 20 of these asserts https://github.com/Microsoft/PTVS/issues/2801)

Then go through the tutorial at https://www.tensorflow.org/get_started/get_started
I can't get through with typing this short section without IntelliSense breaking:
```
import tensorflow as tf
node1 = tf.constant(3.0, dtype=tf.float32)
node2 = tf.constant(4.0)
print(node1, node2)

sess = tf.Session()
print(sess.run([node1, node2]))
tf.Session()

node3 = tf.add(node1, node2)
print('node3: ', node3)
```

When IntelliSense is broken, I check the CPU for devenv.exe and it's at 75-80% (stays like this for as long as I have the patience to wait). The only analyzer process I have running is at 0%.

I can close and reopen the solution and IntelliSense works again, but after going through a couple more lines of the tutorial, it breaks again.

huguesv:
Fix #2802 IntelliSense stops working and devenv.exe CPU usage goes up

I added a bunch of tracing code and found out that when this bug occurred, the code right after TrySetResult wasn't getting executed (because the thread was executing the continuation).

I tested almost all afternoon with and without the fix, and was getting consistent results (ie no repro with the fix, always repro without).

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
@@ -1817,7 +1817,7 @@ namespace Microsoft.PythonTools.Intellisense {
                 }
                 if (info.Task == null) {
                     info.Value = value;
-                    info.Task = new TaskCompletionSource<U>();
+                    info.Task = new TaskCompletionSource<U>(TaskCreationOptions.RunContinuationsAsynchronously);
                 }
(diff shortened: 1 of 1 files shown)


zooba:
👍

PR #2817, opened 2017-07-20
Fix VS crash when error points to a missing file

### LINKED ISSUE #2816: VS crash when error points to a missing file
Steps to reproduce:
1. Create a VS python project in VS 2017 (15.2), Windows 10
2. Add some files to the project
3. Physically delete one file from disk (leave the file listed in the project)
4. Run pylint -> you'll get at least 1 error message about the missing file
5. Double click that error -> VS will say it's missing and then crash

Not sure this isn't a VS bug but I mostly work in Python so reporting here..

huguesv:
Fix #2816 VS crash when error points to a missing file

Here's the fix in action:
```
---------------------------
---------------------------
The document cannot be opened. It has been renamed, deleted or moved.
---------------------------
OK   
---------------------------
```

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/CustomCommand.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/CustomCommand.cs
@@ -383,7 +383,11 @@ namespace Microsoft.PythonTools.Project {
                     try {
                         PythonToolsPackage.NavigateTo(_serviceProvider, document, Guid.Empty, task.Line, task.Column < 0 ? 0 : task.Column);
                     } catch (FileNotFoundException ex) {
+                        // Happens when file was deleted from the project.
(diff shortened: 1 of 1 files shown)


huguesv:
Should I have just caught everything that's not a critical exception instead?

zooba:
Maybe there's a narrower place to handle that error inside the function?

Since this crashes, I'd add a handler that uses the ReportUnhandledException extension method to display the dialog that lets people copy the error for us. But keep the specific message.

huguesv:
VS was catching it and putting up the activity log error message, so not technically a crash.

Trying it again I found that the behavior was slightly different based on whether you delete the file from project or from file explorer, so I adjusted the fix for that.

zooba:
👍

PR #2832, opened 2017-07-25
Fix failing test adapter tests.

huguesv:
Fixes 4 failing tests.

It was trying to determine the module name from the file name `"<string>"`, which was determined to be an invalid path and throwing an exception.


zooba:
👍

PR #2834, opened 2017-07-26
Fix #2833: ptvsd can't be used as a program

### LINKED ISSUE #2833: ptvsd can't be used as a program
The recent refactoring broke `python -m ptvsd`, because it's trying to import some names that have been moved around, such as `DEFAULT_PORT`.

int19h:
Fix imports.


huguesv:
👍

PR #2841, opened 2017-07-27
Fix #2837: Debugger breaks inside ptvsd

### LINKED ISSUE #2837: Debugger breaks inside ptvsd
Test code:
```py
import ptvsd
ptvsd.enable_attach(None)

x = 1
while True:
    x += 1
    print(x)
```
Breakpoint set on `print`. After attaching, the stack is:
```
>	enable_attach in __init__ line 87	Python	Symbols loaded.
 	PythonApplication1 module line 7	Python	Symbols loaded.
```
where `__init__` is `ptvsd\__init__.py`. The line in question is inside `enable_attach`. However, the line on which it's actually broken is the one with `print`. So the breakpoint is hit, which is correct, but the stack is weird.

The real problem here is that `__init__.py` is not in `DONT_DEBUG`.

int19h:
Fix initial frame list when attaching not including enable_attach frame, causing incorrect stack traces if enable_attach is called from another function.

Add missing modules to DONT_DEBUG list.

### DIFF
--- a/Python/Product/PythonTools/ptvsd/attach_server.py
+++ b/Python/Product/PythonTools/ptvsd/attach_server.py
@@ -137,11 +137,9 @@ def enable_attach(secret, address=('0.0.0.0', ptvsd.DEFAULT_PORT), certfile=None
 
     frames = []
     f = sys._getframe()
-    while True:
-        f = f.f_back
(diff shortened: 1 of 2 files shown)


huguesv:
👍

PR #2847, opened 2017-07-27
Fix test result duration

### LINKED ISSUE #2822: After running tests, elapsed time is incorrect
When running Python tests, the first test to run reports its elapsed time as -0:00:00.0015013. Subsequent tests in the same batch always report their elapsed time as zero.

This worked in VS2015.

huguesv:
Fix #2822 After running tests, elapsed time is incorrect

(will wait to merge - this is a 3.2 bug)

### DIFF
--- a/Python/Tests/TestAdapterTests/TestExecutorTests.cs
+++ b/Python/Tests/TestAdapterTests/TestExecutorTests.cs
@@ -209,7 +209,7 @@ namespace TestAdapterTests {
                 AssertUtil.ContainsAtLeast(resultNames, expectedResult.TestCase.FullyQualifiedName);
                 var actualResult = recorder.Results.SingleOrDefault(tr => tr.TestCase.FullyQualifiedName == expectedResult.TestCase.FullyQualifiedName);
                 Assert.AreEqual(expectedResult.Outcome, actualResult.Outcome, expectedResult.TestCase.FullyQualifiedName + " had incorrect result");
-                Assert.IsTrue(actualResult.Duration >= TimeSpan.Zero);
+                Assert.IsTrue(actualResult.Duration >= expectedResult.MinDuration);
             }
         }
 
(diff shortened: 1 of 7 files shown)


Python/Tests/TestAdapterTests/TestExecutorTests.cs
```diff
@@ -209,6 +209,7 @@ public class TestExecutorTests {
                 AssertUtil.ContainsAtLeast(resultNames, expectedResult.TestCase.FullyQualifiedName);
                 var actualResult = recorder.Results.SingleOrDefault(tr => tr.TestCase.FullyQualifiedName == expectedResult.TestCase.FullyQualifiedName);
                 Assert.AreEqual(expectedResult.Outcome, actualResult.Outcome, expectedResult.TestCase.FullyQualifiedName + " had incorrect result");
+                Assert.IsTrue(actualResult.Duration >= TimeSpan.Zero);
```
zooba: Can we have a test that sleeps for a few seconds and assert the duration is over that?

huguesv: Turns out that the duration is not accurate enough when run in the unit tests. It seemed okay in my manual testing in VS, but in the unit tests I'm getting a duration that is slightly shorter than it should be (for example a test that sleeps 0.5s has a duration of 0.46s). It's probably the messaging overhead that is causing it (more overhead at the start than at the end).

My first attempt at fixing this involved calculating the duration on the Python side, and it looks like I may have to do it that way for best accuracy.

zooba: I'm okay with a test that sleeps for 2 seconds and we ensure the duration is greater than 1.5s - it's significantly better than our current state. But if it's relatively easy to do on the Python side then go ahead.

huguesv:
Additional tests and more accurate duration are in.

zooba:
👍

PR #2848, opened 2017-07-27
Enable launch with mixed-Python/C debugger from vcxproj

zooba:
Adds a new extension that can be installed separately to enable F5 Python/C debugging in C++ projects.


Python/Product/VCDebugLauncher/PythonDebugLaunchProvider.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+
+        private async Task<DebugLaunchSettings> CreateLaunchSettingsAsync(DebugLaunchOptions options) {
+            var settings = new DebugLaunchSettings(options) {
+                LaunchOperation = DebugLaunchOperation.CreateProcess,
+                LaunchDebugEngineGuid = DkmEngineId.NativeEng,
+                Project = VsHierarchy
+            };
+            settings.AdditionalDebugEngines.Add(DebugEngine.AD7Engine.DebugEngineGuid);
```
int19h: Is it correct to assume that, aside from this line, we're basically doing the same thing that normal C++ launch does here?

zooba: Pretty much. Unfortunately we need to reimplement a decent chunk of it, but the aim is for it to be the same. If there are useful other properties we can expose then that may be worthwhile, but I haven't done any of those.

PR #2850, opened 2017-07-27
Fixes #2839 Extension module template has wrong WinSDK version

### LINKED ISSUE #2839: Extension module template has wrong WinSDK version
We have updated the default WinSDK installation, but since VC can't figure out that newer SDKs are okay we need to also update the reference in our template.

zooba:
Fixes #2839 Extension module template has wrong WinSDK version
Updates default WinSDK version
Also fixes debug options


huguesv:
👍

PR #2851, opened 2017-07-27
Adds experiment for no-DB interpreter factory.

zooba:
This is not ready to merge just yet, but I'd like to get it in before the next preview release goes out.

Right now it's hidden behind a registry setting. If you set `HKCU\Software\Microsoft\PythonTools\Experimental@NoDatabaseFactory` to `1` and restart VS you'll get the new interpreter that does not use a backing DB at all.

Currently, it's a pretty poor experience mainly due to `import` not working well, but once that's fixed and we have some skeleton files (implicitly supported by inserting another search path) it should be enough for really keen users to try out.


Python/Product/VSInterpreters/PythonInterpreterInformation.cs
```diff
@@ -22,6 +25,27 @@ class PythonInterpreterInformation {
...
+                    if (asInt.GetValueOrDefault() == 0) {
+                        // REG_DWORD but 0 means no experiment
+                        return false;
+                    }
+                } else if (string.IsNullOrEmpty(value as string)) {
+                    // Empty string or no value means no experiment
+                    return false;
+                }
+            }
+            return true;
```
huguesv: You are enabling it by default?

zooba: No, this is interview-question-level code here :)

The `string.IsNullOrEmpty(value as string)` case catches when the key or the value is missing completely and returns false. It also catches when you create an empty string value, or any type of value other than REG_DWORD.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -0,0 +1,89 @@
...
+    class AstPythonInterpreter : IPythonInterpreter {
+        private readonly AstPythonInterpreterFactory _factory;
+        private readonly Dictionary<BuiltinTypeId, IPythonType> _builtinModule;
+        private readonly ConcurrentDictionary<string, IPythonModule> _modules;
+
+        public AstPythonInterpreter(AstPythonInterpreterFactory factory) {
+            _factory = factory;
+            _builtinModule = new Dictionary<BuiltinTypeId, IPythonType>();
+            _modules = new ConcurrentDictionary<string, IPythonModule>();
+            ModuleNamesChanged?.Invoke(this, EventArgs.Empty);
```
huguesv: This is a no-op, ModuleNamesChanged can't be registered before this code is called.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.PythonTools.Parsing;
+
+namespace Microsoft.PythonTools.Interpreter.Ast {
+    class AstPythonInterpreter : IPythonInterpreter {
+        private readonly AstPythonInterpreterFactory _factory;
+        private readonly Dictionary<BuiltinTypeId, IPythonType> _builtinModule;
+        private readonly ConcurrentDictionary<string, IPythonModule> _modules;
+
+        public AstPythonInterpreter(AstPythonInterpreterFactory factory) {
+            _factory = factory;
```
huguesv: throw if null, otherwise later on there will be a NRE.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
zooba: I know. Did it to get rid of the warning-as-error that was keeping me from building

Python/Product/VSInterpreters/PythonInterpreterInformation.cs
huguesv: Okay, that will work.

zooba:
I need to add some tests, but then I think this is at a suitable stage to check in (as an experiment - still not on by default).

There's still a lot of work to do on the scrape script, but since it just generates Python code which we'll parse normally it is fairly easy to do that independently.

zooba:
Tests are sufficient for now to ensure we don't destroy the entire world - feel free to review closely, bearing in mind this is going to be built up over the next few months.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+            return res;
+        }
+
+        public IList<string> GetModuleNames() {
+            return _factory.GetImportableModules().Keys.ToArray();
+        }
+
+        public IPythonModule ImportModule(string name) {
+            IPythonModule mod;
```
huguesv: validate name for null and empty string (there's a call to name.IndexOf which will crash if null)

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            if (!_modules.TryUpdate(name, mod, sentinalValue)) {
+                mod = _modules[name];
+            }
+
+            return mod;
+        }
+
+        public void Initialize(PythonAnalyzer state) {
+            _builtinModule = state.BuiltinModule.InterpreterModule as IBuiltinPythonModule;
```
huguesv: throw if state is null

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -0,0 +1,179 @@
...
+            LanguageVersion = Configuration.Version.ToLanguageVersion();
+
+            _databasePath = options.DatabasePath;
+
+            if (!GlobalInterpreterOptions.SuppressPackageManagers) {
+                try {
+                    var pm = options.PackageManager;
+                    if (pm != null) {
+                        pm.SetInterpreterFactory(this);
+                        pm.InstalledFilesChanged += PackageManager_InstalledFilesChanged;
```
huguesv: There's no unregistration for this event handler that I can see. Maybe this class needs to be disposable.

Python/Product/Analysis/Interpreter/Ast/AstScrapedPythonModule.cs
```diff
@@ -0,0 +1,121 @@
...
+
+namespace Microsoft.PythonTools.Interpreter.Ast {
+    class AstScrapedPythonModule : IPythonModule {
+        private readonly string _filePath;
+        private string _documentation;
+        private readonly Dictionary<string, IMember> _members;
+        private bool _scraped;
+
+        public AstScrapedPythonModule(string name, string filePath) {
+            _documentation = string.Empty;
```
huguesv: Is it missing a `Name = name`?

zooba:
There's nothing left I want to do as part of this PR - would like to get a build tomorrow, so any more feedback?

huguesv:
Feel free to merge.

Issue #2853, opened 2017-07-28
Ctrl-click go to definition


huguesv:
See C# implementation in https://github.com/dotnet/roslyn/pull/18966

PR #2860, opened 2017-07-28
Support Ctrl-click go to definition

huguesv:
Implement #2853 Ctrl-click go to definition

- Need to look into how to test this.
- Need to update VS on our build machine, because it can't handle the INavigable* interfaces.


zooba:
Can we add a reference assembly for the new interfaces? Maybe copy something from Preview?

Python/Product/PythonTools/PythonTools/Navigation/Navigable/NavigableSymbolSource.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            AnalysisEntry entry = null;
+            _entryService?.TryGetAnalysisEntry(_textView, _buffer, out entry);
+            if (entry == null) {
+                return null;
+            }
+
+            foreach (var token in _classifier.GetClassificationSpans(extent.Span)) {
```
zooba: Split this part down into a separate `internal` function - `static` if possible - and we can test that.

huguesv: Okay, cool.

huguesv:
I'll try including the assembly from the VSSDK Common\Assemblies folder to our Dev15.0, and see if that works.  It's Microsoft.VisualStudio.Language.Intellisense.dll.

huguesv:
Build succeeded.

zooba:
👍 

We should pull that file out once the build machines get updated. I'm on the internal list that should be notified when that happens.

Python/Product/PythonTools/PythonTools/Navigation/Navigable/NavigableSymbolSource.cs
```diff
@@ -0,0 +1,103 @@
...
+
+                // Quickly eliminate anything that isn't the right classification.
+                var name = token.ClassificationType.Classification;
+                if (!_classifications.Any(c => name.Contains(c))) {
+                    continue;
+                }
+
+                // Check with the analyzer, which will give us a precise
+                // result, including the source location.
+                var result = await entry.Analyzer.AnalyzeExpressionAsync(entry, _textView, token.Span.Start);
```
zooba: Cancellation is likely for this function, so we should throw here if the token has been cancelled (since this is the longest-running operation).

I suspect if it's cancelled then the result will be ignored anyway, but no point doing extra work. We should also use `ConfigureAwait` on the AnalyzeExpressionAsync call to avoid marshalling to a particular thread on return - the caller will do that if necessary.

huguesv:
I pushed some changes so I can work on a different machine. Not done yet writing the tests.

huguesv:
I've made the requested changes and added some working tests (and some not working yet).
I've removed the check for File.Exists and made it just string.IsNullOrEmpty instead, to make it easier to test. The Navigate() implementation handles missing file already, and it should be rare that the analyzer gives us a file that doesn't exist.

Python/Tests/PythonToolsMockTests/NavigableTests.cs
```diff
@@ -27,44 +26,26 @@
...
-
-        [TestCleanup]
-        public void TestCleanup() {
-            MockPythonToolsPackage.SuppressTaskProvider = false;
-            VsProjectAnalyzer.SuppressTaskProvider = false;
-            _vs.Dispose();
-        }
-
-        [Ignore] // This works in the IDE, need to figure out why it doesn't here
+        [Ignore] // TODO: Figure out why we don't get any definition, this works in IDE
```
zooba: Waiting on classifications may not be the best way to do this - there should be a `WaitForAnalysis` function that waits on full completion.

huguesv: I just tried it, and it did not help. I suppose you need a real interpreter in order to find standard library modules? Right now PythonEditor uses `InterpreterFactoryCreator.CreateAnalysisInterpreterFactory` because I'm not passing a factory.

huguesv: Ah yes, that was it. It's passing now.

PR #2868, opened 2017-07-31
Fixes #2867 Invalid characters in analysis log output

### LINKED ISSUE #2867: Invalid characters in analysis log output
We include analysis log output in test results, but this can sometimes include "invalid" characters that prevent uploading results (in this example, `&#xB;` and `&#xC;`):

`[N] 22955, string.maketrans, &lt;CalledFunctionAnalysisUnit: Name=string.maketrans (38353703), NodeType=FunctionDefinition&gt; def:(&lt;str ' 	\n\r&#xB;&#xC;'&gt;, str)-&gt;{}(&lt;str ' 	\n\r&#xB;&#xC;'&gt;, str)-&gt;{}`

We need to filter these out. Potentially they need to be filtered. Looking at the rest of this string, it may just be that we need to fix our repr for string literals, since `\n\r` are already converted (but this might be specific to logging - I don't recall right now).

zooba:
Fixes #2867 Invalid characters in analysis log output
Replaces unprintable characters.


huguesv:
👍

PR #2884, opened 2017-08-04
Add more ptvsd unit tests.

int19h:


Python/Tests/DebuggerTests/AttachTests.cs
```diff
@@ -865,116 +905,285 @@ void main()
...
+
+                await proc.StartListeningAsync();
+                await attached.Task.WithTimeout(20000, "Failed to attach within 20s");
+                await bpHit.Task.WithTimeout(20000, "Failed to hit breakpoint within 20s");
+            });
+
+            AssertUtil.ArrayEquals(expectedOutput, actualOutput);
+        }
+
+        [TestMethod, Priority(3)]
```
zooba: Do these need to be priority 3? This means they are too flaky and should not be run on CI

int19h: Ah, I didn't realize that's what it is. It was just copied from the existing `ptvsd` test. It has been stable for me so far, but let me do a few runs of this whole batch and verify.

int19h: Okay, so the stability of the `NoWait` test is still flaky. Let me see if I can improve it further (this is the one that gave me the most trouble so far already); if not, then I'll keep this priority for it, but will update the rest.

int19h: It's pretty hard to repro this failure - seems to be timing dependent? - basically maybe 1 out of 20 runs fail on my box here. Pretty much impossible to debug in a straightforward way, as it never fails under debugger.

For the time being, I'll just mark this one test as flaky.

zooba:
👍

PR #2890, opened 2017-08-08
Fix #2797: Import tensorflow module in PTVS is very slow

### LINKED ISSUE #2797: Import tensorflow module in PTVS is very slow
It takes about 10 seconds to finish "Step Over" to this line

"import tensorflow as if"

It take even longer to execute 

"from tensorflow.contrib import learn"

My script behaves normally when executed in command line. Version of tensorflow is 1.2.1.

PTVS 2.2.2 Visual studio Ultimate 2013.

int19h:
Disable tracing while importing stdlib modules.

### DIFF
--- a/Python/Product/PythonTools/ptvsd/debugger.py
+++ b/Python/Product/PythonTools/ptvsd/debugger.py
@@ -927,13 +927,16 @@ class Thread(object):
             return self.trace_func
 
     def handle_call(self, frame, arg):
-        # If we're importing stdlib, don't trace nested calls until we return from the import that started it.
-        if self.is_importing_stdlib:
-            return self.prev_trace_func
-
         f_code = frame.f_code
         co_name = f_code.co_name
         co_filename = f_code.co_filename
+
+        # If we're importing stdlib, don't trace nested calls until we return from the import that started it,
+        # but notify the user if these nested calls end up in non-stdlib code.
+        if self.is_importing_stdlib:
+            if not is_stdlib(path.normcase(co_filename)):
+                debug_output.write('Standard library module invoked user code during import; breakpoints disabled for invoked code.\n')
+            return self.prev_trace_func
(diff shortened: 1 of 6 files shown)


int19h:
Note that this potentially regresses functionality, in that, if an stdlib module invokes some arbitrary code while being imported, and that code happens to be user code, breakpoints in it won't fire. I think this is unlikely and edge case'y enough that the speed gain is well worth it, though.

Python/Product/PythonTools/ptvsd/debugger.py
```diff
@@ -14,7 +14,7 @@
 # See the Apache Version 2.0 License for specific language governing
 # permissions and limitations under the License.
 
-from __future__ import with_statement
+from __future__ import with_statement, print_function
```
int19h: Not actually used anywhere - I used it for print-debugging. But it's one less line to change when inserting prints, and most importantly, now that we need 2.6+ anyway, it can be here without any ill effects.

zooba:
👍 

On that edge case, in most cases I suspect printing a warning is a helpful thing to do if we can. Chances are the import of user code was accidental shadowing rather than deliberate.

As an aside, I'd like to see us start a pure-Python test suite for ptvsd. This seems like something that is easier to test from Python than from C#. Doesn't have to happen now (though it might be worth getting the structure in so it's easier to expand the tests later), but keep thinking about testability as you go.

int19h:
I'm hesitant to add behavior-modifying logic (i.e. printing to console) for a case that is technically legal Python - debugger shouldn't be judgmental, IMO :)

Perhaps make it go to the Debug Output window? This way it's non-intrusive, but it's still there to diagnose problems. Of course, this is a bit more code, since we'll need a new protocol message.

zooba:
Look great!

Issue #2059, opened 2017-01-12
Parser cannot round-trip exec tuple on 2.7


zooba:
Output from test:

```
Seed == 121627391
Testing version V27 D:\Python27_x64\Lib
Mismatch context at 1643:
Expected: ‼¬‼        ns = {'_as_int': _as_int}¬‼        exec('''if 1:¬‼            def func(n):¬‼             
Got     : ‼¬‼        ns = {'_as_int': _as_int}¬‼        exec'''if 1:¬‼            def func(n):¬‼              
Differs :                                                   ^                                                 
Mismatch context at 4:
Expected: exec('''if 1:¬‼            def func(n):¬‼             
Got     : exec'''if 1:¬‼            def func(n):¬‼              
Differs :     ^                                                 
-----
exec('''if 1:
            def func(n):
                if not isinstance(n, int):
                    n = _as_int(n)
                return int(%s)
            ''' % result, ns)
        return ns['func']
    except RuntimeError:
        # Recursion error can be raised in _parse() or exec().
        raise ValueError('plural form expression is too complex')
```

PR #2899, opened 2017-08-10
Fix buffer handling

zooba:
Again... I guess "improve" is the right description.

* centralize ITextBuffer.Properties content to a single instance of PythonTextBufferInfo
 * this instance should never be replaced for a given text buffer
 * all per-buffer items go in here, including classifiers and analysis info, and the events come from this instance
* merge many service imports into PythonEditorServices
 * this should reduce MEF composition time, and makes it easier to get at everything needed
 * there are still many places that should switch to this, but they weren't necessary to make things work right now
* improvements to mock editor tests
 * run the analyzer in-proc (should result in faster tests)
 * disable reparse delay for mock editors
 * explicitly wait for analysis to complete in tests (more tests need this)

I'm going to continue updating tests to be more reliable, but there's enough here to review.


Python/Product/PythonTools/PythonTools/Editor/PythonEditorServices.cs
```diff
@@ -0,0 +1,122 @@
...
+using Microsoft.VisualStudio.Utilities;
+using Microsoft.VisualStudioTools;
+
+namespace Microsoft.PythonTools.Editor {
+    /// <summary>
+    /// Provides centralized access to services used by the editor.
+    /// </summary>
+    /// <remarks>
+    /// MEF services provided by VS should be imported eagerly using the
+    /// [Import] attribute on a public field.
```
DinoV: If you're going to document a pattern here [ImportingConstructor] is a little bit nicer in that it lets you keep fields readonly

Python/Product/PythonTools/PythonTools/Editor/PythonEditorServices.cs
```diff
@@ -0,0 +1,122 @@
...
+    [Export]
+    sealed class PythonEditorServices {
+        [ImportingConstructor]
+        public PythonEditorServices([Import(typeof(SVsServiceProvider))] IServiceProvider site) {
+            Site = site;
+            ComponentModel = Site.GetComponentModel();
+            _errorTaskProvider = new Lazy<ErrorTaskProvider>(CreateTaskProvider<ErrorTaskProvider>);
+            _commentTaskProvider = new Lazy<CommentTaskProvider>(CreateTaskProvider<CommentTaskProvider>);
+            _unresolvedImportSquiggleProvider = new Lazy<UnresolvedImportSquiggleProvider>(CreateImportSquiggleProvider);
+            _analysisEntryService = new Lazy<AnalysisEntryService>(() => ComponentModel.GetService<AnalysisEntryService>());
```
DinoV: Why not just have these come into the importing constructor via Lazy<...> imports?

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntry.cs
```diff
@@ -91,166 +77,62 @@ internal sealed class AnalysisEntry : IDisposable {
...
-
-            // No buffer parser associated with the file yet.  This can happen when
-            // you have a document that is open but hasn't had focus causing the full
-            // load of our intellisense controller.  In that case there is only a single
-            // buffer which is buffer 0.  An easy repro for this is to open a IronPython WPF
-            // project and close it with the XAML file focused and the .py file still open.
-            // Re-open the project, and double click on a button on the XAML page.  The python
-            // file isn't loaded and weh ave no BufferParser associated with it.
-            return 0;
+            // May get null if there is no analysis entry associated with the file yet.
```
DinoV: May get 0 or should this start returning a nullable int?

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -1377,9 +1378,6 @@ private static T FindImmediateChild<T>(HierarchyNode parent)
                     // Add all our files into our analyzer
                     var files = AllVisibleDescendants.OfType<PythonFileNode>().Select(f => f.Url).ToArray();
                     await analyzer.AnalyzeFileAsync(files);
-                    // Ensure any open files that belong to our project are associated
```
DinoV: What happens now when you have a file open that lives in a project and then open that project?

Python/Product/PythonTools/PythonTools/PythonClassifierProvider.cs
```diff
@@ -46,22 +47,13 @@ internal class PythonClassifierProvider : IClassifierProvider {
...
-            IContentTypeRegistryService contentTypeRegistryService,
-            [Import(typeof(SVsServiceProvider))]IServiceProvider serviceProvider,
-            IClassificationTypeRegistryService classificationRegistry,
-            AnalysisEntryService entryService
-        ) {
-            _type = contentTypeRegistryService.GetContentType(PythonCoreConstants.ContentType);
-            _serviceProvider = serviceProvider;
-            _classificationRegistry = classificationRegistry;
-            _entryService = entryService;
+        public PythonClassifierProvider([Import] PythonEditorServices services) {
```
DinoV: I'm pretty sure The [Import] attribute isn't actually necessary here

Python/Product/VSInterpreters/CPythonInterpreterFactoryProvider.cs
```diff
@@ -125,33 +125,41 @@ class CPythonInterpreterFactoryProvider : IPythonInterpreterFactoryProvider {
...
-
-            using (var baseKey = RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Default))
-            using (var root = baseKey.OpenSubKey(PythonPath)) {
-                search.Search(
-                    root,
-                    Environment.Is64BitOperatingSystem ? InterpreterArchitecture.Unknown : InterpreterArchitecture.x86
-                );
-            }
+            PythonRegistrySearch search = null;
+            Dictionary<string, PythonInterpreterInformation> machineFactories = null; ;
```
DinoV: nit: double semi-colon

Python/Product/PythonTools/PythonTools/Navigation/DropDownBarClient.cs
```diff
@@ -72,9 +74,10 @@ class DropDownBarClient : IVsDropdownBarClient {
 
             _serviceProvider = serviceProvider;
             _uiThread = _serviceProvider.GetUIThread();
+            _services = _serviceProvider.GetComponentModel().GetService<PythonEditorServices>();
             _analysisEntry = analysisEntry;
-            textView.TextBuffer.RegisterForParseTree(ParserOnNewParseTree);
```
DinoV: Did these extension methods go away?  I don't see a delete of them anywhere but it seems like the callers are all disappearing?

Python/Product/PythonTools/PythonTools/Editor/PythonEditorServices.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        public readonly IServiceProvider Site;
+
+        #region PythonToolsService
+
+        private PythonToolsService _python;
+
+        internal void SetPythonToolsService(PythonToolsService service) {
```
DinoV: Is this to avoid circular imports?  Why not import Lazy<...> on both sides?  And is this not called sometimes?  I'm not sure why we also have TryGetPythonToolsService

Python/Product/PythonTools/PythonTools/Editor/PythonEditorServices.cs
zooba: For setting up tests it's nicer to not have 100 arguments in the constructor. I'm not so worried about readonliness - this is an internal class. Static analysis could check for it easily enough.

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntry.cs
zooba: It refers to the null that we coerce to 0, but should actually coerce to -1...

Python/Product/PythonTools/PythonTools/Editor/PythonEditorServices.cs
zooba: Didn't realise this was a thing. Works nice.

Python/Product/PythonTools/PythonTools/Editor/PythonEditorServices.cs
zooba: It's hard to guarantee that the GetService will be on the UI thread for this particular one, and definitely preferable to have it be set in the constructor of the service. We may need this class when documents are opened before our package is loaded (because a chain of MEF imports can get us there), and if any of those get the service then we'll trigger the load via TryGetPythonToolsService. Not entirely sure why I wrote it this way, but it keeps the property looking neat.

Python/Product/PythonTools/PythonTools/PythonClassifierProvider.cs
zooba: You're right, I default to explicit. Removed.

Python/Product/PythonTools/PythonTools/Navigation/DropDownBarClient.cs
zooba: Yeah, they're gone. They were in a weird spot in ProjectAnalyzer, IIRC.

zooba:
FWIW, most of the tests/bugs I'm fixing are due to changes in this changeset. Except fixing #2059, which just seemed easy

zooba:
Someone please stop me before I make any more changes :)

Hopefully the test run will be relatively clean after the latest set of changes, and I'll be ready to merge. We should be in a better spot than before.

@DinoV @int19h reviews would be appreciated, though I know there's a lot that's gone on.

Common/Tests/Utilities/FileUtils.cs
```diff
@@ -167,6 +170,14 @@ public static class FileUtils {
             foreach (var newFile in newFiles) {
                 var copyFrom = Path.Combine(sourceDir, newFile);
                 var copyTo = Path.Combine(destDir, newFile);
+
+                if (tryHardLinkFirst) {
```
int19h: Wouldn't it be better to hardlink the whole directory? If hard linking fails for one file, it will probably fail for the others, and conversely, if it succeeds for one, it probably succeeds for all, so I don't think there's any real benefit in trying to hardlink-or-copy on a per-file basis.

zooba: This is mainly used for "copying" binary dependencies for tests into a directory that has about 50% of them already. There are very few subdirectories, so very little benefit there.

Also, once you create a symlink or a junction to a directory (there's no such thing as a directory hardlink to my knowledge) it gets really hard to delete things, whereas file hardlinks just delete normally and remove the reference. Plus we really don't want to modify the originals (I started with setting the read-only bit here, but that applies to all links to the same file and so it broke subsequent builds).

zooba:
The merge is done, but still happy to factor in feedback into new PRs. Just really wanted to get out of the huge one

PR #2915, opened 2017-08-16
Fix stackless schedule callback.

int19h:
Originally https://github.com/Microsoft/PTVS/pull/2249; recreating to account for debugger changes since then.


zooba:
👍

PR #2918, opened 2017-08-17
Fixes #2904 Mixed-mode debugger does not support 3.7

### LINKED ISSUE #2904: Mixed-mode debugger does not support 3.7
We should allow debugging of 3.7, even though it has not yet been released.

Support will obviously lag behind development somewhat, however at this stage it would be very helpful to be able to develop CPython and use our debugger.

zooba:
Fixes #2904 Mixed-mode debugger does not support 3.7
Adds support for version 3.7 to whole system

I don't think I missed any lists of versions - in general everything should go through the `PythonLanguageVersion` enum (which is now easier to maintain).

I checked the diff between 3.6 and 3.7 as well and there don't seem to be any changes needed in the structures just yet, but there's still time :)

Biggest value of this is that when combined with the VC debug launcher we'll be able to actively use this when working on CPython.

### DIFF
--- a/Python/Product/Analysis/Parsing/PythonLanguageVersion.cs
+++ b/Python/Product/Analysis/Parsing/PythonLanguageVersion.cs
@@ -34,7 +34,8 @@ namespace Microsoft.PythonTools.Parsing {
         V33 = 0x0303,
         V34 = 0x0304,
         V35 = 0x0305,
-        V36 = 0x0306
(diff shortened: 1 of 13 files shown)


zooba:
Tested (though not thoroughly yet) with an install from https://www.myget.org/feed/python/package/nuget/pythondaily configured as a custom environment.

PR #2925, opened 2017-08-21
Fixes #2908 Ast interpreter needs to scrape its own builtin types

### LINKED ISSUE #2908: Ast interpreter needs to scrape its own builtin types
Currently we get builtin types from the analyzer, which tries to get them from us. We need to scrape them ourselves first so we have good answers.

zooba:
Fixes #2908 Ast interpreter needs to scrape its own builtin types
Adds builtin type scraper to be run with current environment. (A future change will cache this output, and there is more work to make the default results better on pre-3.6 interpreters or those that cannot be executed.)
Adds initial function walker to do return value analysis

### DIFF
--- a/Python/Product/Analysis/EncodedLocation.cs
+++ b/Python/Product/Analysis/EncodedLocation.cs
@@ -39,16 +39,16 @@ namespace Microsoft.PythonTools.Analysis {
         public readonly object Location;
 
         public EncodedLocation(ILocationResolver resolver, object location) {
+            if (resolver == null && !(location is LocationInfo)) {
+                throw new ArgumentNullException(nameof(resolver));
+            }
+
             Resolver = resolver;
             Location = location;
         }
 
         public override int GetHashCode() {
-            if (Location != null) {
-                return Resolver.GetHashCode() ^ Location.GetHashCode();
-            }
-
-            return Resolver.GetHashCode();
+            return (Resolver?.GetHashCode() ?? 0) ^ (Location?.GetHashCode() ?? 0);
         }
 
         public override bool Equals(object obj) {
@@ -68,6 +68,12 @@ namespace Microsoft.PythonTools.Analysis {
         #endregion
 
         public LocationInfo GetLocationInfo() {
+            if (Location == null) {
+                return null;
+            }
+            if (Resolver == null) {
+                return Location as LocationInfo;
+            }
             return Resolver.ResolveLocation(Location);
         }
     }
--- a/Python/Product/Analysis/Interpreter/Ast/AstAnalysisFunctionWalker.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstAnalysisFunctionWalker.cs
@@ -14,11 +14,10 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Infrastructure;
-using Microsoft.PythonTools.Parsing;
 using Microsoft.PythonTools.Parsing.Ast;
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
@@ -32,8 +31,8 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             NameLookupContext scope,
             FunctionDefinition targetFunction
         ) {
-            _scope = scope;
-            _target = targetFunction;
+            _scope = scope ?? throw new ArgumentNullException(nameof(scope));
+            _target = targetFunction ?? throw new ArgumentNullException(nameof(targetFunction));
             _returnTypes = new List<IPythonType>();
             _overload = new AstPythonFunctionOverload(
                 AstPythonFunction.MakeParameters(_scope.Ast, _target),
--- a/Python/Product/Analysis/Interpreter/Ast/AstAnalysisFunctionWalker.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstAnalysisFunctionWalker.cs
@@ -16,7 +16,9 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Parsing;
 using Microsoft.PythonTools.Parsing.Ast;
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
@@ -24,6 +26,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
         private readonly FunctionDefinition _target;
         private readonly NameLookupContext _scope;
         private readonly List<IPythonType> _returnTypes;
+        private readonly AstPythonFunctionOverload _overload;
 
         public AstAnalysisFunctionWalker(
             NameLookupContext scope,
@@ -32,10 +35,15 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             _scope = scope;
             _target = targetFunction;
             _returnTypes = new List<IPythonType>();
+            _overload = new AstPythonFunctionOverload(
+                AstPythonFunction.MakeParameters(_scope.Ast, _target),
+                _scope.GetLocOfName(_target, _target.NameExpression),
+                _returnTypes
+            );
         }
 
-        public string Documentation { get; private set; }
         public IList<IPythonType> ReturnTypes => _returnTypes;
+        public IPythonFunctionOverload Overload => _overload;
 
         private void GetMethodType(FunctionDefinition node, out bool classmethod, out bool staticmethod) {
             classmethod = false;
@@ -49,7 +57,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             var classmethodObj = _scope.Interpreter.GetBuiltinType(BuiltinTypeId.ClassMethod);
             var staticmethodObj = _scope.Interpreter.GetBuiltinType(BuiltinTypeId.StaticMethod);
             foreach (var d in (_target.Decorators?.Decorators).MaybeEnumerate()) {
-                var m = _scope.GetValueFromExpression(d, NameLookupContext.LookupOptions.Global);
+                var m = _scope.GetValueFromExpression(d);
                 if (m == classmethodObj) {
                     classmethod = true;
                 } else if (m == staticmethodObj) {
@@ -63,7 +71,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             bool classmethod, staticmethod;
             GetMethodType(_target, out classmethod, out staticmethod);
             if (!staticmethod) {
-                self = _scope.LookupNameInScopes("__class__", NameLookupContext.LookupOptions.LocalOnly);
+                self = _scope.LookupNameInScopes("__class__", NameLookupContext.LookupOptions.Local);
                 if (!classmethod) {
                     var cls = self as IPythonType;
                     if (cls == null) {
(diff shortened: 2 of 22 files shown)


Python/Product/Analysis/EncodedLocation.cs
```diff
@@ -39,16 +39,16 @@ struct EncodedLocation : IEquatable<EncodedLocation> {
...
             Location = location;
         }
 
         public override int GetHashCode() {
-            if (Location != null) {
-                return Resolver.GetHashCode() ^ Location.GetHashCode();
-            }
-
-            return Resolver.GetHashCode();
+            return (Resolver?.GetHashCode() ?? 0) ^ (Location?.GetHashCode() ?? 0);
```
int19h: Since we're on C# 7 now:
```cs
(Resolver, Location).GetHashCode()
```

Python/Product/Analysis/Interpreter/Ast/AstPythonFunction.cs
```diff
@@ -40,31 +45,25 @@ class AstPythonFunction : IPythonFunction, ILocatedMember {
...
 
-        private static IEnumerable<IPythonType> MakeReturns(FunctionDefinition def) {
-            yield break;
-        }
-
         public IPythonModule DeclaringModule {get;}
         public IPythonType DeclaringType {get;}
         public string Name { get; }
-        public string Documentation {get;}
+        public string Documentation => _overloads.FirstOrDefault()?.Documentation;
```
int19h: I wonder if this should rather be `_overloads.Select(o => o.Documentation).Where(doc => doc != null).FirstOrDefault()`, so as to get the first available docstring, rather than the first available overload.

Python/Product/Analysis/Interpreter/Ast/AstScrapedPythonModule.cs
```diff
@@ -25,19 +27,27 @@
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
     class AstScrapedPythonModule : IPythonModule {
-        private readonly string _filePath;
+        private readonly string _name, _filePath;
+        private readonly bool _isBuiltins;
```
int19h: Would this make more sense as a derived class and a bunch of virtual/override? It seems like all the `_isBuiltins` checks are pre/post method body, except for building the scraper command line, which can be factored out into its own method.

Python/Product/Analysis/Interpreter/Ast/NameLookupContext.cs
```diff
@@ -0,0 +1,350 @@
...
+                return null;
+            }
+        }
+
+        public IMember GetValueFromExpression(Expression expr) {
+            return GetValueFromExpression(expr, DefaultLookupOptions);
+        }
+
+        public IMember GetValueFromExpression(Expression expr, LookupOptions options) {
+            var ne = expr as NameExpression;
```
int19h: `expr is NameExpression ne` etc?

Python/Product/Analysis/Interpreter/Ast/AstPythonFunction.cs
zooba: This actually needs a generally better implementation that does more intelligent combining of all the overloads. This hack got me through my existing tests - I'll add a TODO (though tbf, most of this stuff has one big TODO on it right now)

Python/Product/Analysis/Interpreter/Ast/AstScrapedPythonModule.cs
zooba: Yeah, I was starting to think that when adding the tests. Probably worth doing now.

Python/Product/Analysis/Interpreter/Ast/NameLookupContext.cs
zooba: You mean the switch statement form? I don't really like the extra indentation it adds, and honestly I'd rather avoid the added cognitive load of the feature, since it's so new and doesn't make it any easier to read.

Python/Product/Analysis/EncodedLocation.cs
zooba: Does that work when the elements are null?

Python/Product/Analysis/Interpreter/Ast/NameLookupContext.cs
int19h: You can also use it in an `if` (as spelled out above), in which case it's just a shorter form of `as` + null check.

Python/Product/Analysis/EncodedLocation.cs
int19h: Yep. `GetHashCode` has to be valid for any value, so tuple has to do whatever magic it does to take care of producing something when some of the elements are null.

The main benefit of using tuples for this, aside from conciseness, is that you get a much better implementation that combines constituent hashes - there are many known scenarios where `^` produces pathologically bad hashes.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -123,17 +131,37 @@ string name
...
-                try {
-                    return ModulePath.FromBasePathAndName(searchPath, name);
-                } catch (ArgumentException) {
+                if (ModulePath.FromBasePathAndName_NoThrow(
+                    searchPath,
+                    name,
+                    null,
+                    null,
+                    out mp,
+                    out isInvalid,
```
int19h: You can now do `out _` for unused out params, to avoid declaring dummy variables.

Python/Product/Analysis/Interpreter/Ast/AstScrapedPythonModule.cs
```diff
@@ -46,19 +56,49 @@ class AstScrapedPythonModule : IPythonModule {
...
                 return _members.Keys.ToArray();
             }
         }
 
+        private void SetBuiltinTypeIds() {
+            lock (_members) {
+                foreach (BuiltinTypeId typeId in Enum.GetValues(typeof(BuiltinTypeId))) {
+                    IMember m;
+                    AstPythonBuiltinType biType;
+                    if (_members.TryGetValue(@"__{typeId}", out m) && (biType = m as AstPythonBuiltinType) != null) {
```
int19h: I think this is supposed to be `"$__{typeId}"?

Python/Product/Analysis/EncodedLocation.cs
zooba: If I get a type-missing error in the IDE when I make it this, which reference am I missing? And is that reference available on .NET Core?

int19h: Ah, I forgot about that! It depends on a NuGet package that provides `System.ValueTuple` (the lightbulb on the error should say as much). The other option is to target .NET 4.7, which has it in the box; but I'm not sure whether VS itself has that as hard dependency, and hence whether we can target it or not.

The NuGet package is available for all .NET targets. This is going to be the general approach for new C# features requiring additional types going forward (they also did it with `INotifyCompletion`, for example).

zooba: Yeah, we can't do 4.7 yet, but I'd be surprised if VS doesn't already include that dependency. Still, not worth it yet just for this one off.

int19h: FWIW, the Analysis library could really benefit from tuples in more conventional ways, seeing how often it uses `out` for multiple return values (that are also quite often discarded).

Either way, up to you.

Python/Product/Analysis/Interpreter/Ast/AstAnalysisFunctionWalker.cs
```diff
@@ -0,0 +1,127 @@
...
+        private readonly List<IPythonType> _returnTypes;
+        private readonly AstPythonFunctionOverload _overload;
+
+        public AstAnalysisFunctionWalker(
+            NameLookupContext scope,
+            FunctionDefinition targetFunction
+        ) {
+            _scope = scope;
+            _target = targetFunction;
+            _returnTypes = new List<IPythonType>();
```
huguesv: throw ArgumentNullException for `_scope` and `_target`

Python/Product/Analysis/Interpreter/Ast/AstAnalysisFunctionWalker.cs
```diff
@@ -0,0 +1,127 @@
...
+            staticmethod = false;
+
+            if (node.IsLambda) {
+                staticmethod = true;
+                return;
+            }
+
+            var classmethodObj = _scope.Interpreter.GetBuiltinType(BuiltinTypeId.ClassMethod);
+            var staticmethodObj = _scope.Interpreter.GetBuiltinType(BuiltinTypeId.StaticMethod);
+            foreach (var d in (_target.Decorators?.Decorators).MaybeEnumerate()) {
```
huguesv: Why do you use parentheses when calling `MaybeEnumerate`? I notice you've done that in a couple of other places.

Python/Product/Analysis/Interpreter/Ast/AstAnalysisWalker.cs
```diff
@@ -47,45 +45,43 @@ class AstAnalysisWalker : PythonWalker {
             Dictionary<string, IMember> members,
             bool includeLocationInfo
         ) {
-            _interpreter = interpreter;
-            _context = _interpreter.CreateModuleContext();
-            _ast = ast;
+            _scope = new NameLookupContext(interpreter, interpreter.CreateModuleContext(), ast, filePath, includeLocationInfo);
```
huguesv: Should check ctor parameters for null

Python/Product/Analysis/Interpreter/Ast/AstPythonFunction.cs
```diff
@@ -23,14 +23,19 @@
...
 
-        public AstPythonFunction(PythonAst ast, IPythonModule declModule, IPythonType declType, FunctionDefinition def, LocationInfo loc, string doc) {
+        public AstPythonFunction(
+            PythonAst ast,
+            IPythonModule declModule,
+            IPythonType declType,
+            FunctionDefinition def,
+            LocationInfo loc
+        ) {
             DeclaringModule = declModule;
```
huguesv: check ctor params for null

Python/Product/Analysis/Interpreter/Ast/NameLookupContext.cs
```diff
@@ -0,0 +1,350 @@
...
+        private readonly Lazy<IPythonModule> _builtinModule;
+
+        public NameLookupContext(
+            IPythonInterpreter interpreter,
+            IModuleContext context,
+            PythonAst ast,
+            string filePath,
+            bool includeLocationInfo
+        ) {
+            Interpreter = interpreter;
```
huguesv: null checks here too, btw you can do them inline with ??

Python/Product/Analysis/Interpreter/Ast/NameLookupContext.cs
huguesv: I agree with both of you. I'm on board with using `if (expr is NameExpression ne)`, and I'm not convinced regarding the new switch features.

Python/Product/Analysis/Interpreter/Ast/AstAnalysisFunctionWalker.cs
zooba: Without it you could end up trying to enumerate over `null`:
`_target.Decorators == null ? null : _target.Decorators.Decorators.MaybeEnumerate()`

With the parens, the `null` get passed to the extension method and we are guaranteed an empty enumerable:
`(_target.Decorators == null ? null : _target.Decorators.Decorators).MaybeEnumerate()`

huguesv:
👍

PR #2930, opened 2017-08-22
Fix font in django project settings

### LINKED ISSUE #2928: Django properties do not scale with high DPI
(Reported at https://developercommunity.visualstudio.com/content/problem/38593/text-field-in-django-project-properties-does-not-s.html)

When using high DPI the fields in the Django project properties panel do not scale.

huguesv:
Fix #2928 Django properties do not scale with high DPI

Changed font to Segoe 8.25pt to match other project settings (the designer changed a bunch of margins automatically, but it all looks fine to me).

image

image


zooba:
👍

PR #2935, opened 2017-08-22
Improve test result stack trace

### LINKED ISSUE #2871: Unit test stack trace is not very good
The stack trace for failed unit tests appears to be upside down (earliest call first), has no module or class information, and includes standard library modules:
image

This trace should have one entry that reads `test_util.Test_util.test_daterange_year` or at worst, `Test_util.test_daterange_year`.

### LINKED ISSUE #2882: Run test never completes on Python 2.5, 2.6, 3.0 and 3.1
Create new Python application
Add new item, Python unit test
Activate Python 2.5 or 2.6 for your project
Open test explorer, run test

Result: Progress bar show testing is in progress, but it never ends, you have to hit cancel.

Depending on the effort, we should either make it work or make it fail fast.

huguesv:
Fix #2871 Unit test stack trace is not very good
Fix #2882 Run test never completes on Python 2.5, 2.6, 3.0 and 3.1

Reuse the debugger QualifiedFunctionNameWalker to improve the method display names.
It's currently shared between debugger and test adapter with a linked file, which I generally want to avoid, but I just wasn't sure which project should have it.

See the bug for more details.


Python/Product/Debugger/Debugger.csproj
```diff
@@ -105,6 +105,7 @@
       <Link>Debugger\Transports\WebSocketStream.cs</Link>
     </Compile>
     <Compile Include="AssemblyInfo.cs" />
+    <Compile Include="QualifiedFunctionNameWalker.cs" />
```
int19h: As far as where this should go - Analysis, perhaps? It feels like a broadly useful thing, and it doesn't have any dependencies.

zooba: Sounds good to me

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
```diff
@@ -160,16 +160,33 @@ def addUnexpectedSuccess(self, test):
         self.sendResult(test, 'passed')
 
     def sendResult(self, test, outcome, trace = None):
+        def is_framework_frame(f):
+            import os.path
+            for lib_path in unittest.__path__:
```
zooba: This won't work on Python 2.6, 3.0 or 3.1.

zooba:
Other than the `is_framework_frame` check being problematic on older versions, it looks good.

huguesv:
Added fix for #2882.
Python 2.5 is still not supported (no json module), but it quickly fails the tests.
Python 2.6, 3.0 and 3.1 were failing before due `TextTestResult` not being available and `TextTestRunner` not having the `resultclass` parameter, and there were some issues in the new code as well `unittest.__path_`, and `sys.version_info.major`.

I've modified the filtering of the call stack slightly, the original implementation was resulting in an empty call stack if a unit test function was at the top of the stack. I've added a test to check that very common case which I had missed.

I've also added validation that we get the import error message, instead of attribute error message when an import fails and prevents the test case from being loaded (for #2938) though I'm not seeing the behavior that the user reported.

We really should run all test executor tests under every interpreter, like we do for REPL, debugger, etc. There's been enough issues on specific versions to be worth it.  I'll work on this now.

huguesv:
Tests for all Python versions added (and they all pass).

Python/Product/TestAdapter/PythonStackTraceParser.cs
```diff
@@ -39,17 +55,68 @@ sealed class PythonStackTraceParser : IStackTraceParser {
...
                         );
                     } else {
-                        yield return new StackFrame(match.Groups[3].Value);
+                        yield return new StackFrame(funcName);
                     }
                 }
             }
         }
+
+        /// <summary>
```
int19h: Can you move that comment to `QualifiedFunctionNameWalker`, since it's now defined far away from these helpers? This helper can then just reference that (as well as the one in debugger).

huguesv:
@zooba this changed quite a bit since you last looked at it

Python/Product/TestAdapter.Executor/TestExecutor.cs
```diff
@@ -532,7 +532,11 @@ sealed class TestRunner : IDisposable {
                         DebugInfo("set " + pythonPath.Key + "=" + pythonPath.Value);
                         DebugInfo(proc.Arguments);
 
-                        _connected.WaitOne();
+                        // If there's an error in the launcher script,
+                        // it will terminate without connecting back.
+                        while (!_connected.WaitOne(1000) && !proc.ExitCode.HasValue) {
```
zooba: I think we can get a WaitHandle for the process, which means we can just wait on both of them with `WaitHandle.WaitAny`.

huguesv: Yep, changing it.

PR #2936, opened 2017-08-23
Fixes #2924 Assert, then unexpected error when using New Window

### LINKED ISSUE #2924: Assert, then unexpected error when using New Window
Open a Python file that has some code in it
From main menu, Window -> New Window
You'll get an assert because the text buffer was already initialized (from BufferParser.InitBuffer)
Close the new window
Move the cursor around in the original 'window' to trigger a tooltip, you'll get an activity log error
If you try to use IntelliSense, you'll get the unexpected error dialog instead

```
System.ArgumentException: Failed to get classifier from buffer
at Microsoft.PythonTools.Intellisense.ReverseExpressionParser..ctor(ITextSnapshot snapshot, ITextBuffer buffer, ITrackingSpan span) in C:\Projects\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\ReverseExpressionParser.cs:line 53
at Microsoft.PythonTools.Language.TextViewFilter.GetDataTipText(TextSpan[] pSpan, String& pbstrText) in C:\Projects\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Navigation\TextViewFilter.cs:line 119
at Microsoft.VisualStudio.Editor.Implementation.ShimQuickInfoSource.TryGetQuickInfoFromFilter(IQuickInfoSession session, TextSpan[] dataBufferTextSpan, String& tipText)
at Microsoft.VisualStudio.Editor.Implementation.ShimQuickInfoSource.AugmentQuickInfoSession(IQuickInfoSession session, IList`1 qiContent, ITrackingSpan& applicableToSpan)
at Microsoft.VisualStudio.Language.Intellisense.Implementation.QuickInfoSession.Recalculate() at Microsoft.VisualStudio.Language.Intellisense.Implementation.QuickInfoSession.Start()
at Microsoft.VisualStudio.Language.Intellisense.Implementation.DefaultQuickInfoController.OnTextView_MouseHover(Object sender, MouseHoverEventArgs e)
at Microsoft.VisualStudio.Text.Editor.Implementation.WpfTextView.RaiseHoverEvents()
```

`IntellisenseController.DisconnectSubjectBuffer` is called when the view is closed, and it does a `PythonTextBufferInfo.TryDispose(subjectBuffer)` which removes the `PythonTextBufferInfo` property, and `ReverseExpressionParser` expects it to be there later on.

I've had the assert and the TryDispose calls commented out while I've been working on Peek Definition, which is also a new window on a shared text buffer. Not sure what the right fix is.

zooba:
Fixes #2924 Assert, then unexpected error when using New Window
Reorders the relationships between our various buffer using/wrapping classes.

Key points:
* nothing should keep `PythonTextBufferInfo` alive except the text buffer itself (circular references are okay)
 * (this is why we now have a sink interface instead of event handlers)
* all files are analyzed on disk, and then we add buffer info if/when we have it

Still some things to fix, tests to run, etc., but the core change that needs reviewing is here.

### DIFF
--- a/Common/Tests/MockVsTests/MockClassifierAggregatorService.cs
+++ b/Common/Tests/MockVsTests/MockClassifierAggregatorService.cs
@@ -42,31 +42,30 @@ namespace Microsoft.VisualStudioTools.MockVsTests {
 
         sealed class AggregatedClassifier : IClassifier, IDisposable {
             private readonly ITextBuffer _buffer;
-            private readonly List<IClassifier> _classifiers;
+            private readonly IReadOnlyList<IClassifier> _classifiers;
 
             public AggregatedClassifier(ITextBuffer textBuffer, IEnumerable<IClassifierProvider> providers) {
                 _buffer = textBuffer;
                 _classifiers = providers.Select(p => p.GetClassifier(_buffer)).ToList();
+                foreach (var c in _classifiers) {
+                    c.ClassificationChanged += Subclassification_Changed;
+                }
+            }
+
+            private void Subclassification_Changed(object sender, ClassificationChangedEventArgs e) {
+                var c = (IClassifier)sender;
+                var refreshSpans = c.GetClassificationSpans(e.ChangeSpan);
+                ClassificationChanged?.Invoke(this, e);
             }
 
             public void Dispose() {
-                foreach (var c in _classifiers.OfType<IDisposable>()) {
-                    c.Dispose();
+                foreach (var c in _classifiers) {
+                    c.ClassificationChanged -= Subclassification_Changed;
+                    (c as IDisposable)?.Dispose();
                 }
             }
 
-            public event EventHandler<ClassificationChangedEventArgs> ClassificationChanged {
-                add {
-                    foreach (var c in _classifiers) {
-                        c.ClassificationChanged += value;
-                    }
-                }
-                remove {
-                    foreach (var c in _classifiers) {
-                        c.ClassificationChanged -= value;
-                    }
-                }
-            }
+            public event EventHandler<ClassificationChangedEventArgs> ClassificationChanged;
 
             public IList<ClassificationSpan> GetClassificationSpans(SnapshotSpan span) {
                 return _classifiers.SelectMany(c => c.GetClassificationSpans(span)).ToList();
--- a/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
+++ b/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
@@ -33,7 +33,7 @@ using Microsoft.VisualStudio.Utilities;
 namespace Microsoft.PythonTools.Editor {
     sealed class PythonTextBufferInfo {
         public static PythonTextBufferInfo ForBuffer(PythonEditorServices services, ITextBuffer buffer) {
-            var bi = buffer.Properties.GetOrCreateSingletonProperty(
+            var bi = (buffer ?? throw new ArgumentNullException(nameof(buffer))).Properties.GetOrCreateSingletonProperty(
                 typeof(PythonTextBufferInfo),
                 () => new PythonTextBufferInfo(services, buffer)
             );
@@ -74,16 +74,6 @@ namespace Microsoft.PythonTools.Editor {
             return view.BufferGraph.GetTextBuffers(_ => true)
                 .Select(b => TryGetForBuffer(b))
                 .Where(b => b != null);
-        }
-
-        public static bool TryDisconnect(object owner, ITextBuffer buffer) {
-            PythonTextBufferInfo bi;
-            if (buffer.Properties.TryGetProperty(typeof(PythonTextBufferInfo), out bi)) {
-                buffer.Properties.RemoveProperty(typeof(PythonTextBufferInfo));
-                //bi?.Disconnect(owner);
-                return true;
-            }
-            return false;
         }
 
         private readonly object _lock = new object();
@@ -323,32 +313,6 @@ namespace Microsoft.PythonTools.Editor {
             LastAnalysisReceivedVersion = ver;
             return true;
         }
-
-        //public Task<AnalysisEntry> WaitForAnalysisEntryAsync(CancellationToken cancellationToken) {
-        //    // Return our current entry if we have one
-        //    var entry = AnalysisEntry;
-        //    if (entry != null) {
-        //        return Task.FromResult(entry);
-        //    }
-
-        //    var tcs = new TaskCompletionSource<AnalysisEntry>();
-        //    if (cancellationToken.CanBeCanceled) {
-        //        cancellationToken.Register(() => tcs.TrySetCanceled());
-        //    }
-
-        //    EventHandler handler = null;
-        //    handler = (s, e) => {
-        //        cancellationToken.ThrowIfCancellationRequested();
-        //        var bi = (PythonTextBufferInfo)s;
-        //        var result = bi.AnalysisEntry;
-        //        if (result != null) {
-        //            bi.OnNewAnalysisEntry -= handler;
-        //            tcs.TrySetResult(result);
-        //        }
-        //    };
-
-        //    return tcs.Task;
-        //}
 
         public bool DoNotParse {
             get => Buffer.Properties.ContainsProperty(BufferParser.DoNotParse);
--- a/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
+++ b/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
@@ -15,6 +15,10 @@
 // permissions and limitations under the License.
 
 using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PythonTools.Infrastructure;
@@ -23,15 +27,21 @@ using Microsoft.PythonTools.Parsing;
 using Microsoft.PythonTools.Repl;
 using Microsoft.VisualStudio.InteractiveWindow;
 using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Utilities;
 
 namespace Microsoft.PythonTools.Editor {
-    sealed class PythonTextBufferInfo : IDisposable {
+    sealed class PythonTextBufferInfo {
         public static PythonTextBufferInfo ForBuffer(PythonEditorServices services, ITextBuffer buffer) {
-            return buffer.Properties.GetOrCreateSingletonProperty(
+            var bi = buffer.Properties.GetOrCreateSingletonProperty(
                 typeof(PythonTextBufferInfo),
                 () => new PythonTextBufferInfo(services, buffer)
             );
+            if (bi._replace) {
+                bi = bi.ReplaceBufferInfo();
+                buffer.Properties[typeof(PythonTextBufferInfo)] = bi;
+            }
+            return bi;
         }
 
         public static PythonTextBufferInfo TryGetForBuffer(ITextBuffer buffer) {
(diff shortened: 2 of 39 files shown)


zooba:
Many tests fail because `PythonEditorServices` requires `MockComponentModel.get_DefaultCompositionService()`, which is currently not implemented.

zooba:
@huguesv Could you take a look at this? I suspect it's holding up your peek PR, and we're all running late for the next preview.

FWIW, the non-UI tests are mostly fine, and the UI tests are totally busted on the test machines (i.e. 0% success rate, instant failure for basically everything - not my fault :) )

Common/Tests/MockVsTests/MockClassifierAggregatorService.cs
```diff
@@ -42,32 +42,31 @@ public class MockClassifierAggregatorService : IClassifierAggregatorService {
...
+                }
             }
 
-            public void Dispose() {
-                foreach (var c in _classifiers.OfType<IDisposable>()) {
-                    c.Dispose();
-                }
+            private void Subclassification_Changed(object sender, ClassificationChangedEventArgs e) {
+                var c = (IClassifier)sender;
+                var refreshSpans = c.GetClassificationSpans(e.ChangeSpan);
```
huguesv: >GetClassificationSpans [](start = 37, length = 22)

The result of this is thrown away. Is there a desired side effect that is achieved by calling this?

zooba: It ensures the token cache is filled for PythonClassifier. The sequence of events is strange, and probably some assumptions based on how VS works need to be fixed, but it goes roughly like this:

* create new text view with text already in it
* classifier attaches and we get classifications to fill the cache (implicit in VS, already explicit elsewhere in our tests)
* analysis entry gets assigned to the buffer, which resets the cache and raises the event (happens in both)
* in VS, and with this call here, the event triggers getting the new classifications, which refills the cache
* we delete the text from the buffer, which deletes the relevant part of the cache

At the last step, if the cache was not refilled then we throw an exception. This exception never seems to occur in actual use, because the aggregator gets the new classifications immediately. However, our test previously was not getting classifications immediately.

The token cache uses "highly optimized" array operations that rely on it being in a certain state. I don't want to touch it just yet, so I set up the mock tests to do the same steps as VS does (which arguably is the right way to do it - if there are problems in the classifier's assumptions, we want them to show up in both VS and mocks).

Python/Product/PythonTools/PythonTools/PythonClassifier.cs
```diff
@@ -19,58 +19,46 @@
...
-            _buffer = buffer;
-
-            _buffer.OnNewAnalysisEntry += NewAnalysisEntry;
-
-            _version = _buffer.LanguageVersion;
         }
 
-        private void NewAnalysisEntry(object sender, EventArgs e) {
-            var analyzer = _buffer.AnalysisEntry?.Analyzer;
+        private Task OnNewAnalysisEntryAsync(PythonTextBufferInfo sender, AnalysisEntry entry) {
```
huguesv: >OnNewAnalysisEntryAsync [](start = 21, length = 23)

Doesn't need to be async

Python/Product/PythonTools/PythonTools/PythonClassifier.cs
```diff
@@ -19,58 +19,46 @@
...
         [ThreadStatic]
         private static Dictionary<PythonLanguageVersion, Tokenizer> _tokenizers;    // tokenizer for each version, shared between all buffers
 
-        internal PythonClassifier(PythonClassifierProvider provider, PythonTextBufferInfo buffer) {
-            buffer.OnChanged += BufferChanged;
-            buffer.OnContentTypeChanged += BufferContentTypeChanged;
-
+        internal PythonClassifier(PythonClassifierProvider provider) {
             _tokenCache = new TokenCache();
             _provider = provider;
```
huguesv: >provider [](start = 24, length = 8)

You could check for null

zooba: It's an internal constructor that gets passed `this` - it'll never be null.

Python/Product/PythonTools/PythonTools/PythonClassifier.cs
```diff
@@ -116,26 +106,17 @@ internal sealed class PythonClassifier : IClassifier {
 
         #region Private Members
 
-        private Dictionary<TokenCategory, IClassificationType> CategoryMap {
-            get {
-                return _provider.CategoryMap;
+        private Task OnTextContentChangedAsync(PythonTextBufferInfo sender, TextContentChangedEventArgs e) {
```
huguesv: >OnTextContentChangedAsync [](start = 21, length = 25)

Doesn't need to be async

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -23,63 +27,90 @@
...
+                .Select(b => TryGetForBuffer(b))
+                .Where(b => b != null);
+        }
+
+        public static bool TryDisconnect(object owner, ITextBuffer buffer) {
             PythonTextBufferInfo bi;
             if (buffer.Properties.TryGetProperty(typeof(PythonTextBufferInfo), out bi)) {
                 buffer.Properties.RemoveProperty(typeof(PythonTextBufferInfo));
-                bi?.Dispose();
+                //bi?.Disconnect(owner);
```
huguesv: >Disconnect [](start = 22, length = 10)

delete?

zooba: 👍

Python/Product/PythonTools/PythonTools/PythonClassifier.cs
zooba: Also doesn't hurt, and it saves changes later if we make it async (it's been refactored out of an async method).

Python/Product/PythonTools/PythonTools/PythonClassifier.cs
zooba: Same as the other one

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -23,63 +27,90 @@
...
 using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Utilities;
 
 namespace Microsoft.PythonTools.Editor {
-    sealed class PythonTextBufferInfo : IDisposable {
+    sealed class PythonTextBufferInfo {
         public static PythonTextBufferInfo ForBuffer(PythonEditorServices services, ITextBuffer buffer) {
-            return buffer.Properties.GetOrCreateSingletonProperty(
+            var bi = buffer.Properties.GetOrCreateSingletonProperty(
```
huguesv: >buffer [](start = 21, length = 6)

The TryGetForBuffer checks for null param, so this one should probably throw ANE rather than NRE.

zooba: 👍

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -124,66 +173,89 @@ sealed class PythonTextBufferInfo : IDisposable {
...
-        public PythonAnalysisClassifier AnalysisClassifier => _analysisClassifier;
-        public PythonAnalysisClassifier GetOrCreateAnalysisClassifier(Func<PythonTextBufferInfo, PythonAnalysisClassifier> creator) => GetOrCreate(ref _analysisClassifier, creator);
+        public T GetOrCreateSink<T>(object key, Func<PythonTextBufferInfo, T> creator) where T : class, IPythonTextBufferInfoEventSink {
+            IPythonTextBufferInfoEventSink sink;
+            if (_eventSinks.TryGetValue(key, out sink)) {
+                return sink as T;
+            }
+            sink = creator(this);
+            if (!_eventSinks.TryAdd(key, sink)) {
+                sink = _eventSinks[key];
```
huguesv: >_eventSinks [](start = 23, length = 11)

Not sure why this would find a sink when a few lines above it didn't find one. Unless the creator callback had code in it that called GetOrCreateSink?

zooba: No telling what could be going on in the creator callback. This is the best pattern for GetOrCreate with a concurrent dictionary (except maybe disposing of IDisposable sinks if we don't use them...)

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -249,31 +324,31 @@ sealed class PythonTextBufferInfo : IDisposable {
...
-                var result = bi.AnalysisEntry;
-                if (result != null) {
-                    bi.OnNewAnalysisEntry -= handler;
-                    tcs.TrySetResult(result);
-                }
-            };
-
-            return tcs.Task;
-        }
+        //public Task<AnalysisEntry> WaitForAnalysisEntryAsync(CancellationToken cancellationToken) {
```
huguesv: >WaitForAnalysisEntryAsync [](start = 37, length = 25)

More code to delete?

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
zooba: Except their Dispose probably removes them from the buffer, and in the context of this race it will likely succeed, so they shouldn't be disposed here. (A finalizer-triggered dispose just shouldn't do the remove, since it's only for unmanaged objects.)

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
zooba: 👍

Python/Product/PythonTools/PythonTools/Intellisense/BufferParser.cs
```diff
@@ -303,31 +294,37 @@ sealed class BufferParser : IDisposable {
...
-            AnalysisEntry entry,
+            VsProjectAnalyzer analyzer,
             IEnumerable<ITextSnapshot> snapshots
         ) {
-            var updates = snapshots.Select(s => GetUpdateForSnapshot(services, s)).Where(u => u != null).ToList();
+            var updates = snapshots
+                .GroupBy(s => PythonTextBufferInfo.TryGetForBuffer(s.TextBuffer)?.AnalysisEntry.FileId ?? -1)
+                .Where(g => g.Key >= 0)
+                .Select(g => Tuple.Create(g.Key, g.Select(s => GetUpdateForSnapshot(services, s)).ToArray()))
+                .Where(u => u != null).ToList();
```
huguesv: >Where(u => u != null) [](start = 17, length = 21)

I'm confused as to what 'u' is now. It used to be the update object. Is that still the case?

zooba: Great catch! This code is somewhat broken now (though I guess we can handle null updates). I'll fix

huguesv: I mean it looks like it's the tuple now, which would never be null?

---
In reply to: [135308418](https://github.com/Microsoft/PTVS/pull/2936#discussion_r135308418) [](ancestors = 135308418)

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -706,39 +706,99 @@ string stdInClientHandle
...
-        /// Creates a new ProjectEntry for the collection of buffers.
-        /// </summary>
-        internal async void ReAnalyzeTextBuffers(BufferParser oldParser) {
-            ITextBuffer[] buffers = oldParser.Buffers;
-            if (buffers.Length > 0) {
-                if (!oldParser.AnalysisEntry.SuppressErrorList) {
-                    _services.ErrorTaskProvider?.ClearErrorSource(oldParser.AnalysisEntry, ParserTaskMoniker);
-                    _services.ErrorTaskProvider?.ClearErrorSource(oldParser.AnalysisEntry, UnresolvedImportMoniker);
-                    _services.CommentTaskProvider?.ClearErrorSource(oldParser.AnalysisEntry, ParserTaskMoniker);
+        internal void ReanalyzeEntry(AnalysisEntry entry) {
```
huguesv: >ReanalyzeEntry [](start = 22, length = 14)

Is anyone calling this? And does this really reanalyze the entry?

zooba: No, it's dead code now, replaced by the "Transfer[File]FromOldAnalyzer" functions (it used to be called when you renamed a source file).

And I also noticed the `OnModulesChanged` event handler is no longer called... need to figure out when we should trigger that.

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -706,39 +706,99 @@ string stdInClientHandle
...
+                if (e == null) {
+                    continue;
                 }
 
-                var monitoredResult = await MonitorTextBufferAsync(buffers[0]);
-                if (monitoredResult?.AnalysisEntry != null) {
-                    for (int i = 1; i < buffers.Length; i++) {
-                        monitoredResult.AddBuffer(buffers[i]);
+                ITextBuffer[] buffers;
+                if (oldBuffers.TryGetValue(e.Path, out buffers)) {
```
huguesv: >TryGetValue [](start = 31, length = 11)

We could start using the inline out variable declaration like this:
if (oldBuffers.TryGetValue(e.Path, out ITextBuffer[] buffers))

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
zooba: Heh, that's been unused for more than a few months now :)

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -774,42 +823,42 @@ string stdInClientHandle
...
-            var snapshot = (entry.AnalysisCookie as SnapshotCookie)?.Snapshot;
-            // File cannot be parsed from disk, so sync the code from the current buffer
-            // This should be rare in normal use, but very frequent in tests
-            if (snapshot != null && (isTemporaryFile || !File.Exists(entry.Path))) {
-                await BufferParser.ParseBuffersAsync(_services, entry, Enumerable.Repeat(snapshot, 1));
-            }
-
-            return bufferParser;
-        }
+        //internal async Task<BufferParser> MonitorTextBufferAsync(ITextBuffer textBuffer, bool isTemporaryFile = false, bool suppressErrorList = false) {
```
huguesv: >MonitorTextBufferAsync [](start = 44, length = 22)

More commented out code here

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
zooba: Sometimes I don't mind that, other times it's weird (mostly because the scope of the declaration is _outside_ the if block, not restricted to inside it, so you can reuse the variable later on - not a fan of that).

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
zooba: 👍

Python/Product/PythonTools/PythonTools/Intellisense/XamlTextViewCreationListener.cs
```diff
@@ -31,46 +33,40 @@ namespace Microsoft.PythonTools.Intellisense {
...
         }
 
         private void TextView_Closed(object sender, EventArgs e) {
             var textView = (ITextView)sender;
 
-            AnalysisEntry entry;
-            if (_entryService.TryGetAnalysisEntry(textView, textView.TextBuffer, out entry)) {
-                entry.Analyzer.BufferDetached(entry, textView.TextBuffer);
-            }
+            //AnalysisEntry entry;
```
huguesv: >AnalysisEntry [](start = 14, length = 13)

TODO?

Python/Tests/Utilities.Python/PythonToolsTestUtilities.cs
```diff
@@ -104,9 +106,29 @@ public static class PythonToolsTestUtilities {
...
-        private static PythonEditorServices CreatePythonEditorServices(IServiceContainer site) {
+        class LazyComponentGetter<T> : Lazy<T> {
+            public LazyComponentGetter(MockComponentModel model) : base(() => (T)model.GetService(typeof(T))) { }
+        }
+
+        private static PythonEditorServices CreatePythonEditorServices(IServiceContainer site, MockComponentModel model) {
             var services = new PythonEditorServices(site);
-            services.ComponentModel.DefaultCompositionService.SatisfyImportsOnce(services);
+
+            //services.ComponentModel.DefaultCompositionService.SatisfyImportsOnce(services);
```
huguesv: >services [](start = 14, length = 8)

delete?

huguesv:
Nothing critical left, 👍 from me

Python/Tests/Utilities.Python/PythonToolsTestUtilities.cs
zooba: I kind of like that it shows what we're emulating... maybe I'll add that to the comment too.

zooba:
@huguesv One more change that is worth a second look - there were codepaths in AnalysisEntryService that would incorrectly return null when they shouldn't. I had to reshuffle some code and duplicate some of it.

This broke XAML support completely. It still seems to be broken after this part is fixed, but it's less related to the current set of changes, and importantly won't affect your work on Peek. So I want to get this much in and then look at XAML.

huguesv:
New changes look ok

PR #2946, opened 2017-08-26
Fixes #2943 XAML has no Intellisense

### LINKED ISSUE #2943: XAML has no Intellisense
Analysis of XAML files broke. Need to fix it

zooba:
Fixes #2943 XAML has no Intellisense
Supports assembly references without correct hint paths
Avoids creating too many analyzers at project load
Ensures references are transferred between analyzers


zooba:
This is not quite finished because, despite the title, there's still no XAML Intellisense. But it seems to be for some weird editor reason, rather than because we are doing something wrong.

The situation is now better enough than our current release that I'd take this before the next build, though it would be nice to also figure out what magical incantation is needed to satisfy the WPF tools.

huguesv:
👍

zooba:
Turns out the WPF team can't properly reproduce the XAML editor issue without this fix, so we'll take it now and come back to the main one.

PR #2974, opened 2017-08-30
Disables task provider completely for mock tests that fail when it is enabled

zooba:


huguesv:
Did you need to also do `VsProjectAnalyzer.SuppressTaskProvider = true;` like in RefactorRenameTests

PR #2979, opened 2017-08-31
Prevent duplicate results in peek definition

### LINKED ISSUE #2973: Peek definition shows duplicate entries
If there's more than one definition, sometimes we get duplicate entries.

```
def my_func(param1, param2 = True):
    print(param1)
    print(param2)

my_func(1)
my_func(2, param2=False)
```

image

huguesv:
Fix #2973 Peek definition shows duplicate entries

I ran all of AnalysisTest test class and there's no regression there.

In the navigable tests, `ParameterDefinition` test was a repro for this bug. The tests now validate all results instead of just first one, so without the fix, that test would now fail.


zooba:
👍

PR #2987, opened 2017-09-01
Convert ML, Kinect and Pyvot VS templates to cookiecutter templates.

### LINKED ISSUE #2788: Move sample templates to cookiecutter
We should take our sample templates and move them to cookiecutter:
* ML templates
* PyVot
* PyKinect

Where appropriate, we should have a VS template that opens the cookiecutter window, either with a search term to find the template or straight to the creation screen (the aim of this is to make people aware of cookiecutter - so I'd prefer to make the search functionality more obvious).

huguesv:
Fix #2788 Move sample templates to cookiecutter

Convert ML, Kinect and Pyvot VS templates to cookiecutter templates.

Add some new functionality to Cookiecutter:
- Control visibility of fields
- Initialize fields with current VS state (value_source)
- Automatically open generated files (open folder, open project, add to solution, add to project) instead of showing a link to open in solution explorer back in the first page

Add a wizard to Python Tools that invokes Cookiecutter via a dte command with parameters. It checks first for the presence of Cookiecutter component, since it's possible to install Python workload without it.

The wizard opens Cookiecutter explorer, lets it perform its default initialization (with no search terms) and then selects the template from the results. If the results do not contain the desired template, then it adds it to a 'Custom' section at the top (preserving all other results). I've done this so we don't have to put pyvot and pykinect templates in the recommended feed.

Converted templates are here, fwlinks are used in the .vstemplate:
https://github.com/Microsoft/python-pykinect-pygame-cookiecutter
https://github.com/Microsoft/python-pyvot-cookiecutter
https://github.com/Microsoft/python-sklearn-classifier-cookiecutter
https://github.com/Microsoft/python-sklearn-clustering-cookiecutter
https://github.com/Microsoft/python-sklearn-regression-cookiecutter

I haven't done the UI improvement around highlighting that they need to hit Next to continue in Cookiecutter Explorer. I had trouble trying to do this with tooltips, and when I eventually get something working it should be relatively small compare to the rest of these changes.


zooba:
👍

PR #2995, opened 2017-09-01
Fixes #16 Tests with decorators do not appear in the test explorer view

### LINKED ISSUE #16: Tests with decorators do not appear in the test explorer view
This is the same as issue # 1969 on the CodePlex issue tracker. https://pytools.codeplex.com/workitem/1969

It has been marked as resolved there, but I still see the problem with the latest stable release 2.1 and the latest dev build April 13th.

If a Python unittest has a decorator, it does not appear in the test explorer. If the decorator is removed, it then appears.

``` python
    @patch.object(os.path, 'isfile')
    def test_A(self):
        ''' Doesn't appear in the test explorer '''
        self.fail("Not implemented")

    def test_B(self):
        ''' This does appear in the test explorer '''
        self.fail("Not implemented")
```

I'm using the mock library, so have hundreds of tests with patch decorators.

zooba:
Fixes #16 Tests with decorators do not appear in the test explorer view
Enhances the code path that extracts test names using analysis to also use the current AST.

### DIFF
--- a/Python/Product/TestAdapter.Analysis/TestAnalysisExtension.cs
+++ b/Python/Product/TestAdapter.Analysis/TestAnalysisExtension.cs
@@ -23,6 +23,7 @@ using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
 using Microsoft.PythonTools.Interpreter.Ast;
+using Microsoft.PythonTools.Parsing.Ast;
 using Microsoft.PythonTools.Projects;
 
 namespace Microsoft.PythonTools.TestAdapter {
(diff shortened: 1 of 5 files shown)


Python/Product/TestAdapter.Analysis/TestAnalysisExtension.cs
```diff
@@ -162,15 +161,30 @@ public class Serialize {
...
-                .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method));
+                .Where(v => v.Value.Any(m => m.MemberType == PythonMemberType.Function || m.MemberType == PythonMemberType.Method))
+                .Select(v => new KeyValuePair<string, LocationInfo>(v.Key, v.Value.SelectMany(av => av.Locations).FirstOrDefault(l => l != null)));
+
+            var analysisTests = methodFunctions.Where(v => v.Key.StartsWith("test"));
+            var analysisRunTest = methodFunctions.Where(v => v.Key.Equals("runTest"));
 
-            var tests = methodFunctions.Where(v => v.Key.StartsWith("test"));
-            var runTest = methodFunctions.Where(v => v.Key.Equals("runTest"));
+            tests = tests?.Union(analysisTests) ?? analysisTests;
```
huguesv: This will produce duplicates in some cases. Like when running TestDiscovererTests.DecoratedTests, which now fails because the expected sequence of 1 element now has 2 elements.

huguesv:
Needs an additional test that ensures that the scenario this is trying to fix will not regress.

Python/Product/TestAdapter.Analysis/TestAnalysisExtension.cs
zooba: Ah. Worked fine before I had to change it all to key value pairs here :)

zooba:
When merged we should remember to update https://developercommunity.visualstudio.com/content/problem/92871/tests-with-decorators-do-not-appear-in-the-test-wi.html

zooba:
@huguesv Minor updates that should address your request.

huguesv:
👍

PR #3001, opened 2017-09-05
Set work directory on python startup

amos402:
1. Some custom python core may load file with relative path before enter to debugger script entry, set the work directory to make them pass.
2. Change a test data srouce file's encoding to utf-8. It seems that the file is latin-1 encoding. For example my default enviroment is gbk it would raise a compile error.


huguesv:
Thanks! That looks okay to me.
@zooba will have a look when he's back next week.

PR #3005, opened 2017-09-06
Fix #3000, #3004

### LINKED ISSUE #3000: C++ nodes not displaying
Getting an error message instead of actual values (building/running CPython from master plus upgrading the projects to build with v141).

image

### LINKED ISSUE #3004: MMD: PyDictKeyEntry not found on 2.7
Mixed-mode debugging is broken on Python 2.7, because it tries to load missing type `PyDictKeyEntry` (it exists in 2.7, but is called `PyDictEntry`).

int19h:
Fix #3000: C++ nodes not displaying

Use module-qualified name for PyObject when determining the object type for [C++ view] node.

Fix #3004: MMD: PyDictKeyEntry not found on 2.7

Re-add StructProxy attributes to look up PyDictEntry instead on 2.7.


zooba:
👍

PR #3019, opened 2017-09-12
Fix #3017, #3018

### LINKED ISSUE #3017: MMD: debugger helper built with /DEBUG:FASTLINK
Getting this error message when trying to use MMD with a fresh CI build:

image

Looks like some VC++ defaults got changed? Probably just need to disable that flag explicitly.

### LINKED ISSUE #3018: MMD: PythonDkm.natvis is not registered correctly
.vsixmanifest says:
```xml
    <Asset Type="NativeVisualizer" Path="PythonDkm.natvis" />
```
But actual path is `Visualizers\PythonDkm.natvis`. This causes it to not be loaded, disabling [Python View] functionality.

int19h:
Fix #3017: MMD: debugger helper built with /DEBUG:FASTLINK

Fix #3018: MMD: PythonDkm.natvis is not registered correctly

Also fixes PythonDeveloper.reg.


Python/Product/PyDebugAttach/PyDebugAttach.vcxproj
```diff
@@ -75,6 +75,8 @@
       <SubSystem>Console</SubSystem>
       <OutputFile>$(TargetPath)</OutputFile>
       <AdditionalDependencies>ws2_32.lib;psapi.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <GenerateDebugInformation Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">DebugFull</GenerateDebugInformation>
+      <FullProgramDatabaseFile Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</FullProgramDatabaseFile>
```
zooba: Any reason to have the conditions in here? We always want these settings, right?

zooba:
👍 when/if you decide to remove the vcxproj conditions.

Python/Product/PyDebugAttach/PyDebugAttach.vcxproj
int19h: No particular reason, it's just the way VS made it when using project properties.

It actually occurs to me that this project doesn't need it, while the one that does (`DebuggerHelper`), I forgot to stage.

PR #3020, opened 2017-09-12
Fixes tabstop on control that should not be tabbable.

zooba:


zooba:
Reported internally by accessibility testers.

huguesv:
👍

PR #3022, opened 2017-09-13
Fixes #3010 Crash collecting output from pip

### LINKED ISSUE #3010: Crash collecting output from pip
We have seen this crash occurring:

```
mscorlib.ni!System.Text.StringBuilder.ExpandByABlock in stringbuilder.cs
at mscorlib.ni!System.Text.StringBuilder.Append in stringbuilder.cs
at mscorlib.ni!System.Text.StringBuilder.AppendHelper in stringbuilder.cs
at mscorlib.ni!System.Text.StringBuilder.Append in stringbuilder.cs
at mscorlib.ni!System.String.Join in string.cs
at Microsoft.PythonTools.VSInterpreters!Microsoft.PythonTools.Interpreter.PipPackageManager
at mscorlib.ni!System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess in TaskAwaiter.cs
at mscorlib.ni!System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification in TaskAwaiter.cs
at Microsoft.PythonTools.VSInterpreters!Microsoft.PythonTools.Interpreter.PipPackageManager
at mscorlib.ni!System.Runtime.CompilerServices.AsyncMethodBuilderCore in AsyncMethodBuilder.cs
at mscorlib.ni!System.Threading.QueueUserWorkItemCallback.WaitCallback_Context in threadpool.cs
at mscorlib.ni!System.Threading.ExecutionContext.RunInternal in executioncontext.cs
at mscorlib.ni!System.Threading.ExecutionContext.Run in executioncontext.cs
at mscorlib.ni!System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem in threadpool.cs
at mscorlib.ni!System.Threading.ThreadPoolWorkQueue.Dispatch in threadpool.cs
at mscorlib.ni!System.Threading._ThreadPoolWaitCallback.PerformWaitCallback in threadpool.cs
???
KERNELBASE!RaiseException in xcpt.c
at clr!RaiseTheExceptionInternalOnly in excep.cpp
at clr!UnwindAndContinueRethrowHelperAfterCatch in excep.cpp
at clr!JIT_New in jithelpers.cpp
at Microsoft.PythonTools.Analysis!Microsoft.PythonTools.Analysis.Analyzer.AnalyzerStatusUpdaterImplementation.ReadAllStatuses in AnalyzerStatusUpdater.cs
at Microsoft.PythonTools.Analysis!Microsoft.PythonTools.Analysis.Analyzer.AnalyzerStatusUpdaterImplementation.ThreadProc in AnalyzerStatusUpdater.cs
at mscorlib.ni!System.Threading.ThreadHelper.ThreadStart_Context in thread.cs
at mscorlib.ni!System.Threading.ExecutionContext.RunInternal in executioncontext.cs
at mscorlib.ni!System.Threading.ExecutionContext.Run in executioncontext.cs
at mscorlib.ni!System.Threading.ExecutionContext.Run in executioncontext.cs
at mscorlib.ni!System.Threading.ThreadHelper.ThreadStart in thread.cs
at clr!CallDescrWorkerInternal in asmhelpers.asm
at clr!CallDescrWorkerWithHandler in callhelpers.cpp
at clr!MethodDescCallSite::CallTargetWorker in callhelpers.cpp
at clr!ThreadNative::KickOffThread_Worker in comsynchronizable.cpp
at clr!ManagedThreadBase_DispatchInner in threads.cpp
at clr!ManagedThreadBase_DispatchMiddle in threads.cpp
at clr!ManagedThreadBase_DispatchOuter in threads.cpp
at clr!ManagedThreadBase_FullTransitionWithAD in threads.cpp
at clr!ThreadNative::KickOffThread in comsynchronizable.cpp
at clr!Thread::intermediateThreadProc in threads.cpp
at kernel32!BaseThreadInitThunk in thread.c
at ntdll!__RtlUserThreadStart in rtlstrt.c
at ntdll!_RtlUserThreadStart in rtlstrt.c
```

Hard to tell exactly what the cause is, but a few places in `PipPackageManager` use `string.Join` to collect many lines of output into a single string. It's entirely possible that for very large output this could get us into an OOM. Might be worth looking at processing the output by lines rather than joining it.

zooba:
Fixes #3010 Crash collecting output from pip
Uses a reader subclass to avoid concatenating large JSON files in memory.

Tests are still coming, but I was so pleased that this worked first time I figured I'd get the PR out so someone can tell me what I missed :)

### DIFF
--- a/Python/Product/VSInterpreters/PipPackageManager.cs
+++ b/Python/Product/VSInterpreters/PipPackageManager.cs
@@ -24,6 +24,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Infrastructure;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.PythonTools.Interpreter {
@@ -458,17 +459,18 @@ namespace Microsoft.PythonTools.Interpreter {
                         try {
                             if ((await proc) == 0) {
                                 if (_pipListHasFormatOption) {
-                                    var json = string.Join(Environment.NewLine, proc.StandardOutputLines);
                                     try {
-                                        var data = JArray.Parse(json);
+                                        var data = JToken.ReadFrom(new JsonTextReader(new StringListReader(proc.StandardOutputLines)));
                                         packages = data
                                             .Select(j => new PackageSpec(j.Value<string>("name"), j.Value<string>("version")))
                                             .Where(p => p.IsValid)
                                             .OrderBy(p => p.Name)
                                             .ToList();
-                                    } catch (Newtonsoft.Json.JsonException ex) {
+                                    } catch (JsonException ex) {
                                         Debug.WriteLine("Failed to parse: {0}".FormatInvariant(ex.Message));
-                                        Debug.WriteLine(json);
+                                        foreach (var l in proc.StandardOutputLines) {
+                                            Debug.WriteLine(l);
+                                        }
                                     }
                                 } else {
                                     packages = proc.StandardOutputLines
@@ -488,7 +490,9 @@ namespace Microsoft.PythonTools.Interpreter {
                         } catch (OperationCanceledException) {
                             // Process failed to run
                             Debug.WriteLine("Failed to run pip to collect packages");
-                            Debug.WriteLine(string.Join(Environment.NewLine, proc.StandardOutputLines));
+                            foreach (var line in proc.StandardOutputLines) {
+                                Debug.WriteLine(line);
+                            }
                         }
                     }
 
--- a/Python/Product/VSInterpreters/StringListReader.cs
+++ b/Python/Product/VSInterpreters/StringListReader.cs
@@ -60,16 +60,11 @@ namespace Microsoft.PythonTools.Interpreter {
         }
 
         public override string ReadLine() {
-            if (_peekBuffer == 0) {
-                _peekBuffer = -1;
-                return string.Empty;
-            }
-
             var r = _current?.ReadLine();
             if (r == null && Next()) {
                 r = _current.ReadLine();
             }
-            if (_peekBuffer > 0) {
+            if (_peekBuffer >= 0) {
                 r = $"{(char)_peekBuffer}{r ?? ""}";
                 _peekBuffer = -1;
             }
--- a/Python/Product/VSInterpreters/StringListReader.cs
+++ b/Python/Product/VSInterpreters/StringListReader.cs
@@ -120,7 +120,17 @@ namespace Microsoft.PythonTools.Interpreter {
         }
 
         public override int ReadBlock(char[] buffer, int index, int count) {
-            return Read(buffer, index, count);
+            int totalCount = 0;
+            while (totalCount < count) {
+                int newCount = Read(buffer, index, count);
+                if (newCount == 0) {
+                    break;
+                }
+                index += newCount;
+                count -= newCount;
+                totalCount += newCount;
+            }
+            return totalCount;
         }
 
         public override string ReadToEnd() {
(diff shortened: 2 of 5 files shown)


Python/Product/VSInterpreters/PipPackageManager.cs
```diff
@@ -458,17 +459,18 @@ CancellationToken cancellationToken
                         try {
                             if ((await proc) == 0) {
                                 if (_pipListHasFormatOption) {
-                                    var json = string.Join(Environment.NewLine, proc.StandardOutputLines);
                                     try {
-                                        var data = JArray.Parse(json);
+                                        var data = JToken.ReadFrom(new JsonTextReader(new StringListReader(proc.StandardOutputLines)));
```
int19h: I wonder if it would be better to change `ProcessOutput` to store the data as a single chunk in the first place (e.g. as `MemoryStream`? so that it can be fed directly to anything that expects a stream or a reader), and reimplement `StandardOutputLines` as lazy iterator over that.

zooba: I think the storing it as the single chunk is the problem. A list of strings can be relocated by the GC, while a chunk approaching GB is going to be hard to allocate (I don't have sufficient info from the crash data, but my guess is that it's a failed large allocation rather than a leak, since it was in the `string.Join`).

If it's still a problem here, we can probably create an adapter between the `Redirector` and `JsonTextReader`, but since it's a push vs. pull model adapter it's going to be much more complicated.

huguesv:
`pip list --format=json` for anaconda 3 (with 157 packages) gives me output that 6640 characters. That's not much, you would need a lot of packages for this to be significant.

Something else had to eat up all the memory and this just happened to be the thing that made it go OOM.

Or maybe there's a bug that causes this code to be executed continuously, like when we were reading code snippets on idle.

zooba:
I agree, it's likely the users were all nearly out of memory already. But since we're being asked to fix the crash by the VS team, and it's easy enough to avoid the triggering allocation, I see no reason not to do it.

Even if the code is running continuously, there's nothing that keeps the memory around that we can derive from the info. We're fixing the only possible bug that exists here.

Python/Product/VSInterpreters/StringListReader.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+            return i;
+        }
+
+        public override int Read(char[] buffer, int index, int count) {
+            if (_current == null) {
+                return 0;
+            }
+            var i = _current.Read(buffer, index, count);
+            if (i == 0 && Next()) {
```
huguesv: If the requested count requires reading from 2 lines, this won't do it. I've seen some stream implementations behaving this way (not filling the requested count, even though there was more data available), and that was unexpected.
Remarks at https://msdn.microsoft.com/en-us/library/97e90x91(v=vs.110).aspx seem to imply that it should always fill the requested count, except when reached EOF.

zooba: I've always understood it to mean "read again", but it looks like `ReadBlock` is intended to have the semantics you describe. I'll update that one

Python/Product/VSInterpreters/StringListReader.cs
```diff
@@ -0,0 +1,155 @@
...
+        public override int Peek() {
+            if (_peekBuffer >= 0) {
+                return _peekBuffer;
+            }
+            _peekBuffer = Read();
+            return _peekBuffer;
+        }
+
+        public override string ReadLine() {
+            if (_peekBuffer == 0) {
```
int19h: What's the special significance of 0? I can't see any place that sets `_peekBuffer` to that value.

zooba: If we read a null byte. Things are going to get weird if we prefix a string with `\0`

zooba: Actually, now I think about it, that doesn't matter. I spent too long in C last week...

PR #3030, opened 2017-09-13
Detects open dialogs at the start of a test and closes them.

zooba:
Detects open dialogs at the start of a test and closes them.
This should prevent one failed test from causing many subsequent failures.


huguesv:
👍

PR #3032, opened 2017-09-14
Fix stdout capture for test tearDown under failure scenario for Python <= 3.1

### LINKED ISSUE #2929: Tests do not setUp/tearDown correctly
(Reported at https://developercommunity.visualstudio.com/content/problem/43362/test-explorer-again.html)

Apparently our test runner does not correctly set up and tear down test classes properly. Need to investigate further.

huguesv:
Fix #2929 Tests do not setUp/tearDown correctly

As mentioned in the bug, setUp and tearDown were done correctly, but the wrong output led you to believe that it wasn't.


Python/Tests/TestAdapterTests/TestExecutorTests.cs
```diff
@@ -486,6 +486,32 @@ class StackFrameComparer : IComparer {
...
+            PrintTestResults(recorder);
+
+            var resultNames = recorder.Results.Select(tr => tr.TestCase.FullyQualifiedName).ToSet();
+            foreach (var expectedResult in expectedTests) {
+                AssertUtil.ContainsAtLeast(resultNames, expectedResult.TestCase.FullyQualifiedName);
+                var actualResult = recorder.Results.SingleOrDefault(tr => tr.TestCase.FullyQualifiedName == expectedResult.TestCase.FullyQualifiedName);
+                Assert.AreEqual(expectedResult.Outcome, actualResult.Outcome, expectedResult.TestCase.FullyQualifiedName + " had incorrect result");
+                var stdOut = actualResult.Messages.Single(m => m.Category == "StdOutMsgs");
+                Assert.IsTrue(stdOut.Text.Contains("doing setUp"), "setUp was not called");
+                Assert.IsTrue(stdOut.Text.Contains("doing tearDown"), "tearDown was not called");
```
zooba: There should be an AssertUtil function to do string matching, which has the advantage of displaying the original string if it doesn't match. Alternatively, put `stdOut.Text` into the message.

zooba:
👍  with whatever change you decide for the test message.

PR #3035, opened 2017-09-14
Disposes of the StringReader to avoid static analysis complaints.

zooba:


zooba:
Build passed, so merging.

PR #3036, opened 2017-09-14
Fix #2201: Web projects are always rebuilt

### LINKED ISSUE #2201: Web projects are always rebuilt
If the project is up to date, we should not rebuild. However, we always rebuild (at least Django and Bottle) web projects.

int19h:
Exclude cloud service files for app service projects.


int19h:
Also verified no regression on Cloud Service deployments with this change.

zooba:
👍 🎉

PR #3039, opened 2017-09-15
Remove Python 2.5 and 3.0 related tests.

huguesv:
Don't waste time running tests on Python 2.5 and 3.0 which we've decided not to support anymore.
Some of them were previously failing (Python 3.0), so removing them helps our test results.


zooba:
If you remove `PythonPaths.Python30` then it should be fairly obvious which other tests are using this version.

huguesv:
I updated the PR to remove the few Python 2.5 tests as well.

zooba:
👍

PR #3043, opened 2017-09-18
Fix #2031: Cannot import subpackages from zip file

### LINKED ISSUE #2031: Cannot import subpackages from zip file
Add a zip file with multiple level of subpackages to your search path.

Try and import the subpackage (`import a.b`). It is unresolved and there are no completions.

int19h:
Fix missing terminating slash when looking up directory entries inside .zip files


zooba:
Oh man, so simple 🤦‍♂️ 

👍

PR #3044, opened 2017-09-18
Extension module template fixes

zooba:
Fixes failure to create extension module template when Python is not installed
Also enables Python debugging by default


huguesv:
👍

PR #3047, opened 2017-09-18
Adds UI option for experimental DB flag.

zooba:
Adds UI option for experimental DB flag.
Updates Options category localized name to just "Python"

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Options/ExperimentalOptions.cs
+++ b/Python/Product/PythonTools/PythonTools/Options/ExperimentalOptions.cs
@@ -1,0 +1,54 @@
+// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using EO = Microsoft.PythonTools.Interpreter.ExperimentalOptions;
+
+namespace Microsoft.PythonTools.Options {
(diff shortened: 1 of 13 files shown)


Python/Product/PythonTools/PythonTools/Options/ExperimentalOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using EO = Microsoft.PythonTools.Interpreter.ExperimentalOptions;
+
+namespace Microsoft.PythonTools.Options {
+    public sealed class ExperimentalOptions {
+        internal ExperimentalOptions(PythonToolsService service) {
```
huguesv: `service` isn't used, did you leave it for possible use by later options?

Python/Product/PythonTools/PythonTools/Options/ExperimentalOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <summary>
+        /// True to use the experimental feature of interpreter factories
+        /// without old-style completion databases.
+        /// </summary>
+        public bool NoDatabaseFactory { get; set; }
+
+        /// <summary>
+        /// True if the current setting is the active setting.
+        /// </summary>
+        public bool IsNoDatabaseFactoryActive => NoDatabaseFactory == EO.NoDatabaseFactory;
```
huguesv: This isn't used?

Python/Product/PythonTools/PythonTools/Options/ExperimentalOptions.cs
zooba: Yeah, and consistency with the other options classes. We only have one experimental option right now and it is system-wide, but future ones could be per-instance, in which case we should use the normal options infrastructure to store the setting.

Python/Product/VSInterpreters/ExperimentalOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using Microsoft.Win32;
+
+namespace Microsoft.PythonTools.Interpreter {
+    public sealed class ExperimentalOptions {
```
huguesv: I think you can make the class static

Python/Product/PythonTools/PythonTools/Options/ExperimentalOptions.cs
zooba: I was considering making the "you need to restart VS" text in the options page conditioned on this state, but decided against it. There's probably no need for this.

Python/Product/VSInterpreters/ExperimentalOptions.cs
zooba: 👍

huguesv:
👍

PR #3048, opened 2017-09-19
Fixes manifest build to ensure hash information is correctly created.

zooba:


zooba:
This will also need to be backported to 3.2 branch.

PR #3050, opened 2017-09-19
Fixes #1575 Rename folder failed to "reparent" folder

### LINKED ISSUE #1575: Rename folder failed to "reparent" folder
Needs further investigation - not sure exactly why this happened.

Basically, I had a folder with some Python files in it and the files were open, then I renamed the folder. Nothing got deleted, but I was unable to see the old or new folders until I closed and reopened the project (even with Show All Files on).

zooba:
Fixes #1575 Rename folder failed to "reparent" folder
Fixes most of the renaming issues.

### DIFF
--- a/Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
+++ b/Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
@@ -1906,7 +1906,7 @@ namespace Microsoft.PythonTools.Intellisense {
             if (Path.GetExtension(path).Equals(".xaml", StringComparison.OrdinalIgnoreCase)) {
                 item = _pyAnalyzer.AddXamlFile(path, null);
 
-            } else if (ModulePath.IsPythonSourceFile(path)) {
+            } else {
                 // assume it's a Python file...
                 string modName;
                 try {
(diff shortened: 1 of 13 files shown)


zooba:
This is not ready to merge yet - I need to get tests running so I can check them, but right now I'm fighting against implicit assembly references failing to resolve.

zooba:
I can't get the shared project tests running at all on my local box, I think they're going to need a restructure to avoid the mixed assembly reference issues I'm hitting constantly. Going to set up a test run to validate this instead, and file a new issue to refactor tests.

Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
```diff
@@ -1906,7 +1906,7 @@ class TextCodeInfo : CodeInfo {
             if (Path.GetExtension(path).Equals(".xaml", StringComparison.OrdinalIgnoreCase)) {
                 item = _pyAnalyzer.AddXamlFile(path, null);
 
-            } else if (ModulePath.IsPythonSourceFile(path)) {
```
huguesv: It's ok to do this for non-python files now?

zooba: Yeah, we'll just treat it like a source file. Basically, we special-case .XAML, and if you tell us to analyze another file type then we'll assume it's source code.

zooba: The trick is going to be in not adding all files to the analyzer, but that'll be on the VS side. This way we can tie it entirely to other options, such as the build action being "Compile".

huguesv: Ok, for now as long as it doesn't crash or cause major issues with text files, etc.

huguesv:
👍

PR #3055, opened 2017-09-20
Remove tests

huguesv:
Remove Azure publishing tests and its automation wrappers which were out of date not long after they were written. If we ever want to automate publishing tests again, we shouldn't do it via UI.

Remove failing LoadFlavoredProject test, which was equivalent to existing LoadWebFlavoredProject passing test (not enough value maintaining both pyproj test files).


zooba:
👍

Issue #3058, opened 2017-09-21
Refactor UI tests


zooba:
The UI tests are a ridiculous mess of assembly references, manual copying, etc.

We should refactor like this:
* Create test VSIXs that can be installed/deployed to Exp
 * this will enable normal assembly resolution, so we don't run into problems with different assemblies being loaded under the same name (and spoiling MEF composition)
* Create and call into the VS instance using DTE
 * now our test assemblies do not need any PTVS references

PR #3059, opened 2017-09-21
UI test refactoring

zooba:
See #3058 

This is a prototype right now for early review. In particular the `UnitTest1.cs` file is representative of what our test modules will end up containing, while the tests become functions embedded into a VSIX.


zooba:
Still to do:

- [ ] Infer/find VS install
- [ ] Remove extra deployment steps (`PythonTestData.Deploy()`)

zooba:
@huguesv If you'd like to get started migrating tests, feel free to merge this - I won't be touching it this morning. Also, the tests that generate projects currently do not work, and hopefully I'll finish those off later today.

My plan is to have separate test runner DLLs for each test DLL, rather than putting them all in the one assembly, but we can take the opportunity to refactor the UI tests more significantly. Right now my main aim is to just get them all running again.

zooba:
@huguesv Actually, let me know before we merge so I can update the test machine configuration.

huguesv:
I've reviewed most of it to be familiar with the changes. Let's merge. 👍

PR #3068, opened 2017-09-26
Updates ProjectGenerator to work in new test model

zooba:
Migrates snippets tests
Adds ability to override project type when statically constructing project definitions.


huguesv:
👍

PR #3070, opened 2017-09-26
Fixes #3067 "Reports" in profiling is not localized

### LINKED ISSUE #3067: "Reports" in profiling is not localized
The name of the "Reports" node in the Profiling explorer window is not localizable.

zooba:
Fixes #3067 "Reports" in profiling is not localized
Moves string literal into resource.

Ideally should be ported to 3.2.1, but we need to wait for the loc team to merge their translations first.


huguesv:
👍

PR #3074, opened 2017-09-27
Make the no DB experiment enabled by default.

zooba:
Make the no DB experiment enabled by default.
Adds UI informing the user and allowing them to disable it.

### DIFF
--- a/Common/Product/SharedProject/Wpf/Controls.xaml
+++ b/Common/Product/SharedProject/Wpf/Controls.xaml
@@ -335,7 +335,8 @@
                 <ControlTemplate TargetType="{x:Type Button}">
                     <ContentPresenter TextElement.Foreground="{TemplateBinding Foreground}" RecognizesAccessKey="True" Height="Auto">
                         <ContentPresenter.Resources>
-                            <Style TargetType="{x:Type TextBlock}">
+                            <Style TargetType="{x:Type TextBlock}" BasedOn="{StaticResource {x:Type TextBlock}}">
+                                <Setter Property="Foreground" Value="{Binding Foreground,RelativeSource={RelativeSource AncestorType=Button}}" />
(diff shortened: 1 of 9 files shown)


Common/Product/SharedProject/Wpf/Controls.xaml
```diff
@@ -335,7 +335,8 @@
                 <ControlTemplate TargetType="{x:Type Button}">
                     <ContentPresenter TextElement.Foreground="{TemplateBinding Foreground}" RecognizesAccessKey="True" Height="Auto">
                         <ContentPresenter.Resources>
```
huguesv: Did you test with all themes? Also, there's a few copy of this file that probably should stay in sync.

zooba: I'll do both of those. Shouldn't affect the themes as it's just making explicit the previously implicit values (which were broken by injecting more styles into the mix...)

Python/Product/EnvironmentsList/Resources.resx
```diff
@@ -429,4 +429,18 @@
   <data name="ToolWindow_ListOfEnvironmentsName" xml:space="preserve">
     <value>Python environments</value>
   </data>
+  <data name="NoDBExtensionDisplayName" xml:space="preserve">
+    <value>IntelliSense [database disabled]</value>
```
huguesv: Since this doesn't change after the user re-enables databases, is there a fixed text that we could use that doesn't appear to be wrong? Not a big deal for a temporary UI though.

zooba: The databases don't reenable immediately anyway. When they restart VS, this item won't be added to the list and the regular "IntelliSense" one will be there.

huguesv:
👍

PR #3077, opened 2017-09-27
Debugger UI tests

huguesv:
Add runners for debugger and attach UI tests.

Add a few helpers to set/reset various options (more can be added as needed).

2 consistent failures in attach tests. They try to update a variable in running code to trigger program exit, which does nothing. Unrelated to the UI tests port.

Running all debugger UI tests results in some failures, but they usually all pass when you run the failed tests again.


zooba:
👍

PR #3089, opened 2017-09-28
Port Django UI tests

huguesv:
All editing tests fail (technically some pass but it's just luck) due to not getting any django classifiers. Looking at the priority that was set on those tests, this matches the results prior to UI test rework.

The debug project test fails due to not being able to locate DebuggerUITests.dll

Some of the project tests due to an assert "failed to analyze <guid>"

Upgraded the django projects under test data by copying from our empty template ("install python support" dialog was popping up on load).

Renamed the django test folders for consistency with their project names (and other folders).


zooba:
👍

PR #3091, opened 2017-09-28
Reduces JSON packet size by omitting default values for common properties

zooba:
Reduces JSON packet size by omitting default values for common properties
Batches test requests to avoid overwhelming analyzer
Improves handling of analysis set edge case

These all popped up while I was working with a large project from a customer.


huguesv:
👍

PR #3097, opened 2017-09-29
Adds retry with VS restart for invalid COM object.

zooba:


zooba:
If you look at our most recent test run, the majority of failures were due to the error that this works around (though I suspect many would have gone on toward failure anyway :) )

huguesv:
👍

PR #3100, opened 2017-09-29
Port REPL UI tests

huguesv:
Implementation of the tests has barely changed, but that's difficult to see in the PR. Pretty much just the addition of the interpreter parameter, and passing it down to the Prepare method.

Test implementation used to be a multitude of derived classes for the convenience of controlling what test to run under what interpreter, but now it's a single class (send tests are separate as they were different enough).

Some class inheritance in the runner in order to control which tests run under what interpreter ;)  but seriously it's not as confusing as before.

It's easier to tell now which tests are covered under multiple interpreters.

image


zooba:
👍

Issue #3046, opened 2017-09-18
IntelliSense doesn't understand library "X"


zooba:
There are a variety of issues relating to analysis not correctly understanding certain libraries. This issue is to collect those reports and spin them off into properly researched and understood tracker items.

New reports will be dup'd against this issue so we have a link to it from here, or can simply be posted here as comments.

- [x] `datetime`
- [x] `datetime.now().`
- [x] `os.path.`
- [x] `with open('', '') as f: f.`
- [x] `re.match` https://github.com/Microsoft/PTVS/issues/4312
- [x] `tensorflow` https://github.com/Microsoft/PTVS/issues/4315
- [x] `time` https://github.com/Microsoft/PTVS/issues/4318
- [x] `unittest` #4319 
- [ ] pandas #4428 #4590 #4001

zooba:
(From https://developercommunity.visualstudio.com/content/problem/119042/pythons-assert-isinstance-not-working-for-type-hin.html)

The `datetime` module in Python 3.6 has a native `_datetime` module but we incorrectly expose its class aliases as instances:

```
# _datetime
date = datetime.date()

# should be
date = datetime.date
```

jcageman:
`datetime.now()` 

gives currently no intellisense.  For example if i want to type datetime.now().year it won't expand. It does suggest datetime.now() when typing datetime.n

HHest:
I had a c:\python27 installation prior to installing VS 2017. After the installation, my IntelliSense could not help with F12 "Go to definition" of any method. There were also other symptoms to indicate that the IntelliSense DB is not working. However, no matter how many times I clicked the "Refresh DB" button, the problem remains.

Out of pure idleness, I decided to update one of my packages that was out of date (pytest to be precise). Lo and behold, IntelliSense started rebuilding all of its DB. After that, F12 started working again. Completions started working. The dropdown box showing methods/classes started working for some files.

Other IntelliSense related features still do not work, however: no class view, methods/classes dropdown box does not work for many files.

If you're having trouble with IntelliSense, try this possible workaround of **forcing a DB refresh via updating a package**. Hope it makes things more usable for you until there is a permanent fix.

huguesv:
The following reported at https://developercommunity.visualstudio.com/content/problem/232480/python-intellisense-improvements.html

Some functions (e.g. django.urls.path) aren't highlighted with the color assigned to a function whilst others are (e.g. django.urls.include).

rondelzer:
[...]

PR #3109, opened 2017-10-02
Fixes module scraping producing invalid code

zooba:
Fixes module scraping producing invalid code by doing repr on container contents
Updates tests to store scraped outputs.


zooba:
This also fixes the `from datetime import date` issue in #3046

huguesv:
👍

PR #3113, opened 2017-10-02
Add loading of dependent VS packages to UI test runner.

huguesv:
Django and Profiling UI tests were having issues loading dependent package assemblies.


zooba:
I'd just made the debugger test DLL included in the VSIX, but I like this approach better. 👍

PR #3115, opened 2017-10-02
Test priorities

### LINKED ISSUE #3122: Cannot round-trip `await await x`
We do not properly handle awaiting the result of an await:

```python
async def f(x):
    await await x
```

### LINKED ISSUE #3123: Cannot roundtrip sublist parameters with default values
Cannot handle this:

```python
def f((a, b) = (1, 2)):
    pass
```

### LINKED ISSUE #3124: Cannot roundtrip exec with both tuple and in
It's valid syntax, though you always get a TypeError from it:

```python
exec(f, g, h) in i, j
```

zooba:
Updates test priorities
Fixes some tests

Fixes #3122 Cannot round-trip `await await x`
Enables an await expression to be an await expression.

Fixes #3123 Cannot round-trip sublist parameters with default values
Adds formatting of default value for sublist parameters and fixes whitespace handling.

Fixes #3124 Cannot round-trip exec with both tuple and in
Enables retaining both tuple and in arguments so they can be round tripped.


Common/Tests/Utilities.UI/UI/AutomationDialog.cs
```diff
@@ -28,6 +30,7 @@ public AutomationDialog(VisualStudioApp app, AutomationElement element)
             : base(element) {
             App = app;
             DefaultTimeout = TimeSpan.FromSeconds(10.0);
+            _isTaskDialog = element.Current.ClassName == "#32770";
```
int19h: This looks like an atom. Is it stable?

zooba: Um, yeah, it's stable, but apparently it's going to be the same for other dialog boxes as well (we were relying on a GetDlgCtrl call succeeding to identify dialog boxes before, but that wasn't working with TaskDialog). But now I've refactored the code that relied on this flag, I can probably just make it more resilient and remove this.

Python/Tests/DjangoUITests/DjangoUITests.csproj
```diff
@@ -146,6 +146,7 @@
     <ProjectReference Include="..\DebuggerUITests\DebuggerUITests.csproj">
       <Project>{82BB3890-C943-421E-8643-7244421E8ADE}</Project>
       <Name>DebuggerUITests</Name>
+      <IncludeInVsix>true</IncludeInVsix>
```
huguesv: >true [](start = 21, length = 4)

you shouldn't need this one with my change.

zooba:
Still adjusting tests, and checked in fixes for some obviously core tests that should pass rather than be deprioritized.

zooba:
Getting >99% pass rate for our two main test sets now, so I'll merge. We should fix or deprioritize P0 tests aggressively for the time being.

PR #3116, opened 2017-10-02
Adds support for multi-value members in AST analysis.

zooba:


zooba:
Still need to do some specific unit tests, but the core functionality is there.

Python/Product/Analysis/Interpreter/Ast/NameLookupContext.cs
```diff
@@ -202,6 +203,23 @@ bool includeLocationInfo
             return null;
         }
 
+        public IEnumerable<IPythonType> GetTypesFromValue(IMember value) {
+            if (value is IPythonMultipleMembers mm) {
+                var seen = new HashSet<IPythonType>();
```
int19h: This looks like `Enumerable.Select(...).Distinct()`.

Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleMembers.cs
```diff
@@ -0,0 +1,68 @@
...
+                return new AstPythonMultipleMembers(new[] { x, y });
+            }
+        }
+
+        public void AddMember(IMember member) {
+            var old = Members;
+            if (!old.Contains(member)) {
+                Members = old.Concat(Enumerable.Repeat(member, 1)).ToArray();
+            } else {
+                Members = new[] { member };
```
huguesv: You are also removing any previous members here, which seem wrong.

zooba: Ah, you're right... made the simplification and didn't think about the changed else case...

zooba: See #3119

PR #3119, opened 2017-10-03
Fixes multiple assignments clearing existing members.

zooba:
Fixes multiple assignments clearing existing members.
Adds test


huguesv:
👍

PR #3121, opened 2017-10-03
Fix send to repl tests

### LINKED ISSUE #3114: Command "Python.SendSelectionToInteractive" is not valid.
`ReplWindowSendUITests` all fail due to this.

Tried to type it in the command window, get the same error and it doesn't appear in IntelliSense.

`Python.ExecuteInInteractive` which is defined right above in .vsct works fine.
`Python.FillCommentParagraph` which is defined right below in .vsct does not work either.

huguesv:
Change 2 command names back to use Python prefix instead of Edit.
No test was using FillCommentParagraph, but I figured it was best to change it anyway. I can take it or leave it.

fix #3114

### DIFF
--- a/Python/Product/PythonTools/PythonTools.vsct
+++ b/Python/Product/PythonTools/PythonTools.vsct
@@ -286,7 +286,7 @@ permissions and limitations under the License.
           <ButtonText>Send to &amp;Interactive</ButtonText>
           <ToolTipText>Sends the current editor selection to a Python Interactive window.</ToolTipText>
           <CanonicalName>.Python.Send Selection To Interactive</CanonicalName>
-          <LocCanonicalName>.Python.Send Selection To Interactive</LocCanonicalName>
+          <LocCanonicalName>Send Selection To Interactive</LocCanonicalName>
         </Strings>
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools.vsct
```diff
@@ -285,8 +285,8 @@ permissions and limitations under the License.
         <Strings>
           <ButtonText>Send to &amp;Interactive</ButtonText>
           <ToolTipText>Sends the current editor selection to a Python Interactive window.</ToolTipText>
-          <CanonicalName>Send Selection To Interactive</CanonicalName>
-          <LocCanonicalName>Send Selection To Interactive</LocCanonicalName>
+          <CanonicalName>.Python.Send Selection To Interactive</CanonicalName>
+          <LocCanonicalName>.Python.Send Selection To Interactive</LocCanonicalName>
```
zooba: Don't add the prefix on the localized name (and if you want, remove it from the one localized name it is already on - the rest look fine)

huguesv:
Ok, removed the prefix from the other one too.

PR #3133, opened 2017-10-05
Fixes non-UI tests

zooba:
Disables task provider in mock tests by default.
Makes MockServiceProvider properly handle lazy creation.

### DIFF
--- a/Common/Tests/Utilities/Mocks/MockServiceProvider.cs
+++ b/Common/Tests/Utilities/Mocks/MockServiceProvider.cs
@@ -22,6 +22,7 @@ using Microsoft.VisualStudio.ComponentModelHost;
 namespace TestUtilities.Mocks {
     public class MockServiceProvider : IServiceProvider, IServiceContainer {
         public readonly Dictionary<Guid, object> Services = new Dictionary<Guid, object>();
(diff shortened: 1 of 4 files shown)


zooba:
I haven't been able to test that the changes will fix the timeouts, since my builds are broken (probably due to a bad VSTS update). It looks like our main CI builds are fine, so merging will get us test results and if they are still bad then I'll demote the tests causing trouble.

Common/Tests/Utilities/Mocks/MockServiceProvider.cs
```diff
@@ -31,32 +32,45 @@ public class MockServiceProvider : IServiceProvider, IServiceContainer {
         public object GetService(Type serviceType) {
             object service;
             Console.WriteLine("MockServiceProvider.GetService({0})", serviceType.Name);
-            Services.TryGetValue(serviceType.GUID, out service);
-            return service;
+            if (Services.TryGetValue(serviceType.GUID, out service)) {
```
huguesv: do you need to handle concurrent access to GetService()?

zooba: Haven't needed to so far, and no evidence that we need it.

huguesv: ok

huguesv:
👍

PR #3135, opened 2017-10-05
Enable go to definition when caret is on the right of identifier

### LINKED ISSUE #2548: Can not goto definition
After select a class name by double click, press f12 have no respones.
Version: 3.0.17114.1
Like this:
default

huguesv:
Fix #2548 Can not goto definition

This final implementation is what I started with initially, then explored the idea of changing ReverseExpressionParser instead, but that wasn't working out.

The logic "check left and right of caret" is nicely expressed the way it is, so I feel alright about it.

I made it conditional because there are other use cases than go to definition that call `GetApplicableExpression` and I didn't want to potentially affect them.

There are no regressions in `CompletionTests` but they are as flaky as they were (run them enough times and they pass).

I did a bunch of manual testing around go to definition for identifiers around operators and such, as well as test editor/debugger tooltips and it all seems fine.


zooba:
👍

PR #3137, opened 2017-10-05
Test fixes

zooba:
Fixes ptvsd deployment, reduces some priorities and avoid out-of-proc analysis in a few more tests.


huguesv:
👍

PR #3140, opened 2017-10-06
Attempts to make WPF proxy-based tests less likely to time out.

zooba:
Attempts to make WPF proxy-based tests less likely to time out.
Protects against crashes due to corrupted registry keys
Skips some regularly failing modules when doing scrape test.


huguesv:
👍

PR #3142, opened 2017-10-06
Yet another attempt to fix test timeouts

zooba:
Yet another attempt to fix test timeouts due to an infinite loop on the finalizer thread.


huguesv:
👍

PR #3144, opened 2017-10-06
Reduce overhead of imports when using AST interpreter

zooba:
In short, does a better job of scraping and importing modules until we really need to get them.

I'm expecting some test failures, but the run is going and I'll deal with them next week.


Python/Tests/DebuggerTests/AttachTests.cs
```diff
@@ -777,6 +777,8 @@ void main()
         [TestMethod, Priority(0)]
         [TestCategory("10s")]
         public virtual async Task AttachWithOutputRedirection() {
+            Version.AssertInstalled();
```
huguesv: You should add this to CheckVersion() at the top of this class instead.

huguesv:
👍

Python/Tests/DebuggerTests/AttachTests.cs
zooba: Oddly enough, this didn't actually fix the problem... still looking into it.

zooba:
@huguesv Made a couple more changes you may want to look over.

huguesv:
👍

zooba:
There are still some test regressions due to this change, but the fixes are going well beyond what this PR should do. Still working on them, but going to merge this and keep the fixes coming.

PR #3146, opened 2017-10-09
Adds argparse snippets.

zooba:
I started writing a script (for another bug) and decided that I was sick of not having snippets for argparse. So here are some snippets for argparse :)

### DIFF
--- a/Python/Product/PythonTools/Snippets/Python/argparse.snippet
+++ b/Python/Product/PythonTools/Snippets/Python/argparse.snippet
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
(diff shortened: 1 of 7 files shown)


Python/Product/PythonTools/Snippets/Python/argparse.snippet
```diff
@@ -0,0 +1,39 @@
...
+          <ID>description</ID>
+          <ToolTip>Short description of your script</ToolTip>
+          <Default>My Python script.</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Python"><![CDATA[def parse_args(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(description="$description$")
+
+    # Use the arg_* snippets to add arguments
+    $end$
```
huguesv: What does $end$ do and why is it not at the end in this one?

zooba: When you don't support surrounding a selection, `$end$` is where the caret should be put when you commit the snippet.

huguesv:
👍

PR #3149, opened 2017-10-10
Prevent NullReferenceException when PYTHONPATH environment variable is not defined

### LINKED ISSUE #3148: NullReferenceException on start debugging when Ignore PYTHONPATH option is off
You get a message box and debugging doesn't start.

```
---------------------------
Microsoft Visual Studio
---------------------------
Object reference not set to an instance of an object.
---------------------------
OK   
---------------------------
```

huguesv:
Fix #3148


zooba:
👍

Issue #3099, opened 2017-09-29
Assert when starting VS with REPL visible


huguesv:
My VS REPL window(s) look like this at startup, and I get an assert while VS finishes initializing (no action on my part).

This interferes with running REPL UI tests because the assert dialog comes up during test.

![image](https://user-images.githubusercontent.com/1696845/31038406-8ae0da78-a52b-11e7-92d9-2c11084b25af.png)

```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
Failed to analyze file Temp.txt
   at Microsoft.PythonTools.Intellisense.IntellisenseController.<ConnectSubjectBufferAsync>d__23.MoveNext() in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\IntellisenseController.cs:line 200
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.InvokeMoveNext(Object stateMachine)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.Run()
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c__DisplayClass4_0.<OutputAsyncCausalityEvents>b__0()
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.ContinuationWrapper.Invoke()
   at System.Runtime.CompilerServices.TaskAwaiter.<>c__DisplayClass11_0.<OutputWaitEtwEvents>b__0()
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.ContinuationWrapper.Invoke()
   at System.Threading.Tasks.SynchronizationContextAwaitTaskContinuation.<>c__DisplayClass6_0.<GetActionLogDelegate>b__0()
   at System.Threading.Tasks.SynchronizationContextAwaitTaskContinuation.<>c.<.cctor>b__8_0(Object state)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
[...]

Issue #2701, opened 2017-06-21
NullReferenceException in CreateInteractiveWindowInternal in REPL UI tests


huguesv:
Not sure yet if this is a potential issue for user, or only for tests.

After we call `windowFactory2.Create`, we immediately try to use `replWindow.InteractiveWindow`, which is sometimes `null`.

```
Result StackTrace:	
at Microsoft.PythonTools.Repl.InteractiveWindowProvider.CreateInteractiveWindowInternal(IInteractiveEvaluator evaluator, IContentType contentType, Boolean alwaysCreate, Int32 id, String title, Guid languageServiceGuid, String replId) in C:\Projects\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Repl\InteractiveWindowProvider.cs:line 256
   at Microsoft.PythonTools.Repl.InteractiveWindowProvider.Create(String replId, Int32 curId) in C:\Projects\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Repl\InteractiveWindowProvider.cs:line 134
   at Microsoft.PythonTools.Repl.InteractiveWindowProvider.OpenOrCreate(String replId) in C:\Projects\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Repl\InteractiveWindowProvider.cs:line 122
   at TestUtilities.UI.Python.PythonReplWindowProxySettings.<>c__DisplayClass4_0.<ActivateInteractiveWindow>b__1() in C:\Projects\huguesv\PTVS\Python\Tests\Utilities.Python\PythonReplWindowProxySettings.cs:line 83
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Microsoft.VisualStudioTools.UIThread.Invoke[T](Func`1 func) in C:\Projects\huguesv\PTVS\Common\Product\SharedProject\UIThread.cs:line 86
[...]

PR #3150, opened 2017-10-10
Respect the clear PYTHONPATH option when creating the REPL launch configuration

### LINKED ISSUE #3014: Use PYTHONPATH option does not work for REPL
(Reported at https://developercommunity.visualstudio.com/content/problem/90795/ignore-python-path.html)

In `PythonReplEvaluatorProvider.GetEnvironmentEvaluator` we need to factor in the option to (not) clear global PYTHONPATH when creating the launch configuration.

Elsewhere we split and add it as search paths, so we should be consistent here.

huguesv:
Fix #3014

### DIFF
--- a/Python/Tests/ReplWindowUITestsRunner/ReplWindowAdvancedUITests.cs
+++ b/Python/Tests/ReplWindowUITestsRunner/ReplWindowAdvancedUITests.cs
@@ -471,6 +471,28 @@ namespace ReplWindowUITestsRunner {
         }
 
         #endregion
+
+        #region Advanced Launch Configuration Tests
+
+        [TestMethod, Priority(2)]
+        [TestCategory("Installed")]
+        public void PythonPathIgnored() {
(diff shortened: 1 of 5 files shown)


Python/Tests/ReplWindowUITestsRunner/ReplWindowAdvancedUITests.cs
```diff
@@ -471,6 +471,28 @@ public abstract class ReplWindowAdvancedUITests {
         }
 
         #endregion
+
+        #region Advanced Launch Configuration Tests
+
+        [TestMethod, Priority(2)]
```
zooba: Are these already flaky? New tests should be Priority(0), and if they aren't reliable then invest the time to make them reliable.

huguesv: Yes, all REPL UI tests are flaky due to #2701 and #3099.

zooba: Okay

PR #3151, opened 2017-10-10
Fixes #2756 Escape to cancel completions also clears REPL text

### LINKED ISSUE #2756: Escape to cancel completions also clears REPL text
If you have code typed in the REPL and completions pop up, which you then cancel by hitting Esc, the current REPL text is also cleared.

We should not be clearing the text in this case.

zooba:
Fixes #2756 Escape to cancel completions also clears REPL text
Intercept Esc command and do not forward if we just closed a session.


huguesv:
👍

PR #3156, opened 2017-10-11
Send to REPL when prior selection in REPL

### LINKED ISSUE #2454: Send to Interactive does not work with selection in REPL
Run some code in the REPL and select the output. Now go back to the editor and send a selection to the REPL.

Because the current insertion point is in a read-only section, the editor does not insert the text. We then execute an empty statement and either advance the caret or switch focus, but no text has been executed.

We should force the insertion point to be within the current input before inserting text.

huguesv:
Fix #2454 

Port back some improvements that were made in RTVS so submission works even when there's selected read-only text.


zooba:
👍

PR #3162, opened 2017-10-12
A range of threading and concurrency fixes

zooba:
Fixes VsProjectAnalyzer to be created safely and awaitable on successful initialization
Changes UIThread to use JoinableTaskContext, creating one for the current thread if the global one does not exist
Change random failures due to out-of-order messages into more sensible errors when arguments are known invalid
Fixes MockVs to create everything on its UI thread and to dispose everything correctly
Disable registry watchers when container is disposed
Validates caret position in refactor rename tests
Avoid ArgumentNullException for member references with no location

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
@@ -97,7 +97,7 @@ namespace Microsoft.PythonTools.Intellisense {
         // Used by tests to avoid creating TaskProvider objects
         internal static bool SuppressTaskProvider = false;
 
-        internal static async Task<VsProjectAnalyzer> CreateDefault(
+        internal static async Task<VsProjectAnalyzer> CreateDefaultAsync(
             PythonEditorServices services,
             IPythonInterpreterFactory factory,
             bool inProcess = false
@@ -107,7 +107,7 @@ namespace Microsoft.PythonTools.Intellisense {
             return analyzer;
         }
 
-        internal static async Task<VsProjectAnalyzer> CreateForProject(
+        internal static async Task<VsProjectAnalyzer> CreateForProjectAsync(
             PythonEditorServices services,
             IPythonInterpreterFactory factory,
             MSBuild.Project project,
@@ -118,7 +118,7 @@ namespace Microsoft.PythonTools.Intellisense {
             return analyzer;
         }
 
-        internal static async Task<VsProjectAnalyzer> CreateForInteractive(
+        internal static async Task<VsProjectAnalyzer> CreateForInteractiveAsync(
             PythonEditorServices services,
             IPythonInterpreterFactory factory,
             string displayName,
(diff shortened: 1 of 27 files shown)


Python/Product/VSInterpreters/CPythonInterpreterFactoryProvider.cs
```diff
@@ -100,7 +118,8 @@ class CPythonInterpreterFactoryProvider : IPythonInterpreterFactoryProvider {
                 // Python key no longer exists, so go back to watching
                 // Software.
                 e.CancelWatcher = true;
-                StartWatching(e.Hive, e.View);
+                _registryTags.Remove(e.Tag);
```
huguesv: Do you need to also do RegistryWatcher.Instance.Remove ?

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -92,19 +92,58 @@ public sealed class VsProjectAnalyzer : ProjectAnalyzer, IDisposable {
...
-            string comment = null,
-            bool outOfProcAnalyzer = true
+            bool inProcess = false
+        ) {
+            var analyzer = new VsProjectAnalyzer(services, factory, true);
+            await analyzer.InitializeAsync(!inProcess, null, null);
+            return analyzer;
+        }
+
+        internal static async Task<VsProjectAnalyzer> CreateForProject(
```
huguesv: >CreateForProject [](start = 54, length = 16)

Should add Async suffix to these new method names

Python/Product/VSInterpreters/CPythonInterpreterFactoryProvider.cs
zooba: No - cancelling via the event removes it from the watcher.

Python/Product/PythonTools/PythonToolsService.cs
```diff
@@ -231,7 +231,7 @@ public sealed class PythonToolsService : IDisposable {
         public VsProjectAnalyzer DefaultAnalyzer {
             get {
                 if (_analyzer == null) {
-                    _analyzer = _container.GetUIThread().Invoke(() => CreateAnalyzer());
+                    _analyzer = _container.GetUIThread().InvokeTask(() => CreateAnalyzerAsync()).WaitAndUnwrapExceptions();
```
huguesv: Could we not have a UIThread wrapper for JoinableTaskFactory.Run ([https://msdn.microsoft.com/en-us/library/dn283807.aspx](https://msdn.microsoft.com/en-us/library/dn283807.aspx)), that would take care of the blocking and not have to call WaitAndUnwrapExceptions here (and everywhere else we need something similar)?

huguesv:
Just those comments. Take your time to test before checking in ;)

Python/Product/PythonTools/PythonToolsService.cs
zooba: We do, but I added it after updating this call and didn't go back. (There are plenty of cases where we need to wait for an async task but don't have the UI thread available, so unfortunately we don't get a complete replacement, but in places where we can use the blocking function we should.)

zooba:
Did my testing before submitting the PR :)

PR #3166, opened 2017-10-16
Replaces use of ReverseExpressionParser for finding expressions with an AST-based approach

zooba:
Replaces use of ReverseExpressionParser for finding expressions with an AST-based approach

This is very much a work in progress, but wanted to put it out there in case anyone has time to review and provide early feedback. I expect tweaks for correct behavior at this stage, but no massive redesign.

Don't sign off on this yet - I'll ping people when it's ready to merge.

### DIFF
--- a/Python/Product/Analysis/Analysis.csproj
+++ b/Python/Product/Analysis/Analysis.csproj
@@ -59,6 +59,7 @@
     <Compile Include="AggregateProjectEntry.cs" />
     <Compile Include="AnalysisDictionary.cs" />
     <Compile Include="AnalysisLogWriter.cs" />
+    <Compile Include="ExpressionFinder.cs" />
     <Compile Include="Interpreter\Ast\AstAnalysisFunctionWalker.cs" />
     <Compile Include="Interpreter\Ast\AstAnalysisWalker.cs" />
     <Compile Include="IHasRichDescription.cs" />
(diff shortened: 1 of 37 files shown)


zooba:
Just burning down the last few test regressions now - I think this is nearly ready. Feel free to take a look, @huguesv @int19h

huguesv:
This looks good to me. I'm not seeing the removal of ReverseExpressionParser implementation? Also, does this fix an open bug?

zooba:
No, turns out we still use it in some places where it makes sense (e.g. quickly determine whether we are in a grouping, which is easier/safer by iterating backwards over tokens than using the AST). The GetExpressionRange() calls should be easily replaceable at this point, though there are still two that I have not done yet. However, these are not the problematic one - it's completions that are most annoying.

I did a quick look for an open bug, but didn't see anything jump out. This is really a combination of technical debt and doing the correct fix for past bugs - we have had plenty of bugs that required tweaking GetExpressionRange() to get it to work in some edge case, and most of those would have been fixed by properly replacing that function with an AST walker.

zooba:
Only remaining use of GetExpressionRange is for GetSignaturesAsync. Since this requires getting the parameter index, it doesn't really have an easy translation to GetExpressionAtPoint, so I'm going to leave it for now and file a bug to do it later.

PR #3172, opened 2017-10-18
Fixes #3169 Do not include full analysis log in diagnostic info

### LINKED ISSUE #3169: Do not include full analysis log in diagnostic info
Analysis log files can get huge, and we are generally only interested in recent events and errors.

We can filter log data when getting the content, especially now we have a regular format. I'd propose:
* drop all events older than 7 days (maybe by rewriting the file on startup?)
* drop non-error events older than 3 days
* writing per-project events to a separate log file

zooba:
Fixes #3169 Do not include full analysis log in diagnostic info
Adds log rotation when it exceeds a certain number of lines.
Makes API changes necessary to enable per-interpreter logs.


Python/Product/Analysis/PythonAnalyzer.cs
```diff
@@ -642,6 +642,8 @@ class LazyModuleEnumerator {
                         yield return new ExportedMemberInfo(mod.Name, name);
                     }
                 }
+
+                yield break;
```
zooba: This change regressed at some point - if we don't break out here (in this `if` block) then we eagerly load every single module available, which adds a lot of log entries.

Python/Product/Analysis/Interpreter/Ast/AstAnalysisWalker.cs
```diff
@@ -153,37 +153,45 @@ bool includeLocationInfo
...
-
-            try {
-                for (int i = 0; i < node.Names.Count; ++i) {
-                    if (!onlyImportModules) {
-                        if (node.Names[i].Name == "*") {
-                            foreach (var member in mod.GetMemberNames(_scope.Context)) {
-                                var mem = mod.GetMember(_scope.Context, member) ?? new AstPythonConstant(
+            var interp = _scope.Context as AstPythonInterpreter;
+            if (interp?.BeginNestedImport() ?? true) {
+                mod.Imported(_scope.Context);
```
huguesv: Should this whole block be under a try/finally to ensure EndNestedImport is called?

zooba: EndNestedImport was removed in the other PR. This one will be re-merged and it will go away here too.

PR #3173, opened 2017-10-18
Fixes #2878 AstAnalysis eagerly imports module members

### LINKED ISSUE #2878: AstAnalysis eagerly imports module members
Because we resolve imported references eagerly, we end up having to analyze a large graph of modules on the first import.

We should lazily import module members and only actually scrape their modules on demand. Otherwise this is unusably slow.

zooba:
Fixes #2878 AstAnalysis eagerly imports module members
Adds ILazyMember interface and implements it for AST analysis

(Note that there are some unrelated changes from #3172 in here too - those should disappear when the other PR is merged and this one gets updated.)


Python/Product/Analysis/Interpreter/Ast/AstNestedPythonModuleMember.cs
```diff
@@ -0,0 +1,64 @@
...
+        private volatile IMember _realMember;
+        private readonly IModuleContext _context;
+
+        public AstNestedPythonModuleMember(
+            string memberName,
+            AstNestedPythonModule module,
+            IModuleContext importContext,
+            LocationInfo importLocation
+        ) {
+            Name = memberName;
```
huguesv: Looks like some/most of these parameters could be validated non-null

PR #3174, opened 2017-10-18
Fixes #2675 Module names in the Modules debugger window are unqualified

### LINKED ISSUE #2675: Module names in the Modules debugger window are unqualified
We use unqualified names in the "Name" column of the Modules tool window - e.g. `abc`, `collections.abc` and `encodings.abc` are all reported as three different entries all named `abc`. You can distinguish them by looking at the module path, but we should really be using properly qualified names as well.

This is particularly bad with respect to the numerous `__init__` entries.

zooba:
Fixes #2675 Module names in the Modules debugger window are unqualified
Includes full module name when we have it
Correctly marks non-user modules
Fixes ptvsd directory not being excluded

### DIFF
--- a/Python/Product/Debugger/Debugger/DebugEngine/AD7Module.cs
+++ b/Python/Product/Debugger/Debugger/DebugEngine/AD7Module.cs
@@ -116,9 +116,6 @@ namespace Microsoft.PythonTools.Debugger.DebugEngine {
             return VSConstants.S_OK;
         }
 
-        // Used to support the JustMyCode features of the debugger.
-        // the sample debug engine does not support JustMyCode and therefore all modules
-        // are considered "My Code"
         int IDebugModule3.IsUserCode(out int pfUser) {
             pfUser = DebuggedModule.IsUserCode ? 1 : 0;
             return VSConstants.S_OK;
--- a/Python/Product/Debugger/Debugger/DebugEngine/AD7Module.cs
+++ b/Python/Product/Debugger/Debugger/DebugEngine/AD7Module.cs
@@ -16,6 +16,8 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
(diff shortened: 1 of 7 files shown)


Python/Product/Debugger/Debugger/DebugEngine/AD7Module.cs
```diff
@@ -110,7 +120,7 @@ class AD7Module : IDebugModule2, IDebugModule3 {
         // the sample debug engine does not support JustMyCode and therefore all modules
         // are considered "My Code"
         int IDebugModule3.IsUserCode(out int pfUser) {
-            pfUser = 1;
+            pfUser = DebuggedModule.IsUserCode ? 1 : 0;
```
int19h: Can you also get rid of the comment above, while you're at it?

Python/Product/Debugger/Debugger/LegacyDebuggerProtocol.cs
```diff
@@ -405,6 +406,10 @@ public sealed class ModuleLoadEvent : Event {
 
             public int moduleId;
             public string moduleFileName;
+            [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+            public string moduleName;
+            [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+            public int isStdLib;
```
int19h: bool?

zooba: Kept blowing up in deserialization for some reason I couldn't figure out. I think the JSON serialization is more forgiving for missing/null int than bool.

zooba: Tried changing it back to `bool` and it worked, so I guess I messed something up last time.

PR #3182, opened 2017-10-19
Prevent virtual env error related to deferred project save

### LINKED ISSUE #1429: Corrupted environment after saving a Flask project
### Problem
1. Start Visual Studio
2. Create a new Flask web project
3. Click on "install into a virtual environment" and then click on Create. At this point Visual Studio downloads and installs all the required dependencies and creates a virual environment
4. Press F5 and the server starts and the browser can navigate through the 3 default pages. Great!
5. Press Ctrl+Shift+S to save the project and click on Save to save to the default folder
6. Press F5 and the server does not start. An error message about the environment being incorrectly configured or missing appears. I tried the same steps a few times, and sometimes at this point instead of the error message I get a Python crash.

If I close and restart VS and reopen the project just saved, the solution explorer shows the environment env previously created, but there are no modules in it.
### Workaround

Create a new environment:
1. Close and reopen Visual Studio and reload the project
2. Right click on the corrupted env environment and delete it
3. Right click on the Python Environment tree element and add an environment

huguesv:
Fix #1429 

Discovered that the virtual env tests hadn't been ported yet, as I wanted to verify the impact of the bug fix on them. Turns out that since we don't create the projects via new project dialog, the "Save new projects" option doesn't apply to them so there is no need to prevent/check for the save dialog.

Most tests previously were using `virtualenv` but are now using `venv`. I spent more than 2 days trying to find/fix the issue in `virtualenv` where the env and executable was created successfully, but pip/wheel was failing to install. The only error indication was a -1 exit code. I decided to cut my losses and switch them all to `venv` which succeeds much more reliably.


Python/Tests/Core.UI/VirtualEnvUITests.cs
```diff
@@ -54,52 +50,28 @@ public class VirtualEnvTests {
...
-            } finally {
-                dis?.Dispose();
-            }
-        }
-
-        static DefaultInterpreterSetter Init3(PythonVisualStudioApp app, bool installVirtualEnv = false) {
-            var dis = app.SelectDefaultInterpreter(
+        static DefaultInterpreterSetter InitPython3(PythonVisualStudioApp app) {
+            return app.SelectDefaultInterpreter(
+                PythonPaths.Python36 ?? PythonPaths.Python36_x64 ??
```
zooba: Add Python37 here as well. When the test machines come back up I'll put 3.7.0a2 on them.

PR #3188, opened 2017-10-19
Fixes #2869 Go to definition fails for named argument

### LINKED ISSUE #2869: Go to definition fails for named argument for instance method call
Affected test: NavigableTests.NamedArgumentDefinition

```
class MyClass(object):
    def my_class_func1(self, param2 = True):
        pass

    def my_class_func2(self, param3 = True):
        pass

def my_func(param3 = True):
    pass

my_func(param3=False)

obj = MyClass()
obj.my_class_func1(param2=False)
obj.my_class_func2(param3=False)
```

If you press F12 on `param2` in `obj.my_class_func1(param2=False)`, you'll get Cannot go to definition.

If you press F12 on `param3` in `obj.my_class_func2(param3=False)`, it will go to `my_func` instead of `my_class_func2`.

zooba:
Fixes #2869 Go to definition fails for named argument
Enable GTD for argument names.
Fixes bug in FindScopes where we were selecting function scopes incorrectly.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/TaskProvider.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/TaskProvider.cs
@@ -24,6 +24,7 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using Microsoft.PythonTools.Editor.Core;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
 using Microsoft.PythonTools.Parsing;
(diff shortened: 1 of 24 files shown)


zooba:
It looks like most of the other navigable tests are currently failing too. If nobody reviews this today, then I'll work on them tomorrow as part of this change (since it's probably going to be related issues).

zooba:
@huguesv Worth another quick look. This is ready to merge now, barring any feedback

Python/Product/PythonTools/PythonTools/Intellisense/TaskProvider.cs
```diff
@@ -85,19 +86,22 @@ LocationTracker spanTranslator
...
         public void CreateSquiggleSpan(SimpleTagger<ErrorTag> tagger) {
-            if (_rawSpan.Length <= 0 || _spanTranslator == null) {
+            if (_rawSpan.Start <= _rawSpan.End || _spanTranslator == null) {
                 return;
             }
 
-            SnapshotSpan target = _spanTranslator.TranslateForward(
-                new Span(_rawSpan.Start.Index, _rawSpan.Length)
-            );
+            // TODO: Map between versions rather than using the current snapshot
```
huguesv: Did you mean to get back to that before checkin or later?

huguesv:
Still looks ok

Python/Product/PythonTools/PythonTools/Intellisense/TaskProvider.cs
zooba: Later

PR #3193, opened 2017-10-20
Relocate parethenses for proper tab completion

mkolovic:
Putting the parentheses around `object` in the `<Default>` tag caused them to be overwritten by the tab completion, resulting in a broken class definition like:

```python

class NewFloatfloat:
    pass
```

Placing the parentheses in the `<Code>`  tag we ensure that tab completion only highlights `object` for replacement, and leaves the parentheses in place resulting in a proper class definition like:

```python
class NewFloat(float):
    pass
```

### DIFF
--- a/Python/Product/PythonTools/Snippets/Python/class.snippet
+++ b/Python/Product/PythonTools/Snippets/Python/class.snippet
@@ -21,10 +21,10 @@
         <Literal>
           <ID>bases</ID>
           <ToolTip>The base class list for the class</ToolTip>
-          <Default>(object)</Default>
+          <Default>object</Default>
         </Literal>      
       </Declarations>
-      <Code Language="Python"><![CDATA[class $name$$bases$:
+      <Code Language="Python"><![CDATA[class $name$($bases$):
     $selected$]]></Code>
     </Snippet>  
   </CodeSnippet>


zooba:
Unfortunately, leaving them out of the tag makes it more difficult to delete the bases entirely, resulting in it being harder to create a correct Python 3 class definition.

We'll take this proposal as a vote for changing that, but without stronger evidence that _most_ users want to specify bases rather than remove them, we're not prepared to make the change globally.

zooba:
Unless we can figure out a way for delete to also remove the parentheses... might be possible?

mkolovic:
I didn't take into account that in Python 3 all classes inherit from `object` implicitly.

As it stands, users wanting explicit inheritance from `object` just have to hit enter after filling out `ClassName`, so this code snippet benefits them the most, in particular Python 2 users who have to use explicit inheritance. Users wanting implicit inheritance from `object` have to hit the backspace and erase the `(object)` portion. Users wanting to replace `object` have to hit backspace, then add the parentheses back along with the bases. I see now why the current state is an optimal balance for Python 2 and Python 3 users under the assumption that inheriting from `object` is the most common use case.

I would like to try making the next tab highlight only `object` for replacement, before cycling back to `ClassName`. So the sequence of highlighting would be `ClassName`, `(object)`, `object`. That way all users could replace or delete the parts they want. I made an attempt at this within the `class.snippet` XML file but I suspect it may not be possible solely from there. I would be interested in pursuing this but I would need some guidance. If this solution is not viable I am fine with the current state of this code snippet.

Thanks for taking a look at this.

huguesv:
With the existing snippet, you're not really saving any keystroke if you delete the base (unless you really wanted that `pass`).
Personally, I'd rather have 2 separate optimized snippets `class` and `classb`.

zooba:
`cl<Tab><Tab>` is shorter than `class<space>`, though since we're not using the bases variable anywhere else in the snippet we may as well remove it completely and let people type it as part of the name.

zooba:
Closing this PR, as we're not going to take it as-is, but we should keep thinking about ways to improve our code generation features like snippets.

PR #3212, opened 2017-10-23
Fixes #3076 Some tests do not appear

### LINKED ISSUE #3076: Some tests do not appear in VS 2017 that do appear in 2015.
Our project has a number of tests that use the Python VCR library to record HTTP responses. In VS2015, these tests all appear in the test explorer. However, in VS2017, these tests no longer appear--only regular Python unit tests do. This seriously inhibits our ability to debug into these tests without going back to the old version of VS.

### LINKED ISSUE #3211: GetTypeFromValue() doesn't handle AstPythonBoundMethod type
Just got that assert:
`Debug.Fail("Unhandled type() value: " + value.GetType().FullName);`

zooba:
Fixes #3076 Some tests do not appear
Adds correct MRO to scraped analysis types.
Fixes handling of multiple definitions of the same class
Fixes minor crash when tool tips time out.

Fixes #3211 GetTypeFromValue() doesn't handle AstPythonBoundMethod type 
Adds IPythonBoundFunction handling

### DIFF
--- a/Python/Product/Analysis/Interpreter/Ast/AstPythonTuple.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstPythonTuple.cs
@@ -1,0 +1,48 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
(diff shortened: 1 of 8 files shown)


Python/Product/Analysis/Interpreter/Ast/AstPythonTuple.cs
```diff
@@ -0,0 +1,48 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.PythonTools.Interpreter.Ast {
+    class AstPythonSequence : IPythonSequenceType {
```
huguesv: Should name the file AstPythonSequence.cs

Python/Product/Analysis/Interpreter/Ast/AstPythonTuple.cs
```diff
@@ -0,0 +1,48 @@
...
+        private readonly IPythonType _sequenceType;
+
+        public AstPythonSequence(
+            IPythonType sequenceType,
+            IPythonModule declaringModule,
+            IEnumerable<IPythonType> contents
+        ) {
+            _sequenceType = sequenceType;
+            IndexTypes = contents.ToArray();
```
huguesv: throw ANE for null contents

PR #3215, opened 2017-10-24
LEGO: Pull request from lego/hb_839_20171024024228999 to master

csigs:
LEGO: Pull request from lego/hb_839_20171024024228999 to master with localized lcls


huguesv:
This looks wrong. It's reverting some localized resources back to what they used to be.
For example, I updated the english resource for `ImportCoverageCommandFileFilter` a few weeks ago to `coverage.py XML (*.xml)|*.xml|All Files (*.*)|*.*`.
That was already translated and is now going back to its previous value?

zooba:
Agreed. I think this is because I built the v3.2 branch into the main drop location and the loc team picked it up. I'm fixing the build now, so it shouldn't happen again.

PR #3219, opened 2017-10-25
Fixes #3218 Use return annotation for AST analysis

### LINKED ISSUE #3218: Use return annotation for AST analysis
This is a subset of #82 just to enable using return annotations when doing AST analysis. This will enable stdlib and installed package analysis to do more return types.

zooba:
Fixes #3218 Use return annotation for AST analysis
Adds basic (builtin/global) return annotation parsing


Python/Product/Analysis/Interpreter/Ast/AstTypeAnnotationConverter.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.PythonTools.Parsing.Ast;
+
+namespace Microsoft.PythonTools.Interpreter.Ast {
+    class AstTypeAnnotationConverter : TypeAnnotationConverter<IPythonType> {
+        private readonly NameLookupContext _scope;
+
+        public AstTypeAnnotationConverter(NameLookupContext scope) {
+            _scope = scope;
```
huguesv: throw if null

PR #3235, opened 2017-10-27
Fixes #3232 Classifications do not cover whole file

### LINKED ISSUE #3232: Classifications do not cover whole file
Sometimes while editing you will lose all syntax highlighting from a certain point to the end of the file. We should not do this.

This is likely due to changes in #3224

zooba:
Fixes #3232 Classifications do not cover whole file
Moves in-proc tokenization into PythonTextBufferInfo
Switches some users of the classifier to the correct token source
Fixes line number management of token cache (i.e. don't trust the tokens)
Removes index comparisons from SourceLocation
Fixes explicit/incorrect uses of three argument SourceLocation constructor

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
+++ b/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
@@ -655,7 +655,7 @@ namespace Microsoft.PythonTools.Editor {
 
                 var afterState = lineTokenization.State;
 
-                // stop if we visted all affected lines and the current line has no tokenization state
+                // stop if we visited all affected lines and the current line has no tokenization state
                 // or its previous state is the same as the new state.
                 if (lineNo > lastLine && (beforeState == null || beforeState.Equals(afterState))) {
                     break;
--- a/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
+++ b/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
@@ -90,6 +90,7 @@ namespace Microsoft.PythonTools.Editor {
         private Tokenizer _tokenizer;
         private readonly TokenCache _tokenCache;
 
+        private readonly bool _hasChangedOnBackground;
         private bool _replace;
 
         private PythonTextBufferInfo(PythonEditorServices services, ITextBuffer buffer) {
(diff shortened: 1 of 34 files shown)


zooba:
I bit the bullet and centralized tokenization/caching, which I've been worrying about having to do for a while. It means we can stop using the classifier and use the PythonTextBufferInfo in a variety of places (which is better because we control the lifetime of the text buffer info, so it'll clean up certain race conditions with VS). But there's different information available, so it's a bit more work to translate them all.

This is probably also the last piece we need to get off ReverseExpressionParser. But I'm starting to think it may need a bit of a rewrite from what's here in order to keep snapshots in sync. Feel free to review, but there may be more changes before it can be merged.

zooba:
This is probably ready for review now. I'm running the test suite against it to try and flush out issues before merging though. But it's essentially all internal refactoring with no visible impact (though the use of the new event for tokenization should improve typing responsiveness a bit).

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -455,6 +579,94 @@ sealed class PythonTextBufferInfo {
...
+            for (int lineNo = firstLine; lineNo < span.Snapshot.LineCount; ++lineNo) {
+                var line = span.Snapshot.GetLineFromLineNumber(lineNo);
+
+                var beforeState = lineTokenization.State;
+
+                tokenCache[lineNo] = lineTokenization = TokenizeLine(tokenizer, line, beforeState);
+
+                var afterState = lineTokenization.State;
+
+                // stop if we visted all affected lines and the current line has no tokenization state
```
huguesv: np typo 'visted'

Python/Product/PythonTools/PythonTools/TokenCache.cs
```diff
@@ -54,13 +53,136 @@ internal struct LineTokenization : ITag {
...
+        /// 0-based index where the token starts on the line.
+        /// </summary>
+        public int Column;
+        /// <summary>
+        /// Number of characters included in the token.
+        /// </summary>
+        public int Length;
+    }
+
+    public struct TrackingTokenInfo {
```
huguesv: hard to tell from here, does this have to be public?

zooba: Strictly no, but then we have a stack of internal API on PythonTextBufferInfo. We can do that, but I think there's value in that API being public. But maybe it's better just to hide the whole thing for now...

huguesv: PythonTextBufferInfo isn't public either. I wouldn't worry about making this public if it's not necessary.

zooba: Ah, you're right. Guess I misread the "sealed" as "public" at some point.

PR #3254, opened 2017-10-31
Fix debugger test failures

huguesv:
Don't add the ptvsd_launcher.py parent folder to the PREFIXES array of standard library folders. The ptvsd files are already in the DONT_DEBUG array of files to never debug.

### DIFF
--- a/Python/Product/PythonTools/ptvsd/debugger.py
+++ b/Python/Product/PythonTools/ptvsd/debugger.py
@@ -370,7 +370,7 @@ def get_thread_from_id(id):
 def should_send_frame(frame):
     return (frame is not None and
             frame.f_code not in DEBUG_ENTRYPOINTS and
-            path.normcase(frame.f_code.co_filename) not in DONT_DEBUG)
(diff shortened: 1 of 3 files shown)


zooba:
Do we know why that line was added in the first place?

huguesv:
You said "Fixes ptvsd directory not being excluded" but I'm not sure what were the repro steps for that.
https://github.com/Microsoft/PTVS/commit/2b56096fbd001bc1ea5acb9397acf77873f50c16

zooba:
Ah, I was trying to hide the ptvsd modules from the modules window. I bet they're showing up again now, or they may be showing up as "user code" - I don't think we filter these using `DONT_DEBUG`.

huguesv:
Yeah they show up in modules window. Since we pretend that ptvsd doesn't exist when debugging (always removed from call stack) then we probably don't want to skip sending the module loaded event for them (DONT_DEBUG) as well, all the time?

zooba:
I think what you're saying is "we should skip sending module load events for items in DONT_DEBUG"? If so, yeah, let's just do that.

PR #3255, opened 2017-10-31
Enables full line highlighting for breakpoints and call stack locations.

zooba:
Enables full line highlighting for breakpoints and call stack locations.
Prevents setting breakpoints on empty lines.
Make LocationToIndex and IndexToLocation functions public
Make LocationToIndex deliberately support columns well beyond the line length.

The LocationToIndex and IndexToLocation functions were always meant to become public - it's the only reliable way to translate between the two formats.

### DIFF
--- a/Python/Product/Debugger/Debugger/PythonProcess.cs
+++ b/Python/Product/Debugger/Debugger/PythonProcess.cs
@@ -490,9 +490,6 @@ namespace Microsoft.PythonTools.Debugger {
 
         internal SourceSpan GetStatementSpan(string filename, int line, int column = 0) {
             var ast = GetAst(filename);
-            if (ast == null) {
-                return new SourceSpan(new SourceLocation(line, 1), new SourceLocation(line, 1));
-            }
             int len = 0;
             if (ast != null) {
                 int start = ast.LocationToIndex(new SourceLocation(line, 1));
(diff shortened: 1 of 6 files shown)


Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -488,6 +488,21 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
             return ast;
         }
 
+        internal SourceSpan GetStatementSpan(string filename, int line, int column = 0) {
```
huguesv: Remove column param that is unused?

Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -488,6 +488,21 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
...
+            var ast = GetAst(filename);
+            if (ast == null) {
+                return new SourceSpan(new SourceLocation(line, 1), new SourceLocation(line, 1));
+            }
+            int len = 0;
+            if (ast != null) {
```
huguesv: ast won't be null here

Python/Product/Debugger/Debugger/PythonProcess.cs
zooba: Planning for future use - if we switch to an AST walk (which might make sense) and start doing opcode lookups then the column will be relevant. (Might not be available, but we'll deal with that when it comes.)

PR #3258, opened 2017-11-01
Fixes #1639 Support PEP 515 syntax

### LINKED ISSUE #1639: Support PEP 515 syntax
[PEP 515](https://www.python.org/dev/peps/pep-0515/) adds support for underscores within numeric literals.

These are ignored by the parser, but we need to round-trip them correctly.

zooba:
Fixes #1639 Support PEP 515 syntax
Makes tokenizer skip underscores in numbers
Adds post-parsing error reporting


zooba:
Still need more tests - open to suggestions for formats that should be handled.

The post-parse error reporting lets us not break an entire script because of a misplaced underscore, but we will inform users about tokens that will fail to parse. I think this is better than handling it all in the tokenizer, since it means you won't get 1000 errors while typing a number.

zooba:
I'm happy with the coverage now, feel free to review.

Python/Product/Analysis/Parsing/Tokenizer.cs
```diff
@@ -1200,30 +1212,32 @@ public Tokenizer(PythonLanguageVersion version, ErrorSink errorSink = null, Toke
                     case 'L':
                         MarkTokenEnd();
 
-                        if (_langVersion.Is3x()) {
-                            ReportSyntaxError(new IndexSpan(_tokenEndIndex - 1, 1), "invalid token", ErrorCodes.SyntaxError);
+                        if (Verbatim) {
```
huguesv: Is there code somewhere else that will report the syntax error on 3.x (for invalid 'L')?

Python/Product/Analysis/Parsing/Tokenizer.cs
```diff
@@ -1307,28 +1321,37 @@ public Tokenizer(PythonLanguageVersion version, ErrorSink errorSink = null, Toke
...
-                        return new ConstantValueToken(ParseFloat(GetTokenString()));
+                        return new ConstantValueToken(ParseFloat(tokenStr));
                 }
             }
         }
 
         private Token ReadExponent(bool leftIsFloat = false) {
+            BufferBack();
+            MarkTokenEnd();
+            var beforeEStr = GetTokenString();
```
huguesv: unused variable?

Python/Tests/Analysis/ParserTests.cs
```diff
@@ -853,6 +853,46 @@ public class ParserTests {
         }
 
         [TestMethod, Priority(0)]
+        public void Literals36() {
```
huguesv: Maybe add a 3.5 test that verify we report errors?

Python/Product/Analysis/Parsing/Tokenizer.cs
zooba: Yeah, was a nice but ultimately insufficient idea to validate better.

Python/Product/Analysis/Parsing/Tokenizer.cs
zooba: No, I'll add it (and some tests)

PR #3272, opened 2017-11-02
Fixes #3269 Changes in search paths are not automatically noticed by the analyzer

### LINKED ISSUE #3269: Changes in search paths are not automatically noticed by the analyzer
We need to set up file watchers on search paths and reset the analyzer whenever the contents change.

zooba:
Fixes #3269 Changes in search paths are not automatically noticed by the analyzer
Adds file watcher for search paths in the IDE. The notifications flow into the analyzer and cause it to rescan search paths as necessary.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/SearchPathManager.cs
+++ b/Python/Product/PythonTools/PythonTools/SearchPathManager.cs
@@ -17,13 +17,48 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Microsoft.PythonTools.Infrastructure;
+using Microsoft.VisualStudio;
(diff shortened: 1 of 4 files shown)


Python/Product/PythonTools/PythonTools/SearchPathManager.cs
```diff
@@ -17,14 +17,49 @@
...
+        public void Dispose() {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        ~SearchPathManager() {
+            Dispose(false);
+        }
+
+        protected void Dispose(bool disposing) {
```
int19h: `virtual` (otherwise the Dispose pattern doesn't really make sense)

zooba:
This might be worth backporting to 15.5

PR #3275, opened 2017-11-03
LEGO: Pull request from lego/hb_839_20171103084157355 to master

csigs:
LEGO: Pull request from lego/hb_839_20171103084157355 to master with localized lcls


huguesv:
There's one string that I noticed (the coverage.py one) whose translation is deleted in PTB/RUS/TRK.

zooba:
Let's assume that it will come back before we need to release.

At some point we need to go through the product ourselves in every language and check every string, since the loc team won't do that anymore... not looking forward to that...

PR #3280, opened 2017-11-03
Fixes #3248 and #3269 for v3.2.1

### LINKED ISSUE #3248: Intellisense fails to scrape modules from other projects
Upgraded to VS2017 15.4.2 today, was pleased that error highlighting has been fixed. However, modules still fail to import from other projects.

Steps to reproduce: 
-  Install VS2017 15.4.2 with Python Support
-  Create new solution with project inside (here, its called PythonApplication1)
-  Create another project in the solution (here, its called other_module), with other_project.py inside.
-  Add to PythonApplication1's _Search Paths_ the directory of other_module.
-  Attempt to import other module.
-  Observe failure of Intellisense to recognize other_module.
-  Refreshing DB does nothing. Adding other_module to _References_ does nothing.

Screenshots of various behaviors observed:

Importing modules from other projects was supported at one time, so I assume this is a bug. I still have one computer with VS2017 _15.3_ on it, everything works great in that version.

Here is the gist of the python tools diagnostic
[https://gist.github.com/joseortiz3/def9d14386aa82ec176401d88cff1013](https://gist.github.com/joseortiz3/def9d14386aa82ec176401d88cff1013)

My setup: Anaconda 5.0.0, Python 3.6, VS2017 Community install with Python Tools option without optional Python 3.6

zooba:
Fixes #3248 and #3269 for v3.2.1.
Merges other changes from 3.2

### DIFF
--- a/Python/Setup/signlayout.proj
+++ b/Python/Setup/signlayout.proj
@@ -124,8 +124,6 @@
 
   <Target Name="_UpdateLayoutWithSignedBinaries"
           DependsOnTargets="_GetBinariesInLayout"
-          Inputs="%(_BinariesToCopy.FullPath)"
-          Outputs="%(_BinariesToCopy.TargetPath)"
           AfterTargets="SignFiles">
     <Copy SourceFiles="%(_BinariesToCopy.FullPath)" DestinationFiles="%(_BinariesToCopy.TargetPath)" />
   </Target>
--- a/Python/Product/Analysis/Interpreter/Ast/AstAnalysisWalker.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstAnalysisWalker.cs
@@ -30,7 +30,8 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
 
         private readonly List<AstAnalysisFunctionWalker> _postWalkers;
 
-        private IMember _noneInst;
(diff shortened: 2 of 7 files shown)


Python/Setup/signlayout.proj
```diff
@@ -124,8 +124,6 @@
 
   <Target Name="_UpdateLayoutWithSignedBinaries"
           DependsOnTargets="_GetBinariesInLayout"
```
huguesv: What are those removals? They are still in master.

Python/Setup/signlayout.proj
```diff
@@ -124,8 +124,6 @@
...
-          Inputs="%(_BinariesToCopy.FullPath)"
-          Outputs="%(_BinariesToCopy.TargetPath)"
           AfterTargets="SignFiles">
```
huguesv: I mean here

Python/Setup/signlayout.proj
zooba: Yeah, we should remove them from master as well. Just haven't ported it that far forward yet.

Removing these means that when the signed files come back with a modified date earlier than the originals (which happened to me yesterday), we will still overwrite the originals. Otherwise, we end up with unsigned files in the bundle.

PR #3281, opened 2017-11-03
Fixes breakpoints on first and last line

### LINKED ISSUE #3270: Breakpoint on first line is not hit
New Python application
```
print('hello world')
print('hello world')
```
Put breakpoint on line 1 and start debugging.
Result: no break

Put breakpoint on line 2 and start debugging.
Result: break on line 2

~9 tests in DebugProjectUITests are failing due to this.

huguesv:
Fix #3270 

Also fixes EnumModules test.

### DIFF
--- a/Python/Product/Debugger/Debugger/PythonProcess.cs
+++ b/Python/Product/Debugger/Debugger/PythonProcess.cs
@@ -492,9 +492,14 @@ namespace Microsoft.PythonTools.Debugger {
             var ast = GetAst(filename);
             int len = 0;
             if (ast != null) {
-                int start = ast.LocationToIndex(new SourceLocation(line, 1));
-                int end = ast.LocationToIndex(new SourceLocation(line, int.MaxValue));
-                len = end - start;
+                try {
+                    int start = ast.LocationToIndex(new SourceLocation(line, 1));
+                    int end = ast.LocationToIndex(new SourceLocation(line, int.MaxValue));
+                    len = end - start;
+                } catch (OverflowException) {
+                    // This happens when we request the end of the last line
+                    len = 0;
+                }
             }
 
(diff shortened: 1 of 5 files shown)


huguesv:
image

Python/Product/Debugger/Debugger/PythonProcess.cs
```diff
@@ -492,9 +492,14 @@ public PythonProcess(PythonLanguageVersion languageVersion, string exe, string a
...
             if (ast != null) {
-                int start = ast.LocationToIndex(new SourceLocation(line, 1));
-                int end = ast.LocationToIndex(new SourceLocation(line, int.MaxValue));
-                len = end - start;
+                try {
+                    int start = ast.LocationToIndex(new SourceLocation(line, 1));
+                    int end = ast.LocationToIndex(new SourceLocation(line, int.MaxValue));
+                    len = end - start;
+                } catch (OverflowException) {
+                    // This happens when we request the end of the last line
```
zooba: Can we handle this in `LocationToIndex` rather than here? I'd rather get a sensible return value for requesting the end of the line (I assume that's where the error is coming from - the `end - start` bit ought to be fine)

huguesv: What's a sensible value in that case? -1 ?

zooba: And to be clear, I'd rather detect it with comparisons before doing arithmetic than just doing it and catching the exception.

zooba: The very last index of the file.

huguesv:
I ran most non-ui tests and although some are failing they don't appear to be regressions from my changes.

PR #3296, opened 2017-11-06
Reduce the amount of text in test output

huguesv:
Capture output and only print it if there's a test failure.

This eliminates most output from attach tests and debugger tests.

I can do something similar (separately) for analysis tests if this looks ok.

### DIFF
--- a/Python/Product/Ipc.Json/DebugTextWriter.cs
+++ b/Python/Product/Ipc.Json/DebugTextWriter.cs
@@ -14,48 +14,25 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-using System;
 using System.Diagnostics;
 using System.IO;
 using System.Text;
 
 namespace Microsoft.PythonTools.Ipc.Json {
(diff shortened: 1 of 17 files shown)


zooba:
LGTM, but would like to see @int19h's opinion as well in case he has any better ideas or tricks.

Python/Product/Ipc.Json/DebugTextWriter.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public override long Seek(long offset, SeekOrigin origin) {
+            throw new NotImplementedException();
+        }
+
+        public override void SetLength(long value) {
+            throw new NotImplementedException();
+        }
+
+        public override void Write(byte[] buffer, int offset, int count) {
```
int19h: In theory, this might get a partial write that is not valid UTF-8 (e.g. one or the other boundary is in the middle of an encoded character). I think it would be better to ditch the stream, and just implement this logic in the writer instead, since you're getting Unicode text there.

huguesv:
Thanks @int19h, I made the change.

PR #3305, opened 2017-11-08
Reduce test output for AstAnalysisTests

huguesv:
Also some test fixes

CodeFlow may be better at showing the actual changes - most lines just have increased indentation

Edit: I can confirm that CodeFlow is much better for this one

### DIFF
--- a/Python/Tests/Analysis/AstAnalysisTests.cs
+++ b/Python/Tests/Analysis/AstAnalysisTests.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


Python/Tests/Analysis/AstAnalysisTests.cs
```diff
@@ -399,37 +445,43 @@ public class AstAnalysisTests {
...
-                foreach (var err in errors) {
-                    Console.WriteLine(err);
-                }
-                Assert.AreEqual(0, errors.Count(), "Parse errors occurred");
+                try {
+                    var fact = (AstPythonInterpreterFactory)analysis.Analyzer.InterpreterFactory;
+                    var interp = (AstPythonInterpreter)analysis.Analyzer.Interpreter;
+
+                    var mod = interp.ImportModule(interp.BuiltinModuleName);
+                    // TODO: CreateAnalysis specifies UseExistingCache = false,
```
huguesv: If printing this is important, I can save it and print it out in cleanup if test fails. If not important, I can delete the code.

zooba: Yeah, we want this if the test fails.

Python/Tests/Analysis/AstAnalysisTests.cs
```diff
@@ -399,37 +445,43 @@ public class AstAnalysisTests {
...
-                foreach (BuiltinTypeId v in Enum.GetValues(typeof(BuiltinTypeId))) {
-                    var type = interp.GetBuiltinType(v);
-                    Assert.IsNotNull(type, v.ToString());
-                    Assert.IsInstanceOfType(type, typeof(AstPythonBuiltinType), $"Did not find {v}");
-                }
+                    // Ensure we can get all the builtin types
+                    foreach (BuiltinTypeId v in Enum.GetValues(typeof(BuiltinTypeId))) {
+                        var type = interp.GetBuiltinType(v);
+                        Assert.IsNotNull(type, v.ToString());
+                        Assert.IsInstanceOfType(type, typeof(AstPythonType), $"Did not find {v}");
```
zooba: If the type is not `AstPythonBuiltinType`, the test should fail.

Python/Tests/Analysis/AstAnalysisTests.cs
zooba: The `ImportModule` call above generates this file.

PR #3321, opened 2017-11-10
Enables better detection of installed Win10 SDK.

zooba:


zooba:
Feel free to merge after approval.

huguesv:
This means we'll build with Windows 10 SDK instead of Windows 8? Does the Windows 10 SDK drop support of OSes that we still need?

huguesv:
I guess if the rest of VS builds with Win 10 SDK, then that works for us too.

zooba:
The SDK is actually fully backwards compatible, so it's not a big deal. The bigger problem is that VS can only automatically find the Win 8 SDK, and you need to specify a specific Win10 SDK (which is pointless...)

This code is very similar to what I contributed to CPython to solve the same problem.

zooba:
@huguesv Can you merge this once you've merged your own? I'd rather the test fix go in first.

PR #3323, opened 2017-11-11
More assert prevention under tests

huguesv:


Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -540,7 +540,12 @@ SafePipeHandle stdInClientHandle
 
         private Connection StartSubprocessConnection(string comment, out AnalysisProcessInfo proc) {
             var libAnalyzer = typeof(AP.FileChangedResponse).Assembly.Location;
-            var psi = new ProcessStartInfo(libAnalyzer, "/interactive /comment \"" + comment + "\"");
+#if DEBUG
+            var inTests = Debug.Listeners["Microsoft.PythonTools.AssertListener"] != null;
```
zooba:
```
var testOption = "/unittest";
#else
string testOption = "";
```
Thought? Your call.

PR #3330, opened 2017-11-13
NRE in SendToReplCommand

### LINKED ISSUE #3331: NullReferenceException while using SendToReplCommand
In `EditFilterQueryStatus` the null check for `activeView` occurs after attempting to access its property. That can lead to a NRE.

karthiknadig:
Fix #3331 : removing the line where the NRE occurred. The result of that check is never used.
Testing this process out with a NRE fix.


zooba:
Is there an issue? If so, make sure you have "Fixes #<issue number>" in the message

zooba:
See #3329 for an example of how I do my messages

PR #3333, opened 2017-11-14
VirtualEnvUITests fixes

huguesv:
InstallUninstallPackage, InstallGenerateRequirementsTxt, CreateInstallRequirementsTxt have been failing consistently on the test machine, all in AbortIfNotReady like this:
```
Test method PythonToolsUITestsRunner.VirtualEnvUITests.CreateInstallRequirementsTxt threw exception: 
TestRunnerInterop.TestFailedException: AssertFailedException: Assert.Fail failed. Unhandled exception in Microsoft.PythonTools.Project.AddVirtualEnvironmentOperation.Run (E:\A\_work\15\s\Python\Product\PythonTools\PythonTools\Project\AddVirtualEnvironmentOperation.cs:113)
System.InvalidOperationException: Operation cannot be started because the environment is not configured.
at Microsoft.PythonTools.Interpreter.PipPackageManager.<AbortIfNotReady>d__24.MoveNext()
```

Not sure the additional logging will tell us anything we don't know, it will probably just print out the import error, but I don't have a better idea.

The other change, in CreateInstallRequirementsTxt, was needed for the test to pass locally.

### DIFF
--- a/Python/Product/VSInterpreters/PipPackageManager.cs
+++ b/Python/Product/VSInterpreters/PipPackageManager.cs
@@ -184,8 +184,24 @@ namespace Microsoft.PythonTools.Interpreter {
                 )) {
                     try {
                         IsReady = (await proc == 0);
+#if DEBUG
+                        if (!IsReady) {
+                            Trace.WriteLine("import pip failed");
+                            Trace.WriteLine("stdout:");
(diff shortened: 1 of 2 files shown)


huguesv:
This isn't going to work.

Issue #3306, opened 2017-11-08
System.InvalidOperationException: uninitialized token cache


huguesv:
I created a new python application, and got this activity log error after the new project was loaded.

```

System.InvalidOperationException: uninitialized token cache at  
Microsoft.PythonTools.TokenCache.get_Map() in  C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\TokenCache.cs:line  181 at 
Microsoft.PythonTools.TokenCache.set_Item(Int32 line, LineTokenization  value) in  C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\TokenCache.cs:line  234 at  
Microsoft.PythonTools.Editor.PythonTextBufferInfo.GetLineTokenization(ITextSnapshotLine  line, Tokenizer& tokenizer) in  C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Editor\PythonTextBufferInfo.cs:line  648 at  Microsoft.PythonTools.Editor.PythonTextBufferInfo.<GetTrackingTokens>d__74.MoveNext()  in  C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Editor\PythonTextBufferInfo.cs:line  479 at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext() at  System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext() at  
System.Collections.Generic.List`1..ctor(IEnumerable`1 collection) at  
System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source) at  Microsoft.PythonTools.PythonClassifier.GetClassificationSpans(SnapshotSpan span)  in  C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\PythonClassifier.cs:line  72 
at  Microsoft.VisualStudio.Text.Classification.Implementation.ClassifierTagger.<GetTags>d__5.MoveNext()  
at  Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1.<GetTagsForBuffer>d__39.MoveNext()  --- End of stack trace from previous location where exception was thrown --- at 
 Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception  exceptionObject)
--

```

huguesv:
[...]

PR #3339, opened 2017-11-14
#3306 System.InvalidOperationException: uninitialized token cache 

MikhailArkhipov:
- Move most of the cache handling code into TokensCache and make it private
- Leave locking inside the token cache only, in 3 internal methods
- In GetTrackingTokens add check for snapshot version change during enumeration
- Remove some condition verification from now-private methods as capacity is controlled/set by the caller.


Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -473,12 +472,21 @@ sealed class PythonTextBufferInfo {
...
+            // Cache may change from background thread during enumeration
+            // so make sure we track that snapshot version is still the same.
+            var spanSnapshot = span.Snapshot;
+
+            Tokenizer tokenizer = null;
             for (int line = firstLine; line <= lastLine; ++line) {
-                var lineTokenization = GetLineTokenization(span.Snapshot.GetLineFromLineNumber(line), ref tokenizer);
+                var currentBufferSpanshot = span.Snapshot.TextBuffer.CurrentSnapshot;
+                if (spanSnapshot.Version != currentBufferSpanshot.Version) {
+                    yield break;
```
zooba: We probably need to abort more noisily here so that callers can restart with the new snapshot. At least auto-indent assumes that the end of iteration means we've reached the start of the file (and known state) - that will no longer be valid with this.

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -412,7 +411,7 @@ sealed class PythonTextBufferInfo {
         /// </summary>
         public IEnumerable<TrackingTokenInfo> GetTokens(ITextSnapshotLine line) {
             Tokenizer tokenizer = null;
-            var lineTokenization = GetLineTokenization(line, ref tokenizer);
+            var lineTokenization = _tokenCache.GetLineTokenization(line, () => tokenizer ?? (tokenizer = MakeTokenizer()));
```
zooba: Should we just pass `Lazy<Tokenizer>` instead? And maybe change `MakeTokenizer` to return Lazy?

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
zooba: Then again, it shouldn't actually matter here, as we do all the remapping required via tracking spans. But perhaps anyone iterating over these tokens will want to abort anyway if the content changes?

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
MikhailArkhipov: Yep, makes sense

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
MikhailArkhipov: It is probably better to lock the cache for the duration as it was doing before

zooba: Agreed, but that will probably have to be done with LockedEnumerable and make sure that all callers properly dispose the enumerator (I think they're all using `foreach` so it should be fine).

Python/Product/PythonTools/PythonTools/TokenCache.cs
```diff
@@ -159,140 +159,225 @@ struct TrackingTokenInfo {
...
-                EnsureCapacity(snapshot.LineCount);
-
-                foreach (var change in e.Changes) {
-                    var endLine = snapshot.GetLineNumberFromPosition(change.NewEnd) + 1;
-                    if (change.LineCountDelta > 0) {
-                        InsertLines(endLine - change.LineCountDelta, change.LineCountDelta);
-                    } else if (change.LineCountDelta < 0) {
-                        DeleteLines(endLine, Math.Min(-change.LineCountDelta, snapshot.LineCount - endLine));
+                if (_map == map) {
+                    if(_useCount == 0) {
```
zooba: Nit: a few misformatted spaces here and there. Won't call them all out.

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -472,31 +473,20 @@ sealed class PythonTextBufferInfo {
...
-                foreach (var token in lineTokenization.Tokens.MaybeEnumerate()) {
-                    if (line < firstLine || line > lastLine) {
-                        continue;
-                    }
-                    if (line == firstLine && token.Column + token.Length < startCol) {
-                        continue;
+            // We need current state of the cache since it can change from a background thread
+            using (var cacheSnapshot = _tokenCache.GetSnapshot()) {
+                for (int line = firstLine; line <= lastLine; ++line) {
+                    var lineTokenization = cacheSnapshot.GetLineTokenization(span.Snapshot.GetLineFromLineNumber(line), GetTokenizerLazy());
```
zooba: Calling `GetTokenizerLazy()` each time defeats the purpose - the tokenizer will always be the same for this entire function (maybe it makes sense to pass it into `GetSnapshot` and store it there?)

MikhailArkhipov: OK

PR #3342, opened 2017-11-14
Some IntelliSense improvements in Debug REPL

huguesv:
Fixes the existing completions so that they are correct when the user changes the current frame. Also includes parameters in the completions, not just the locals.

Return completions when an expression (text) is supplied when executing in current frame.

I'm hesitant to do more at this time given our planned debugger improvements, and the consequences it may have on the debug repl implementation.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Repl/PythonDebugProcessReplEvaluator.cs
+++ b/Python/Product/PythonTools/PythonTools/Repl/PythonDebugProcessReplEvaluator.cs
@@ -275,10 +275,12 @@ namespace Microsoft.PythonTools.Repl {
         }
 
         public override CompletionResult[] GetMemberNames(string text) {
-            if (string.IsNullOrEmpty(text) && _currentFrameLocals != null) {
+            if (_currentScopeName == CurrentFrameScopeFixedName && string.IsNullOrEmpty(text) && _currentFrameLocals != null) {
                 return _currentFrameLocals.ToArray();
             };
-            // TODO: Implement child members
+
+            // TODO: Implement child members for current frame, as well as
+            // module members / child members
             return Array.Empty<CompletionResult>();
         }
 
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools/Repl/PythonDebugProcessReplEvaluator.cs
```diff
@@ -275,18 +276,54 @@ public PythonDebugProcessReplEvaluator(IServiceProvider serviceProvider, PythonP
...
+        private async Task<CompletionResult[]> GetMemberNamesAsync(string text) {
+            // TODO: implement support for getting members for module scope,
+            // not just for current frame scope
+            if (_currentScopeName == CurrentFrameScopeFixedName) {
+                var frame = GetFrames().SingleOrDefault(f => f.FrameId == _frameId);
+                if (frame != null) {
+                    AutoResetEvent completion = new AutoResetEvent(false);
+                    PythonEvaluationResult result = null;
+
+                    var expression = string.Format(CultureInfo.InvariantCulture, "str(dir({0}))", text ?? "");
```
zooba: Maybe do `':'.join(dir({0}))` as the expression and then split on `:`?

zooba: Also, do we need to take into account not evaluating calls? I think that's handled already, but I don't recall exactly...

Python/Product/PythonTools/PythonTools/Repl/PythonDebugProcessReplEvaluator.cs
```diff
@@ -275,18 +276,54 @@ public PythonDebugProcessReplEvaluator(IServiceProvider serviceProvider, PythonP
...
             return new OverloadDoc[0];
         }
 
+        private async Task<CompletionResult[]> GetMemberNamesAsync(string text) {
+            // TODO: implement support for getting members for module scope,
+            // not just for current frame scope
+            if (_currentScopeName == CurrentFrameScopeFixedName) {
+                var frame = GetFrames().SingleOrDefault(f => f.FrameId == _frameId);
+                if (frame != null) {
+                    AutoResetEvent completion = new AutoResetEvent(false);
```
zooba: Use a `using` block or a `ManualResetEventSlim` object.

PR #3345, opened 2017-11-15
Experimental conda support

huguesv:
I broke out of bug jail for a few hours to bring this up to date and have it disabled by default using experimental options registry keys.

- If a global Anaconda installation is found, it uses it to automatically detect Conda environments.
- When conda is available for an environment, it uses it as the package manager.

There are some TODOs that explain what remains to be done, as well as potential upcoming changes to conda that will require some work on our part (our contact will let us know about them when ready).

I had to make some changes to the package manager interfaces and UI in order to make it more generic (remove any reference to PyPI). Hopefully they don't break anything, since that's not behind the experimental flags.

I hadn't gotten to this yet, but I wanted to make it so that an environment can have multiple package managers, so that the UI can display all of them (pip, conda). This is part of the reason I made the experimental options for env detection and conda package manager separate for now (so you can still use pip if you want or if our conda package manager causes issues).

FYI my first attempt at implementing this had the discovery of root and user-created conda environments in the same factory. It was a much more complicated change, which required refactoring a bunch of code in the existing 'regular' factory for reuse. It introduced ambiguity between environments detected by our 'regular' factory and 'conda' factory, which I was partially able to resolve by introducing priorities (but not completely solve - add custom environment was still broken at that point). I realized that our MEF architecture for plugin discoverers is working well for adding environments from different sources, but having multiple factories try to discover from the same source just to be able to provide a different package manager doesn't work out.

We can wait as long as we want to merge this, I was just tired of having it sit in a branch and not be visible.

### DIFF
--- a/Python/Product/Analysis/Interpreter/IPackageManager.cs
+++ b/Python/Product/Analysis/Interpreter/IPackageManager.cs
@@ -52,19 +52,19 @@ namespace Microsoft.PythonTools.Interpreter {
         event EventHandler IsReadyChanged;
 
         /// <summary>
-        /// Name to display for the extension in the environments window.
+        /// Localized name to display for the extension in the environments window.
         /// Example: 'Packages (PyPI)'
         /// </summary>
         string ExtensionDisplayName { get; }
 
         /// <summary>
-        /// Name of the index where packages are fetched from.
+        /// Localized name of the index where packages are fetched from.
         /// Example: 'PyPI'
         /// </summary>
         string IndexDisplayName { get; }
 
         /// <summary>
-        /// Watermark for the search query text box.
+        /// Localized watermark text for the search query text box.
         /// Example: 'Search PyPI and installed packages'
         /// </summary>
         string SearchHelpText { get; }
@@ -78,7 +78,7 @@ namespace Microsoft.PythonTools.Interpreter {
         /// <summary>
         /// Returns if the specified package is allowed to be uninstalled.
         /// </summary>
-        bool CanBeUninstalled(PackageSpec package);
+        bool CanUninstall(PackageSpec package);
 
         /// <summary>
         /// Prepares the package manager for use. This only needs to be called
--- a/Python/Product/Analysis/Interpreter/IPackageManager.cs
+++ b/Python/Product/Analysis/Interpreter/IPackageManager.cs
@@ -52,6 +52,35 @@ namespace Microsoft.PythonTools.Interpreter {
         event EventHandler IsReadyChanged;
 
         /// <summary>
+        /// Name to display for the extension in the environments window.
+        /// Example: 'Packages (PyPI)'
+        /// </summary>
+        string ExtensionDisplayName { get; }
+
+        /// <summary>
+        /// Name of the index where packages are fetched from.
+        /// Example: 'PyPI'
+        /// </summary>
+        string IndexDisplayName { get; }
+
+        /// <summary>
+        /// Watermark for the search query text box.
+        /// Example: 'Search PyPI and installed packages'
+        /// </summary>
+        string SearchHelpText { get; }
+
+        /// <summary>
+        /// Returns a description for the command that appears in search result.
+        /// Example: 'pip install mypackage from PyPI'
+        /// </summary>
+        string GetInstallCommandDisplayName(string searchQuery);
+
+        /// <summary>
+        /// Returns if the specified package is allowed to be uninstalled.
+        /// </summary>
+        bool CanBeUninstalled(PackageSpec package);
+
+        /// <summary>
         /// Prepares the package manager for use. This only needs to be called
         /// if <see cref="IsReady"/> is false. After successful completion,
         /// <see cref="IsReady"/> should be true.
--- a/Python/Product/Analysis/Interpreter/NoPackageManager.cs
+++ b/Python/Product/Analysis/Interpreter/NoPackageManager.cs
@@ -28,10 +28,20 @@ namespace Microsoft.PythonTools.Interpreter {
         public bool IsReady => true;
         public IPythonInterpreterFactory Factory => null;
 
+        public string ExtensionDisplayName => string.Empty;
+
+        public string IndexDisplayName => string.Empty;
+
+        public string SearchHelpText => throw new NotImplementedException();
+
         public event EventHandler InstalledFilesChanged { add { } remove { } }
         public event EventHandler InstalledPackagesChanged { add { } remove { } }
 
         public event EventHandler IsReadyChanged { add { } remove { } }
+
+        public bool CanBeUninstalled(PackageSpec package) {
+            return true;
+        }
 
         public Task<PackageSpec> GetInstalledPackageAsync(string name, CancellationToken cancellationToken) {
             return Task.FromResult(new PackageSpec());
@@ -79,6 +89,10 @@ namespace Microsoft.PythonTools.Interpreter {
 
         public Task<PackageSpec> GetInstallablePackageAsync(PackageSpec package, CancellationToken cancellationToken) {
             return Task.FromResult(new PackageSpec());
+        }
+
+        public string GetInstallCommandDisplayName(string searchQuery) {
+            throw new NotImplementedException();
         }
     }
 }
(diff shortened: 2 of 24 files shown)


Python/Product/Analysis/Interpreter/IPackageManager.cs
```diff
@@ -52,6 +52,35 @@ public interface IPackageManager {
...
+        /// <summary>
+        /// Returns a description for the command that appears in search result.
+        /// Example: 'pip install mypackage from PyPI'
+        /// </summary>
+        string GetInstallCommandDisplayName(string searchQuery);
+
+        /// <summary>
+        /// Returns if the specified package is allowed to be uninstalled.
+        /// </summary>
+        bool CanBeUninstalled(PackageSpec package);
```
zooba: `CanUninstall`

Python/Product/Analysis/Interpreter/IPackageManager.cs
```diff
@@ -52,6 +52,35 @@ public interface IPackageManager {
         event EventHandler IsReadyChanged;
 
         /// <summary>
+        /// Name to display for the extension in the environments window.
+        /// Example: 'Packages (PyPI)'
+        /// </summary>
+        string ExtensionDisplayName { get; }
```
zooba: I assume this is localized text? Do we need an unlocalized version (e.g. for settings)?

Python/Product/Analysis/Interpreter/NoPackageManager.cs
```diff
@@ -80,5 +90,9 @@ public sealed class NoSuppression : IDisposable {
         public Task<PackageSpec> GetInstallablePackageAsync(PackageSpec package, CancellationToken cancellationToken) {
             return Task.FromResult(new PackageSpec());
         }
+
+        public string GetInstallCommandDisplayName(string searchQuery) {
+            throw new NotImplementedException();
```
zooba: Don't like throwing `NotImplementedException` from real code. This doesn't seem like a good candidate for `NotSupportedException`, so probably should just return an empty string.

Python/Product/Common/Strings.resx
```diff
@@ -2126,4 +2126,40 @@ Expand selection to the full expression?</value>
...
+    <value>Search PyPI and installed packages</value>
+  </data>
+  <data name="CondaExtensionCondaInstall" xml:space="preserve">
+    <value>"conda install {0}"</value>
+    <comment>Do not translate.
+{0} is the search query</comment>
+  </data>
+  <data name="CondaExtensionCondaInstallFrom" xml:space="preserve">
+    <value> from {0}</value>
+    <comment>{0} is name of package index. This string is appended to the string resource CondaExtensionCondaInstall</comment>
```
zooba: I don't remember why (presumably) I did the strings separately like this in the first place, but we shouldn't be concatentating strings. If sometimes we want to skip the "from X" part, we should have two complete strings `"pip install X"` and `"pip install X" from Y`

Python/Product/VSInterpreters/CondaEnvironmentFactoryProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        private void EnsureInitialized() {
+            lock (this) {
+                if (!_initialized) {
+                    _initialized = true;
+                    _environmentsTxtPath = Path.Combine(
+                        Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+                        ".conda",
+                        "environments.txt"
+                    );
```
zooba: Historically, we've found this kind of code to be pretty fragile. Should handle any errors that occur and quietly ignore the environments.txt file if we can't get a path.

Python/Product/VSInterpreters/CondaEnvironmentFactoryProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+                    }
+                }
+            }
+        }
+
+        private async void _envsTxtWatcherTimer_Elapsed(object state) {
+            try {
+                _envsTxtWatcherTimer.Change(Timeout.Infinite, Timeout.Infinite);
+            } catch (ObjectDisposedException) {
+            }
```
zooba: If we're disposed, do we still want to discover?

Python/Product/VSInterpreters/CondaEnvironmentFactoryProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+            var interpreterPath = Path.Combine(prefixPath, CondaEnvironmentFactoryConstants.ConsoleExecutable);
+            var windowsInterpreterPath = Path.Combine(prefixPath, CondaEnvironmentFactoryConstants.WindowsExecutable);
+
+            InterpreterArchitecture arch = InterpreterArchitecture.Unknown;
+            Version version = null;
+
+            if (File.Exists(interpreterPath)) {
+                using (var output = ProcessOutput.RunHiddenAndCapture(
+                    interpreterPath, "-c", "import sys; print('%s.%s' % (sys.version_info[0], sys.version_info[1]))"
+                )) {
```
zooba: In theory, we could query `conda` for this information, right?

Python/Tests/Utilities.Python/MockPackageManager.cs
```diff
@@ -20,6 +20,16 @@ public class MockPackageManager : IPackageManager {
         public event EventHandler InstalledPackagesChanged;
         public event EventHandler IsReadyChanged { add { throw new NotImplementedException(); } remove { } }
 
+        public string ExtensionDisplayName => string.Empty;
+
+        public string IndexDisplayName => string.Empty;
+
+        public string SearchHelpText => string.Empty;
+
+        public string GetInstallCommandDisplayName(string searchQuery) => string.Empty;
```
zooba: We probably want better values for these, in case we use MockPackageManger instances in UI tests (which I suspect we may, as we fake some Environment Window tests).

karthiknadig:
Looks like some of the fixes here to `ExperimentalOptions` are needed for my work.

zooba:
@karthiknadig Probably. Unless @huguesv says he'll get this wrapped up today (which I wouldn't expect), maybe just steal his changes :)

Python/Tests/Utilities.Python/MockPackageManager.cs
huguesv: It's only ever used in a TextBlock that appears in the list of results. It's easy enough to change it, I'll make it match what our pip package manager returns.

Python/Product/VSInterpreters/CondaEnvironmentFactoryProvider.cs
huguesv: We could query the list of packages using conda.exe, and see which version of python is installed, but that would be much slower than the above code. Also, I was curious and checked the file properties on conda.exe and turns out there's no version resources at all on it (not that I would want to use that here but I was hoping to use it somewhere else)

Python/Product/Analysis/Interpreter/IPackageManager.cs
huguesv: Not at this time, but I see what you're saying. I can add one but it would be unused for now.

Python/Product/VSInterpreters/CondaEnvironmentFactoryProvider.cs
zooba: You're right. Ultimately less reliable too.

I'm pretty sure conda.exe is just a standard script wrapper generated by setuptools or a similar library (distlib is common), so it's not going to have any normal features. It's on my list of "projects I'd love to have time to create"

Python/Product/VSInterpreters/CondaEnvironmentFactoryProvider.cs
huguesv: That was probably an unnecessary precaution given that nobody disposes of the Timer (the factory provider is not disposable). If it was to become disposable, I agree it makes no sense to discover. I'll move it inside the try block.

Python/Product/Analysis/Interpreter/IPackageManager.cs
zooba: Let's just clarify that it should be localized in the comment.

huguesv:
I made the requested changes and added UI for experimental options.

zooba:
👍

zooba:
Let's add a surprise feature then 🎉

PR #3349, opened 2017-11-16
Fixes #1645 Cannot parse trailing commas in function definitions

### LINKED ISSUE #1645: Cannot parse trailing commas in function definitions
The following code should parse correctly:

``` python
        # Check trailing commas are permitted in funcdef argument list
        def f(a,): pass
        def f(*args,): pass
        def f(**kwds,): pass
        def f(a, *args,): pass
        def f(a, **kwds,): pass
        def f(*args, b,): pass
        def f(*, b,): pass
        def f(*args, **kwds,): pass
        def f(a, *args, b,): pass
        def f(a, *, b,): pass
        def f(a, *args, **kwds,): pass
        def f(*args, b, **kwds,): pass
        def f(*, b, **kwds,): pass
        def f(a, *args, b, **kwds,): pass
        def f(a, *, b, **kwds,): pass
```

Currently we report errors because of the trailing commas.

zooba:
Fixes #1645 Cannot parse trailing commas in function definitions
Rewrites function definition parser to be more reliable. We now parse all syntax valid in any version and then raise errors later.
Fixes formatting/round-trip of parameters.
Fixes finding variables to rename.
Ensures more error messages are returned from analyzer initialization.

### DIFF
--- a/Python/Product/Analysis/Parsing/Parser.cs
+++ b/Python/Product/Analysis/Parsing/Parser.cs
@@ -20,9 +20,11 @@ using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Numerics;
 using System.Text;
 using System.Text.RegularExpressions;
+using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.Parsing.Ast;
 
 namespace Microsoft.PythonTools.Parsing {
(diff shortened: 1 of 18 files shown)


Python/Product/Analysis/Parsing/Parser.cs
```diff
@@ -1986,338 +1992,268 @@ struct Name {
...
-                }
-            }
-            return MakeTupleOrExpr(list, itemWhiteSpace, trailingComma, parenFreeTuple);
-        }
+                    ReportSyntaxError(e.StartIndex, e.EndIndex, "invalid sublist parameter");
+                }
+            }
+        }
+
+        //private Expression LEGACY_ParseSublistParameter(HashSet<string> names) {
```
huguesv: Just checking that you want to keep this around

PR #3356, opened 2017-11-16
Fix elevation check for conda package manager.

### LINKED ISSUE #3354: Correctly detect when we should elevate to install conda packages
The existing logic in the package manager ui doesn't work with conda.

I've described what we need to do in the comments of CondaPackageManager.ShouldElevate.

huguesv:
Fix #3354

### DIFF
--- a/Python/Product/VSInterpreters/CondaPackageManager.cs
+++ b/Python/Product/VSInterpreters/CondaPackageManager.cs
@@ -164,42 +164,38 @@ namespace Microsoft.PythonTools.Interpreter {
             }
         }
 
-        private Task<bool> ShouldElevate(IPackageManagerUI ui, string operation) {
-            // Apply the same logic as conda.gateways.disk.test.prefix_is_writable()
-            // Check for errors when opening in append mode for first file that matches:
-            // - ./conda-meta/history
-            // - ./conda-meta/conda*.json
-            // - ./conda-meta/*.json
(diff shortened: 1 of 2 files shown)


Python/Product/VSInterpreters/CondaPackageManager.cs
```diff
@@ -164,21 +165,41 @@ class CondaPackageManager : IPackageManager, IDisposable {
...
+                    using (var fileStream = new FileStream(filePath, FileMode.Append)) {
+                        result = Task.FromResult(false);
+                    }
+                }
+            } catch (ArgumentException) {
+            } catch (IOException) {
+            } catch (NotSupportedException) {
+            } catch (UnauthorizedAccessException) {
+                result = Task.FromResult(true);
+            } catch (SecurityException) {
```
int19h: This shouldn't be needed (or should count as inconclusive), since it's only raised for CAS, and doesn't have anything to do with filesystem permissions.

huguesv:
I just realized that this doesn't respect the "always run pip as administrator" option (it only does when the check is inconclusive - ui.ShouldElevateAsync that checks it)

I'll change it to call that first, and only if that returns false then go with our check.

I guess I'll rename the option to "always run pip and conda as administrator".

zooba:
Make it "always run package managers as Administrator"

PR #3364, opened 2017-11-18
Refactor analyzer dependencies.

zooba:


zooba:
Yes, this has some breaking public API changes. There isn't really a good workaround here - type forwarding requires circular dependencies. I'm hoping the Preview releases will flush out any real-world problems and we can evaluate them then - eventually porting to .NET Core is going to require these breaks anyway, so it's either now or later.

zooba:
@int19h Could you take a look at this for me? The conflict is just a moved file - I'll make sure the other changes get into it before merging.

PR #3365, opened 2017-11-18
Experimental setting and adapter launcher for VSCode based debugger 

karthiknadig:
If the experimental flag is enabled and the debugger is not available this is the error you will see:
image


Python/Product/Debugger/DebugAdapterLauncher.cs
```diff
@@ -0,0 +1,45 @@
...
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using Microsoft.PythonTools.DkmDebugger;
```
int19h: What's coming from this namespace? It shouldn't be needed for pure Python debugging.

Also, can you sort the usings?

Python/Product/PythonTools/PythonTools/Debugger/DebugLaunchHelper.cs
```diff
@@ -84,40 +85,82 @@ public static class DebugLaunchHelper {
             return SubstitutionPattern.Replace(
                 str,
                 m => {
```
int19h: No need for braces or `return` now, either.

Python/Product/PythonTools/PythonTools/Debugger/DebugLaunchHelper.cs
```diff
@@ -84,40 +85,82 @@ public static class DebugLaunchHelper {
...
+            var pyService = provider.GetPythonToolsService();
+            return string.Join(";", GetGlobalDebuggerOptions(pyService)
+                .Concat(GetLaunchConfigurationOptions(config))
+                .Where(s => !string.IsNullOrEmpty(s))
+                .Select(s => s.Replace(";", ";;"))
+            );
+        }
+
+        private static string GetLaunchJsonForVsCodeDebugAdapter(IServiceProvider provider, LaunchConfiguration config) {
+            bool nativeDebug = config.GetLaunchOption(PythonConstants.EnableNativeCodeDebugging).IsTrue();
```
int19h: This should never be true on this code path. Basically, if native debugging is enabled, we should be using the old engine (which in that case just bails out and lets Concord handle it).

Python/Product/PythonTools/SharedProject/Debugger/ProvideDebugAdapterAttribute.cs
```diff
@@ -0,0 +1,129 @@
...
+             * This appears in VS in several places.  For example:
+             *     -The "Select Code Type" dialog for choosing which debugger to attach to a process(if Attach is supported)
+             *     -The "Debugging" column of the "Processes" tool window
+             */
+            engineKey.SetValue("Name", _name);
+
+            /*
+             * Path to the debug adapter executable
+             */
+            engineKey.SetValue("Adapter", @"$PackageFolder$\DebugAdapter.exe");
```
int19h: Do we actually need this, given that we'll have a custom `IAdapterLauncher` that provides the streams directly, without another process?

Python/Product/PythonTools/SharedProject/Debugger/ProvideDebugAdapterAttribute.cs
```diff
@@ -0,0 +1,129 @@
...
+             * This appears in (e.g.) the "Language" column of the Stack Trace tool window.
+             */
+            engineKey.SetValue("Language", _languageName);
+            engineKey.SetValue("LanguageId", _languageId);
+
+            /* 
+             * Adapter launcher registration 
+             */
+            var adapterKey = context.CreateKey($"CLSID\\{_adapterLauncherCLSID}");
+            adapterKey.SetValue("Assembly", "Microsoft.PythonTools.Debugger");
```
int19h: Might be a good idea to use `typeof(...).FullName` here, to make this refactoring-proof.

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -0,0 +1,68 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using Microsoft.VisualStudio.Debugger.Interop.VSCodeDebuggerHost;
+
+namespace Microsoft.PythonTools {
```
zooba: Can these go in the Debugger namespace?

Python/Product/Debugger/Debugger.csproj
```diff
@@ -46,6 +46,11 @@
     <Reference Include="envdte80, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
     <Reference Include="Microsoft.VisualStudio.ComponentModelHost, Version=$(VSTarget).0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="Microsoft.VisualStudio.Debugger.Engine, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.VisualStudio.Debugger.Interop.VSCodeDebuggerHost, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.VisualStudio.Debugger.InteropA, Version=9.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
+      <HintPath>..\..\packages\Microsoft.VisualStudio.Debugger.InteropA.9.0.21022\lib\Microsoft.VisualStudio.Debugger.InteropA.dll</HintPath>
+      <Private>True</Private>
```
zooba: Neither of these properties should be needed. We default all references to non-private (since they mostly are), interop types should be embedded, and all the nuget packages get added to the assembly search path automatically.

Setting any of them in the project file is likely to break that (and I know this was what VS added - just clarifying why we need to override them).

Python/Product/Debugger/Debugger.csproj
```diff
@@ -239,6 +245,7 @@
       <SubType>Designer</SubType>
     </None>
     <Content Include="DkmDebugger\PythonDkm.natvis" />
+    <None Include="packages.config" />
```
zooba: Don't add nuget dependencies directly to the project. Add them to the `Build\15.0\packages.config` file and run `PreBuild.ps1` again.

Python/Product/PythonTools/PythonTools.csproj
```diff
@@ -692,6 +692,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="PkgCmdID.cs" />
     <Compile Include="PythonTools\WebProject\PythonWebPropertyPage.cs" />
+    <Compile Include="SharedProject\Debugger\ProvideDebugAdapterAttribute.cs" />
```
zooba: Don't add files to the SharedProject folder - it isn't real (it's actually under `Common` rather than `Python`, and all its files are linked in).

This one can go under `PythonTools\Debugger`.

Python/Product/PythonTools/PythonTools/Debugger/DebugLaunchHelper.cs
```diff
@@ -83,41 +84,79 @@ public static class DebugLaunchHelper {
...
+            JObject jsonObj = new JObject {
+                ["exe"] = config.GetInterpreterPath(),
+                ["curDir"] = string.IsNullOrEmpty(config.WorkingDirectory) ? PathUtils.GetParent(config.ScriptName) : config.WorkingDirectory,
+                ["remoteMachine"] = "",
+                ["args"] = GetArgs(config),
+                ["options"] = GetOptions(provider, config),
+                ["env"] = envArray
+            };
+            
+            // Note: these are optional, but sepcial. These override the pkgdef version of the adapter settings
```
zooba: "special"

Python/Product/PythonTools/PythonTools/Options/ExperimentalOptions.cs
```diff
@@ -61,5 +64,11 @@ public sealed class ExperimentalOptions {
         /// True to use conda for package management when available.
         /// </summary>
         public bool UseCondaPackageManager { get; set; }
+
+        /// <summary>
+        /// True to use new Ptvs debugger built to run on VS Code Debug Adapter Host
```
zooba: Always "PTVS"

zooba: And rarely needs to be named within our own project - this could just read "Use a debugger built to run on VS Code Debug Adapter Host"

Python/Product/PythonTools/PythonTools/Options/PythonExperimentalOptionsControl.Designer.cs
```diff
@@ -26,9 +26,10 @@ partial class PythonExperimentalOptionsControl {
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(PythonExperimentalOptionsControl));
             this._noDatabaseFactory = new System.Windows.Forms.CheckBox();
             this.tableLayoutPanel2 = new System.Windows.Forms.TableLayoutPanel();
-            this._mustRestartLabel = new System.Windows.Forms.Label();
```
zooba: Double-check that the tab order has not changed in this dialog.

Python/Product/VSInterpreters/ExperimentalOptions.cs
```diff
@@ -23,13 +23,16 @@ public static class ExperimentalOptions {
         internal const string NoDatabaseFactoryKey = "NoDatabaseFactory";
         internal const string AutoDetectCondaEnvironmentsKey = "AutoDetectCondaEnvironments";
         internal const string UseCondaPackageManagerKey = "UseCondaPackageManager";
+        internal const string UseVsCodeDebuggerKey = "UseVsCodeDebugger";
         internal static readonly Lazy<bool> _noDatabaseFactory = new Lazy<bool>(GetNoDatabaseFactory);
         internal static readonly Lazy<bool> _autoDetectCondaEnvironments = new Lazy<bool>(GetAutoDetectCondaEnvironments);
         internal static readonly Lazy<bool> _useCondaPackageManager = new Lazy<bool>(GetUseCondaPackageManager);
+        internal static readonly Lazy<bool> _useVsCodeDebugger = new Lazy<bool>(GetUseNewPtvsDebugger);
```
zooba: GetUseVsCodeDebugger

Python/Product/VSInterpreters/ExperimentalOptions.cs
```diff
@@ -23,13 +23,16 @@ public static class ExperimentalOptions {
...
 
         public static bool GetNoDatabaseFactory() => GetBooleanFlag(NoDatabaseFactoryKey, defaultVal: true);
         public static bool GetAutoDetectCondaEnvironments() => GetBooleanFlag(AutoDetectCondaEnvironmentsKey, defaultVal: false);
         public static bool GetUseCondaPackageManager() => GetBooleanFlag(UseCondaPackageManagerKey, defaultVal: false);
+        public static bool GetUseNewPtvsDebugger() => GetBooleanFlag(UseVsCodeDebuggerKey, defaultVal: false);
```
zooba: GetUseVsCodeDebugger

Python/Product/Debugger/Debugger.csproj
int19h: We don't embed any interop assemblies in this project (note the references below). I don't think that's intentional, though.

zooba: Yeah, you're right. Embedding interop types isn't actually the default.

General point is, use the default settings until they fail.

Python/Product/PythonTools/PythonTools/Intellisense/ExtractMethodSuggestedActionSourceProvider.cs
```diff
@@ -82,7 +82,9 @@ class ExtractMethodSuggestedActionSource : ISuggestedActionsSource {
                 if (MethodExtractor.CanExtract(_view) == true) {
                     return new SuggestedActionSet[] {
                     new SuggestedActionSet(
+                        requestedActionCategories.First(),
```
zooba: Unrelated change? Should safely handle `InvalidOperationException` if you're going to use `First()`.

karthiknadig: I am going to undo this one. I was not able to build because VS kept complaining that the version that is used was Obsolete.

PR #3368, opened 2017-11-20
Add experimental basic ptvsd-pydevd support to launcher.

int19h:
Adds a new command line switch to skip the bundled ptvsd package.

### DIFF
--- a/Python/Product/PythonTools/ptvsd_launcher.py
+++ b/Python/Product/PythonTools/ptvsd_launcher.py
@@ -28,41 +28,30 @@ import os.path
 import sys
 import traceback
 
-try:
-    ptvs_lib_path = os.path.dirname(__file__)
-    sys.path.insert(0, ptvs_lib_path)
-    import ptvsd.debugger as vspd
-except:
-    traceback.print_exc()
-    print('''
-Internal error detected. Please copy the above traceback and report at
-https://go.microsoft.com/fwlink/?LinkId=293415
-
-Press Enter to close. . .''')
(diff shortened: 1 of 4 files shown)


Python/Product/PythonTools/ptvsd_launcher.py
```diff
@@ -28,42 +28,31 @@
...
 port_num = int(sys.argv[2])
 debug_id = sys.argv[3]
-debug_options = vspd.parse_debug_options(sys.argv[4])
+debug_options = set([opt.strip() for opt in sys.argv[4].split(',')])
+
 del sys.argv[0:5]
 
+# Use bundled ptvsd or not?
+bundled_ptvsd = True
+if sys.argv and sys.argv[0] == '-g':
```
int19h: @karthiknadig - this is the option that you'll need to give to the launcher when experimental flag is checked.

Python/Tests/DjangoUITests/DjangoUITests.csproj
```diff
@@ -176,5 +176,8 @@
       <SubType>Designer</SubType>
     </None>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
int19h: These seem to be some weird VS thing to indicate that it's a test project. Not sure why it suddenly needs them now, but it keeps re-adding it, and so far as I can tell, it's otherwise harmless.

PR #3383, opened 2017-11-23
Fixes IPC tests to prevent lingering processes.

zooba:
This ought to be enough to prevent our test runs from timing out due to that VSTest bug.


zooba:
Worked on my test run, so merging without review.

PR #3395, opened 2017-11-27
Debug Adapter fixes

karthiknadig:


Python/Product/Common/Infrastructure/StringExtensions.cs
```diff
@@ -97,5 +97,7 @@ public static class StringExtensions {
             }
             return str;
         }
+
+        public static string AddQuotes(this string str) => "\"" + str + "\"";
```
AlexanderSher: `$"\"{str}\""` - does it really need an extension method?

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -15,41 +15,112 @@
...
-        private DebugAdapterProcess(ProcessStartInfo psi) {
+        private const int _debuggerConnectionTimeout = 5000; // 5000 ms
+        private const int _connectionCloseTimeout = 5000; // 5000 ms
+
+        private Process _process;
+        private readonly Guid _processGuid;
+        private PythonDebugOptions _debugOptions;
+        private string _interpreterOptions;
+        private int _listenerPort = -1;
+        private Stream _stream;
```
AlexanderSher: How this stream get closed?

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -15,41 +15,112 @@
...
+        private void StartProcess(string launchJson) {
+            InitializeListenerSocket();
+
+            var json = JObject.Parse(launchJson);
+            var exe = json["exe"].Value<string>();
+            var args = json["args"].Value<string>();
+            var cwd = json["cwd"].Value<string>();
+            ParseOptions(json["options"].Value<string>());
+
+            List<string> argsList = new List<string> {
```
AlexanderSher: `var argsList = new List<string> {`

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -64,5 +135,122 @@ public sealed class DebugAdapterProcess : ITargetHostProcess {
...
+            List<string> res = new List<string>();
+            int lastStart = 0;
+            for (int i = 0; i < options.Length; i++) {
+                if (options[i] == ';') {
+                    if (i < options.Length - 1 && options[i + 1] != ';') {
+                        // valid option boundary
+                        res.Add(options.Substring(lastStart, i - lastStart));
+                        lastStart = i + 1;
+                    } else {
+                        i++;
```
AlexanderSher: Why do you need this?

Python/Product/Debugger/DebugAdapterProcess.cs
karthiknadig: On Process exit, we close the socket and the stream.

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -15,41 +15,112 @@
...
+            var json = JObject.Parse(launchJson);
+            var exe = json["exe"].Value<string>();
+            var args = json["args"].Value<string>();
+            var cwd = json["cwd"].Value<string>();
+            ParseOptions(json["options"].Value<string>());
+
+            List<string> argsList = new List<string> {
+                string.IsNullOrWhiteSpace(_interpreterOptions) ? "" : _interpreterOptions,
+                PythonToolsInstallPath.GetFile("ptvsd_launcher.py").AddQuotes(),
+                cwd.Trim('\\').AddQuotes(),
```
int19h: This code should use `ProcessOutput` if possible. If not, then we need to take the code for argument quoting out of it, and make that usable on its own. The quoting is more involved than what's currently done here (consider what happens if the name of the current directory itself has a quote...). 

I'm fine merging this as is for now, just so that we get the working bits, and then clean it up in a separate commit. But then it shouldn't be defining an extension method, since that'll be going away anyway.

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -15,41 +15,112 @@
...
+                WorkingDirectory = cwd,
+                RedirectStandardError = true,
+                RedirectStandardInput = false,
+                RedirectStandardOutput = false,
+                UseShellExecute = false,
+                CreateNoWindow = false,
+            };
+
+            var env = json["env"].Value<JArray>();
+            if (env.Count > 0) {
```
int19h: This check shouldn't be necessary (empty collection is valid to iterate over). Is it meant to be a null check? Or was there something else inside the `if`?

Python/Product/Debugger/DebugAdapterProcess.cs
int19h: This code is taken as is from `AD7Engine`. This particular `if` just skips over any redundant `;`, I think.

Python/Product/Common/Infrastructure/StringExtensions.cs
karthiknadig: I can remove this.

---
In reply to: [153326133](https://github.com/Microsoft/PTVS/pull/3395#discussion_r153326133) [](ancestors = 153326133)

Python/Product/Debugger/DebugAdapterProcess.cs
karthiknadig: the check is not necessary.

---
In reply to: [153330672](https://github.com/Microsoft/PTVS/pull/3395#discussion_r153330672) [](ancestors = 153330672)

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -64,5 +135,122 @@ public sealed class DebugAdapterProcess : ITargetHostProcess {
...
+                res.Add(options.Substring(lastStart, options.Length - lastStart));
+            }
+            return res.ToArray();
+        }
+
+        private void ParseOptions(string options) {
+            foreach (var optionSetting in SplitOptions(options)) {
+                var setting = optionSetting.Split(new[] { '=' }, 2);
+                if (setting.Length == 2) {
+                    switch (setting[0]) {
```
AlexanderSher: should we trim spaces?

int19h: This code is taken from elsewhere in the product as is, so it behaves exactly as the other side. Note the comment above saying it should all be refactored into a single reusable class.

AlexanderSher: Ok, got it

PR #3402, opened 2017-11-28
Fix build for 15.6

### LINKED ISSUE #3398: Build broken on 15.6.0 Preview 1
```
PythonNavigateToItemProvider.cs(140,59,140,68): error CS0618: 'MatchKind' is obsolete: 'MatchKind is being replaced with PatternMatch and PatternMatchKind. Please remove all references and replace them with PatternMatch objects.'
```

huguesv:
Fix #3398 

I added the assemblies that are needed to build so that build doesn't break on previous versions, though it won't work very well (exception when you try to use IntelliSense). At least that will give us some time to update the build and test VMs.

I got the result from the build machine, and that didn't work (I can build locally on 15.4.5).

```

C:\Program Files   (x86)\Microsoft Visual   Studio\2017\Enterprise\MSBuild\Microsoft\VisualStudio\v15.0\VSSDK\Microsoft.VsSDK.targets(979):C:\Program   Files (x86)\Microsoft Visual   Studio\2017\Enterprise\MSBuild\Microsoft\VisualStudio\v15.0\VSSDK\Microsoft.VsSDK.targets(979,5):   Error : CreatePkgDef : error : ReflectionTypeLoadException: Unable to load   one or more of the requested types. Retrieve the LoaderExceptions property   for more information. at   
System.Reflection.RuntimeModule.GetTypes(RuntimeModule module) at   
System.Reflection.RuntimeModule.GetTypes() at
   System.Reflection.Assembly.GetTypes() at   
Microsoft.VisualStudio.Tools.CreatePkgDef.ProcessAssembly(String fileName,   Hive hive, PkgDefContext context, Boolean register, RegistrationMode mode) in   
E:\A\_work\2\s\src\product\vssdk\tools\CreatePkgDef\CreatePkgDef.cs:line 327   at 
Microsoft.VisualStudio.Tools.CreatePkgDef.DoCreatePkgDef(InputArguments   inputArguments) in   
E:\A\_work\2\s\src\product\vssdk\tools\CreatePkgDef\CreatePkgDef.cs:line 201   at 
Microsoft.VisualStudio.Tools.CreatePkgDef.Main(String[] arguments) in   E:\A\_work\2\s\src\product\vssdk\tools\CreatePkgDef\CreatePkgDef.cs:line 93   Could not load file or assembly 'System.Collections.Immutable,   Version=1.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of   its dependencies. The system cannot find the file specified. Could not load   file or assembly 'System.Collections.Immutable, Version=1.2.1.0,   Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.   The system cannot find the file specified.
--
Process   'msbuild.exe' exited with code '1'.

```


huguesv:
Ok, that error also happens locally if you delete the BuildOutput folder before building. On subsequent build (even rebuild), there's no error. Also when the error happens, the assembly it complains about (system.collections.immutable.dll v1.2.1.0) is present in the binaries folder. Not sure what's going on.

zooba:
System.Collections.Immutable is a dependency that we only need because you copied the assembly into the references folder. Ideally we can avoid having to copy the assembly entirely, otherwise we'll need to bring in its dependencies too.

huguesv:
I submitted https://github.com/Microsoft/PTVS/pull/3410 which disables deprecation warning.

PR #3404, opened 2017-11-28
Update executable name for IronPython 2.7.8 and above.

slide:
In IronPython 2.7.8, we will be changing the file names for the 32-bit specific and Any CPU executables. ipy.exe will now be Any CPU and ipy32.exe will be the 32-bit (x86) executable. This is to make the Any CPU the default build.


huguesv:
These probably need to change too?
https://github.com/Microsoft/PTVS/blob/ec073b7536983feb5638e8575cd24d1118ff563c/Python/Product/IronPython/Interpreter/IronPythonInterpreterFactory.cs#L50

huguesv:
Thanks! Others can comment but LGTM. We'll need you to sign the CLA before we can merge.

PR #3407, opened 2017-11-29
Support to profile python via an "external" profiler (initially VTune)

rgesteve:
Initial payload for the external process to forward workload description to the profiler, handles data collection and transformation of results in a form understandable by PTVS.  Invocation and consumption to the service from PTVS is still not included.   Please consider this work in flux.


zooba:
We'd certainly like to see the builds aligned with the rest of the project, as that's the easiest way to make sure everything is built, signed, and localized correctly. Once you're closer to merging I can help make sure that's right, but an easy starting point is to copy an existing .csproj file.

Our build doesn't support nuget packages embedded in project files. In general, avoid needing too many packages, and add them to `Build\15.0\packages.config` and re-run `Build\Prebuild.ps1` to install them.

We'll also eventually normalize the formatting against the rest of our project, particularly the header comments and newlines. Feel free to do those whenever you like.

rgesteve:
@afaruqi is submitting an updated version of this code.

Issue #507, opened 2015-06-23
Commit characters other than enter/tab should not apply within a word


zooba:
Type `x = d`, then go to before the `d` and type `int(`. The result should be `x = int(d` but is actually `x = InterruptedError(` because we committed `intd`.

When the caret is within the applicable span and the completion is being committed with anything other than enter/tab/mouse, we should dismiss rather than commit.

zooba:
Won't get back to this for a few days, but here's my plan:
* add a flag to `IntellisenseController.TriggerCompletionSession` to indicate manual/auto trigger
* change to create and start the session in two steps
* attach a flag to the session indicating trigger type
* use the flag when getting the applicable span to determine whether the span can extend to the right of the caret (manual) or not (auto)

MikhailArkhipov:
I can take this

MikhailArkhipov:
a) applicable span is calculated from the classifier and in `i|d` it gets 'id' which then grows and `d` is eaten by to completion commit.
b) C# does not bring completion when typing in front of the existing construct, it requires space.

So options are 

a) If we still want completion, determine applicable span from the previous snapshot and/or tell completion invoked via `Ctrl+J` from typing: the first replaces construct, the latter does not.

b) Suppress completion line C# does.

RTVS does both (a) and (b) but it uses AST rather than classifier so it does not take into account just typed text.

MikhailArkhipov:
https://github.com/Microsoft/PTVS/commit/deccfe5b202d9c9e9e9edbaeb46d10ed1f2435a1

PR #3409, opened 2017-11-29
#507 Commit characters other than enter/tab should not apply

MikhailArkhipov:
- When in auto-complete mode make applicable span to be left part of the token to the current caret position.
- In explicit mode (`Ctrl+J`) use complete token for replace

Discussed this with @DinoV 
No entirely sure about naming, there is `ForceIntellisense` flag but looks that also `completeWord` term is used for the mode.


huguesv:
If you haven't already, you should run the PythonToolsMockTests.EditorTests. If there's no regression, it probably means we need to add test(s) for this scenario.

FYI on my machine without your changes, they all pass except for `EnterCommitsCompleteNoNewLine`:
`Assert.Fail failed. Expected at least {__class__}, didn't find {__class__}. All: {}`

MikhailArkhipov:
Will do

MikhailArkhipov:
Pending merge of https://github.com/Microsoft/PTVS/issues/3398

MikhailArkhipov:
`CompletionWithLongDocString` - passing on re-run. Might be that some random strings are not valid?
`BuiltinOverrideCompletions` - fails in master
`FromOSPathImportCompletions2x` - fails in master
`FromOSPathImportCompletions3x` - fails in master

Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
```diff
@@ -876,22 +876,29 @@ private class ExpressionCompletionWalker : PythonWalker {
...
-            } else {
-                if (commitByDefault.HasValue) {
-                    foreach (var s in session.CompletionSets.OfType<FuzzyCompletionSet>()) {
-                        s.CommitByDefault = commitByDefault.GetValueOrDefault();
-                    }
+                return;
+            }
+
+            if (completeWord) {
+                session.SetCompleteWordMode();
```
zooba: Don't we need to set this before the first round of completions are calculated? (When we chatted I suggested splitting the TriggerCompletion call into separate "create completion" and "start completion" calls, so we can set these properties in the middle.)

MikhailArkhipov: Unfortunately, session is available only after `broker.TriggerCompletion()` returns. So I use both property and the `IntellisenseController.ForceCompletions` in the extension.

So we want to split Trigger into `ICompletionBroker.CreateCompletion` and the start right?

zooba: Yeah. That's all that Trigger actually does, as well as handling failure to start safely. It's only 3-4 LoC.

MikhailArkhipov: I've done that, pending running tests

PR #3422, opened 2017-11-30
Fixes #82 Use type hints

### LINKED ISSUE #82: Use type hints in Python 3.x to infer variable types if possible
Even though Python ignores type hints at runtime and they can be pretty much any expression, we should use them to infer variable types in a function if we possibly can.

See [PEP 484](https://www.python.org/dev/peps/pep-0484/)

(Migrated from https://pytools.codeplex.com/workitem/394)

zooba:
This is a big change that started off adding automatic support for typeshed and then disabled it, because it wasn't as useful as we hoped, but now includes a lot of support for `typing` both within user code and .pyi files in search paths.

There's still a bit more to come, but it's really just filling out the gaps. The core functionality is pretty solid in my opinion. Suggestions for tests that are missing are welcome.


Python/Product/Analysis/AnalysisSet.cs
```diff
@@ -455,6 +455,69 @@ public static class AnalysisSet {
             return set;
         }
 
+        /// <summary>
+        /// Splits values in the set according to the predicate. Returns true
+        /// if there are any values in <paramref name="trueSet"/> on return.
+        /// </summary>
+        public static bool Split(this IAnalysisSet set, Func<AnalysisValue, bool> predicate, out IAnalysisSet trueSet, out IAnalysisSet falseSet) {
```
huguesv: It seems like we have 2 classes for public IAnalysisSet extension methods. They should probably be all moved to AnalysisSetExtensions.cs.

Python/Product/Analysis/AnalysisSet.cs
```diff
@@ -455,6 +455,69 @@ public static class AnalysisSet {
...
+                empty = CreateUnion(uc);
+            } else {
+                empty = Create();
+            }
+
+            if (set.Count == 0) {
+                trueSet = falseSet = empty;
+                return false;
+            } else if (set is AnalysisValue av) {
+                if (predicate(av)) {
```
huguesv: >predicate [](start = 20, length = 9)

this parameter wasn't validated non-null

Python/Product/Analysis/Values/TypingTypeInfo.cs
```diff
@@ -0,0 +1,329 @@
...
+                        di.AddTypes(
+                            n,
+                            _unit,
+                            ToInstance(args[0]),
+                            None
+                        );
+                        return di;
+                    }) as DictionaryInfo)?.GetKeysView(_unit);
+                case "ValuesView":
+                    if (args.Count < 1) {
```
huguesv: Count won't be 0, that's checked at the top of this method

Python/Product/Analysis/Values/TypingTypeInfo.cs
```diff
@@ -0,0 +1,329 @@
...
+                            }
+                            var seq = t.MakeFromIndexes(n, Entry);
+                            seq.AddTypes(_unit, args.Select(ToInstance).ToArray());
+                            return seq;
+                        });
+                    } catch (KeyNotFoundException) {
+                        return null;
+                    }
+                case "KeysView":
+                    if (args.Count < 1) {
```
huguesv: Count won't be 0, that's checked at the top of this method

Python/Product/Analysis/Values/TypingTypeInfo.cs
```diff
@@ -0,0 +1,329 @@
...
+                        if (name == "ItemsView") {
+                            return (d as DictionaryInfo).GetItemsView(_unit);
+                        }
+                        return d;
+                    } catch (KeyNotFoundException) {
+                        return null;
+                    }
+
+                case "Callable":
+                    if (args.Count == 0) {
```
huguesv: Count won't be 0, that's checked at the top of this method

Python/Product/Analysis/Values/TypingTypeInfo.cs
```diff
@@ -0,0 +1,329 @@
...
+                case "Tuple": return ClassInfo[BuiltinTypeId.Tuple];
+                case "Container": return ClassInfo[BuiltinTypeId.List];
+                case "ItemsView": return ClassInfo[BuiltinTypeId.DictItems];
+                case "Iterable": {
+                        var p = new ProtocolInfo(Entry);
+                        p.AddReference(_node, _unit);
+                        p.AddProtocol(new IterableProtocol(p, AnalysisSet.Empty));
+                        return p;
+                    }
+                case "Iterator": {
```
huguesv: >Iterator [](start = 22, length = 8)

case looks identical to above, merge?

Python/Product/PythonTools/PythonTools/Options/PythonAdvancedEditorOptionsControl.Designer.cs
```diff
@@ -171,6 +173,12 @@ partial class PythonAdvancedEditorOptionsControl {
             this.tableLayoutPanel1.Controls.Add(this._miscOptionsGroupBox, 0, 2);
             this.tableLayoutPanel1.Name = "tableLayoutPanel1";
             // 
+            // contextMenuStrip1
```
huguesv: what is this change for?

Python/Product/Core/Properties/AssemblyInfo.cs
```diff
@@ -41,3 +41,6 @@
 [assembly: ProvideCodeBase(AssemblyName = "Microsoft.PythonTools.VSCommon", CodeBase = "Microsoft.PythonTools.VSCommon.dll", Version = AssemblyVersionInfo.StableVersion)]
 [assembly: ProvideCodeBase(AssemblyName = "Microsoft.PythonTools.VSInterpreters", CodeBase = "Microsoft.PythonTools.VSInterpreters.dll", Version = AssemblyVersionInfo.StableVersion)]
 [assembly: ProvideCodeBase(AssemblyName = "Microsoft.PythonTools.Workspace", CodeBase = "Microsoft.PythonTools.Workspace.dll", Version = AssemblyVersionInfo.StableVersion)]
+
+// HACK: Helps running on 15.5
+[assembly: ProvideCodeBase(AssemblyName = "Microsoft.VisualStudio.Debugger.Interop.VSCodeDebuggerHost", CodeBase = "Microsoft.VisualStudio.Debugger.Interop.VSCodeDebuggerHost.dll")]
```
huguesv: Probably don't want to merge that, right?

zooba: Correct, especially since it isn't even necessary with the other fix I put in.

Python/Product/PythonTools/PythonTools/Options/PythonAdvancedEditorOptionsControl.Designer.cs
zooba: I added and then removed some UI from this control, which will have modified the auto-generated code. Haven't manually reverted this one to the old revision yet.

Python/Product/Analysis/Values/TypingTypeInfo.cs
zooba: We'd switch between `IterableProtocol` and `IteratorProtocol` then, but otherwise they are the same. I can merge - it isn't that ugly.

Python/Product/Analysis/AnalysisSet.cs
zooba: The difference is that this one has set methods, while the other broadcasts the methods over the values. That isn't entirely clear, so I'll add internal documentation to the class to clarify.

zooba: Actually, I think it is clear enough from the class comments already. I'll rename the other class to "AnalysisValueSetExtensions"

zooba:
I'm going to hold off merging until after we have confirmed for the next release. This doesn't need to go in that quickly.

PR #3435, opened 2017-12-05
FromTextBuffer always returned null when using _buffer.

### LINKED ISSUE #3427: "end" tags not appearing in Django completion
After an opening "if", "for" tag we should offer the closing end tag in completions. It seems to have disappeared.

karthiknadig:
Fixes #3427

### DIFF
--- a/Python/Product/Django/Intellisense/DjangoCompletionSource.cs
+++ b/Python/Product/Django/Intellisense/DjangoCompletionSource.cs
@@ -65,13 +65,13 @@ namespace Microsoft.PythonTools.Django.Intellisense {
             var artifactText = doc.HtmlEditorTree.ParseTree.Text.GetText(artifact.InnerRange.Start, artifact.InnerRange.Length);
             artifact.Parse(artifactText);
 
-            ITrackingSpan applicableSpan;
-            var completionSet = GetCompletionSet(session.GetOptions(_serviceProvider), _analyzer, artifact.TokenKind, artifactText, artifact.InnerRange.Start, triggerPoint, out applicableSpan);
+            var completionSet = GetCompletionSet(session.GetOptions(_serviceProvider), _analyzer, artifact.TokenKind, artifactText, artifact.InnerRange.Start, triggerPoint, out _);
             completionSets.Add(completionSet);
         }
 
         protected override IEnumerable<DjangoBlock> GetBlocks(IEnumerable<CompletionInfo> results, SnapshotPoint triggerPoint) {
-            var doc = HtmlEditorDocument.FromTextBuffer(_buffer);
+            var buffers = _buffer.GetContributingBuffers().Where(b => b.ContentType.IsOfType(TemplateHtmlContentType.ContentTypeName));
+            var doc = HtmlEditorDocument.FromTextBuffer(buffers.FirstOrDefault() ?? _buffer);
             if (doc == null) {
                 yield break;
             }


Python/Product/Django/Intellisense/DjangoCompletionSource.cs
```diff
@@ -65,13 +65,13 @@ public DjangoCompletionSource(IGlyphService glyphService, VsProjectAnalyzer anal
...
 
-            ITrackingSpan applicableSpan;
-            var completionSet = GetCompletionSet(session.GetOptions(_serviceProvider), _analyzer, artifact.TokenKind, artifactText, artifact.InnerRange.Start, triggerPoint, out applicableSpan);
+            var completionSet = GetCompletionSet(session.GetOptions(_serviceProvider), _analyzer, artifact.TokenKind, artifactText, artifact.InnerRange.Start, triggerPoint, out ITrackingSpan applicableSpan);
             completionSets.Add(completionSet);
         }
 
         protected override IEnumerable<DjangoBlock> GetBlocks(IEnumerable<CompletionInfo> results, SnapshotPoint triggerPoint) {
-            var doc = HtmlEditorDocument.FromTextBuffer(_buffer);
+            var buffers = _buffer.GetContributingBuffers().Where(b => b.ContentType.TypeName == TemplateHtmlContentType.ContentTypeName);
```
AlexanderSher: @zooba , are there any rules about string comparison/equality?

Python/Product/Django/Intellisense/DjangoCompletionSource.cs
```diff
@@ -65,13 +65,13 @@ public DjangoCompletionSource(IGlyphService glyphService, VsProjectAnalyzer anal
             var artifactText = doc.HtmlEditorTree.ParseTree.Text.GetText(artifact.InnerRange.Start, artifact.InnerRange.Length);
             artifact.Parse(artifactText);
 
-            ITrackingSpan applicableSpan;
-            var completionSet = GetCompletionSet(session.GetOptions(_serviceProvider), _analyzer, artifact.TokenKind, artifactText, artifact.InnerRange.Start, triggerPoint, out applicableSpan);
+            var completionSet = GetCompletionSet(session.GetOptions(_serviceProvider), _analyzer, artifact.TokenKind, artifactText, artifact.InnerRange.Start, triggerPoint, out ITrackingSpan applicableSpan);
```
huguesv: We should be able to use `out _`

Python/Product/Django/Intellisense/DjangoCompletionSource.cs
```diff
@@ -65,13 +65,13 @@ public DjangoCompletionSource(IGlyphService glyphService, VsProjectAnalyzer anal
...
             completionSets.Add(completionSet);
         }
 
         protected override IEnumerable<DjangoBlock> GetBlocks(IEnumerable<CompletionInfo> results, SnapshotPoint triggerPoint) {
-            var doc = HtmlEditorDocument.FromTextBuffer(_buffer);
+            var buffers = _buffer.GetContributingBuffers().Where(b => b.ContentType.TypeName == TemplateHtmlContentType.ContentTypeName);
+            var doc = HtmlEditorDocument.FromTextBuffer(buffers.Count() > 0 ? buffers.First() : _buffer);
```
huguesv: `buffers.FirstOrDefault() ?? _buffer`

Python/Product/Django/Intellisense/DjangoCompletionSource.cs
zooba: For ContentType there's an "IsOfType" or similar method that we should use rather than a string comparison.

PR #3444, opened 2017-12-06
Fix launch when startup file is missing or not set

### LINKED ISSUE #3434: Debugging with no startup file doesn't produce expected error dialog
Create a new python application
Delete the startup file
F5
(Correct) Result: You'll get an error saying no startup file

In project properties, change the debug launch mode to anything except standard Python launcher and then try again.
Result: Either a blinking console window followed by a debugger progress bar dialog, or a Python interactive in a console.
Expected: "no startup file" error like in the standard Python launcher

huguesv:
Fix #3434 

I had previously fixed this but only for the default launcher and didn't realize that some of the others, like web launcher, were still broken.

Slightly improve the error message when the startup file is set but the file doesn't exist.

### DIFF
--- a/Common/Tests/Utilities.UI/UI/VisualStudioApp.cs
+++ b/Common/Tests/Utilities.UI/UI/VisualStudioApp.cs
@@ -566,6 +566,13 @@ namespace TestUtilities.UI {
             }
         }
 
+        /// <summary>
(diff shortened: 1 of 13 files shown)


zooba:
Please hold this checkin until I merge the test fixes.

zooba:
Go ahead whenever you're ready

PR #3462, opened 2017-12-11
More test fixes

zooba:
Skips assertion that seems to be leaking text onto stdout.
Adds handling for error messages besides failing with another error.


zooba:
I have tests running for this, since I'm not 100% sure it'll be a complete fix. But since the issue only repros on the test machines it is pretty hard to track down.

zooba:
That last PR basically fixes the major issues. There are still errors in the IPC stream, but at least we get successful test runs.

PR #3464, opened 2017-12-11
Fixes static analysis complaints

zooba:


Common/Product/SharedProject/UIThread.cs
```diff
@@ -47,6 +49,23 @@ class UIThread : UIThreadBase {
             }
         }
 
+        public void Dispose() {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        ~UIThread() {
```
AlexanderSher: There are no unmanaged resources, you don't need a destructor.

zooba: Maybe, but this is to keep static analysis from complaining, so I'm covering all bases.

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -101,8 +109,22 @@ public sealed class DebugAdapterProcess : ITargetHostProcess {
...
+                if (connection.Wait(_debuggerConnectionTimeout)) {
+                    var socket = connection.Result;
+                    if (socket != null) {
+                        _stream = new NetworkStream(connection.Result, ownsSocket: true);
+                    }
+                } else {
+                    Debug.WriteLine("Timed out waiting for debuggee to connect.", nameof(DebugAdapterProcess));
+                }
+            } catch (AggregateException ex) {
+                Debug.WriteLine("Error waiting for debuggee to connect {0}".FormatInvariant(ex.InnerException ?? ex), nameof(DebugAdapterProcess));
```
AlexanderSher: `_listenerPort` isn't reset to `-1` in case of exception. Is that ok?

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -101,8 +109,22 @@ public sealed class DebugAdapterProcess : ITargetHostProcess {
...
             _process.Start();
             _process.BeginErrorReadLine();
-            if (!_connectedEvent.WaitOne(_debuggerConnectionTimeout)){
-                Debug.WriteLine("Timed out waiting for debuggee to connect.", nameof(DebugAdapterProcess));
+
+            try {
+                if (connection.Wait(_debuggerConnectionTimeout)) {
+                    var socket = connection.Result;
+                    if (socket != null) {
+                        _stream = new NetworkStream(connection.Result, ownsSocket: true);
```
AlexanderSher: Is it possible that `_process` has exited before `_stream` is set?

Python/Product/Debugger/DebugAdapterProcess.cs
zooba: Yeah, an instance of this class is not reused if it fails.

Python/Product/Debugger/DebugAdapterProcess.cs
zooba: Yes, and there are a few other (unlikely) race conditions here. Karthik was going to significantly refactor this class for the next release and switch to our more reliably ProcessOutput class.

AlexanderSher: Ok

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -60,6 +62,27 @@ class CondaEnvironmentFactoryProvider : IPythonInterpreterFactoryProvider {
...
+        }
+
+        ~CondaEnvironmentFactoryProvider() {
+            Dispose(false);
+        }
+
+        protected virtual void Dispose(bool disposing) {
+            if (!_isDisposed) {
+                _isDisposed = true;
+                if (_envsTxtWatcher != null) {
```
AlexanderSher: `_envsTxtWatcherTimer` and `_envsTxtWatcher` are accessed both inside and outside lock statement. Is that ok?

zooba: Didn't see the lock. No, it's not okay, I'll fix.

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -38,21 +38,29 @@ public sealed class DebugAdapterProcess : ITargetHostProcess {
...
             _processGuid = Guid.NewGuid();
         }
 
+        public void Dispose() {
+            if (_stream != null) {
+                _stream.Dispose();
+            }
+            if (_process != null && !_process.HasExited) {
+                _process.Kill();
+                _process.Dispose();
```
huguesv: Should we always call `_process.Dispose()`, as long as it's not null? For example, it might have been killed in `StartProcess`

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -60,6 +62,27 @@ class CondaEnvironmentFactoryProvider : IPythonInterpreterFactoryProvider {
             _globalProvider = globalProvider;
         }
 
+        public void Dispose() {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        ~CondaEnvironmentFactoryProvider() {
+            Dispose(false);
```
AlexanderSher: If we are going to add destructors to every type that implements `IDisposable`, that would create a lot of useless garbage code. Can't we just silence this rule?

zooba: It's a valid rule, and worth it to implement the pattern properly at the time when we add the Dispose method.

Python/Product/Debugger/DebugAdapterProcess.cs
zooba: Yes, you're right.

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
AlexanderSher: There are two good practices about managed and unmanaged resources:
1. Managed object shouldn't own both managed and unmanaged resources. Since finalizer should never access managed objects that can be disposed or finalized, it may have any meaningful logic only if the type is a wrapper for unmanaged resources.
2. Managed object that owns unmanaged resource should own only one resource. This basically means that any managed object that wraps unmanaged resource can be derived from `SafeHandle` and doesn't require its own finalizer implementation.

So in a real application, this rule is valid for an extremely rare case, when you need to create a class to wrap unmanaged resource and you can't use SafeHandle or any of its derived types. In all other cases, manually implemented finalizer won't actually do anything useful.

zooba: That's fine. I'm okay with a little bit of unnecessary code to avoid missing cases where it becomes necessary and to make sure we have good examples of the complete pattern.

AlexanderSher: > to avoid missing cases where it becomes necessary and to make sure we have good examples of the complete pattern

The problem is that it isn't a good pattern. Here are just some of the issues that come with it:
- Performance overhead for tracking object lifetime, putting it into finalizer queue, etc.
- Object can be put into finalizer queue even before its constructor is completed, so you can't put any initialization logic in constructor
- `Dispose(bool)` can be called concurrently from 2 threads, so it has to be thread-safe even for types that are never used in multi-threaded environment (this isn't checked by FxCop).
- Derived class should always call base class finalization logic, so `Dispose(bool)` override should always contain `try/finally`, (also isn't checked by FxCop):
```cs
protected void Dispose(bool disposing) {
    try {
        // Dispose logic
    } finally {
        base.Dispose(disposing)
    }
}
```

And you hardly find a single case in whole PTVS where you need this pattern. CA1063 rule was added into FxCop in .Net 1.0 and was never removed from there. The article https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose contains so many coding issues and even bugs that I can write a separate blog post about it, but the main problem is this:

```cs
      if (disposing) {
         // Free any other managed objects here.
         //
      }
      
      // Free any unmanaged objects here.
      //
```

Originally, this code piece looked like this (https://msdn.microsoft.com/en-us/library/aa720161(v=vs.71).aspx):

```cs
         // If disposing equals true, dispose all managed 
         // and unmanaged resources.
         if(disposing)
         {
            // Dispose managed resources.
            Components.Dispose();
         }
         // Release unmanaged resources. If disposing is false, 
         // only the following code is executed.
         CloseHandle(handle);
         handle = IntPtr.Zero;
         // Note that this is not thread safe.
         // Another thread could start disposing the object
         // after the managed resources are disposed,
         // but before the disposed flag is set to true.
         // If thread safety is necessary, it must be
         // implemented by the client.
```
As you can see, MSDN documentation had originally introduced this anti-pattern. In .Net 2.0, most (if not all) BCL classes were rewritten to use `SafeHandle` approach (https://referencesource.microsoft.com/#mscorlib/system/runtime/interopservices/safehandle.cs) and never mix managed and unmanaged again. BCL still has some leftovers of this pattern being implemented, but they are for backward compatibility or just not fully cleaned up (for instance, see https://referencesource.microsoft.com/#mscorlib/system/io/streamwriter.cs that overrides `Dispose(bool)` but has no finalizer in base class). So instead of CA1063, FxCop should complain about mix of managed and unmanaged resources in one class, but it is too primitive for this kind of analysis.

The funny thing is, FxCop has a rule about empty finalizers: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1821-remove-empty-finalizers. Too bad it isn't smart enough to analyze call chain and find finalizers that has code but do nothing.

zooba:
@AlexanderSher FYI, I backed out the changes to the dispose pattern. FxCop still hates it, and the current ruleset is a VS requirement.

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -38,21 +38,31 @@ public sealed class DebugAdapterProcess : ITargetHostProcess {
...
             _processGuid = Guid.NewGuid();
         }
 
+        public void Dispose() {
+            if (_stream != null) {
+                _stream.Dispose();
+            }
+            if (_process != null) {
+                if (!_process.HasExited) {
+                    _process.Kill();
```
int19h: This can throw (e.g. if there's a race between `HasExited` and `Kill`). Since we're in `Dispose`, probably a good idea to catch/discard.

Issue #3469, opened 2017-12-12
Switch to new Navigate To APIs (PatternMatch/PatternMatchKind)


huguesv:
Our current code has the deprecation warnings disabled. We need to move over to the new APIs.

This was started in this PR https://github.com/Microsoft/PTVS/pull/3402 but abandoned since we didn't have access to the NuGet package that includes the new APIs.

PR #3474, opened 2017-12-12
Updates the nuget packages we use for build

zooba:


zooba:
Can't merge without #3469

huguesv:
LGTM. Looks like a slight reduction in functionality (regex search) but if it was popular they would have implemented it I suppose.

zooba:
Yeah, no real choice I'm afraid. Still need to make sure it builds though

PR #3476, opened 2017-12-12
Stop warning the user that breakpoints are disabled while importing stdlib

### LINKED ISSUE #3475: Debugger warnings kill import performance
Split from https://github.com/Microsoft/PTVS/issues/3408

I also get a ton of messages on the output screen before the program loads that weren't there before in PTVS:

"Standard library module invoked user code during import; breakpoints disabled for invoked code."

If you use the classifier cookiecutter project and start debuging with a breakpoint on the read_table line (which means it imports pandas, numpy and matplotlib):

- It takes over 120-150 secs to reach breakpoint with 1000s of warning in output window
- It takes 5-10 secs if the warning is removed

huguesv:
Fix #3475 

This was causing really bad performance and 1000s of warnings to be displayed during import of pandas/numpy/matplotlib.

Updated the check so that it's no longer incorrectly displaying for non-user code, and so it doesn't display it more than once per thread.

### DIFF
--- a/Python/Product/PythonTools/ptvsd/debugger.py
+++ b/Python/Product/PythonTools/ptvsd/debugger.py
@@ -776,6 +776,7 @@ class Thread(object):
         self.django_stepping = None
         self.is_sending = False
         self.is_importing_stdlib = False
+        self.is_importing_stdlib_warned = False
 
         # stackless changes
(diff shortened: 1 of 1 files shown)


pherrman:
Thank you so much!!!

PR #3480, opened 2017-12-13
Disables HTML support in Django to remove dependency on web editor.

zooba:


huguesv:
If this wasn't temporary, I would say use the constant for output window guid and use a localizable string for the warning, but I don't think that's necessary here.

What about template debugging? Do you need to disable DjangoDebuggerTests too? Does template debugging need the stuff that's being disabled?

zooba:
Agreed - it'll be pulled forward, but mostly so we can keep the ifdefs. The warning message should become an info bar.

zooba:
This also disables template debugging, so I've disabled the test.

PR #3486, opened 2017-12-14
Update DebugAdapterHost assembly.

huguesv:


int19h:
Is it a post-Preview1 VS change?

huguesv:
Yes Preview 2.0. Steve provided the file, but I looked on my machine and I have the same one.

Issue #3350, opened 2017-11-16
Allow multiple package managers per environment


huguesv:
UI work for environments window to display more than one package manager should be pretty minimal, we can just instantiate multiple copies of the same UI and pass it a different package manager.

IPythonInterpreterFactory needs to have an array of IPackageManager instead of a single one (or in addition to the existing property).

Install Package command in solution explorer can bring up the UI for the first package manager.

zooba:
When I was designing the package manager API, I thought the best design would be for IPythonInterpreterFactory to have one _default_ package manager, but allow the UI to use an alternative one.

That should still be possible. Package managers are given the factory instance, and then shouldn't care about what the factory thinks its package manager is.

The `InterpreterListToolWindow.List_ViewCreated` method can add as many extension providers as it likes.

huguesv:
Environment window will have additional package manager in the drop down so the user can pick which one to use there.

Here are features/commands that target the default package manager:
a. Install from Requirements command
b. New project task dialog that appears when requirements.txt is detected (when you pick install into my global env)
c. Generate Requirements command
d. Solution explorer package list
e. Remove package command from solution explorer
f. Install package command from solution explorer just brings up the default package mgr in environment window, user can easily switch to another one

Some of these work more or less with conda as the default package manager:

There's no `bottle` conda package, and you can easily run into that error with b).

When generating requirements for conda env, you'll end up with the conda package list, which will likely fail if someone tries to install that using pip (and vice-versa).

huguesv:
All operations dealing with requirements.txt have been changed to always use pip, so I feel this is all done.

PR #3487, opened 2017-12-14
Enable multiple package managers in environment window

huguesv:
Small incremental improvement that lets you use either pip or conda in the environment window (when conda is detected).

I've avoided changing the interface, which wasn't technically necessary to support this particular enhancement. I think this will have to be revisited (I don't like it much).

Other enhancements may require access to the alternate package manager, which may mean we need it stored on the factory interface. See my latest notes #3350 for various things that come up when pip is not the default package manager.


Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -70,6 +70,7 @@ InterpreterFactoryCreationOptions options
                     if (pm != null) {
                         pm.SetInterpreterFactory(this);
                         pm.InstalledFilesChanged += PackageManager_InstalledFilesChanged;
+                        // TODO: do we need to worry about the non-primary package managers?
```
zooba: No. The point of this one is to be the primary package manager (mainly for detecting "do we need to rescan for intellisense")

Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
```diff
@@ -244,6 +240,16 @@ sealed class InterpreterListToolWindow : ToolWindowPane {
             }
         }
 
+        private void AddPepExtension(EnvironmentView view, IPythonInterpreterFactory factory, IPackageManager packageManager) {
```
zooba: `AddPipExtension`

Python/Product/VSInterpreters/PackageManager/PackageManagers.cs
```diff
@@ -0,0 +1,33 @@
...
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.PythonTools.Interpreter {
+    public static class PackageManagers {
```
zooba: We should have an extension methods class somewhere, right? Can we have `GetAllPackageManagers` that returns all of them?

huguesv:
Thanks, updated.

zooba:
The changes are good, but let's chat about refactoring package manager management more before merging it.

huguesv:
The change was integrated in #3500

PR #3489, opened 2017-12-14
Updates build dependencies.

zooba:
Updates build dependencies.
Removes obsolete configuration for VS 2015.


huguesv:
With the addition of the text packages, you must be able to delete the dlls in Dev15.0?
We could do the same for Microsoft.VisualStudio.Language.NavigateTo.Interfaces?

PR #3491, opened 2017-12-14
Merge changes from stabilisation branch

zooba:


zooba:
Importantly, this also includes a change to re-enable the HTML editor extensions, that was not in the 15.6 P2 branch.

zooba:
And it has an extra fix that enables us to build.

PR #3500, opened 2017-12-16
Major analysis refactor

zooba:
I recommend not using the GitHub review page for this one :)

This is the first part of refactoring legacy code out of the analysis engine and injecting it where needed. The Analyzer project is going to be the project where all of this code goes, and I'll create a new one in stage two for the a new analysis host.

This also switches the tests to use the new analysis engine rather than completion DBs. All those tests should be good now, but there is going to be more fallout that I haven't tidied up yet. I'll try to get to anything blocking in this PR, but plan on deferring most to the next stage that will include splitting the legacy tests into their own assemblies.


huguesv:
>             // Use the NoPackageManager instance so that we get a

comment out-of-date

---
Refers to: Python/Product/Analyzer/Interpreter/LegacyDB/PythonInterpreterFactoryWithDatabase.cs:239 in 30eeb99. [](commit_id = 30eeb99a4b49cf9d2fb50f3db6b6aa3e4ca5a914, deletion_comment = False)

huguesv:
>             return new PythonTypeDatabase(null, BaselineDatabasePath, isDefaultDatabase: true);

This will initialize _factory to null, and this code will NRE:
        public Version LanguageVersion {
            get {
                return _factory.Configuration.Version;
            }
        }

---
Refers to: Python/Product/Analyzer/Interpreter/LegacyDB/PythonTypeDatabase.cs:366 in 30eeb99. [](commit_id = 30eeb99a4b49cf9d2fb50f3db6b6aa3e4ca5a914, deletion_comment = False)

huguesv:
>         IEnumerable<IPackageManager> GetPackageManagers(IPythonInterpreterFactory factory);

Some package manager lifetime questions:

Every time this gets called, new package managers are created. So should the caller call dispose on all of them?

Also, since package managers are only created on demand, and they are responsible for watching file system, does this mean that if the environment window has never been opened, that the interpreter factory will not be notified of changes if user installs a package via command-line?

---
Refers to: Python/Product/VSInterpreters/Interpreter/IInterpreterOptionsService.cs:49 in 30eeb99. [](commit_id = 30eeb99a4b49cf9d2fb50f3db6b6aa3e4ca5a914, deletion_comment = False)

huguesv:
>         IEnumerable<IPackageManager> GetPackageManagers(IPythonInterpreterFactory factory);

Oh I see that the interpreter node in solution explorer will have a reference to the first package manager and enables notifications, so projects should be ok.

---
In reply to: [352518487](https://github.com/Microsoft/PTVS/pull/3500#issuecomment-352518487) [](ancestors = 352518487)

---
Refers to: Python/Product/VSInterpreters/Interpreter/IInterpreterOptionsService.cs:49 in 30eeb99. [](commit_id = 30eeb99a4b49cf9d2fb50f3db6b6aa3e4ca5a914, deletion_comment = False)

huguesv:
>             foreach (var c in x) {

what if x was null?

---
Refers to: Python/Product/Analysis/OverloadResult.cs:66 in 30eeb99. [](commit_id = 30eeb99a4b49cf9d2fb50f3db6b6aa3e4ca5a914, deletion_comment = False)

zooba:
>         IEnumerable<IPackageManager> GetPackageManagers(IPythonInterpreterFactory factory);

Yes. I need to document this better, but essentially you shouldn't need to Dispose() a package manager until you've called EnableNotifications (and then disposal should be the same as DisableNotifications)

---
In reply to: [352523366](https://github.com/Microsoft/PTVS/pull/3500#issuecomment-352523366) [](ancestors = 352523366,352518487)

---
Refers to: Python/Product/VSInterpreters/Interpreter/IInterpreterOptionsService.cs:49 in 30eeb99. [](commit_id = 30eeb99a4b49cf9d2fb50f3db6b6aa3e4ca5a914, deletion_comment = False)

zooba:
>             foreach (var c in x) {

We check at the start of the function

---
In reply to: [352525302](https://github.com/Microsoft/PTVS/pull/3500#issuecomment-352525302) [](ancestors = 352525302)

---
Refers to: Python/Product/Analysis/OverloadResult.cs:66 in 30eeb99. [](commit_id = 30eeb99a4b49cf9d2fb50f3db6b6aa3e4ca5a914, deletion_comment = False)

huguesv:
>     static class Conda {

The whole class is unused, might as well delete it.

---
Refers to: Python/Product/PythonTools/PythonTools/Project/Conda.cs:27 in 23ed888. [](commit_id = 23ed88870dbfaed710f67d488939c714159e1388, deletion_comment = False)

Common/Tests/MockVsTests/MockVs.cs
```diff
@@ -482,9 +489,10 @@ out _monSelCookie
 
         private CompositionContainer CreateCompositionContainer() {
             var container = new CompositionContainer(CachedInfo.Catalog);
-            container.ComposeExportedValue<MockVs>(this);
-            var batch = new CompositionBatch();
 
+            var batch = new CompositionBatch();
+            batch.AddExportedValue(this);
+            batch.AddExportedValue(typeof(SComponentModel).FullName, this);
             container.Compose(batch);
```
zooba: This (and related SComponentModel changes) is totally the wrong fix - it should only be a VS service.

huguesv:
>                     _packageManager.InstalledPackagesChanged -= InstalledPackagesChanged;

Don't know if you've updated this in response to my other comment, but call Dispose/DisableNotifications here?

---
Refers to: Python/Product/PythonTools/PythonTools/Project/InterpretersNode.cs:110 in 23ed888. [](commit_id = 23ed88870dbfaed710f67d488939c714159e1388, deletion_comment = False)

PR #3504, opened 2017-12-19
Removes Common dependency from analysis

zooba:


Python/Product/Analysis/Infrastructure/PathEqualityComparer.cs
```diff
@@ -0,0 +1,145 @@
...
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+
+namespace Microsoft.PythonTools.Analysis.Infrastructure {
+    class PathEqualityComparer : IEqualityComparer<string> {
```
huguesv: >PathEqualityComparer [](start = 10, length = 20)

I might not have seen it yet, but this class deserves some tests.

PR #3519, opened 2018-01-05
Add event for when the analyzer is initializing.

huguesv:


Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -110,6 +110,13 @@ public sealed class VsProjectAnalyzer : ProjectAnalyzer, IDisposable {
             IPythonInterpreterFactory factory,
             bool inProcess = false
         ) {
+            services.Python?.Logger?.LogEvent(PythonLogEvent.AnalysisInitializing, new AnalysisInitialize() {
```
zooba: @huguesv  Can you add catch-all handlers around these? I'd hate to have something break because of logging

huguesv: done: https://github.com/Microsoft/PTVS/pull/3527

PR #3530, opened 2018-01-06
Analysis refactor part 3

zooba:
This is not yet ready to merge, but is getting fairly close. I've done enough of the changes that I want to for this (already huge) PR, so I'm focusing on stabilizing it and getting it merged now.

@MikhailArkhipov With this, you should have enough LSP in the Analysis project to just build that project as a DLL and wrap it up in your own host (that instantiates the `Server` class and passes through messages to it) - shouldn't need any other dependencies. It likely won't be functional for you right away given the hard failure for unexpected messages, but feel free to start making whatever tweaks are needed there to get it running.

Signature help is the next item that needs migrating. I'll do that in the next PR - this one is big enough.


huguesv:
After looking at the changes for some time I haven't seen anything objectionable.

I pulled your branch and started testing locally because I figure I have more chances of finding issues that way than staring at various functions with a bunch of "index management" that I need to execute in my head.

First project load was ok, but second gave me this assert:
```
---------------------------
---------------------------
Unhandled exception in Microsoft.PythonTools.Editor.PythonTextBufferInfo.InvokeSinks (C:\Projects\GitHub\zooba\PTVS\Python\Product\PythonTools\PythonTools\Editor\PythonTextBufferInfo.cs:282)

System.NullReferenceException: Object reference not set to an instance of an object.

   at Microsoft.PythonTools.Intellisense.VsProjectAnalyzer.<>c__DisplayClass87_0.<<GetMissingImportsAsync>b__1>d.MoveNext() in C:\Projects\GitHub\zooba\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\ProjectAnalyzer.cs:line 820

   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()

   at Microsoft.PythonTools.Intellisense.VsProjectAnalyzer.<EnsureSingleRequest>d__155`2.MoveNext() in C:\Projects\GitHub\zooba\PTVS\Python\Produ......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------
```

huguesv:
I've gotten this assert a few times while editing:
```
---------------------------
---------------------------
Buffer content mismatch



   at Microsoft.PythonTools.Intellisense.BufferParser.ValidateBufferContents(IEnumerable`1 snapshots, FileUpdateResponse response) in C:\Projects\GitHub\zooba\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\BufferParser.cs:line 233

   at Microsoft.PythonTools.Intellisense.BufferParser.<ParseBuffersAsync>d__43.MoveNext() in C:\Projects\GitHub\zooba\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\BufferParser.cs:line 279

   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.InvokeMoveNext(Object stateMachine)

   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.Run()

   at System.Runtime.CompilerServices.AsyncMethodBuilderC......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------
```

huguesv:
I've seen refactor-rename work in one scenario, but several others lead to the command not doing anything, with no preview, like this:
image

huguesv:
If I start by creation a new application and write some code, I don't get any assert but IntelliSense doesn't give any completions.
If I save that project, close and reload solution, I then get a bunch of mismatch asserts, but IntelliSense seem to work better.
I've tried the "missing import" code fix feature, but it doesn't show up as available, with either of the above scenarios.

huguesv:
FYI I tested the latest commits for a bit. Extract method is doing better. I still get the buffer content mismatch (but maybe less frequently) and also the save as dialog.

zooba:
@huguesv Awesome, thanks! I think I've fixed the update ordering issue (though we can probably now lose an update completely, so I'll need a custom event for "I'm out of sync send me the whole document again").

Extract Method should be back in action again - I accidentally reversed a type check condition that broke everything. Got a bit more minor cleanup coming, but it's worth grabbing new sources and seeing how it works. (My non-UI test numbers are higher than master, but that's probably because I fixed that invalid header bug. Working on the UI tests today.)

zooba:
I think I've got the buffer consistency issues with that last commit (I was grabbing the wrong version number from the message), so now I'll let the tests run through and work on adding more to help validate the LSP interfaces.

huguesv:
Sorry, but not quite :( I just took your commits and rebuilt. I opened my existing project and just added a `import sys` and pressed enter at the end of my file and got the assert.

image

Python/Product/Analysis/Infrastructure/PathUtils.cs
```diff
@@ -318,5 +318,87 @@ static class PathUtils {
...
+            return null;
+        }
+
+        /// <summary>
+        /// Normalizes and returns the provided path.
+        /// </summary>
+        /// <exception cref="ArgumentException">
+        /// If the provided path contains invalid characters.
+        /// </exception>
+        public static string NormalizePath(string path) {
```
AlexanderSher: We have `Microsoft.PythonTools.Infrastructure.PathUtils.NormalizePath`. How this one is different?

zooba: The Analysis project has no dependencies, so it can't access that.

We can change the existing one to redirect here and only have one implementation (I *think* everything that uses Common also uses Analysis, though it isn't required, and arguably these helpers should be private since the Analysis API is going to be public/stable), but I've mainly just been copy-pasting into the new namespace.

Python/Product/Analysis/AnalysisLogWriter.cs
```diff
@@ -222,36 +223,12 @@ public struct LogItem {
...
-            return null;
-        }
-
         private TextWriter OpenLogFile() {
             if (string.IsNullOrEmpty(_outputFile)) {
                 return null;
             }
             var enc = new UTF8Encoding(false);
-            var stream = OpenRawLogFile(true, true);
+            var stream = PathUtils.OpenWithRetry(_outputFile, FileMode.Append, FileAccess.Write, FileShare.Read);
```
AlexanderSher: Why are we trying to get file access several times?

Python/Product/Analysis/DocumentBuffer.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public void Reset(int version, string content) {
+            Version = version;
+            Text.Clear();
+            Text.Append(content);
+        }
+
+        public void Update(IEnumerable<DocumentChangeSet> changes) {
+            foreach (var change in changes) {
+                Update(change);
```
AlexanderSher: If one of the updates fails with exception, the others won't be processed. Is that ok?

Python/Product/Analysis/AnalysisLogWriter.cs
zooba: Because we've seen random IO failures in the past that work on retry (presumably AV or disk tools, but it's near impossible to diagnose).

Python/Product/Analysis/DocumentBuffer.cs
zooba: Yep. We do need to add an "out of sync" message for the server to raise in this case though (and some others), as there's no way to return failure from a document changed event.

Python/Product/Analysis/DocumentBuffer.cs
```diff
@@ -0,0 +1,94 @@
...
+            if (Version >= 0) {
+                if (changes.FromVersion < Version) {
+                    return;
+                } else if (changes.FromVersion > Version) {
+                    throw new InvalidOperationException("missing prior versions");
+                }
+            }
+
+            int lastStart = int.MaxValue;
+            var lineLoc = SplitLines(Text.ToString()).ToArray();
```
AlexanderSher: Not sure how critical is this, but allocating new string for indexing is not required.

Python/Product/Analysis/Parsing/Tokenizer.cs
```diff
@@ -2663,6 +2663,7 @@ public enum NewLineKind {
         CarriageReturnLineFeed
     }
 
+    [DebuggerDisplay("NewLineLocation({_endIndex}, {_kind})")]
     public struct NewLineLocation : IComparable<NewLineLocation> {
         private int _endIndex;
         private NewLineKind _kind;
```
AlexanderSher: These fields should be made `readonly`

Python/Product/Analysis/DocumentBuffer.cs
```diff
@@ -0,0 +1,94 @@
...
+            foreach (var change in changes.Changes) {
+                if (change.WholeBuffer) {
+                    Text.Clear();
+                    if (!string.IsNullOrEmpty(change.InsertedText)) {
+                        Text.Append(change.InsertedText);
+                    }
+                    continue;
+                }
+
+                int start = NewLineLocation.LocationToIndex(lineLoc, change.ReplacedSpan.Start, Text.Length);
```
AlexanderSher: Are the `NewLineLocation` instances still valid after `Text` is changed during first iteration?

Python/Product/Analysis/DocumentBuffer.cs
zooba: I think StringBuilder is missing IndexOf[Any], and I didn't want to rewrite it. I'll mark it as a potential optimization and we can do it in its own PR.

Python/Product/Analysis/DocumentBuffer.cs
zooba: Changes have to be ordered in reverse order (according to LSP, but we also check on the next line), so the locations after the change are invalid, but we shouldn't try to use them. They're valid enough for the check to work.

huguesv:
image

System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: currentPosition 
at Microsoft.VisualStudio.Text.Tracking.TrackPositionBackwardInTime(PointTrackingMode trackingMode, Int32 currentPosition, ITextVersion currentVersion, ITextVersion targetVersion)
at Microsoft.PythonTools.Intellisense.LocationTracker.Translate(SourceLocation loc, Int32 fromVersion, Int32 toVersion) in C:\Projects\GitHub\zooba\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\LocationTracker.cs:line 269
at Microsoft.PythonTools.PythonAnalysisClassifier.GetClassificationSpans(SnapshotSpan span) in C:\Projects\GitHub\zooba\PTVS\Python\Product\PythonTools\PythonTools\PythonAnalysisClassifier.cs:line 171
at Microsoft.VisualStudio.Text.Classification.Implementation.ClassifierTagger.<GetTags>d__5.MoveNext()
at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1.<GetTagsForBuffer>d__39.MoveNext()
at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)

huguesv:
Completions appear when you're defining a function with multiple parameters. I compared with master branch, and that doesn't happen there, so it's a regression from this PR.

image

zooba:
Completion triggering is going to change again - LSP embeds all of its logic behind a single message, whereas we use a range of messages depending on context. So that will all have to be reimplemented on the other side anyway.

For now, I'll make sure we don't trigger normal completions on space. That should get us back to at least sensible behaviour, even if it doesn't match the old behaviour 100%.

Python/Product/Analysis/DocumentBuffer.cs
AlexanderSher: Sorry, I don't get it.
Imagine we have string `123456` and two changes: 
- remove range `[1..3]`: `123456 -> 156`
- add text `789` into position `2`: `156 -> 157896`

If they are handled in reverse order, we will get:
- add text `789` into position `2`: `123456 -> 127893456`
- remove range `[1..3]`: `127893456 -> 193456`

AlexanderSher: Also, I think reverse order isn't guaranteed. Changes are reversed in `BufferParser.GetUpdatesForSnapshot`, but in `OutOfProcProjectAnalyzer.UpdateContent` they are simply merged:
```cs
                    changes.AddRange(fileChange.changes.Select(c => new LS.TextDocumentContentChangedEvent {
                        range = new SourceSpan(
                            new SourceLocation(c.startLine, c.startColumn),
                            new SourceLocation(c.endLine, c.endColumn)
                        ),
                        text = c.newText
                    }));
```

zooba: Reverse order of start location, not reverse order of change. I don't recall if there's a non-overlapping requirement, or if it just works out that VS doesn't send overlapping changes in the same update, but this area is *woefully* underspecified and I don't think we have much of a choice other than assuming that our clients (VS/VSC) behave sensibly.

For VS, the changes are guaranteed to be in increasing start location when we get them, so reversing the order to match LSP is fine.

Python/Product/Analysis/ExpressionFinder.cs
```diff
@@ -162,37 +188,29 @@ private class NormalExpressionWalker : ExpressionWalker {
...
+                foreach (var b in node.Bases.MaybeEnumerate()) {
+                    b.Walk(this);
                 }
                 node.Body?.Walk(this);
 
-                return true;
+                return false;
             }
 
             public override bool Walk(Arg node) {
```
AlexanderSher: So this one always return `false`?

Python/Product/Analysis/ModulePath.cs
```diff
@@ -768,6 +776,73 @@ out errorParameter
...
+            isMissing = false;
+
+            if (!PathEqualityComparer.Instance.StartsWith(sourceFile, basePath)) {
+                return false;
+            }
+
+            if (isPackage == null) {
+                isPackage = f => !string.IsNullOrEmpty(GetPackageInitPy(f));
+            }
+            if (getModule == null) {
```
AlexanderSher: `getModule` isn't used anywhere

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -36,34 +39,48 @@ namespace Microsoft.PythonTools.Analysis {
...
-        private PythonAst _tree;
-        private ModuleAnalysis _currentAnalysis;
+    internal sealed class ProjectEntry : IPythonProjectEntry, IAggregateableProjectEntry, IDocument {
         private AnalysisUnit _unit;
-        private int _analysisVersion;
-        private Dictionary<object, object> _properties = new Dictionary<object, object>();
         private ManualResetEventSlim _curWaiter;
         private int _updatesPending, _waiters;
+        private readonly SortedDictionary<int, DocumentBuffer> _buffers;
         internal readonly HashSet<AggregateProjectEntry> _aggregates = new HashSet<AggregateProjectEntry>();
```
AlexanderSher: Can be made private

AlexanderSher:
>         private void ClearScope() {

Not used

---
Refers to: Python/Product/Analysis/ProjectEntry.cs:279 in 0903354. [](commit_id = 09033543fa7e7a21cb72e36012275dd03ed2b7eb, deletion_comment = False)

huguesv:
I'm unable to get completions on project-local packages. In picture below, I get completions on `os`, `class1` but not `package1`. I've tried a number of projects and reloads, but it never worked.
image

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -280,60 +283,31 @@ where fileName.StartsWith(pathPrefix)
...
 
-        internal PythonAnalyzer ProjectState {
-            get { return _projectState; }
-        }
+        internal PythonAnalyzer ProjectState { get; private set; }
 
-        public PythonAst Tree {
-            get { return _tree; }
-        }
+        public PythonAst Tree { get; private set; }
```
AlexanderSher: This property is accessed under lock inside `ProjectEntry` but is available without lock `ModuleAnalysis` and several other places

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -36,34 +39,48 @@ namespace Microsoft.PythonTools.Analysis {
...
+            Uri documentUri,
+            IAnalysisCookie cookie
+        ) {
+            ProjectState = state;
+            ModuleName = moduleName ?? "";
+            DocumentUri = documentUri ?? MakeDocumentUri(filePath);
+            FilePath = filePath;
+            Cookie = cookie;
+            MyScope = new ModuleInfo(ModuleName, this, state.Interpreter.CreateModuleContext());
+            _unit = new AnalysisUnit(Tree, MyScope.Scope);
```
AlexanderSher: I think tree is always `null` here

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -342,15 +316,101 @@ where fileName.StartsWith(pathPrefix)
...
+                return null;
+            }
+        }
+
+        internal static Stream EncodeToStream(StringBuilder text, Encoding encoding, int chunkSize = 4096) {
+            if (text.Length < chunkSize) {
+                return new MemoryStream(encoding.GetBytes(text.ToString()));
+            }
+
+            var ms = new MemoryStream(encoding == Encoding.Unicode ? text.Length * 2 : text.Length);
```
AlexanderSher: I'm not sure that `==` is safe. I'd rather use `Equals`.

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -342,15 +316,101 @@ where fileName.StartsWith(pathPrefix)
...
+                    var s = PathUtils.OpenWithRetry(FilePath, FileMode.Open, FileAccess.Read, FileShare.Read);
+                    version = s == null ? -1 : 0;
+                    return s;
+                }
+                version = -1;
+                return null;
+            }
+        }
+
+        internal static Stream EncodeToStream(StringBuilder text, Encoding encoding, int chunkSize = 4096) {
```
AlexanderSher: I think it can be private

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -342,15 +316,101 @@ where fileName.StartsWith(pathPrefix)
...
-
         internal void Enqueue() {
             _unit.Enqueue();
         }
 
         public void AggregatedInto(AggregateProjectEntry into) {
             _aggregates.Add(into);
         }
+
+        public IEnumerable<int> DocumentParts => _buffers.Keys.AsLockedEnumerable(_buffers);
```
AlexanderSher: This is pretty much the same as `lock(this)`. Any client code that calls `DocumentParts.GetEnumerator()` acquires a lock that controls access to private implementation.

Python/Product/Analysis/LanguageServer/ParseQueue.cs
```diff
@@ -0,0 +1,209 @@
...
+            return false;
+        }
+
+        public async Task<IAnalysisCookie> Enqueue(IDocument doc, PythonLanguageVersion languageVersion) {
+            TaskCompletionSource<IAnalysisCookie> created = null;
+            var tcs = _parsing.GetOrAdd(doc, e => { return created = new TaskCompletionSource<IAnalysisCookie>(); });
+            if (created == null) {
+                // Do not start parsing until the previous one has completed
+                await tcs.Task;
+                tcs = new TaskCompletionSource<IAnalysisCookie>();
```
AlexanderSher: If this method is called three times in a row for the same doc, second and third calls will await for the first one and then will run concurrently. You probably need an async ARE here.

huguesv:
In one session, I lost all IntelliSense / editor tooltips. It worked again after reloading.
In most sessions, I got many "buffer content mismatch" asserts. For a while, I thought we were rid of those but then I mentioned to Alex that things looked relatively good and it jinxed it.
It looks like the analyzer process doesn't stop when you quit VS? I don't know how frequently that happens.
image

Python/Product/Analysis/LanguageServer/ParseQueue.cs
AlexanderSher: Or you can simplify this by using `AddOrUpdate`
```cs
public Task<IAnalysisCookie> Enqueue(IDocument doc, PythonLanguageVersion languageVersion) {
    return _parsing.AddOrUpdate(doc, 
        d => Task.Run(() => ParseWorker(d, languageVersion)),
        (d, task) => task.ContinueWith(t => ParseWorker(d, languageVersion)));
}
```
and get rid of the `TaskCompletionSource` altogether.

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -153,7 +167,7 @@ internal sealed class ProjectEntry : IPythonProjectEntry, IAggregateableProjectE
             }
 
             lock (this) {
-                _analysisVersion++;
+                AnalysisVersion++;
```
AlexanderSher: This is the only place where `AnalysisVersion` is accessed under lock

AlexanderSher:
>                 return new StreamReader(new PartiallyReadStream(readBytes, stream), encoding);

`PartiallyReadStream` doesn't override `Dispose`. Who closes the `stream`?

---
Refers to: Python/Product/Analysis/Parsing/Parser.cs:5031 in 0903354. [](commit_id = 09033543fa7e7a21cb72e36012275dd03ed2b7eb, deletion_comment = False)

Python/Product/Analysis/Infrastructure/ProcessHelper.cs
```diff
@@ -126,11 +138,12 @@ sealed class ProcessHelper : IDisposable {
...
 
-            await Task.WhenAll(
-                _seenNullOutput.WaitAsync(cancellationToken),
-                _seenNullError.WaitAsync(cancellationToken)
-            );
-            cancellationToken.ThrowIfCancellationRequested();
+            if (!_process.HasExited) {
+                await _seenNullOutput.WaitAsync(cancellationToken).ConfigureAwait(false);
+                await _seenNullError.WaitAsync(cancellationToken).ConfigureAwait(false);
+                cancellationToken.ThrowIfCancellationRequested();
```
AlexanderSher: Do you really want to throw OCE when process has completed it's work?

Python/Product/Analysis/DocumentBuffer.cs
zooba: Within each snapshot, which is why we recalculate the line locations between each one.

Python/Product/Analysis/ExpressionFinder.cs
zooba: `false` here means don't continue walking down this branch. In this case, it's because we've already walked it (the `node.Body.Walk(this)` above).

Python/Product/Analysis/ProjectEntry.cs
zooba: I haven't touched aggregates at all in this PR and I'd rather not start now.

Python/Product/Analysis/ProjectEntry.cs
zooba: When used correctly (which is to enumerate this sequence and use the values to index into `_buffers`) it's going to be necessary to lock the outer object on all uses. This way, we get a single lock that lasts for the entire `foreach` and prevents races.

Either way requires the caller to use the object correctly. Not disposing an `IDisposable` is more likely to be done correctly than locking the object everywhere.

Python/Product/Analysis/LanguageServer/ParseQueue.cs
zooba: Oh that's nice. I didn't know you got the previous value in AddOrUpdate.

Python/Product/Analysis/Infrastructure/ProcessHelper.cs
zooba: Semantically, the decision has been made to cancel, which presumably means we also want to cancel any subsequent work as well. Though since we've just completed a cross-process task without cancellation and are not marshalling back to the original thread here, it's very unlikely that a cancellation will occur between the previous wait and this check. (I suspect the call has stuck around from a previous iteration of this code... took a few goes to get `Process` termination right.)

Python/Product/Analysis/ProjectEntry.cs
zooba: Reading Tree doesn't require the lock - Cookie and _updatesPending do. GetTreeAndCookie reads under the lock to ensure you get matched tree and cookie, but directly accessing tree is not guaranteed to be the latest version.

Python/Product/Analysis/ProjectEntry.cs
zooba: It needs tests (guess that's my job this morning)

Python/Product/Analysis/ProjectEntry.cs
zooba: It's almost the only place where it's modified. I'll lock the other one, but most of the reads are fine.

zooba:
`PartiallyReadStream` wraps a user argument, so it should be disposed by the original caller.

Python/Product/Analysis/DocumentBuffer.cs
AlexanderSher: So `changes.ToVersion - changes.FromVersion == 1`?

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -342,15 +320,118 @@ where fileName.StartsWith(pathPrefix)
...
+        }
+
+        public TextReader ReadDocument(int part, out int version) {
+            var stream = ReadDocumentBytes(part, out version);
+            if (stream == null) {
+                version = -1;
+                return null;
+            }
+            try {
+                var sr = Parser.ReadStreamWithEncoding(stream, ProjectState.LanguageVersion);
```
AlexanderSher: `ReadStreamWithEncoding` wraps `stream` with `PartiallyReadStream` and then `stream` is set to `null`, so when `TextReader` is disposed, underlying stream will stay alive.

Python/Product/Analysis/LanguageServer/ParseQueue.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        private Task<IAnalysisCookie> Parse(IDocument doc, PythonLanguageVersion languageVersion, Task<IAnalysisCookie> previousTask) {
+            if (previousTask != null && previousTask.Status == TaskStatus.Running) {
+                var callingThread = Thread.CurrentThread;
+                var res = previousTask.ContinueWith(t => {
+                    // We need to cancel immediately if we're running on the same thread
+                    // This can happen if the task completes before we call ContinueWith,
+                    // though the RunContinuationsAsynchronously should protect against this too
+                    if (Thread.CurrentThread == callingThread) {
```
AlexanderSher: Delegate passed to `ContinueWith` may be executed in the caller thread only if `TaskContinuationOptions.ExecuteSynchronously` is explicitly passed to `ContinueWith`. Otherwise, task will be queued into scheduler. Otherwise, you can safely call `ContinueWith` for completed task.

zooba: The debugger disagrees with you, though with the RunContinuationsAsynchronously it seems to work. Still, better to have the safety net in there than deal with bug reports later on (which I can guarantee will be vague performance related issues)

AlexanderSher: > The debugger disagrees with you

That's something. Any specific repro?

Python/Product/Analysis/DocumentBuffer.cs
zooba: According to the LSP spec, yes. For VS, we expect them to match snapshot version numbers, so the increment is only a fallback and we prefer the explicit value if it's provided. VS Code apparently assumes that updates are synchronous and never dropped.

AlexanderSher: Ok, got it.

Python/Product/Analysis/Intellisense/AnalysisQueue.cs
```diff
@@ -64,12 +63,42 @@ sealed class AnalysisQueue : IDisposable {
...
 
+        public Task WaitForCompleteAsync(AnalysisPriority priority = AnalysisPriority.None) {
+            var task = new WaitAnalyzableTask();
+            Enqueue(task, priority);
+            return task.Task;
+        }
+
+        public event EventHandler<UnhandledExceptionEventArgs> UnhandledException;
+
+        public TaskScheduler Scheduler { get; private set; }
```
AlexanderSher: Is it still in use?

Python/Product/Analysis/Intellisense/AnalysisQueue.cs
```diff
@@ -145,10 +174,15 @@ sealed class AnalysisQueue : IDisposable {
             }
         }
 
+        public void WaitForActivity(int millisecondsTimeout) {
```
AlexanderSher: Method isn't used anywhere, and as a result, `_activityEvent` isn't used as well.

Python/Product/Analysis/Intellisense/LineInfo.cs
```diff
@@ -0,0 +1,81 @@
...
+using Microsoft.PythonTools.Parsing;
+
+namespace Microsoft.PythonTools.Intellisense {
+    class LineInfo {
+        public int Start, Length;
+        /// <summary>
+        /// 0-based line number
+        /// </summary>
+        public int LineNo;
+        public NewLineKind LineBreak;
```
AlexanderSher: Why these fields aren't read-only?

AlexanderSher: Also, I believe this type can be a struct.

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
```diff
@@ -0,0 +1,159 @@
...
+                List<int> lineInfo;
+                if (!oldLineMapping.TryGetValue(GetOldText(i), out lineInfo)) {
+                    oldLineMapping[GetOldText(i)] = lineInfo = new List<int>();
+                }
+                lineInfo.Add(i);
+            }
+
+            int curOldLine = 0;
+            for (int curNewLine = 0; curOldLine < _oldLines.Count && curNewLine < _newLines.Count; curOldLine++) {
+                if (GetOldText(curOldLine) == GetNewText(curNewLine)) {
```
AlexanderSher: You can do
```cs
                var oldLine = _oldLines[curOldLine];
                var newLine = _newLines[curNewLine];
                if (oldLine.Length == newLine.Length && string.CompareOrdinal(_oldCode, oldLine.Start, _newCode, newLine.Start, newLine.Length) == 0) {
                    curNewLine++;
                    continue;
                }
```
That's a bit more code, but it is faster and no string allocation for equal lines

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
```diff
@@ -0,0 +1,159 @@
...
+                }
+
+                if (!replaced) {
+                    ReplaceLines(edits, curOldLine, _oldLines.Count, startNewLine, _newLines.Count);
+                    curOldLine = _oldLines.Count;
+                    break;
+                }
+            }
+
+            if (curOldLine < _oldLines.Count) {
```
AlexanderSher: Do you need to handle remaining `_newLines` as well?

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
```diff
@@ -0,0 +1,159 @@
...
+            for (int curNewLine = 0; curOldLine < _oldLines.Count && curNewLine < _newLines.Count; curOldLine++) {
+                if (GetOldText(curOldLine) == GetNewText(curNewLine)) {
+                    curNewLine++;
+                    continue;
+                }
+
+                bool replaced = false;
+                // replace starts on this line, figure out where it ends...
+                int startNewLine = curNewLine;
+                for (curNewLine += 1; curNewLine < _newLines.Count; curNewLine++) {
```
AlexanderSher: If `curOldLine` line is removed, there will be no matching in the `oldLineMapping`, and as a result single line deletion will be treated as a replacement of the whole remaining text.

AlexanderSher: Sorry, my original comment was incorrect.

But there are still some cases when a simple change can result in a replacement of a big chunk.
For instance, there is a subsequence of `n+1` lines `A..NA`, and user adds a new line to the beginning of it: `ZA..NA`. Loop will start with `curNewLine == 1`, `GetNewText(curNewLine)` will return line `A`, `lines` will contain `0` and `n`, and as a result `A..NA` will be replaced with `ZA`. The remaining comparison will most likely end up with full replacement.

I'm not sure how often such thing may happen and how critical is it at all.

Python/Product/Analysis/Intellisense/ProjectEntryMap.cs
```diff
@@ -47,13 +48,13 @@ sealed class ProjectEntryMap : IEnumerable<KeyValuePair<string, IProjectEntry>>
                 // ids are 1 based
                 id = _ids.Count;
             }
-            _projectFiles[filename] = node;
+            _projectFiles[new Uri(filename, UriKind.RelativeOrAbsolute)] = node;
```
AlexanderSher: Is it ok to have both relative and absolute paths in the same map?

Python/Product/Analysis/LanguageServer/CallbackEventArgs.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
+    public abstract class CallbackEventArgs<T> : EventArgs where T : struct {
+        private TaskCompletionSource<object> _task;
+
+        internal CallbackEventArgs(TaskCompletionSource<object> task) {
+            _task = task;
+        }
+
+        public T? @params { get; set; }
```
AlexanderSher: `Params`?

Python/Product/Analysis/LanguageServer/DiagnosticsErrorSink.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.PythonTools.Analysis.Infrastructure;
+using Microsoft.PythonTools.Parsing;
+
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
+    class DiagnosticsErrorSink : ErrorSink {
+        private readonly string _source;
+        private readonly List<Diagnostic> _diagnostics;
+        private readonly IReadOnlyList<KeyValuePair<string, DiagnosticSeverity>> _taskCommentMap;
+
+        public DiagnosticsErrorSink(string source, List<Diagnostic> diagnostics, IReadOnlyDictionary<string, DiagnosticSeverity> taskCommentMap = null) {
```
AlexanderSher: `DiagnosticsErrorSink` is called in two places and both times `diagnostics = new List<Diagnostic>()`, so you can drop this parameter.

Python/Product/Analysis/LanguageServer/Messages.cs
```diff
@@ -0,0 +1,302 @@
...
+            Data["Code"] = code;
+        }
+
+        public LanguageServerException(int code, string message, Exception innerException) : base(message, innerException) {
+            Data["Code"] = code;
+        }
+    }
+
+    public struct InitializeResult {
+        public ServerCapabilities? capabilities;
```
AlexanderSher: `struct` with mutable field. `capabilities` can be made `readonly` and initialized from constructor

Python/Product/Analysis/LanguageServer/Messages.cs
```diff
@@ -0,0 +1,302 @@
...
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
+    public struct InitializeParams {
```
AlexanderSher: struct with mutable fields

Python/Product/Analysis/LanguageServer/ServerBase.cs
```diff
@@ -0,0 +1,232 @@
...
+
+        public virtual async Task Shutdown() { }
+
+        public virtual async Task Exit() { }
+
+        public virtual void CancelRequest() {
+            Volatile.Read(ref _lock)?.Cancel();
+        }
+
+        public virtual async Task<MessageActionItem?> ShowMessageRequest(ShowMessageRequestParams @params) {
```
AlexanderSher: Not used anywhere

Python/Product/Analysis/LanguageServer/ServerBase.cs
```diff
@@ -0,0 +1,232 @@
...
+        public event EventHandler<LogMessageEventArgs> OnLogMessage;
+
+        protected void LogMessage(MessageType type, string message) => OnLogMessage?.Invoke(this, new LogMessageEventArgs { type = type, message = message });
+
+        public event EventHandler<TelemetryEventArgs> OnTelemetry;
+
+        protected void Telemetry(TelemetryEventArgs e) => OnTelemetry?.Invoke(this, e);
+
+        public event EventHandler<RegisterCapabilityEventArgs> OnRegisterCapability;
+        protected Task RegisterCapability(RegistrationParams @params) {
```
AlexanderSher: `RegisterCapability`, `UnregisterCapability` and related events aren't used. Is it for the future functionality?

Python/Product/Analysis/LanguageServer/VolatileCounter.cs
```diff
@@ -0,0 +1,96 @@
...
+        private TaskCompletionSource<int> _waitForChange;
+        private TaskCompletionSource<object> _waitForZero;
+
+        public int Count => Volatile.Read(ref _count);
+
+        public bool IsZero => Count == 0;
+
+        private sealed class DecrementOnDispose : IDisposable {
+            public VolatileCounter _self;
+            public void Dispose() => _self.Decrement();
```
AlexanderSher: You can ensure that `Decrement` is called only once: `public void Dispose() => Interlocked.Exchange(ref _self, null)?.Decrement();`

Python/Product/Analysis/LanguageServer/VolatileCounter.cs
```diff
@@ -0,0 +1,96 @@
...
+            int newCount = Interlocked.Increment(ref _count);
+            Interlocked.Exchange(ref _waitForChange, null)?.SetResult(newCount);
+            return newCount == 1;
+        }
+
+        /// <summary>
+        /// Decrements the counter
+        /// </summary>
+        /// <returns></returns>
+        public bool Decrement() {
```
AlexanderSher: Public increment/decrement is usually an error-prone. Unless there is a memory-critical code path where additional IDisposable instance can make a problem, I think it is safer to make both these methods private.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -0,0 +1,841 @@
...
+        }
+
+        private IDisposable GetDocumentParseCounter(IDocument doc, out int count) {
+            VolatileCounter counter;
+            lock (_pendingParse) {
+                if (!_pendingParse.TryGetValue(doc, out counter)) {
+                    _pendingParse[doc] = counter = new VolatileCounter();
+                    // Automatically remove counter from the dictionary when it
+                    // reaches zero.
+                    counter.WaitForZeroAsync().ContinueWith(t => {
```
AlexanderSher: For new `VolatileCounter`, `WaitForZeroAsync` will return completed task because `IsZero == true`.
Because of the lock, `ContinueWith` will be executed right after counter is incremented, `if (counter.IsZero` check will fail and `doc` will be referenced forever.

Python/Product/Analysis/LanguageServer/VolatileCounter.cs
```diff
@@ -0,0 +1,96 @@
...
+                tcs = Interlocked.CompareExchange(ref _waitForChange, tcs, null) ?? tcs;
+            }
+            return tcs.Task;
+        }
+
+        public Task WaitForZeroAsync() {
+            if (IsZero) {
+                return Task.CompletedTask;
+            }
+            var tcs = Volatile.Read(ref _waitForZero);
```
AlexanderSher: If while current thread is on this line another thread decrements `_count`, it is possible that `_waitForZero.SetResuls` will never be called.

Python/Product/Analysis/LanguageServer/ServerBase.cs
```diff
@@ -0,0 +1,232 @@
...
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
+    public abstract class ServerBase {
+        private RequestLock _lock;
+
+        private sealed class RequestLock : IDisposable {
+            public readonly ServerBase Owner;
+            public readonly CancellationTokenSource CancellationTokenSource;
+
+            public CancellationToken Token => CancellationTokenSource.Token;
+            public void Cancel() => CancellationTokenSource.Cancel();
```
AlexanderSher: If `Cancel` is called after `Dispose`, there will be `ObjectDisposedException`

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -0,0 +1,841 @@
...
+                        }
+                    });
+                }
+                var res = counter.Incremented();
+                count = counter.Count;
+                return res;
+            }
+        }
+
+        private async void EnqueueItem(IDocument doc, AnalysisPriority priority = AnalysisPriority.Normal) {
```
AlexanderSher: `async void`

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -993,7 +1025,53 @@ SafePipeHandle stdInClientHandle
...
+            Interlocked.Increment(ref _parsePending);
+            var response = await SendRequestAsync(
+                new AP.AddFileRequest {
+                    uri = uri,
+                    isTemporaryFile = isTemporaryFile,
+                    suppressErrorLists = suppressErrorList
+                }
+            ).ConfigureAwait(false);
+
+            return FinishAnalyze(
```
AlexanderSher: Should `FinishAnalyze` be called in `finally` block?

Python/Product/PythonTools/PythonTools/Intellisense/BufferParser.cs
```diff
@@ -294,48 +259,86 @@ IEnumerable<ITextSnapshot> snapshots
...
             }
 
+            var needRetry = new List<ITextSnapshot>();
             foreach (var task in tasks) {
                 var res = await task.Item2;
 
-                if (res != null) {
-                    Debug.Assert(res.failed != true);
+                if (res?.failed ?? false) {
+                    Interlocked.Decrement(ref analyzer._parsePending);
```
AlexanderSher: Do we need a `finally` block somewhere here to ensure that `_parsePending` is decremented in case of exception?

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -288,6 +308,31 @@ sealed class PythonTextBufferInfo {
...
+            if (tcs != null) {
+                return tcs.Task;
+            }
+            tcs = new TaskCompletionSource<AnalysisEntry>();
+            tcs = Interlocked.CompareExchange(ref _waitingForEntry, tcs, null) ?? tcs;
+            entry = AnalysisEntry;
+            if (entry != null) {
+                tcs.TrySetResult(entry);
+            } else if (cancellationToken.CanBeCanceled) {
+                cancellationToken.Register(() => tcs.TrySetResult(null));
```
AlexanderSher: `TrySetCanceled`?

Python/Product/PythonTools/PythonTools/PythonAnalysisClassifier.cs
```diff
@@ -99,10 +98,14 @@ private class ClassificationComparer : IEqualityComparer<AP.AnalysisClassificati
...
-                return x.start == y.start && x.length == y.length && x.type == y.type;
+                return x.startLine == y.startLine &&
+                    x.startColumn == y.startColumn &&
+                    x.endLine == y.endLine &&
+                    x.endColumn == y.endColumn &&
+                    x.type == y.type;
             }
 
-            public int GetHashCode(AP.AnalysisClassification obj) => obj.start;
+            public int GetHashCode(AP.AnalysisClassification obj) => obj.startLine << 8 + obj.startColumn;
```
AlexanderSher: Why `endLine`, `endColumn` and `type` are skipped?

Python/Product/PythonTools/PythonTools/OutliningTaggerProvider.cs
```diff
@@ -171,30 +168,19 @@ internal class OutliningTagger : ITagger<IOutliningRegionTag>, IPythonTextBuffer
...
-            internal static TagSpan GetTagSpan(ITextSnapshot snapshot, int start, int end, int headerIndex = -1) {
+            internal static TagSpan GetTagSpan(SnapshotPoint start, SnapshotPoint end) {
                 TagSpan tagSpan = null;
+                var snapshot = start.Snapshot;
                 try {
-                    // if the user provided a -1, we should figure out the end of the first line
-                    if (headerIndex < 0) {
-                        headerIndex = snapshot.GetLineFromPosition(start).End.Position;
+                    SnapshotPoint hintEnd = end;
+                    if (start.GetContainingLine().LineNumber + 5 < hintEnd.GetContainingLine().LineNumber) {
```
AlexanderSher: Why `5`?

AlexanderSher:
>             _server.OnPublishDiagnostics += OnPublishDiagnostics;

These events aren't unsubscribe in `Dispose`

---
Refers to: Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs:69 in 60f3aca. [](commit_id = 60f3aca16f217cbb3d867961b4f1f5db26375501, deletion_comment = False)

AlexanderSher:
>                     (oldExtension as IDisposable)?.Dispose();

`oldExtension` can be disposed outside of the critical section. Not sure how critical is it.

---
Refers to: Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs:270 in 60f3aca. [](commit_id = 60f3aca16f217cbb3d867961b4f1f5db26375501, deletion_comment = False)

huguesv:
I tested it for another hour or so.

Feature-wise, completions seem to work as expected, including when importing packages defined in the project.

Unfortunately, completions don't work for very long. Within 5 to 10 mins of editing, all IntelliSense stops working. That's if I'm not running under the debugger, because if I do that I usually get stuck in an infinite wait at some point. I guess if I'm not debugging that's when IntelliSense stop working.

image

I've been trying to find what causes it, but there's no simple instruction for it. Sometimes it's after I type really fast. Sometimes it's after I switch to a different file in the editor then come back. One time I added a unit test, ran it then after that wasn't getting completions. Reloading the solution gets things working again.

Python/Product/Analysis/Intellisense/LineInfo.cs
zooba: No idea, I didn't write this class. But you're right, it does only seem to be treated as an integral value, so it can be a struct.

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
zooba: I'm hesitant to change this code right now - it can be changed in its own PR (this PR has just moved it)

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
zooba: File a new issue for this - I'm not going to start messing with this algorithm right now.

Python/Product/Analysis/Intellisense/ProjectEntryMap.cs
zooba: This entire class is unused now - I'll delete it.

Python/Product/Analysis/LanguageServer/CallbackEventArgs.cs
zooba: Matching the spec here, using the prefix for keywords. If anything, I'd change it to `params_`, but it's actually easier to type with the `@` at the start.

Python/Product/Analysis/LanguageServer/DiagnosticsErrorSink.cs
zooba: That actually changes the semantics away from what I want this class to be. I'll change the argument to `Action<Diagnostic>` and remove the (unused) Diagnostics property.

Python/Product/Analysis/LanguageServer/CallbackEventArgs.cs
AlexanderSher: Ok

Python/Product/Analysis/LanguageServer/Messages.cs
zooba: These are the public interface for the LSP implementation and I want them to be as close as possible to JSON objects for ease of mapping.

Python/Product/Analysis/LanguageServer/ServerBase.cs
zooba: Part of the LSP spec

Python/Product/Analysis/LanguageServer/ServerBase.cs
zooba: Part of the LSP spec

Python/Product/Analysis/LanguageServer/Messages.cs
AlexanderSher: Ok, got it.

Python/Product/Analysis/LanguageServer/VolatileCounter.cs
zooba: So review the usage carefully (it's needed for ParseQueue when we increment before scheduling the worker and decrement on failure or completion from a different thread). In this case, I believe it's simpler and clearer than passing around a disposable.

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Ah, that's why they weren't being removed. Good catch!

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
AlexanderSher: Ok. https://github.com/Microsoft/PTVS/issues/3566

Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
```diff
@@ -1,2381 +1,2139 @@
...
-                    doc = sig.Documentation,
-                    parameters = sig.Parameters.Select(
-                        param => new AP.Parameter() {
-                            name = param.Name,
-                            defaultValue = param.DefaultValue,
-                            optional = param.IsOptional,
-                            doc = param.Documentation,
-                            type = param.Type,
-                            variables = param.Variables?.Select(MakeReference).Where(r => r != null).ToArray()
-
```
AlexanderSher: These events aren't unsubscribed in `Dispose`

AlexanderSher: `async void`

AlexanderSher: `oldExtension` can be disposed outside of the critical section. Not sure how critical is it.

Python/Product/Analysis/LanguageServer/ServerBase.cs
zooba: Yes. You shouldn't call `CancelRequest` outside of an `AllowRequestCancellation` state or your object has been disposed.

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Correct. This is fire-and-forget by design.

Python/Product/Analysis/LanguageServer/ServerBase.cs
AlexanderSher: Ok

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
zooba: No. This is a straight refactor splitting one method into two.

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
zooba: By design. You need to check for null anyway, so no point forcing all the callers to also check for exceptions and treat them identically to null.

Python/Product/PythonTools/PythonTools/Intellisense/BufferParser.cs
zooba: Any exception coming out of the Send call is irrecoverable.

AlexanderSher: Ok

Python/Product/PythonTools/PythonTools/PythonAnalysisClassifier.cs
zooba: Because it's just a hash code, and the start location should be unique anyway.

Python/Product/PythonTools/PythonTools/OutliningTaggerProvider.cs
zooba: Ask the original author. I suspect it's to prevent unnecessarily large tooltips.

Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
zooba: Doesn't matter, neither object lives beyond the disposal.

zooba: We're already crashing by the time we get here.

zooba:
@huguesv Can you enable the connection logs for `Analyzer` and `ProjectAnalyzer` and when you see this, send the logs to me?

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
AlexanderSher: That's a bit unusual. For every awaited async method that accepts `cancellationToken`, caller expects that the code below canceled await won't be executed, no matter how deep in the call stack the actual handler of cancellation is placed. 
Here it is actually not true. While immediate caller can check for `null`, other methods in call stack has no idea about it.
Right now, none of the callers provide real `cancellationToken`, but it may change in future.

Python/Product/Analysis/LanguageServer/ParseQueue.cs
AlexanderSher: Ok, I was able to get an exception when `Thread.CurrentThread == callingThread`. There is no `ParseQueue.Parse` in the call stack, just continuation executed in the same `ThreadPool` thread the task with `ParseQueue.Parse` was executed before. This is normal - `ThreadPoolTaskScheduler` doesn't start new thread for each task, they are reused.

I don't have a repro, but here is what I think has happened:
1. `previousTask.Status == TaskStatus.Running` was true
2. `ParseQueue.Parse` has returned `res` (which wasn't canceled at that moment)
3. `res` was saved into `_parsing` and returned from `ParseQueue.Enqueue`
4. Since task is canceled, `ParseWorker` and `ProjectEntry.UpdateTree` won't be executed
5. `AnalysisQueue.Worker` will wait for `_workEvent` forever. This is the reasons (or at least one of the reasons) for the hanging @huguesv has found.

AlexanderSher: So, if you have seen a continuation delegate and `ParseQueue.Parse` being in the same call stack without `TaskContinuationOptions.ExecuteSynchronously` explicitly specified, I believe it should be filed as a bug in Task implementation, cause such behavior is unexpected (synchronously executed continuation is an ad-hoc scenario).

If it is just the thread ids that match, then I think this method should be reduced to 
```cs
        private Task<IAnalysisCookie> Parse(IDocument doc, PythonLanguageVersion languageVersion, Task<IAnalysisCookie> previousTask) 
            => previousTask?.ContinueWith(t => ParseWorker(doc, languageVersion), CancellationToken.None, TaskContinuationOptions.None, TaskScheduler.Default)
               ?? Task.Factory.StartNew(() => ParseWorker(doc, languageVersion), CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default);
```

`TaskScheduler.Default` ensures that task won't be inlined because of some unexpected logic in the `TaskScheduler.Current`. You can use `TaskCreationOptions.RunContinuationsAsynchronously` and `TaskContinuationOptions.RunContinuationsAsynchronously` to ensure that **all** continuations of these tasks are executed as separate thread pool work items.

AlexanderSher:
>             if (entry.Analysis != null) {

`GetPythonEntry` can return null, so here we can get `NRE`.

---
Refers to: Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs:1430 in 60f3aca. [](commit_id = 60f3aca16f217cbb3d867961b4f1f5db26375501, deletion_comment = False)

AlexanderSher:
>             if (entry.Analysis != null) {

`entry` isn't checked for null

---
Refers to: Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs:1431 in ada8e2a. [](commit_id = ada8e2a4a3182866a8c6a2b73672c5cb2e8ef5ec, deletion_comment = False)

AlexanderSher:
>             var tree = entry.Tree;

`entry` isn't checked for `null`

---
Refers to: Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs:2076 in ada8e2a. [](commit_id = ada8e2a4a3182866a8c6a2b73672c5cb2e8ef5ec, deletion_comment = False)

AlexanderSher:
>             var tree = entry.Tree;

`tree` isn't used later

---
Refers to: Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs:2076 in ada8e2a. [](commit_id = ada8e2a4a3182866a8c6a2b73672c5cb2e8ef5ec, deletion_comment = False)

AlexanderSher:
>             if (entry.Analysis != null) {

`entry` isn't checked for `null`

---
Refers to: Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs:1431 in ada8e2a. [](commit_id = ada8e2a4a3182866a8c6a2b73672c5cb2e8ef5ec, deletion_comment = False)

PR #3557, opened 2018-01-12
Adds tests for the experimental debugger

karthiknadig:


Python/Tests/DebuggerUITests/DebugProjectUITests.cs
```diff
@@ -40,202 +41,263 @@ public class DebugProjectUITests {
...
 
         /// <summary>
         /// Tests using a custom interpreter path that is relative
         /// </summary>
-        public void DebugPythonCustomInterpreter(VisualStudioApp app, DotNotWaitOnNormalExit optionSetter) {
-            var sln = app.CopyProjectForTest(@"TestData\RelativeInterpreterPath.sln");
-            var project = app.OpenProject(sln, "Program.py");
-            var interpreterPath = Path.Combine(PathUtils.GetParent(sln), "Interpreter.exe");
-            File.Copy(PythonPaths.Python27.InterpreterPath, interpreterPath, true);
+        public void DebugPythonCustomInterpreter(VisualStudioApp app, bool useVsCodeDebugger, DotNotWaitOnNormalExit optionSetter) {
```
int19h: Is the `optionSetter` argument actually used anywhere? It doesn't look like it was, even before the change. Ditto in other methods here.

Maybe get rid of that class altogether, if this was the only place where it was used?

huguesv: Test runner inspects the method parameters, creates an instance of that class and passes it in. After execution, it casts to IDisposable and calls Dispose on it. So it's used, but you don't see the code.

Python/Tests/DebuggerUITests/DebugProjectUITests.cs
```diff
@@ -835,6 +1068,21 @@ public class DotNotWaitOnNormalExit : PythonDebuggingGeneralOptionsSetter {
             Assert.AreEqual(mode, app.Dte.Debugger.CurrentMode);
         }
 
+        internal static void KillPythonProcess(TimeSpan delay) {
+            // This is needed as a workaround for the bug https://github.com/Microsoft/PTVS/issues/3547. In these tests we did not use
+            // ProcessScope, because there were cases where the test would time out before ProcessScope.Dispose was called. We needed this
+            // to make it easier to test the actual functnality of the experimental debugger.
+            HashSet<int> alreadyRunning = new HashSet<int>(System.Diagnostics.Process.GetProcessesByName("python").Select(p => p.Id));
```
huguesv: np: could have used `var alreadyRunning`, and `functnality` typo above

PR #3560, opened 2018-01-16
Cookiecutter virtual environment creation fixes

### LINKED ISSUE #3485: Cookiecutter fails to install the project after a cookiecutter template was cloned
In latest Visual studio 2017I go to File --> New -->--> From Cookiecutter. Then I select an arbitrary template and click Run or Next. I see in the output window that the repository is cloned, but then I get an error, e.g.

```
----- Cloning template 'audreyr/cookiecutter-pypackage' -----
Cloning into 'cookiecutter-pypackage'...
----- Successfully cloned template 'audreyr/cookiecutter-pypackage' to 'C:\Users\pavol\.cookiecutters\cookiecutter-pypackage' -----
----- Installing cookiecutter -----
C:\tools\Anaconda3\python.exe returned an exit code of 1.
----- Failed to install cookiecutter -----
```

huguesv:
Fix #3485 

When selecting a Python interpreter for use with cookiecutter, prefer CPython over others.

When creating a virtual environment fails, try again using --without-pip option (workaround for Anaconda), then install pip.

Delete the virtual environment folder before creating a new one, to ensure that broken virtual environments are repaired (such as when the base environment was removed/moved).

Also, send all output from environment creation and installation of packages to output window to help diagnose future issues.

### DIFF
--- a/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
+++ b/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
@@ -71,15 +71,73 @@ namespace Microsoft.CookiecutterTools.Model {
 
         public async Task CreateCookiecutterEnv() {
             // Create a virtual environment using the global interpreter
-            var interpreterPath = _interpreter.InterpreterExecutablePath;
-            var output = ProcessOutput.RunHiddenAndCapture(interpreterPath, "-m", "venv", _envFolderPath);
+            try {
+                await CreateVenv();
+            } catch (ProcessException ex) when (ex.Result.ExitCode == 1) {
+                // Create fails on some Anaconda due to venv failing to install pip.
+                // Try again by installing pip ourselves.
+                await CreateVenvWithoutPipThenInstallPip();
+            }
+        }
 
-            await WaitForOutput(interpreterPath, output);
+        private async Task CreateVenv() {
+            if (Directory.Exists(_envFolderPath)) {
+                _redirector.WriteLine(Strings.InstallingCookiecutterDeleteEnv.FormatUI(_envFolderPath));
+                Directory.Delete(_envFolderPath, true);
+            }
+
+            _redirector.WriteLine(Strings.InstallingCookiecutterCreateEnv.FormatUI(_envFolderPath));
+            var output = ProcessOutput.Run(
+                _interpreter.InterpreterExecutablePath,
+                new[] { "-m", "venv", _envFolderPath },
+                null,
+                null,
+                false,
+                _redirector
+            );
+            await WaitForOutput(_interpreter.InterpreterExecutablePath, output);
+        }
+
+        private async Task CreateVenvWithoutPipThenInstallPip() {
+            if (Directory.Exists(_envFolderPath)) {
+                _redirector.WriteLine(Strings.InstallingCookiecutterDeleteEnv.FormatUI(_envFolderPath));
+                Directory.Delete(_envFolderPath, true);
+            }
+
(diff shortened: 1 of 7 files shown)


Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -71,15 +71,73 @@ class CookiecutterClient : ICookiecutterClient {
...
+            } catch (ProcessException ex) when (ex.Result.ExitCode == 1) {
+                // Create fails on some Anaconda due to venv failing to install pip.
+                // Try again by installing pip ourselves.
+                await CreateVenvWithoutPipThenInstallPip();
+            }
+        }
 
-            await WaitForOutput(interpreterPath, output);
+        private async Task CreateVenv() {
+            if (Directory.Exists(_envFolderPath)) {
```
AlexanderSher: Directory may be deleted right after you've checked for it. It is more reliable to just call `Directory.Delete` and catch `DirectoryNotFoundException`.

Python/Product/Cookiecutter/Model/CookiecutterClientProvider.cs
```diff
@@ -38,11 +38,21 @@ static class CookiecutterClientProvider {
...
-                .OrderByDescending(x => x.Configuration.Version)
-                .FirstOrDefault(x => x.Configuration.Version >= new Version(3, 3));
-            return compatible != null ? new CookiecutterPythonInterpreter(compatible.Configuration.InterpreterPath) : null;
+                .Where(x => x.Configuration.Version >= new Version(3, 3))
+                .OrderByDescending(x => x.Configuration.Version);
+
+            // Prefer a CPython installation if there is one because
+            // some Anaconda installs have trouble creating a venv.
+            var cpython = all
+                .Where(x => x.Vendor.IndexOf("Python Software Foundation", StringComparison.InvariantCultureIgnoreCase) == 0);
```
AlexanderSher: `StringComparison.OrdinalIgnoreCase`

Python/Product/Cookiecutter/Model/CookiecutterClientProvider.cs
```diff
@@ -38,11 +38,21 @@ static class CookiecutterClientProvider {
...
         private static CookiecutterPythonInterpreter FindCompatibleInterpreter() {
             var interpreters = PythonRegistrySearch.PerformDefaultSearch();
-            var compatible = interpreters
+            var all = interpreters
                 .Where(x => File.Exists(x.Configuration.InterpreterPath))
-                .OrderByDescending(x => x.Configuration.Version)
-                .FirstOrDefault(x => x.Configuration.Version >= new Version(3, 3));
-            return compatible != null ? new CookiecutterPythonInterpreter(compatible.Configuration.InterpreterPath) : null;
+                .Where(x => x.Configuration.Version >= new Version(3, 3))
+                .OrderByDescending(x => x.Configuration.Version);
```
AlexanderSher: `all` got enumerated twice. You can call `ToList()` or `ToArray()` at the end to avoid it.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
huguesv: I still want to check it exists because I only want to show the message if we'll try to delete it. I'll add the catch so that we don't fail if it no longer exists by the time we delete.

AlexanderSher: Ok

PR #3571, opened 2018-01-18
Fixes #3570 Buffer mismatches

### LINKED ISSUE #3570: Buffer mismatches
Sometimes while typing the two text models can still get out of sync.

I'll figure it out.

zooba:
Currently, this fixes multiple parses being started simultaneously for the same document. I still need to verify that this was the underlying cause, but it's not too early to start reviewing this.

In short, the problem we had before is that `ConcurrentDictionary.AddOrUpdate` does not guarantee that it will only call one of the factory functions, so we could end up with two parsers for the one counter increment, as well as two conflicting results that _may_ have been causing incorrect responses and hence comparisons leading to the assertion.

### DIFF
--- a/Python/Product/Analysis/LanguageServer/ParseQueue.cs
+++ b/Python/Product/Analysis/LanguageServer/ParseQueue.cs
@@ -51,10 +51,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
 
             var task = new ParseTask(this, doc, languageVersion);
             try {
-                var c = _parsing.AddOrUpdate(doc, task, (d, prev) => prev?.ContinueWith(task) ?? task);
-                if (c == task) {
-                    task.Start();
-                }
+                _parsing.AddOrUpdate(doc, task, (d, prev) => task.ContinueAfter(prev)).Start();
                 return task.Task;
             } finally {
                 task.DisposeIfNotStarted();
@@ -127,10 +124,10 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
             private readonly IPythonParse _parse;
 
             private readonly TaskCompletionSource<IAnalysisCookie> _tcs;
-            private ParseTask _next;
+            private Task<IAnalysisCookie> _previous;
 
             // State transitions:
-            //  UNSTARTED -> QUEUED     when passed to ContinueWith
+            //  UNSTARTED -> QUEUED     when Start() called and _previous is set
             //  UNSTARTED -> DISPOSED   when DisposeIfNotStarted() is called
             //  UNSTARTED -> STARTED    when Start() is called
             //  QUEUED    -> STARTED    when Start(QUEUED) is called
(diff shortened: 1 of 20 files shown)


Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -84,86 +82,74 @@ IAnalysisCookie cookie
...
-            lock (this) {
-                if (_updatesPending > 0) {
-                    _updatesPending--;
+        private readonly ManualResetEventSlim _pendingParse = new ManualResetEventSlim(true);
+        private long _expectedParse;
+
+        private class ActivePythonParse : IPythonParse {
+            private readonly ProjectEntry _entry;
+            private readonly long _expected;
+            private bool _completed;
```
AlexanderSher: Has to be either volatile or accessed under lock

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -84,86 +82,74 @@ IAnalysisCookie cookie
...
-
-            lock (this) {
-                _waiters--;
-                if (_waiters == 0 &&
-                    Interlocked.CompareExchange(ref _sharedWaitEvent, _curWaiter, null) != null) {
-                    _curWaiter.Dispose();
-                }
-                _curWaiter = null;
+        public IPythonParse WaitForCurrentParse(int timeout = -1) {
+            _pendingParse.Wait(timeout);
```
AlexanderSher: You need to check if `Wait` has returned `true` or `false`

Python/Product/Analysis/LanguageServer/ParseQueue.cs
```diff
@@ -32,181 +32,223 @@ class ParseQueue {
...
+            private void DisposeWorker() {
+                Debug.Assert(Volatile.Read(ref _state) == DISPOSED);
+
+                var next = Interlocked.Exchange(ref _next, null);
+                _queue._parsingInProgress.Decrement();
+                _queue._parsing.TryUpdate(_document, next, this);
+                next?.Start(QUEUED);
+            }
+
+            public ParseTask ContinueWith(ParseTask nextTask) {
```
AlexanderSher: `ContinueWith` is invoked in inside `updateValueFactory`, which means that you don't know how how many times it is called and which of results will be used.
For example, if there are two concurrent requests, and `_parsing` already has a value, `ContinueWith` may be called 3 times:
- Invocation of  `updateValueFactory` for the first request: `ContinueWith` changes `_next`  to the `ParseTask` from first request
- Invocation of  `updateValueFactory` for the second request: `ContinueWith` finds that `_next` is already set and changes `_next` of the the `ParseTask` from first request to the `ParseTask` from second request
- Second invocation processed faster than the first one, so dictionary accepts `ParseTask` from second request as a new value
- Another invocation of the `updateValueFactory` for the first request: `ContinueWith` changes `_next`  of the the `ParseTask` from second request to the `ParseTask` from first request

As a result, we have a loop.

Python/Product/Analysis/ProjectEntry.cs
zooba: Either that or change the definition to say "return the latest after the delay", which is the usual behaviour anyway. But returning `null` here is what the previous version of this did and it helps improve logging.

Python/Product/Analysis/ProjectEntry.cs
zooba: Each instance of this class is only used on a single thread - it's the call back into ProjectEntry that needs to be controlled.

Python/Product/Analysis/LanguageServer/ParseQueue.cs
zooba: Ah, it's worse than that. They just use a `while (true)` loop and keep calling either factory until something succeeds. I think I can make this safer though.

AlexanderSher: I think I've managed to fix it with smaller change: https://github.com/AlexanderSher/PTVS/commit/15fab6679bdbc759c44e32389a6f8bcf96b8ccb8

Python/Product/Analysis/LanguageServer/ParseQueue.cs
```diff
@@ -32,181 +32,224 @@ class ParseQueue {
...
+                _parse = (_document as IPythonProjectEntry)?.BeginParse();
+
+                _tcs = new TaskCompletionSource<IAnalysisCookie>();
+            }
+
+            public Task<IAnalysisCookie> Task => _tcs.Task;
+
+            public void DisposeIfNotStarted() {
+                if (Interlocked.CompareExchange(ref _state, DISPOSED, UNSTARTED) == UNSTARTED) {
+                    _parse?.Dispose();
```
AlexanderSher: This code is unreachable. `parsing.AddOrUpdate` will always return current task,  `Start` is always called before `DisposeIfNotStarted` and it changes `_state` to either `QUEUED` or `STARTED` before any exception can occur.

zooba: There are a few ways that it could be reached (all very exceptional), so I'd rather leave it there. But I have streamlined the transitions a bit.

Python/Product/Analysis/LanguageServer/ParseQueue.cs
```diff
@@ -32,181 +32,225 @@ class ParseQueue {
...
-                        }
-                        complete = true;
-                        result = UpdateTree(pyEntry, buffers);
-                    } finally {
-                        if (!complete) {
-                            AbortParsingTree(pyEntry);
-                        }
+                        previous.ContinueWith(StartAfterTask);
+                        return Task;
+                    } catch (ObjectDisposedException) {
```
AlexanderSher: `ContinueWith` doesn't throw `ObjectDisposedException` in .Net 4.5 or later, even if you have disposed a `Task` (which you actually shouldn't: https://blogs.msdn.microsoft.com/pfxteam/2012/03/25/do-i-need-to-dispose-of-tasks/).

Python/Product/Analysis/LanguageServer/ParseQueue.cs
AlexanderSher: > There are a few ways that it could be reached (all very exceptional), so I'd rather leave it there. 

Sure, but can you give an example? I can see it happening only if there is a bug in `ConcurrentDictionary` that it either crashes with exception or returns a value that is different from what has been added or updated.

Python/Product/Analysis/LanguageServer/ParseQueue.cs
zooba: They need to update the documentation :)

PR #3574, opened 2018-01-19
Fixes #3573 Unexpected error analysis/debug interactive

### LINKED ISSUE #3573: Unexpected error analysis/debug interactive
In design mode, I closed my output window which brought my debug interactive window to the front, and produced this error. I was able to repro that a second time.

[Window Title]
devenv.exe

[Main Instruction]
An unexpected error occurred

[Content]
Please press Ctrl+C to copy the contents of this dialog and report this error to our issue tracker.

[^] Hide details  [Close]

[Expanded Information]
```
Build: 3.0.18018.1

System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.PythonTools.Intellisense.AnalysisEntryService.TryGetAnalyzer(ITextBuffer textBuffer, ProjectAnalyzer& analyzer, String& filename)
   at Microsoft.PythonTools.Intellisense.IntellisenseController.<AnalyzeBufferAsync>d__22.MoveNext()
   at Microsoft.PythonTools.Intellisense.IntellisenseController.<ConnectSubjectBufferAsync>d__23.MoveNext()
   at Microsoft.PythonTools.Infrastructure.VSTaskExtensions.<HandleAllExceptions>d__5.MoveNext()
```

zooba:
Fixes #3573 Unexpected error analysis/debug interactive
Ensures DocumentUri is checked for null

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Repl/PythonCommonInteractiveEvaluator.cs
+++ b/Python/Product/PythonTools/PythonTools/Repl/PythonCommonInteractiveEvaluator.cs
@@ -184,7 +184,13 @@ namespace Microsoft.PythonTools.Repl {
         }
 
         public virtual Uri DocumentUri { get => _documentUri; protected set => _documentUri = value; }
-        public virtual Uri NextDocumentUri() => new Uri(DocumentUri, $"#{++_nextDocumentIndex}");
+        public virtual Uri NextDocumentUri() {
+            var d = DocumentUri;
+            if (d != null) {
+                return new Uri(d, $"#{++_nextDocumentIndex}");
+            }
+            return null;
+        }
 
         internal void WriteOutput(string text, bool addNewline = true) {
             var wnd = CurrentWindow;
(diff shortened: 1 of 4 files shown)


Python/Product/PythonTools/PythonTools/Repl/PythonCommonInteractiveEvaluator.cs
```diff
@@ -184,7 +184,13 @@ internal abstract class PythonCommonInteractiveEvaluator :
         }
 
         public virtual Uri DocumentUri { get => _documentUri; protected set => _documentUri = value; }
-        public virtual Uri NextDocumentUri() => new Uri(DocumentUri, $"#{++_nextDocumentIndex}");
+        public virtual Uri NextDocumentUri() {
+            var d = DocumentUri;
+            if (d != null) {
+                return new Uri(d, $"#{++_nextDocumentIndex}");
+            }
+            return null;
```
huguesv: I'm trying to find out what will happen when this returns `null`.

`PythonTextBufferInfo.DocumentUri` would return `null`.

That's used in `IntellisenseController.AnalyzerBufferAsync` and passed to `ProjectAnalyzer.AnalyzeFileAsync`, which will throw ANE,

It will be caught with `HandleAllException` in `IntellisenseController..ConnectSubjectBuffer` but that's going to log the error?

Probably not the right thing to do since debug repl is not an unexpected/exceptional case? We should avoid calling into `AnalyzeFileAsync` in the first place?

zooba: Probably. If we don't have a document URI then we really shouldn't be analyzing anything. The tricky part is making sure that it doesn't get cached and never updated when you start debugging. I'll take a closer look this afternoon.

zooba:
Turns out it wasn't a good idea to return the active frame here, as we currently have no concept of "analysis in this document without editing it". So I just switched back to the default (auto-generated) URI and handle the `null` case.

PR #3586, opened 2018-01-19
Fix browse filter to enable browse button in add reference dialog.

### LINKED ISSUE #3567: Add Reference... is missing browse tab
(from https://developercommunity.visualstudio.com/content/problem/181732/add-reference-is-missing-browse-tab.html)

I've made a IoT Core Python Background Application. In order to use the GPIO board on my Rasperry Pi, i need to add some references...

MS-documentation

•Right-click on the References node in the Python project
•Click Add Reference…
•Click on Browse tab
•Download pywindevices content from here
•Navigate to ARM for RPi2 or win32 for MBM
[...]

When i press "Add References..." the add references window pops up, but is missing the browse tab

huguesv:
Fix #3567

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
@@ -1500,7 +1500,7 @@ namespace Microsoft.PythonTools.Project {
 
         protected override string AddReferenceExtensions {
             get {
-                return null;
+                return Strings.AddReferenceExtensions;
             }
         }
 


huguesv:
Hmm, browse really only supports assembly references. Anything that fails to be recognized as an assembly (like .pyd files) is ignored/not added as a reference. We used to have code that falls back to creating a ComReferenceNode when it's determined not to be an assembly, but that's been #ifdef FALSE.

zooba:
Yeah, we need to make sure this does something useful when enabled.

There’s a thread with the IoT team going on to see what they need that will inform any change here. I haven’t looked at the PR, but it should be possible to only enable this for UWP projects and then at least most of our users won’t ever notice.

huguesv:
Closing this since it's either insufficient or unnecessary.

PR #3588, opened 2018-01-20
Environment list tests fixes

huguesv:
ChangeDefault still fails occasionally but the others are doing better now.


Python/Product/EnvironmentsList/ToolWindow.xaml.cs
```diff
@@ -483,6 +481,12 @@ await StartRefreshDBAsync(view)
                     _condaEnvironmentView = null;
                 }
                 _onlineHelpView = EnvironmentView.CreateOnlineHelpEnvironmentView();
+
+                if (synchronous) {
+                    Dispatcher.Invoke(FirstUpdateEnvironments);
+                } else {
+                    Dispatcher.InvokeAsync(FirstUpdateEnvironments).Task.DoNotWait();
```
zooba: Not a big deal, but the way I've approached this in other places is to keep the task around as an internal variable and wait on it in the test.

zooba:
@huguesv Any reason not to merge this?

huguesv:
I was still thinking if I would improve on this but there's more important stuff to do. I fixed the conflict using the online tool (first time I do that). I'll merge.

PR #3589, opened 2018-01-20
Fixes #2575 No error highlighting for NameErrors

### LINKED ISSUE #2575: No error highlighting for clear "NameError"s?
Suppose I have a script like this:

'''
#!/usr/bin/env python

a = some_variable
'''
That will raise a NameError without fail, yet this kind of very simple error isn't highlighted, and (bafflingly) doesn't even show up when running pylint in VS2017 Preview (or VS2015 for that matter.) Is there something I'm missing, or is this kind of error highlighting just not here? If not, consider this a feature request--I know not everyone likes to use this sort of thing all the time, but for many of us, it helps our productivity immensely.

zooba:
Fixes #2575 No error highlighting for NameErrors
Adds diagnostic message for variables that are created on read by our analyzer
Also adds default module variables (`__name__`, etc.) as these were causing an unfortunate number of warnings on otherwise good code.

When tested against an old version of the Azure SDK, the only false positive is for `import os.path` not resolving, which is unrelated to this change.


zooba:
The majority of the change is adding a way to get errors out from within the analyzer, which was not there before.

I also fixed a minor but annoying bug where the AST analyzer was deleting its own default DB.

zooba:
This is no longer targeting the release branch, so no reason not to merge.

Python/Product/PythonTools/PythonTools/Intellisense/TaskProvider.cs
```diff
@@ -928,61 +937,97 @@ string sourceFile
...
+        public int IsCustomColumnReadOnly(ref Guid guidView, uint iCustomColumnIndex, out int pfReadOnly) {
+            pfReadOnly = 1;
+            return VSConstants.S_OK;
+        }
+
+        public int GetHierarchy(out IVsHierarchy ppProject) {
+            if (_projectHierarchyIsNull || ProjectHierarchy != null) {
+                ppProject = ProjectHierarchy;
+            } else if (!string.IsNullOrEmpty(SourceFile)) {
+                ppProject = ProjectHierarchy = _serviceProvider.GetProjectFromFile(SourceFile);
```
huguesv: This caching will break if someone changes `SourceFile` after this is called (its setter is public so it can technically be done). We could change its setter to private or make it a read-only field. `ProjectHierarchy` also does not need to be public.

Python/Product/Analysis/Analyzer/ExpressionEvaluator.cs
```diff
@@ -135,10 +136,11 @@ internal class ExpressionEvaluator {
             }
 
             var res = ProjectState.BuiltinModule.GetMember(node, _unit, name);
-            if (createIn != null && !res.Any()) {
-                var refs = createIn.CreateVariable(node, _unit, name, addRef);
+            if (!_unit.ForEval && !res.Any()) {
+                ProjectState.AddDiagnostic(node, _unit, "unknown variable '{0}'".FormatUI(name), LanguageServer.DiagnosticSeverity.Warning, "used-before-assignment");
```
huguesv: It would be nice to consolidate all the error/warning names and descriptions in static class(es). They will be easier to review if they're all in one place.

Python/Product/PythonTools/PythonTools/Intellisense/TaskProvider.cs
zooba: This is very much a "don't do that" situation - it's nice to be able to initialize the optional members of class using the assignment syntax rather than having a huge constructor. But in this case, it's not an optional member, so it can be read only.

zooba:
New tests are fine - there are still failures in this branch that we haven't forward-ported from 15.6 yet. I'll get onto that later today.

PR #3600, opened 2018-01-22
Fixes #3592 Migrate signature help to language server

### LINKED ISSUE #3592: Migrate signature help to language server
Subtask of #3484.

zooba:
Fixes #3592 Migrate signature help to language server
Also migrates finding references, which was needed to be able to provide enough information back to VS to keep rename working.

Still enhancing a couple of tests.


Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -361,6 +380,154 @@ public sealed class Server : ServerBase, IDisposable {
             return item;
         }
 
+        public override async Task<SignatureHelp> SignatureHelp(TextDocumentPositionParams @params) {
```
AlexanderSher: This method is 100% sync. Is it by design?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -361,6 +380,154 @@ public sealed class Server : ServerBase, IDisposable {
...
+            var uri = @params.textDocument.uri;
+            GetAnalysis(@params.textDocument, @params.position, @params._version, out var entry, out var tree);
+
+            if (_traceLogging) {
```
AlexanderSher: If we have `MessageType` parameter that controls severity, why do you need to check for `_traceLogging`?

zooba: The parameter doesn't decide whether the event gets raised or not, only what the client does with it once it has received it. Trace logging is for extremely verbose messages that we don't even want to send in most cases, or they'll clog up the IPC channel and hurt performance (particularly in remote scenarios).

Python/Product/Analysis/LanguageServer/Server.cs
zooba: It is right now, but in the future it may not be. This function is public stable API, so keeping it a Task even if it doesn't matter yet is fine, and historically we've missed enough `Task.FromResult(null)` calls to cause crashes that appear to come from other places in code, so leaving it `async` is also safer. I'm not concerned about the performance implications of this particular call.

Python/Product/PythonTools/PythonTools/Navigation/Peek/PythonPeekResultSource.cs
```diff
@@ -49,43 +49,26 @@ internal sealed class PythonPeekResultSource : IPeekResultSource {
...
                 location.FilePath,
-                startLine,
-                startColumn,
-                endLine,
-                endColumn,
-                location.Line - 1,
-                location.Column - 1,
-                location.Line - 1,
-                location.Column - 1,
+                (location.DefinitionSpan?.Start.Line ?? location.Span.Start.Line) - 1,
```
AlexanderSher: you can introduce `var definitionSpan = location.DefinitionSpan ?? location.Span)` and simplify code by getting rid of null checks

Python/Tests/Analysis/LanguageServerTests.cs
```diff
@@ -309,6 +309,44 @@ class MyClass:
...
+def f(a : int, b: int): pass
+def f(x : str, y: str): pass
+def f(a = 2, b): pass
+
+");
+
+            await AssertSignature(s, mod, 0, 2,
+                new string[] { "f(a:int=)", "f(a:int=2,b:int=)", "f(x:str=,y:str=)" },
+                new string[0]
+            );
```
AlexanderSher: Do you need to call `await s.UnloadFileAsync(mod);` at the end of test?

zooba: No, we throw the server away completely.

zooba:
Given this no longer targets the release branch, I believe it's ready to merge.

Python/Product/Analysis/LanguageServer/Server.cs
AlexanderSher: Ok, then how about introducing a helper method for it:
```cs
private void LogIfTraceEnabled(string message) {
    if (_traceLogging) {
        LogMessage(MessageType.Log, message);
    }
}
```

zooba: Sure. I can do that

zooba:
Just waiting on a clean test run after the big merge.

PR #3605, opened 2018-01-23
Fixes #3601 Updates the default stable version to 15.0 and default version to 15.7

### LINKED ISSUE #3601: Update assembly versions to 15.0
We have significant breaking changes coming in the next release, so we should take this opportunity to completely close down any APIs that we don't want to be public and update our assembly version.

It probably makes the most sense to jump straight to 15.0 and match the rest of VS.

zooba:
Same as #3602 but for master.


zooba:
Build is clean; merging.

Issue #3301, opened 2017-11-07
SocketException when installing package with elevation


JeffJQ:
[Window Title]
Visual Studio - Python support

[Main Instruction]
An unexpected error occurred

[Content]
Please press Ctrl+C to copy the contents of this dialog and report this error to our issue tracker.

[^] Hide details  [Close]

[Expanded Information]
```
Build: 3.0.17297.4

System.Net.Sockets.SocketException (0x80004005): Only one usage of each socket address (protocol/network address/port) is normally permitted
   at System.Net.Sockets.Socket.DoBind(EndPoint endPointSnapshot, SocketAddress socketAddress)
   at System.Net.Sockets.Socket.Bind(EndPoint localEP)
   at System.Net.Sockets.TcpListener.Start(Int32 backlog)
   at Microsoft.PythonTools.Infrastructure.ProcessOutput.RunElevated(String filename, IEnumerable`1 arguments, String workingDirectory, IEnumerable`1 env, Redirector redirector, Boolean quoteArgs, Boolean elevate, Encoding outputEncoding, Encoding errorEncoding)
   at Microsoft.PythonTools.Infrastructure.ProcessOutput.Run(String filename, IEnumerable`1 arguments, String workingDirectory, IEnumerable`1 env, Boolean visible, Redirector redirector, Boolean quoteArgs, Boolean elevate, Encoding outputEncoding, Encoding errorEncoding)
   at Microsoft.PythonTools.Interpreter.PipPackageManager.<InstallAsync>d__35.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.PythonTools.Interpreter.PipPackageManager.<InstallAsync>d__35.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PythonTools.EnvironmentsList.PipExtensionProvider.<InstallPackage>d__38.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #3624, opened 2018-01-24
Fixes #3616 RunElevated ran out of ports

### LINKED ISSUE #3616: RunElevated ran out of ports
[Window Title]

[Main Instruction]
An unexpected error occurred

[Content]
Please press Ctrl+C to copy the contents of this dialog and report this error to our issue tracker.

[^] Hide details  [Close]

[Expanded Information]
```
Build: 3.0.17320.1

System.Net.Sockets.SocketException (0x80004005): Only one usage of each socket address (protocol/network address/port) is normally permitted
   at System.Net.Sockets.Socket.DoBind(EndPoint endPointSnapshot, SocketAddress socketAddress)
   at System.Net.Sockets.Socket.Bind(EndPoint localEP)
   at System.Net.Sockets.TcpListener.Start(Int32 backlog)
   at Microsoft.PythonTools.Infrastructure.ProcessOutput.RunElevated(String filename, IEnumerable`1 arguments, String workingDirectory, IEnumerable`1 env, Redirector redirector, Boolean quoteArgs, Boolean elevate, Encoding outputEncoding, Encoding errorEncoding)
   at Microsoft.PythonTools.Infrastructure.ProcessOutput.Run(String filename, IEnumerable`1 arguments, String workingDirectory, IEnumerable`1 env, Boolean visible, Redirector redirector, Boolean quoteArgs, Boolean elevate, Encoding outputEncoding, Encoding errorEncoding)
   at Microsoft.PythonTools.Interpreter.PipPackageManager.<InstallAsync>d__35.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.PythonTools.Interpreter.PipPackageManager.<InstallAsync>d__35.MoveNext()
   at Microsoft.PythonTools.EnvironmentsList.PipExtensionProvider.<InstallPackage>d__38.MoveNext()
   at Microsoft.PythonTools.EnvironmentsList.PipExtension.<UpgradePackage_Executed>d__12.MoveNext()
```

zooba:
Fixes #3616 RunElevated ran out of ports
Adds better helper method for selecting free port.


zooba:
Also fixes #3301

Python/Product/Common/Infrastructure/SocketUtils.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Net.NetworkInformation;
+using System.Runtime.ExceptionServices;
+
+namespace Microsoft.PythonTools.Infrastructure {
+    static class SocketUtils {
+        private static readonly Random _freePortRandom = new Random();
+
+        public static T WithFreePort<T>(Func<int, T> action, int minimumPort = 49152, int maximumPort = 65536, bool excludeActiveConnections = true, bool randomStart = true) {
```
AlexanderSher: Ok, I'll try to explain how this method works, correct me if I'm wrong:
1. It takes range of ports `[minimumPort, maximumPort)`, action that has to succeed on one of the ports, and exclusion rules
2. If `excludeActiveConnections  == true`, it excludes ports that are not free, otherwise it iterates through the whole range
3. if `randomStart == true`, it starts iteration from some random number in the range up to its end and then continues from its beginning, otherwise it starts iteration from `minimumPort`
4. If `action` doesn't throw exception during first iteration, method successfully completes, otherwise it throws exception from the last unsuccessful iteration
5. If no suitable port was found, it throws `InvalidOperationException("No free ports")`.

If everything is correct, I don't think it is better than simple iteration:
```cs
static Random _rand = new Random();
public static TcpListener StartForRandomPort(int minimumPort = 49152, int maximumPort = 65536, bool randomStart = true) {
    var range = maximumPort - minimumPort;
    var start = _rand.Next(range);
    for (var i = 0; i < range; ++i) {
        var port = (i + start) % range + minimumPort; 
        var listener = new TcpListener(IPAddress.Loopback, port);
        try {
            listener.Start();
            return listener;
        } catch (SocketException ex) when (ex.SocketErrorCode == SocketError.AddressAlreadyInUse) {
            listener.Stop();
        }
    }
    throw new InvalidOperationException("No free ports");
}
```

List of busy ports from `IPGlobalProperties` become invalid as soon as we get it, and it may be faster to check several ports explicitly than getting a list of all busy ports.

Also, we probably should handle the case when all ports are taken.

zooba: For this one particular case it makes sense to do it as a loop, but we actually have a few places where we do almost exactly this with different intentions - mostly to start debugging a web server project. In this case, we don't want to start listening ourselves because someone else will (and we also want to limit the number of attempts or it could take ages to exhaust all the possibilities...). And excluding busy ports (assuming that most are long-term busy, which TCPView seems to suggest) is going to be much quicker than launching a process that starts listening.

The callback also makes it easier to test in isolation, without other processes interfering. And disabling the in-use port check is mainly for testing too.

I didn't switch all the other uses over now, because it isn't necessary to solve this particular issue. Just planning for the future - do a code search for `GetFreePort` for examples of other places that could use this.

AlexanderSher: I have found 5 places where `IPGlobalProperties` is used in product code:
- `ProcessOutput.GetFreePort` in `Common` - used in `ProcessOutput.RunElevated` to create `TcpListener`
- `PythonWebLauncher.GetTestServerPort` - used to get port number
- `TestExecuter.GetFreePort` - used to get port number
- `ProcessOutput.GetFreePort` in `Cookiecutter` - used in `ProcessOutput.RunElevated` to create `TcpListener`
- `WslPythonLauncher.StartWithDebugger` - used to get port number

So in two cases it is a  `TcpListener` instance, and in three cases it is a port, which can be acquired from `TcpListener`:
```cs
public static int GetFreePort(int minimumPort = 49152, int maximumPort = 65536, bool randomStart = true)
    var tcpListener = StartForRandomPort(minimumPort, maximumPort, randomStart);
    var port = ((IPEndPoint)tcpListener.LocalEndpoint).Port;
    tcpListener.Stop();
    return port; 
}
```

Now, I've run BenchmarkDotNet for two code pieces:
```cs
var netProp = System.Net.NetworkInformation.IPGlobalProperties.GetIPGlobalProperties();
return netProp.GetActiveTcpConnections().Select(p => p.LocalEndPoint.Port).Union(netProp.GetActiveTcpListeners().Select(p => p.Port)).First();
```
and 
```cs
while (true) {
    var port = _rand.Next(49152, 65536);
    var tcpListener = new TcpListener(IPAddress.Loopback, port);
    try {
        tcpListener.Start();
        return port;
    } catch (SocketException ex) { 
    } finally {
        tcpListener.Stop();
    }
}
```

The first one is at least 17 times slower than the second one (average is 20). So using `IPGlobalProperties` is better only after 18 attempts of testing random port with `TcpListener`. Or, in case of simple linear iteration, randomly picking a port at the beginning of sequence of at least 18 busy ports. This is not impossible, but probability is very close to zero. 

Both implementations are fast enough, but one of them is three times shorter than the other and doesn't require callbacks.

zooba: In the cases where we want a port, can we assume the port will be available soon enough after we stop the listener? I don't know how quickly it works.

zooba:
If @AlexanderSher has no further comments then I'll merge.

PR #3628, opened 2018-01-25
Fixes #1810 Merge ImportWizard with ProjectWizards

### LINKED ISSUE #1810: Tool discovery and suggestions for special files
**Scenario:** I created a flask project from a cookiecutter template that uses a bunch of azure services and Docker containers (several templates out there seem to have docker-compose.yml files). 

**Feature:** After creating the project, Cookiecutter displays links to additional tools relevant to the project e.g. if it detects a docker-compose.yml file, it could give a link/option to download Docker Tools for VS. This will not only delight the user, but also give us currency for driving traffic to other products. Similarly, if there is an azuredeploy.json (like in Brett’s Web App template), we could display an option to deploy web app tools. 

**Potential Implementation:** Have a heuristic that looks for certain files to figure out what tooling might be needed. A more heavy weight approach could involve scanning the config files and imports. After detecting what tools are needed, we could display link to their marketing pages, or show an install button that’s hooked into in to the VSIX installation subsystem.

zooba:
Also removes unnecessary VS version from end of assembly versions.

(Note that the big warning about changing the ProjectWizards version is from when we used to keep these in the GAC. As that's no longer the case, it doesn't matter.)


zooba:
The only entry point into this is through the project template - all other ways of triggering the wizard go directly to it. And none of the localizations are required once the DLL is removed.

PR #3643, opened 2018-01-26
Partial fix to the error message after program exit with VSC debugger

### LINKED ISSUE #3537: VS Code debugger error message when closing the console
Start debugging a small script that prints to the console, no need for breakpoints
Console stays up at the end leaving only option to close the window
When you do that you get:
```
---------------------------
Microsoft Visual Studio
---------------------------
The debug adapter exited unexpectedly.
---------------------------
OK   
---------------------------
```

or sometimes:
```
---------------------------
Microsoft Visual Studio
---------------------------
Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.



An existing connection was forcibly closed by the remote host
---------------------------
OK   
---------------------------


```

karthiknadig:
Fixes #3537
This is partial fix because there is a change needed in VSC debugger where we fire `exitedEvent` with correct error code on program exit.

### DIFF
--- a/Python/Product/Debugger/DebugAdapterProcessStream.cs
+++ b/Python/Product/Debugger/DebugAdapterProcessStream.cs
@@ -1,0 +1,63 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


Python/Product/Debugger/DebugAdapterProcessStream.cs
```diff
@@ -0,0 +1,63 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.IO;
+using System.Net.Sockets;
+
+namespace Microsoft.PythonTools.Debugger {
+    public sealed class DebugAdapterProcessStream : Stream {
```
zooba: Does it need to be public?

PR #3647, opened 2018-01-26
Fixes #3646 Assertion on nested sequences in type annotation

### LINKED ISSUE #3646: Assertion on nested sequences in type annotation
When we finalize type annotations with nested sequences, we try to store the value against the same node. This is not allowed (because it has the same kind), so we need to come up with a better approach.

zooba:
Fixes #3646 Assertion on nested sequences in type annotation
Switches type annotation evaluations from real types to ProtocolInfo instances
Also fixes generator evaluation

This means we don't have to be as careful about minimizing instantiation, as the values will be fully recreated when we re-evaluate the annotation. So we can dispense with the node value storage completely and just store the entire evaluated annotation.

### DIFF
--- a/Python/Product/Analysis/Analyzer/DDG.cs
+++ b/Python/Product/Analysis/Analyzer/DDG.cs
@@ -165,9 +165,8 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
             foreach (var left in node.Left) {
                 if (left is ExpressionWithAnnotation annoExpr && annoExpr.Annotation != null) {
                     var annoType = _eval.EvaluateAnnotation(annoExpr.Annotation);
-                    var annoInst = annoType?.GetInstanceType();
-                    if (annoInst?.Any() == true) {
-                        _eval.AssignTo(node, annoExpr.Expression, annoInst);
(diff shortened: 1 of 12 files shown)


zooba:
Turns out, when you fix the highlighting for type annotation types, you find all sorts of issues.

One still remaining is that we don't correctly (more like magically) handle the difference between string literals that are supposed to be string literals, and string literals that are meant to be forward references. This means that the below annotation ends up with two definitions - one with the correct name and one without:

```
n1 : NamedTuple('n1', [('x', int), ['y', str]]) = …
```

Previously it worked because we'd only resolve it once, and `n1` didn't exist at that stage. Now it gets reevaluated and found, so we end up with a value rather than a string literal. The same is going to apply for 'x' and 'y', so we probably need to force annotation evaluators to do the lazy name lookup - we should still be able to parse the string literal eagerly and break it down if it parses.

zooba:
@huguesv Want me to fix the name confusion from my last comment? Or file it separately and deal with it later?

huguesv:
I'd say file it and merge this.

zooba:
Filed as #3667

PR #3656, opened 2018-01-29
Adds missing code analysis rules and resolves existing issues.

zooba:
This enables most of the code analysis rules we are supposed to be meeting in order to release. The majority of the fixes are for string comparisons (all comparisons apart from `==` and `!=` default to `CurrentCulture`, and the invariant culture doesn't define comparison - only conversion - so we should use ordinal for non-UI comparisons) and `XmlDocument` (which FxCop can't properly detect, so the messages are fixed and suppressed).

I suppressed the rule requiring you to pass a `CultureInfo` to `string.Format`, which prevents use of `$""` strings. Bear in mind that `$""` strings default to the current culture and not invariant, though for most cases this is okay and we don't need to mangle our code too much just to satisfy the static analyzer.

### DIFF
--- a/Build/xTVS.ruleset
+++ b/Build/xTVS.ruleset
@@ -11,7 +11,7 @@
     <Rule Id="CA1033" Action="None" /> <!-- Explicit interface implementation on base classes -->
     <Rule Id="CA1060" Action="Warning" />
     <Rule Id="CA1061" Action="Warning" />
-    <Rule Id="CA1063" Action="Warning" />
+    <Rule Id="CA1063" Action="None" /> <!-- Implement IDisposable properly -->
     <Rule Id="CA1065" Action="None" /> <!-- Throw from property accessors -->
     <Rule Id="CA1301" Action="Warning" />
     <Rule Id="CA1400" Action="Warning" />
--- a/Build/xTVS.ruleset
+++ b/Build/xTVS.ruleset
@@ -96,5 +96,117 @@
     <Rule Id="CA2226" Action="Warning" />
     <Rule Id="CA2227" Action="Warning" />
     <Rule Id="CA2239" Action="Warning" />
+
+    <Rule Id="CA1304" Action="Error" /> <!-- specify CultureInfo -->
+    <Rule Id="CA1305" Action="None" />  <!-- provide IFormatProvider -->
+    <Rule Id="CA1307" Action="Error" /> <!-- specify StringComparison -->
+    <Rule Id="CA1309" Action="Warning" /> <!-- use StringComparison.Ordinal -->
+
+    <Rule Id="CA900" Action="Error" />
+
+    <Rule Id="CA2001" Action="Error" />
+    <Rule Id="CA2103" Action="Error" />
+    <Rule Id="CA2107" Action="Error" />
+    <Rule Id="CA2109" Action="Error" />
+    <Rule Id="CA2110" Action="None" />
+    <Rule Id="CA2118" Action="Error" />
+    <Rule Id="CA2135" Action="Error" />
+    <Rule Id="CA2136" Action="Error" />
+    <Rule Id="CA2139" Action="Error" />
+    <Rule Id="CA2142" Action="Error" />
+    <Rule Id="CA2143" Action="Error" />
+    <Rule Id="CA2144" Action="Error" />
+    <Rule Id="CA2145" Action="Error" />
+    <Rule Id="CA2151" Action="Error" />
+    <Rule Id="CA2153" Action="Error" />
+    <Rule Id="CA5122" Action="Error" />
+    <Rule Id="CA5350" Action="Error" />
+    <Rule Id="CA5351" Action="Error" />
+    <Rule Id="CA5352" Action="Error" />
+    <Rule Id="CA5353" Action="Error" />
+    <Rule Id="CA5354" Action="Error" />
+    <Rule Id="CA5355" Action="Error" />
+    <Rule Id="CA5356" Action="Error" />
+    <Rule Id="CA5357" Action="Error" />
+    <Rule Id="CA3001" Action="Error" />
+    <Rule Id="CA3002" Action="None" />  <!-- XSS detection -->
+    <Rule Id="CA3003" Action="Error" />
+    <Rule Id="CA3004" Action="Error" />
+    <Rule Id="CA3005" Action="Error" />
+    <Rule Id="CA3006" Action="Error" />
+    <Rule Id="CA3007" Action="Error" />
+    <Rule Id="CA3008" Action="Error" />
+    <Rule Id="CA3009" Action="Error" />
+    <Rule Id="CA3010" Action="Error" />
+    <Rule Id="CA3011" Action="Error" />
+    <Rule Id="CA3012" Action="Error" />
+    <Rule Id="CA3101" Action="None" />
+    <Rule Id="CA3102" Action="Error" />
+    <Rule Id="CA3103" Action="Error" />
+    <Rule Id="CA3104" Action="Error" />
+    <Rule Id="CA3105" Action="Error" />
+    <Rule Id="CA3106" Action="Error" />
+    <Rule Id="CA3107" Action="Error" />
+    <Rule Id="CA3108" Action="Error" />
+    <Rule Id="CA3109" Action="Error" />
+    <Rule Id="CA3110" Action="Error" />
+    <Rule Id="CA3111" Action="Error" />
+    <Rule Id="CA3112" Action="Error" />
+    <Rule Id="CA3113" Action="Error" />
+    <Rule Id="CA3114" Action="Error" />
+    <Rule Id="CA3115" Action="Error" />
+    <Rule Id="CA3116" Action="Error" />
+    <Rule Id="CA3117" Action="Error" />
+    <Rule Id="CA3118" Action="Error" />
+    <Rule Id="CA3119" Action="Error" />
+    <Rule Id="CA3120" Action="Error" />
+    <Rule Id="CA3121" Action="Error" />
+    <Rule Id="CA3122" Action="Error" />
+    <Rule Id="CA3123" Action="Error" />
+    <Rule Id="CA3124" Action="Error" />
+    <Rule Id="CA3125" Action="Error" />
+    <Rule Id="CA3127" Action="Error" />
+    <Rule Id="CA3129" Action="Error" />
+    <Rule Id="CA3130" Action="Error" />
+    <Rule Id="CA3131" Action="Error" />
+    <Rule Id="CA3132" Action="Error" />
+    <Rule Id="CA3133" Action="Error" />
+    <Rule Id="CA3134" Action="Error" />
+    <Rule Id="CA3135" Action="Error" />
+    <Rule Id="CA3137" Action="Error" />
+    <Rule Id="CA3138" Action="Error" />
+    <Rule Id="CA3139" Action="Error" />
+    <Rule Id="CA3140" Action="Error" />
+    <Rule Id="CA3141" Action="Error" />
+    <Rule Id="CA3142" Action="Error" />
+    <Rule Id="CA3143" Action="Error" />
+    <Rule Id="CA3145" Action="Error" />
+    <Rule Id="CA3146" Action="Error" />
+    <Rule Id="CA3147" Action="Error" />
+    <Rule Id="CA3050" Action="Error" />
+    <Rule Id="CA3051" Action="None" />
+    <Rule Id="CA3052" Action="None" />
+    <Rule Id="CA3053" Action="Error" />
+    <Rule Id="CA3054" Action="Error" />
+    <Rule Id="CA3055" Action="Error" />
+    <Rule Id="CA3056" Action="Error" />
+    <Rule Id="CA3057" Action="Error" />
+    <Rule Id="CA3058" Action="Error" />
+    <Rule Id="CA3059" Action="Error" />
+    <Rule Id="CA3060" Action="Error" />
+    <Rule Id="CA3061" Action="Error" />
+    <Rule Id="CA3062" Action="Error" />
+    <Rule Id="CA3063" Action="Error" />
+    <Rule Id="CA3064" Action="Error" />
+    <Rule Id="CA3065" Action="Error" />
+    <Rule Id="CA3066" Action="Error" />
+    <Rule Id="CA3067" Action="Error" />
+    <Rule Id="CA3068" Action="Error" />
+    <Rule Id="CA3069" Action="Error" />
+    <Rule Id="CA3070" Action="Error" />
+    <Rule Id="CA3071" Action="Error" />
+    <Rule Id="CA3072" Action="Error" />
+    <Rule Id="CA3073" Action="Error" />
+    <Rule Id="CA3074" Action="Error" />
   </Rules>
 </RuleSet>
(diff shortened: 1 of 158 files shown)


Build/xTVS.ruleset
```diff
@@ -96,5 +96,117 @@
     <Rule Id="CA2226" Action="Warning" />
     <Rule Id="CA2227" Action="Warning" />
     <Rule Id="CA2239" Action="Warning" />
+
+    <Rule Id="CA1304" Action="Error" /> <!-- specify CultureInfo -->
+    <Rule Id="CA1305" Action="None" />  <!-- provide IFormatProvider -->
```
AlexanderSher: Can you also switch `CA1063` to `None`?

Python/Product/Analysis/Infrastructure/StringExtensions.cs
```diff
@@ -16,15 +16,68 @@
...
+            @"\{(\d+)",
+            RegexOptions.IgnorePatternWhitespace,
+            TimeSpan.FromSeconds(1)
+        );
+
+        private static void ValidateFormatString(string str, int argCount) {
+            foreach (Match m in SubstitutionRegex.Matches(str)) {
+                int index = int.Parse(m.Groups[1].Value, CultureInfo.InvariantCulture);
+                if (index >= argCount) {
+                    Debug.Fail(string.Format(CultureInfo.InvariantCulture, "Format string expects more than {0} args.\n\n{1}", argCount, str));
```
AlexanderSher: `string.Format` already throws `FormatException` when string expects more arguments than provided. Why do you need to check manually?

Python/Product/Analysis/Infrastructure/StringExtensions.cs
```diff
@@ -16,15 +16,68 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
 using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.PythonTools.Analysis.Infrastructure {
     static class StringExtensions {
```
AlexanderSher: Why not make this class `public`? It can be useful everywhere in PTVS.

Python/Product/Analysis/Analyzer/DDG.cs
```diff
@@ -282,7 +283,7 @@ internal class DDG : PythonWalker {
                         userMod.Imported(_unit);
 
                         foreach (var varName in userMod.GetModuleMemberNames(GlobalScope.InterpreterContext)) {
-                            if (!varName.StartsWith("_")) {
+                            if (!varName.StartsWith("_", StringComparison.Ordinal)) {
```
AlexanderSher: Since we have `StringExtensions`, you can extract `StartsWithOrdinal` method

AlexanderSher:
One of the possible solution for culture in string interpolation is to use static methods with short name:
```cs
public static class CultureFormattableString {
    public static string I(FormattableString fs) => fs.ToString(CultureInfo.InvariantCulture);
    public static string C(FormattableString fs) => fs.ToString(CultureInfo.CurrentCulture);
}
```

Usage:
```cs
using static Microsoft.PythonTools.Infrastructure.CultureFormattableString;
```
```cs
var invariantCultureString = I($"This string with value {1+1} is formatted with invariant culture");
var currentCultureString = C($"This string with value {2+2} is formatted with current thread culture");
```
It will be also easier to update if [this](https://github.com/dotnet/csharplang/issues/177) approach is accepted.

Python/Product/Analysis/AnalysisValueSetExtensions.cs
```diff
@@ -270,7 +271,7 @@ class DotsLastStringComparer : IComparer<string> {
                     return -1;
                 }
 
-                return x.CompareTo(y);
+                return string.Compare(x, y, StringComparison.CurrentCultureIgnoreCase);
```
AlexanderSher: We can extract `CompareCurrentIgnoreCase` extension method

Python/Product/Analysis/AnalysisSet.cs
```diff
@@ -600,8 +601,8 @@ sealed class UnionComparer : IEqualityComparer<AnalysisValue>, IEqualityComparer
 #if FULL_VALIDATION
             var z2 = y.UnionMergeTypes(x, Strength);
             if (!object.ReferenceEquals(z, z2)) {
-                Validation.Assert(z.UnionEquals(z2, Strength), string.Format("{0}\n{1} + {2} => {3}\n{2} + {1} => {4}", Strength, x, y, z, z2));
-                Validation.Assert(z2.UnionEquals(z, Strength), string.Format("{0}\n{1} + {2} => {3}\n{2} + {1} => {4}", Strength, y, x, z2, z));
+                Validation.Assert(z.UnionEquals(z2, Strength), $"{Strength}\n{x} + {y} => {z}\n{y} + {x} => {z2}");
```
AlexanderSher: If you replace `Validation.Assert` with:
```cs
        public static void Assert(bool expression, FormattableString message) {
            if (!expression) {
                try {
                    throw new ValidationException(message.ToString(CultureInfo.InvariantCulture));
                } catch (ValidationException ex) {
                    Console.Error.WriteLine(ex.ToString());
                }
            }
        }
```
you can save a lot of memory, cause `FormattableString` won't try to build a new string unless asserted expression is invalid.

Build/xTVS.ruleset
zooba: Sure. It's not in the required list for releasing.

Python/Product/Analysis/Infrastructure/StringExtensions.cs
zooba: The Debug.Fail will show up more prominently while we're working (FormatException will likely be swallowed and logged somewhere, which adds a few more steps to find it, and this also only reports mismatched argument count rather than other formatting errors).

It's also leftover from when we used to check for unused arguments as well. I don't remember exactly why we took that out, but I do remember it caught a lot of bugs. @huguesv - any memory of this?

Python/Product/Analysis/Infrastructure/StringExtensions.cs
zooba: It's InternalsVisibleTo. `public` API is public/stable for end users, especially in the `Analysis` project which we know will have multiple users.

Python/Product/Analysis/Analyzer/DDG.cs
zooba: Oh nice idea. This one comes up all the time.

Python/Product/Analysis/AnalysisValueSetExtensions.cs
zooba: I don't think we use that one often enough to be worth extracting.

Python/Product/Analysis/AnalysisSet.cs
zooba: This code is disabled almost all the time - you only turn it on when looking for a particular kind of bug.

Python/Product/Analysis/Infrastructure/StringExtensions.cs
AlexanderSher: Ok, got it

Python/Tests/Analysis/AnalysisTest.Perf.cs
```diff
@@ -325,7 +325,7 @@ internal sealed class FileStreamContentProvider : StreamContentProvider {
 
         private static void CollectFiles(string dir, List<string> files, ISet<string> excludeDirectories = null) {
             foreach (string file in Directory.GetFiles(dir)) {
-                if (file.EndsWith(".py", StringComparison.OrdinalIgnoreCase)) {
+                if (file.EndsWithOrdinal(".py", ignoreCase: true)) {
```
AlexanderSher: `EndsWithOrdinal`  with parameter or `EndsWithOrdinalIgnoreCase`?

Python/Product/VSInterpreters/RegistryWatcher.cs
```diff
@@ -485,24 +486,24 @@ enum RegNotifyChange : uint {
...
         private static RegistryHive ParseRegistryKey(string key, out string subkey) {
             int firstPart = key.IndexOf('\\');
-            if (firstPart < 0 || !key.StartsWith("HKEY", StringComparison.InvariantCultureIgnoreCase)) {
+            if (firstPart < 0 || !key.StartsWithOrdinal("HKEY", ignoreCase: true)) {
                 throw new ArgumentException("Invalid registry key: " + key, "key");
             }
             var hive = key.Remove(firstPart);
             subkey = key.Substring(firstPart + 1);
-            if (hive.Equals("HKEY_CURRENT_USER", StringComparison.InvariantCultureIgnoreCase)) {
+            if (hive.Equals("HKEY_CURRENT_USER", StringComparison.OrdinalIgnoreCase)) {
```
AlexanderSher: `EqualsOrdinalIgnoreCase`

Python/Product/PythonTools/PythonTools/Project/CustomCommand.cs
```diff
@@ -762,61 +762,61 @@ class CommandStartInfo {
...
             get {
-                return PythonCommandTask.ExecuteInConsolePause.Equals(ExecuteIn, StringComparison.InvariantCultureIgnoreCase);
+                return PythonCommandTask.ExecuteInConsolePause.Equals(ExecuteIn, StringComparison.OrdinalIgnoreCase);
             }
         }
 
         public bool ExecuteHidden {
             get {
-                return PythonCommandTask.ExecuteInNone.Equals(ExecuteIn, StringComparison.InvariantCultureIgnoreCase);
+                return PythonCommandTask.ExecuteInNone.Equals(ExecuteIn, StringComparison.OrdinalIgnoreCase);
```
AlexanderSher: `EqualsOrdinalIgnoreCase`

Python/Product/Common/Infrastructure/PathUtils.cs
```diff
@@ -643,14 +643,14 @@ static class PathUtils {
             bool fullPaths = true
         ) {
             var queue = new Queue<string>();
-            if (!root.EndsWith("\\")) {
+            if (!root.EndsWith("\\", StringComparison.Ordinal)) {
```
AlexanderSher: `EndsWithOrdinal`

Common/Product/SharedProject/Automation/VSProject/OAReferences.cs
```diff
@@ -88,7 +88,7 @@ private Reference FindByName(string stringIndex)
         public Reference Add(string bstrPath)
         {
             // ignore requests from the designer which are framework assemblies and start w/ a *.
-            if (string.IsNullOrEmpty(bstrPath) || bstrPath.StartsWith("*"))
+            if (string.IsNullOrEmpty(bstrPath) || bstrPath.StartsWith("*", StringComparison.Ordinal))
```
AlexanderSher: `StartsWithOrdinal`

Common/Product/SharedProject/Misc/NativeMethods.cs
```diff
@@ -904,7 +904,7 @@ public struct _NETRESOURCE {
 
                 // UNC Paths will start with \\?\.  Remove this if present as this isn't really expected on a path.
                 var pathString = path.ToString();
-                return pathString.StartsWith(@"\\?\") ? pathString.Substring(4) : pathString;
+                return pathString.StartsWith(@"\\?\", StringComparison.Ordinal) ? pathString.Substring(4) : pathString;
```
AlexanderSher: `StartsWithOrdinal`

Common/Product/SharedProject/ProjectResources.cs
```diff
@@ -213,21 +213,21 @@ internal class SR {
...
-                Debug.Assert(result.IndexOf("{0}") < 0, "Resource string '" + value + "' requires format arguments.");
+                Debug.Assert(result.IndexOf("{0}", StringComparison.Ordinal) < 0, "Resource string '" + value + "' requires format arguments.");
                 return result;
             }
 
             Debug.WriteLineIf(
-                Enumerable.Range(0, args.Length).Any(i => result.IndexOf(string.Format("{{{0}", i)) < 0),
-                string.Format("Resource string '{0}' does not use all {1} arguments", value, args.Length)
+                Enumerable.Range(0, args.Length).Any(i => result.IndexOf(string.Format(CultureInfo.InvariantCulture, "{{{0}", i), StringComparison.Ordinal) < 0),
+                string.Format(CultureInfo.InvariantCulture, "Resource string '{0}' does not use all {1} arguments", value, args.Length)
```
AlexanderSher: `FormatInvariant`

Common/Product/SharedProject/ProjectConfig.cs
```diff
@@ -473,7 +473,7 @@ internal abstract class ProjectConfig :
         /// <returns>S_OK if the method succeeds, otherwise an error code</returns>
         public virtual int QueryDebugLaunch(uint flags, out int fCanLaunch) {
             string assembly = this.project.GetAssemblyName(this.ConfigName);
-            fCanLaunch = (assembly != null && assembly.ToUpperInvariant().EndsWith(".exe", StringComparison.OrdinalIgnoreCase)) ? 1 : 0;
+            fCanLaunch = (assembly != null && assembly.EndsWith(".exe", StringComparison.OrdinalIgnoreCase)) ? 1 : 0;
```
AlexanderSher: >EndsWith [](start = 55, length = 8)

`EndsWithOrdinal`

Common/Product/SharedProject/CommonFileNode.cs
```diff
@@ -53,7 +52,7 @@ public CommonFileNode(CommonProjectNode root, ProjectElement e)
         public bool IsFormSubType {
             get {
                 string result = this.ItemNode.GetMetadata(ProjectFileConstants.SubType);
-                if (!String.IsNullOrEmpty(result) && string.Compare(result, ProjectFileAttributeValue.Form, true, CultureInfo.InvariantCulture) == 0)
+                if (!String.IsNullOrEmpty(result) && string.Compare(result, ProjectFileAttributeValue.Form, StringComparison.OrdinalIgnoreCase) == 0)
```
AlexanderSher: `string.CompareOrdinal`

Python/Product/Analysis/Infrastructure/StringExtensions.cs
huguesv: sorry, no memory

Python/Tests/Analysis/AnalysisTest.Perf.cs
zooba: The argument matches other string functions that don't take a culture info object.

Python/Product/VSInterpreters/RegistryWatcher.cs
zooba: Don't see the benefit here, there aren't so many places where we do this.

Common/Product/SharedProject/Automation/VSProject/OAReferences.cs
zooba: Files under `Common` directory don't have access to things we declare in `Microsoft.PythonTools` namespace.

zooba:
I believe that's all the comments addressed. I don't see enough uses to justify `CompareOrdinal` or `EqualsOrdinalIgnoreCase` extension methods, and AFAIK the `SharedProject` folder is still used independently of the `Microsoft.PythonTools` namespace and can't use anything we declare there.

Issue #3662, opened 2018-01-29
Format Document corrupts specific code


AlexanderSher:
1. Create new file
2. Paste the following code:
```python
import sys
import abc

#ABC

foo()
foo()

x = 34;

y = x;
goo(y)
x = 34;
```
3. Edit -> Advanced -> Format Document

**Result**:
```python
import sys
import abc

#ABC
foo()
foo()

x = 34;
```

AlexanderSher:
May be related to #3566

PR #3678, opened 2018-01-31
Fix #3566: Improve LinePreservingCodeReplacer.ReplaceCode

### LINKED ISSUE #3566: Improve LinePreservingCodeReplacer.ReplaceCode
Replacer produces lots of garbage strings and doesn't correctly handle some of the cases, for instance:
*User adds a new line a subsequence of n+1 lines A..NA: ZA..NA. Loop will start with curNewLine == 1, GetNewText(curNewLine) will return line A, lines will contain 0 and n, and as a result A..NA will be replaced with ZA. The remaining comparison will most likely end up with full replacement.*

Consider implementing dynamic programming LCS algorithm with trimmed start/end.

AlexanderSher:
This also fixes https://github.com/Microsoft/PTVS/issues/3662 (at least it correctly works with repro sample), but I'm not sure if we should merge this into 15.6 or into master.

### DIFF
--- a/Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
+++ b/Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
@@ -15,9 +15,10 @@
 // permissions and limitations under the License.
 
 using System;
+using Microsoft.PythonTools.Analysis;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.PythonTools.Analysis;
+using System.Text;
 using Microsoft.PythonTools.Parsing;
 
 namespace Microsoft.PythonTools.Intellisense {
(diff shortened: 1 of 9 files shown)


huguesv:
Looks like this is only used by FormatCode, so worse case is we regress some more in that area, best case some broken formatting now works as expected and doesn't delete user code.

You should run CodeFormatterTests.* and add new ones for the code snippets in #3662. Looking at the tests, it's hard for me to feel confident even if they pass. Have you run this on some large files?

AlexanderSher:
Thanks for the hint on tests, I've found one issue (result of copy-paste). 
I've tried it on `test_descr.py`. I haven't noticed issues in the algorithm itself, though I've noticed that we change code in comments. Not sure if it is the right thing to do.

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
```diff
@@ -26,124 +27,89 @@ namespace Microsoft.PythonTools.Intellisense {
...
+        }
+
+        private DocumentChange Insert(int oldStart, int newStart, int newEnd) {
+            var startIndex = _newLines[newStart].Info.Start;
+            var length = _newLines[newEnd].Info.EndIncludingLineBreak - startIndex;
+            var newSource = _newLines[newStart].Source;
+            var insertText = new StringBuilder(newSource, startIndex, length, length)
+                .Replace(NewLineKind.LineFeed.GetString(), _newLine)
+                .Replace(NewLineKind.CarriageReturn.GetString(), _newLine)
+                .Replace(NewLineKind.CarriageReturnLineFeed.GetString(), _newLine)
```
zooba: Not sure that this is going to do what you probably hope when `_newLine` == `\r\n`.

```
'\r\n'.Replace('\n', '\r\n')
'\r\r\n'.Replace('\r', '\r\n')
'\r\n\r\n\n'.Replace('\r\n', '\r\n')
`\r\n\r\n\n'
```

Probably better to iterate over `newSource` with FindNewLine and extract and recombine each line individually.

Python/Product/Analysis/Intellisense/LongestCommonSequence.cs
```diff
@@ -0,0 +1,258 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.PythonTools.Intellisense {
+    internal class LongestCommonSequence<T> {
```
zooba: Do we gain anything by not making this exclusively for `string`?

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
```diff
@@ -26,124 +27,89 @@ namespace Microsoft.PythonTools.Intellisense {
...
+                .ToString();
+
+            var insertLocation = _oldLines[oldStart].Info.SourceEndIncludingLineBreak;
```
zooba: Don't know if it matters (probably not), but this location could be different from `_oldLines[oldStart + 1].Info.SourceStart` if any edits race. I don't think they can, but semantically you may prefer one over the other.

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
AlexanderSher: Nice catch

Python/Product/Analysis/Intellisense/LongestCommonSequence.cs
AlexanderSher: 1. It is not actually `string`. It works with `LineInfo`. Splitting input code into strings is a waster of memory
2. It is much easier to test algorithm on chars.

Python/Product/Analysis/Intellisense/LinePreservingCodeReplacer.cs
AlexanderSher: It is actually should be `_oldLines[oldStart].Info.SourceStart` for all lines except the last one, which should be `_oldLines[oldStart-1].Info.SourceEndIncludingLineBreak`. Is there any formatting rule that can insert the line?

zooba:
I think there's a rule to split semicolon separated statements, and one to wrap long lines. But I think they’re fine if the formatted code has new lines, it’s if there’s a race (for example, if inserting a snippet triggered format).

huguesv:
Merging with Alex's permission.

PR #3681, opened 2018-01-31
Fixes #3660 Imports are not resolved for reopened project

### LINKED ISSUE #3660: Imports are not resolved for reopened project
1. Create new bottle project, create virtual environment.
1. Open `app.py`
1. Close VS
1. Reopen VS, open previously created project

**Actual**: imports aren't highlighted
image

If you close and reopen file, everything works.

zooba:
Fixes #3660 Imports are not resolved for reopened project
Avoids some potential deadlocks on import due to missing ConfigureAwait(false)
Handles case where a buffer detects an expired analyzer but does not replace it
Adds (compile-time optional) trace logging to PythonTextBufferInfo
Ensures stderr collection from analyzer does not fill up memory


huguesv:
I was testing this and at one point all IntelliSense stopped working. I sent you an email with project, logs, screenshot.

huguesv:
I've opened a bottle project, and made no edits at all. At first, colorization of modules was fine, but as I opened, scrolled, closed and reopened files, I ended up with no colorization of modules.

It's almost as if opening a new file (routes.py) affects colorization of already opened file (app.py).

I easily gets stuff like this:
image

zooba:
It looks like something may be happening to put classifications off by one line, and now I wonder how many of the "no classifications" are actually "I can't see the classifications because the next line is black".

I can't reproduce this trivially, but perhaps I can track it down by review.

zooba:
Since I can't repro it hasn't triggered for me, but I just added some extra validation to detect when we've calculated line info that doesn't match the actual snapshot. Outside of that, I don't see anything that would put us out by a line.

Sending me connection logs when the highlighting gets out of place would be great - that will show whether it's on the analyzer side or the IDE side.

huguesv:
Just updated to the latest + this PR, I double checked that I do have your commits.
Opening my bottle project and opening, closing files, took less than 2 mins to repro.
I'll email the connection logs.
image

zooba:
I think this will fix it. The logs are clean, so it's likely confusion in getting the response from the analyzer. My guess is that `ITextVersion` may compare equal even if they belong to different buffers, so by making sure we only have one active classifications request per entry instead, we shouldn't run into trouble when they race.

huguesv:
I'm trying it.

Issue #3611, opened 2018-01-24
Mixed-mode debugging doesn't work with 64-bit Python


huguesv:
I tried with 15.6 preview 2, and the latest 15.6 build, and I'm getting the same results.

Create new python application
Project settings/debug/enable native code debugging
Set interpreter to Python 3.6
Write some code and put breakpoints
Start debugging

Result: VS goes to debugging mode and app runs, but doesn't stop at breakpoints. Breakpoint icon says it won't be hit because symbols haven't been loaded.

I also tried with Python 2.7.13 64-bit, where I downloaded the pdbs and added the folder to debugger symbols settings (before doing this I got the task dialog telling me to download them).

Am I missing a step?

zooba:
See if the debugger helper DLLs are loaded and have symbols, and if so, try to set a breakpoint in one of them (grab the source code from a build).

Maybe someone has changed the default build settings again and we're getting non-debuggable DLLs.

huguesv:
Symbols for debugger.helper loaded, same for python.exe. I've set breakpoints in every function in trace.cpp, they appear red with no warning, but don't get hit.

huguesv:
I checked that on master branch, debug build of PTVS.

int19h:
What happens if you just do Break All? In particular, what does it show in the call stack and in the modules list?

huguesv:
![image](https://user-images.githubusercontent.com/1696845/35357628-f98ded1a-0108-11e8-9b62-db9190319275.png)

int19h:
So the Python debugger isn't running at all (no .py files in Modules, and python27.dll frames in stack trace). 

Hard to say what's going wrong exactly, but the breakage is certainly much broader than just breakpoints. I'll take a look.

int19h:
Seems to be specific to 64-bit Python - 32-bit works fine.

int19h:
The root cause is that the breakpoint that the debugger sets inside `OnInitialized` (to detect when helper DLL is ready to go) is not hit in 64-bit builds.

The breakpoint is set and enabled properly in this code in `LocalComponent`:
```cs
                    bool isInitialized = moduleInstance.GetExportedStaticVariable<ByteProxy>("isInitialized").Read() != 0;
                    if (isInitialized) {
                        OnHelperDllInitialized(moduleInstance);
                    } else {
                        DkmRuntimeBreakpoint initBP = null;
[...]

PR #3693, opened 2018-02-02
Minfix #3611: Mixed-mode debugging doesn't stop at breakpoints

int19h:
Minfix for 15.6 only - disable VSCode debug adapter to avoid missing assembly reference issues.


int19h:
Note that this is targeting the 15.6rel branch - it will not be cherry-picked to master.

Python/Product/Debugger/Debugger.csproj
```diff
@@ -46,7 +46,6 @@
     <Reference Include="envdte80, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
     <Reference Include="Microsoft.VisualStudio.ComponentModelHost, Version=$(VSTarget).0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="Microsoft.VisualStudio.Debugger.Engine, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
-    <Reference Include="Microsoft.VisualStudio.Debugger.DebugAdapterHost.Interfaces, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
```
huguesv: Searching for `Microsoft.VisualStudio.Debugger.DebugAdapterHost.Interfaces` yields results in Core.csproj and PythonTools.csproj

I guess that will pull the dll in the .vsix, is it better to remove those references too?

PR #3694, opened 2018-02-02
LEGO: Pull request from lego/hb_839_20180202094034278 to 15.6rel

csigs:
LEGO: Pull request from lego/hb_839_20180202094034278 to 15.6rel with localized lcls


zooba:
@huguesv In case you're unsure, we ought to reinsert with this change (not necessarily for preview though), so plan on one more.

PR #3709, opened 2018-02-06
Cookiecutter fixes

### LINKED ISSUE #3661: Cookiecutter progress bar should not move the button when it shows up

### LINKED ISSUE #3707: Cookiecutter should respect create solution folder option in new project dialog
When using New project dialog with ML templates,, user can check create solution folder option but we ignore it.

Looking at the code it's trying to do something special in that scenario, we just need to not do that.

huguesv:
Fix #3707 Cookiecutter should respect create solution folder option in new project dialog
Fix #3661 Cookiecutter progress bar should not move the button when it shows up

image

(similar fix for the second page also included)

### DIFF
--- a/Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
+++ b/Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
@@ -78,11 +78,11 @@ namespace Microsoft.CookiecutterTools.ViewModel {
 
         private string _statusMessage, _statusHelp;
         private ImageMoniker _statusImage;
-        private Visibility _statusVisibility = Visibility.Collapsed;
+        private Visibility _statusVisibility = Visibility.Hidden;
         private Visibility _statusImageVisibility = Visibility.Collapsed;
         private Visibility _spinningStatusImageVisibility = Visibility.Collapsed;
-        private Visibility _statusProgressVisibility = Visibility.Collapsed;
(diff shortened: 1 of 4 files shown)


Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -273,9 +273,9 @@ class CookiecutterViewModel : INotifyPropertyChanged {
             _statusImageVisibility = image.HasValue && !progressVisible ? Visibility.Visible : Visibility.Collapsed;
             _spinningStatusImageVisibility = image.HasValue && progressVisible ? Visibility.Visible : Visibility.Collapsed;
             bool v = (visible ?? !string.IsNullOrEmpty(message));
-            _statusVisibility = v ? Visibility.Visible : Visibility.Collapsed;
-            _statusProgressVisibility = progressVisible ? Visibility.Visible : Visibility.Collapsed;
-            _statusUpdateProgressVisibility = updateProgressVisible ? Visibility.Visible : Visibility.Collapsed;
+            _statusVisibility = v ? Visibility.Visible : Visibility.Hidden;
```
AlexanderSher: Those are fields. WPF won't be notified about updated values.

huguesv: The PropertyChanged notifications are just a few lines below.

PR #3710, opened 2018-02-06
Conda environment detection fix

### LINKED ISSUE #3659: Create conda environment doesn't appear in list after creation for the first ever created env
Install Anaconda
In Environments window, create new conda environment

Result: After it is created successfully, it doesn't appear in the environment list

We watch for changes to the `environments.txt` file only if that file exists, which it doesn't for a new user until they create their very first environment. Subsequent creations/deletions are fine. We should set up the file system watcher even when the file doesn't exist. Note: its parent folder won't exist either.

```
PS C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64> conda -V
conda 4.4.8
```

huguesv:
Fix #3659 Create conda environment doesn't appear in list after creation for the first ever created env

If the `.conda` folder doesn't exist, we can't setup a watcher on that folder, so we create one for the parent folder (which is the user home folder) and when we get notification that `.conda` is created, we then start watching that folder.

### DIFF
--- a/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
+++ b/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
@@ -48,7 +48,9 @@ namespace Microsoft.PythonTools.Interpreter {
         private readonly CPythonInterpreterFactoryProvider _globalProvider;
         private readonly bool _watchFileSystem;
         private FileSystemWatcher _envsTxtWatcher;
-        private Timer _envsTxtWatcherTimer;
+        private FileSystemWatcher _condaFolderWatcher;
+        private Timer _envsWatcherTimer;
+        private string _environmentsTxtFolder;
         private string _environmentsTxtPath;
 
         [ImportingConstructor]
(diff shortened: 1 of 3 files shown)


Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -96,28 +101,22 @@ class CondaEnvironmentFactoryProvider : IPythonInterpreterFactoryProvider, IDisp
                     _initialized = true;
                     doDiscover = true;
                     try {
-                        _environmentsTxtPath = Path.Combine(
+                        _environmentsTxtFolder = Path.Combine(
                             Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
```
zooba: Should we make this a parameter and add a test for this scenario? (So we don't have to mess with `UserProfile` on the test machines)

huguesv: I don't want it to mess with the folder on my machine either.

I'd be surprised if we can tell conda to use a different path though, I'll have to check.

I could easily enough make the factory provider use an alternate path, have the test create the environments.txt file in that path, and then check that this triggered a rediscovery. There won't be a InterpretersChanged event triggered because `conda list` will return the same results as before. But I could add an internal DiscoveryStarted event and have the test expect that instead.

huguesv:
Added tests.

PR #3714, opened 2018-02-08
Fixes #3706 Type hint isn't properly recognized for complex types

### LINKED ISSUE #3706: Type hint isn't properly recognized for complex types
1. Paste the following code:
```python
from typing import Tuple, Dict

def f(x: Tuple[Tuple[str, int], Dict[str, str]]):
    pass
```
2. Type `x[0].` on the line before `pass`

**Actual**
Intellisense for `Dict`
image

**Expected**
Intellisense for `Tuple`

zooba:
Fixes #3706 Type hint isn't properly recognized for complex types
Fixes handling of tuple type annotation
Improves handling of protocol object descriptions
Improves unary operation on numeric literals.

### DIFF
--- a/Python/Product/Analysis/Values/ProtocolInfo.cs
+++ b/Python/Product/Analysis/Values/ProtocolInfo.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
@@ -40,7 +41,7 @@ namespace Microsoft.PythonTools.Analysis.Values {
             DeclaringModule = declaringModule;
             DeclaringVersion = declaringModule?.AnalysisVersion ?? -1;
             _references = new ReferenceDict();
-            State = state;
+            State = state ?? throw new ArgumentNullException(nameof(state));
         }
 
         internal PythonAnalyzer State { get; }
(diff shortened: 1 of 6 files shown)


Python/Product/Analysis/Values/ProtocolInfo.cs
```diff
@@ -284,33 +285,42 @@ class ProtocolInfo : AnalysisValue, IHasRichDescription {
...
-                    }
-                    foreach (var d in p.GetRichDescription()) {
-                        yield return d;
+
+            var res = new List<KeyValuePair<string, string>>();
+            var namespaces = _protocols.OfType<NamespaceProtocol>().ToArray();
+            var other = _protocols.Except(names).Except(namespaces).ToArray();
+
+            var fallbackName = other.Select(p => p.Name).FirstOrDefault(n => !string.IsNullOrEmpty(n)) ?? "<unknown>";
+            if (!string.IsNullOrEmpty(fallbackName)) {
```
AlexanderSher: It looks like `fallbackName` can't be null or empty

zooba: Highly unlikely, true, but this will cover any future changes that may leave it empty.

Python/Product/Analysis/Values/TypingTypeInfo.cs
```diff
@@ -435,13 +446,31 @@ sealed class TypingTypeInfoFinalizer {
...
+                    if (description.Any()) {
+                        description.Add(new KeyValuePair<string, string>(WellKnownRichDescriptionKinds.Comma, ", "));
+                    }
+                    description.AddRange(p.GetRichDescription());
                 }
+
+                res.AddProtocol(new TupleProtocol(res, tupleItem));
+            }
+
+            name = namedTupleName?.GetConstantValueAsString().FirstOrDefault() ?? "tuple";
```
huguesv: use `namedTupleName?.GetConstantValueAsString()?.FirstOrDefault()` because `null.FirstOrDefault()` causes an ANE.

zooba: `GetConstantValueAsString()` will never return null, and if `namedTupleName` is null then the rest of that expression is skipped and it goes straight to the `??`.

huguesv: Ah, there's 2 GetConstantValueAsString and I looked at the wrong one...

PR #3721, opened 2018-02-09
Port TaskObserver into PTVS

AlexanderSher:
This approach helps tests catch exceptions in tasks that aren't awaited or run too for long.

For example, imagine the following code:
```cs
public void Run() {
    RunThenFailAsync().DoNotWait();
    RunForeverAsync().DoNotWait();
}

private async Task RunThenFailAsync() {
    await Task.Delay(1000);
    throw new InvalidOperationException();
}

private async Task RunForeverAsync() {
    while (true) {
        await Task.Delay(500);
    }
}
```

If we create a simple test for it, it will successfully pass:
```cs
[TestMethod]
public void SomeTypeTest() {
    new SomeType().Run();
}
```
image

But, with `TaskObserver` enabled:
```cs
[TestInitialize]
public void Initialize() => MSTestEnvironment.TestInitialize(5);

[TestCleanup]
public void Cleanup() => MSTestEnvironment.TestCleanup();

[TestMethod]
public void SomeTypeTest() {
    new SomeType().Run();
}
```
we will get both tasks in the output:
```
Test Failed - SomeTypeTest
Message: One task that has been started during test run is still not completed:

   at void SomeType.Run() in C:\Temp\SomeType.cs:line 29
   at void Tests.SomeTypeTest() in C:\Temp\Tests.cs:line 44
   at object RuntimeMethodHandle.InvokeMethod(object target, object[] arguments, Signature sig, bool constructor)
   at object System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(object obj, object[] parameters, object[] arguments)
   at object System.Reflection.RuntimeMethodInfo.Invoke(object obj, BindingFlags invokeAttr, Binder binder, object[] parameters, CultureInfo culture)

Also, one not awaited tasks has failed:

InvalidOperationException: Operation is not valid due to the current state of the object.
   at async Task SomeType.RunThenFailAsync() in C:\Temp\SomeType.cs:line 34
   --- Task stack trace: ---
   at void SomeType.Run() in C:\Temp\SomeType.cs:line 28
   at void Tests.SomeTypeTest() in C:\Temp\Tests.cs:line 44
   at object RuntimeMethodHandle.InvokeMethod(object target, object[] arguments, Signature sig, bool constructor)
   at object System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(object obj, object[] parameters, object[] arguments)
   at object System.Reflection.RuntimeMethodInfo.Invoke(object obj, BindingFlags invokeAttr, Binder binder, object[] parameters, CultureInfo culture)
```

### DIFF
--- a/Common/Tests/Utilities/MSTestEnvironment.cs
+++ b/Common/Tests/Utilities/MSTestEnvironment.cs
@@ -1,0 +1,63 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Common = Microsoft.PythonTools.Infrastructure;
+using Analysis = Microsoft.PythonTools.Analysis.Infrastructure;
+
+namespace TestUtilities {
+    public sealed class MSTestEnvironment : Common.ITestEnvironment, Analysis.ITestEnvironment {
+        private static MSTestEnvironment _instance;
+
+        public static void Initialize() {
+            _instance = new MSTestEnvironment();
+            Analysis.TestEnvironment.Current = _instance;
+            Common.TestEnvironment.Current = _instance;
+        }
+
+        public static void TestInitialize(int secondsTimeout = 30) => _instance.BeforeTestRun(secondsTimeout);
+        public static void TestCleanup() => _instance.AfterTestRun();
+
+        private readonly AsyncLocal<TaskObserver> _taskObserver = new AsyncLocal<TaskObserver>();
+
+        public bool TryAddTaskToWait(Task task) {
+            var taskObserver = _taskObserver.Value;
+            if (taskObserver == null) {
+                return false;
+            }
+            taskObserver.Add(task);
+            return true;
+        }
+        
+        private void BeforeTestRun(int secondsTimeout) {
+            if (_taskObserver.Value != null) {
+                throw new InvalidOperationException("AsyncLocal<TaskObserver> reentrancy");
+            }
+
+            _taskObserver.Value = new TaskObserver(secondsTimeout);
+        }
+
+        private void AfterTestRun() {
+            try {
+                _taskObserver.Value?.WaitForObservedTask();
+            } finally {
+                _taskObserver.Value = null;
+            }
+        }
+    }
+}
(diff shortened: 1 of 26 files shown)


Common/Tests/Utilities/MSTestEnvironment.cs
```diff
@@ -0,0 +1,63 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Common = Microsoft.PythonTools.Infrastructure;
+using Analysis = Microsoft.PythonTools.Analysis.Infrastructure;
```
zooba: This should go in `TestUtilities.Python.Analysis` in order to take dependencies on PythonTools assemblies. (Or we should move the entire TestUtilities project out of the Common directory - I'm 99% sure nobody else is using this project now, but at one point a couple of external companies were using it.)

Python/Tests/Core/AssemblySetup.cs
```diff
@@ -0,0 +1,26 @@
...
+// permissions and limitations under the License.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TestUtilities;
+
+namespace PythonToolsTests {
+    [TestClass]
+    public sealed class AssemblySetup {
+        [AssemblyInitialize]
+        public static void Initialize(TestContext testContext) => MSTestEnvironment.Initialize();
```
zooba: We'll also need something similar in the test runner interop module for the UI tests.

Common/Tests/Utilities/MSTestEnvironment.cs
AlexanderSher: It is dependent only by projects in `Test` folder:

image

Do you want me to move it?

Python/Tests/Core/AssemblySetup.cs
AlexanderSher: I don't get it. `TestUtilities.UI` references `TestUtilities` directly, can't we just call `MSTestEnvironment.Initialize()` in the `HostedPythonToolsTestRunner` constructor?

Common/Tests/Utilities/MSTestEnvironment.cs
zooba: Within our repo, this is true, but there are other projects besides PTVS that use this. If we move it, they have to stop using it, whereas if we just move the new additions into the Python-specific project then it doesn't affect the Common one.

Python/Tests/Core/AssemblySetup.cs
zooba: Yes. I don't see that change in this PR.

zooba: But it needs to be called on the VS side, not the MSTest side.

Common/Tests/Utilities/MSTestEnvironment.cs
AlexanderSher: Ok. So `TestUtilities.Python.Analysis` or `TestUtilities.Python`?

Python/Tests/Core/AssemblySetup.cs
AlexanderSher: Ok

Common/Tests/Utilities/MSTestEnvironment.cs
zooba: TestUtilities.Python.Analysis (which is the "non-UI" test project... the naming is not great :) )

AlexanderSher:
Ping

Python/Tests/Utilities.Python.Analysis/MSTestEnvironment.cs
```diff
@@ -0,0 +1,29 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using Common = Microsoft.PythonTools.Infrastructure;
+using Analysis = Microsoft.PythonTools.Analysis.Infrastructure;
+
+namespace TestUtilities.Python {
+    public sealed class MSTestEnvironment : TestEnvironment, Common.ITestEnvironment, Analysis.ITestEnvironment {
+        public static void Initialize() {
+            var isntance = new MSTestEnvironment();
```
zooba: *instance

PR #3725, opened 2018-02-09
Add UI test for mixed-mode debugger.

### LINKED ISSUE #3713: Add a few UI smoke tests for mixed-mode debugger
We should have some non-glass tests that run in VS UI.
Maybe one breakpoint test that is run on 4 configs (2.7, 3.6, 32-bit, 64-bit).

huguesv:
Fix #3713 

This is sufficient to catch the regression with 64-bit mixed-mode debugging.

We can add more later, for example a test that involves C++ code ;)

### DIFF
--- a/Python/Tests/DebuggerUITests/MixedModeDebugProjectUITests.cs
+++ b/Python/Tests/DebuggerUITests/MixedModeDebugProjectUITests.cs
@@ -1,0 +1,79 @@
+// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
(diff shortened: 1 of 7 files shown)


Python/Tests/DebuggerUITests/MixedModeDebugProjectUITests.cs
```diff
@@ -0,0 +1,79 @@
...
+            // For simplicity, require the test machine to have the .pdb files
+            // next to their binaries.
+            return File.Exists(Path.ChangeExtension(interpreter.Configuration.InterpreterPath, "pdb"));
+        }
+
+        private static PythonVersion FindInterpreter(string pythonVersion) {
+            // Examples of values for pythonVersion (match field names on PythonPaths):
+            // "Python36"
+            // "Python36|Python36_x64"
+            var interpreter = pythonVersion.Split('|')
```
int19h: If we always split these into elements, perhaps just make it an array of strings instead of a string? Or is it used somewhere else as is?

zooba: I think it's really just a convenience for us - these values are provided in the UITestRunner side as a string literal and converted automatically into PythonVersion objects on the test side.

huguesv: Our UI test runner can only pass on simple data type, it can't handle `string[]` parameters.

PR #3738, opened 2018-02-13
Fix display of UI and IntelliSense in Visual Studio 2017

jmartens:
It seems that in Visual Studio 2017 I need to explicitly add PresentationFramework 4.0.0.0 as a reference to Python projects to display the UI in the editor window and get IntelliSense working in XAML.

#2943


zooba:
I think the `WindowBase` reference should be updated too, but then we'll take this 👍

huguesv:
Note that without the HintPath, it doesn't work.

zooba:
I don't see any difference with or without the hintpath, and unfortunately we can't easily embed those in the template or preserve them in the project. All of these assemblies should be resolvable by default though.

PR #3742, opened 2018-02-14
Fix #3611: Mixed-mode debugging doesn't work with 64-bit Python

### LINKED ISSUE #3611: Mixed-mode debugging doesn't work with 64-bit Python

int19h:
Split regular and mixed-mode debugger into separate assemblies, to avoid missing dependencies at runtime.


int19h:
It has been a while since I had to add a new assembly - what are the other necessary changes to package it properly, if any?

Python/Product/Debugger.Concord/Guids.cs
```diff
@@ -0,0 +1,56 @@
+// Python Tools for Visual Studio
```
huguesv: Is this copy of Guids.cs needed? It looks like you already link the version at ../debugger/guids.cs

Python/Product/Debugger/Guids.cs
```diff
@@ -34,7 +34,10 @@ static class Guids {
         public static readonly Guid UnknownPythonModuleGuid = new Guid("42A6F911-9997-4504-861C-91015BDCE588");
         public static readonly Guid CppLanguageGuid = new Guid("3A12D0B7-C26C-11D0-B442-00A0244A1DD2");
 
-        public const string PythonExceptionCategoryId = "EC1375B7-E2CE-43E8-BF75-DC638DE1F1F9";
+		public const string PythonDebugEngineId = "EC1375B7-E2CE-43E8-BF75-DC638DE1F1F9";
```
huguesv: wrong indentation

Python/Product/Debugger.Concord/Guids.cs
int19h: Yep, just something I overlooked.

Python/Product/PythonTools/PythonTools.csproj
```diff
@@ -59,6 +59,11 @@
       <Name>Microsoft.PythonTools.Debugger</Name>
       <IncludeInVSIX>True</IncludeInVSIX>
     </ProjectReference>
+    <ProjectReference Include="..\Debugger.Concord\Debugger.Concord.csproj">
+      <Project>{515F701E-336D-4D58-AF4B-E976BC33C957}</Project>
+      <Name>Microsoft.PythonTools.Debugger</Name>
```
huguesv: Microsoft.PythonTools.Debugger.Concord

Python/Product/Core/Core.csproj
```diff
@@ -101,6 +101,10 @@
       <Project>{DECC7971-FA58-4DB0-9561-BFFADD393BBD}</Project>
       <Name>Microsoft.PythonTools.Debugger</Name>
     </ProjectReference>
+    <ProjectReference Include="..\Debugger.Concord\Debugger.Concord.csproj">
+      <Project>{515F701E-336D-4D58-AF4B-E976BC33C957}</Project>
+      <Name>Microsoft.PythonTools.Debugger.Concord</Name>
+    </ProjectReference>
```
zooba: This should be unnecessary since there is a reference via PythonTools.csproj. Or you can remove the PythonTools.csproj reference if it isn't needed in that assembly.

PR #3758, opened 2018-02-20
Analysis performance improvements

AlexanderSher:
Performance was tested using modified `TestMutateStdLib` tests. Improvement under profiler is 5-7 percent.

### DIFF
--- a/Python/Product/Analysis/Parsing/Ast/DecoratorStatement.cs
+++ b/Python/Product/Analysis/Parsing/Ast/DecoratorStatement.cs
@@ -27,9 +27,7 @@ namespace Microsoft.PythonTools.Parsing.Ast {
             _decorators = decorators;
         }
 
-        public IList<Expression> Decorators {
-            get { return _decorators; }
-        }
+        public Expression[] Decorators => _decorators;
 
         public override void Walk(PythonWalker walker) {
             if (walker.Walk(this)) {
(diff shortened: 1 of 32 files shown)


Python/Product/Analysis/Parsing/Ast/DecoratorStatement.cs
```diff
@@ -27,9 +27,7 @@ public class DecoratorStatement : Statement {
             _decorators = decorators;
         }
 
-        public IList<Expression> Decorators {
-            get { return _decorators; }
-        }
+        public Expression[] Decorators => _decorators;
```
zooba: This breaks a static analysis rule about returning references to arrays through public API

Python/Product/Analysis/Parsing/Ast/PythonAst.cs
```diff
@@ -117,9 +114,25 @@ public sealed class PythonAst : ScopeStatement {
...
+            if (_attributes.TryGetValue(from, out var fromAttrs)) {
+                var toAttrs = new Dictionary<object, object>(fromAttrs.Count);
+                foreach (var nodeAttr in fromAttrs) {
+                    toAttrs[nodeAttr.Key] = nodeAttr.Value;
+                }
+                _attributes[to] = toAttrs;
+            }
+        }
+
+        public void SetAttributes(Dictionary<Node, Dictionary<object, object>> attributes) {
```
zooba: This doesn't need to start as public API.

PR #3763, opened 2018-02-20
Launches the browser if there is a web url present in the options.

### LINKED ISSUE #3732: Debugging Flask/Django using VSC Debugger does not start the browser

karthiknadig:
fixes #3732

### DIFF
--- a/Python/Product/Debugger/DebugAdapterProcess.cs
+++ b/Python/Product/Debugger/DebugAdapterProcess.cs
@@ -22,10 +22,13 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Sockets;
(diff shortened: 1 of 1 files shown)


karthiknadig:
Currently, this uses the experimental debugger to launch the browser. If we want to use some other engine let me know.

zooba:
I think we normally do this through the launcher classes - find the `PythonWebLauncher` class.

karthiknadig:
for `debug` case in the old debugger we launched the browser from AD7Engine's `OnConnected` event. The  `PythonWebLauncher` 's StartBrowser is used in start without debugging mode. So, for this PR I used the way we do in the AD7Engine.

PR #3764, opened 2018-02-21
Lazy function evaluation

### LINKED ISSUE #3675: No IntelliSense on return type with type hint
```
def func() -> int:
    pass

a = func()
a.
```

image

zooba:
This is an idea I've been working on for years and I think I've got it mostly there. Along with significantly improved performance and memory usage in many circumstances, this will also fix #3675 among others.

In brief, we currently create separate analysis units and scopes for each call to a function, using some heuristics to avoid doing this forever (the old call chains implemented "if the call comes through these same call sites assume the parameters are probably the same type"). Before that, we tried to separate them based on parameter types.

With this new approach, we use placeholder types in parameters for normal analysis, and store all the incoming values in a separate dictionary. When a placeholder ("lazy") type escapes, it is resolved to all its possible values. When we are handling a call to the function, if there is a lazy type in the known return value then we substitute based on the current argument set and return that.

(If the arguments may be used in a closure then we revert back to creating a separate call for each set of arguments. Unlike in the past, we only do this for free variables, and if calls come in with different sets then they get different results.)

Still a work in progress because not all the existing tests pass. Some of them codify the old behavior in ways that require the tests to be changed (i.e. we wanted to detect changes to known "failures", which now succeed) but I haven't finished going through them all yet. But feel free to review and make any suggestions, or grab the code and give it a try.

Known issues:
 - [x] Decorators/wrapped functions
 - [x] Returning sequences containing argument values
 - [x] Improvements to typing module support
 - [x] Script for broader analysis testing


zooba:
This is a big and scary PR right now because I've merged in #3844 to get some extra testing on it.

I'm also doing more type hint work in this PR, as we've been told to finish it up for the current release and the changes here were already a huge improvement in our ability to handle certain hints correctly. So there'll be more work coming here, but it should be mostly in the form of tests (in particular, some scripted way to run the analyzer against lots of packages to flush out crashes, etc.).

If we can finish up #3844 soon then this one shouldn't have as many changes, and it'll be time to start reviewing.

huguesv:
Please add AnalysisMemoryTester to our skip verification files.

huguesv:
I ran the full analysis script and got an assert while analyzing pandas (0.20.3) for Anaconda 3 64-bit

```
---------------------------
---------------------------
Accessing Mro before SetBases has been called



   at Microsoft.PythonTools.Interpreter.Ast.AstPythonType.get_Mro() in C:\Projects\GitHub\huguesv\PTVS\Python\Product\Analysis\Interpreter\Ast\AstPythonType.cs:line 91

   at Microsoft.PythonTools.Interpreter.Ast.AstPythonType.GetMember(IModuleContext context, String name) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\Analysis\Interpreter\Ast\AstPythonType.cs:line 199

   at Microsoft.PythonTools.Interpreter.Ast.NameLookupContext.GetValueFromMember(MemberExpression expr, LookupOptions options) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\Analysis\Interpreter\Ast\NameLookupContext.cs:line 224

   at Microsoft.PythonTools.Interpreter.Ast.NameLookupContext.GetValueFromExpression(Expression expr, LookupOptions options) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\Analysis\Interpreter\Ast\NameLookupContext.cs:line 184

   at Microsoft.PythonTools.Interpreter.Ast.NameLookupContext.GetValueFromExpression(Expression expr) in C:\Projects\GitHub\huguesv\......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------
```

zooba:
This is now ready for review. Apologies for the huge-ness, but most of it has been tracking down fixes for broken tests, including many that broke in merges (which I assume were broken elsewhere, but can't be sure).

I'm doing one more full pass over Anaconda, since there were a couple of changes that may affect overall stability, but I'm expecting everything to be pretty much ready.

Python/Product/Analysis/AnalysisValue.cs
```diff
@@ -322,32 +326,48 @@ public class AnalysisValue : IAnalysisSet, ICanExpire {
...
-        /// match the value used in <see cref="UnionEquals"/>.
-        /// </para>
-        /// </remarks>
+        internal virtual IAnalysisSet Resolve(AnalysisUnit unit, ResolutionContext context) {
+            return this;
+        }
+
+        #endregion
+
+            #region Union Equality
```
huguesv: indentation

Python/Tests/fullanalysistest.py
```diff
@@ -0,0 +1,57 @@
+from __future__ import print_function
+
+import os
+import re
+import subprocess
+import sys
+
+if sys.version_info[0] != 3:
+    import threading
```
MikhailArkhipov: Should it be `== 2` ?  Ver 4 prob will be like 3

Python/Tests/Analysis/FunctionBodyAnalysisTests.cs
```diff
@@ -0,0 +1,132 @@
+﻿using System;
```
huguesv: missing header

Python/Product/Analysis/AnalysisHashSet.cs
```diff
@@ -89,6 +88,19 @@ public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable, IEqualityComparer<
             Union(enumerable);
         }
 
+        private BucketSet Buckets {
+            get {
+                lock (this) {
```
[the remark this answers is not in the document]
zooba: Nobody should ever lock on an `IAnalysisSet` instance, since they're (nearly) immutable. And we create a lot of these instances, so every byte counts.

Ultimately, many users will still run out of memory during analysis, but we don't need to encourage it (we also don't really have a good sense of the distribution of memory usage by the analyzer - might be an interesting telemetry point to collect)

Python/Tests/Analysis/TypeAnnotationTests.cs
```diff
@@ -339,5 +350,257 @@ private class StringConverter : TypeAnnotationConverter<string> {
...
+    return 'Hello ' + name
+", 
+                new[] {
+                    "greeting:greeting(name:str=)->[str]"
+                }
+            );
+        }
+
+        [TestMethod, Priority(0)]
+        public void TypingModuleDocumentationExample_2() {
```
huguesv: Fails:
```Assert.AreEqual failed. Expected:<scale(scalar:float=,vector:list, list[float]=)->[list,list[float]]>. Actual:<scale(scalar:float=,vector:list[float]=)->[list,list[float]]>.```

File: Python/Tests/Analysis/TypeAnnotationTests.cs [hunk 6]
```diff
@@ -29,6 +29,17 @@
 namespace AnalysisTests {
     [TestClass]
     public class TypeAnnotationTests : BaseAnalysisTest {
+        [TestInitialize]
```
huguesv: TypingModuleProtocolAnalysis fails:
```
Assert.Fail failed. Expected description of 'test-module.i' was 'iterable'. Actual was '<unknown>' or '<unknown>'
```

huguesv:
Additional failures in AnalysisTest/StdLibAnalysisTest:

BuiltinMethodSignatures
```
Assert.AreEqual failed. Expected:<1>. Actual:<0>. Expected one parameter for const.append
```

CartesianClosures
```
Assert.Fail failed. ContainsExactly failed.

Expected:
{Int}

Actual:
{Int, Bytes}
```

CartesianStarArgs (expected)

Generator2x
```
Assert.Fail failed. ContainsExactly failed.

Expected:
{Int}

Actual:
{Int, Bytes}
```

Generator3x
```
Assert.Fail failed. ContainsExactly failed.

Expected:
{Int}

Actual:
{Int, Unicode}
```

OverrideCompletions
```
Assert.AreEqual failed. Expected:<2>. Actual:<1>. Expected: <
self
object
>.
Actual: <
self
>.
```

SuperclassMemberReferencesCrossModule (expected)

MikhailArkhipov:
LGTM. I've seen error mentioned by @huguesv on Mac so it may not be related to changed.

AlexanderSher:
Please don't merge it right now, I need to look a bit more into it

Python/Product/Analysis/AnalysisHashSet.cs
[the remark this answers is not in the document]
AlexanderSher: > Nobody should ever lock on an `IAnalysisSet` instance, since they're (nearly) immutable. 

Unless someone does and you miss it on code review. I'm not sure I understand the purpose of this lock. It should be enough to declare `_buckets` volatile.

Python/Product/Analysis/AnalysisHashSet.cs
```diff
@@ -505,6 +564,20 @@ struct Bucket {
...
+                Buckets = capacity > 0 ? new Bucket[capacity] : Array.Empty<Bucket>();
+                Count = count;
+                if (count > capacity) {
+                    throw new ArgumentOutOfRangeException(nameof(count));
+                }
+            }
+
+            public readonly Bucket[] Buckets;
+            public int Count;
+            public int Capacity => Buckets?.Length ?? 0;
```
AlexanderSher: `Capacity` can also be set in constructor

Python/Product/Analysis/AnalysisHashSet.cs
```diff
@@ -505,6 +564,20 @@ struct Bucket {
...
+            public BucketSet(int capacity, int count = 0) {
+                Buckets = capacity > 0 ? new Bucket[capacity] : Array.Empty<Bucket>();
+                Count = count;
+                if (count > capacity) {
+                    throw new ArgumentOutOfRangeException(nameof(count));
+                }
+            }
+
+            public readonly Bucket[] Buckets;
+            public int Count;
```
AlexanderSher: Structs shouldn't have mutable fields

Python/Product/Analysis/AnalysisHashSet.cs
```diff
@@ -377,35 +432,38 @@ public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable, IEqualityComparer<
...
                         // The hash code should not change, but if it does, we
                         // need to keep things consistent
                         Debug.Fail("Hash code changed when merging AnalysisValues");
                     }
                     Thread.MemoryBarrier();
-                    buckets[index].Key = _removed;
-                    _count--;
-                    return AddOne(buckets, newKey, newHc);
+                    buckets.Buckets[index].Key = _removed;
+                    buckets.Count -= 1;
```
AlexanderSher: Is this method `AddOne` supposed to be thread-safe?

Python/Product/Analysis/AnalysisHashSet.cs
```diff
@@ -265,96 +323,93 @@ public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable, IEqualityComparer<
...
+            var buckets = Buckets;
+            var res = AddOne(ref buckets, key, _comparer);
+            Buckets = buckets;
+            return res;
+        }
 
-                if (Add(_buckets, key)) {
-                    _count++;
+        private static bool AddOne(ref BucketSet buckets, AnalysisValue key, IEqualityComparer<AnalysisValue> comparer) {
+            if (buckets.Buckets == null) {
```
AlexanderSher: `buckets.Buckets` isn't checked for `null` in many places. Is it safe?

Python/Product/Analysis/AnalysisHashSet.cs
```diff
@@ -228,23 +268,41 @@ public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable, IEqualityComparer<
...
+            res.Buckets = newBuckets;
             return res;
         }
 
         public bool SetEquals(IAnalysisSet other) {
+            if (ReferenceEquals(this, other)) {
+                return true;
+            }
+
+            if (Comparer == other.Comparer) {
```
AlexanderSher: What is the scenario when comparers aren't equal?

Python/Product/Analysis/AnalysisHashSet.cs
```diff
@@ -377,35 +432,38 @@ public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable, IEqualityComparer<
...
+            if (buckets.Buckets[addIndex].Key == null || buckets.Buckets[addIndex].Key == _removed) {
+                // Removal has been counted already
+                buckets.Count += 1;
+            } else if (!buckets.Buckets[addIndex].Key.IsAlive) {
+                // Remove/add means no change te count
+            } else {
+                buckets.Count += 1;
             }
-            buckets[addIndex].HashCode = hc;
+            buckets.Buckets[addIndex].HashCode = hc;
```
AlexanderSher: What's the point of calling `Thread.MemoryBarrier();` after this line?

Python/Product/Analysis/ResolutionContext.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public FunctionInfo Caller { get; set; }
+        public ArgumentSet CallArgs {
+            get => _callArgs ?? LazyCallArgs?.Value ?? default(ArgumentSet);
+            set => _callArgs = value;
+        }
+        public Lazy<ArgumentSet> LazyCallArgs { get; set; }
+        public Node CallSite { get; set; }
+        public bool ResolveFully { get; set; }
+        public int ResolveDepth { get; set; } = 10;
```
AlexanderSher: `ResolveDepth` is never changed.

Python/Product/Analysis/ResolutionContext.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Diagnostics;
+using Microsoft.PythonTools.Analysis.Values;
+using Microsoft.PythonTools.Parsing.Ast;
+
+namespace Microsoft.PythonTools.Analysis {
+    sealed class ResolutionContext {
+        private ArgumentSet? _callArgs;
+        private int _depth;
+
+        public static ResolutionContext Empty => new ResolutionContext();
```
AlexanderSher: I can't find any usage of this property

Python/Product/Analysis/ResolutionContext.cs
```diff
@@ -0,0 +1,60 @@
...
+        public static ResolutionContext Complete => new ResolutionContext { ResolveFully = true };
+
+        public FunctionInfo Caller { get; set; }
+        public ArgumentSet CallArgs {
+            get => _callArgs ?? LazyCallArgs?.Value ?? default(ArgumentSet);
+            set => _callArgs = value;
+        }
+        public Lazy<ArgumentSet> LazyCallArgs { get; set; }
+        public Node CallSite { get; set; }
+        public bool ResolveFully { get; set; }
```
AlexanderSher: `ResolveFully` can be made readonly

Python/Product/Analysis/Values/IterableInfo.cs
```diff
@@ -202,5 +186,89 @@ public IterableValue(VariableDef[] indexTypes, BuiltinClassInfo seqType, Node no
...
+            }
+
+            if (newTypes.Length == 0 || IndexTypes.Length == 0) {
+                return false;
+            }
+
+            var resolvedTypes = new IAnalysisSet[IndexTypes.Length];
+            bool anyChange = false;
+
+            if (Push()) {
```
AlexanderSher: `ResolveIndexTypes` calls `Push()` instead of `context.Push()`. Is it correct?

Python/Product/Analysis/ResolutionContext.cs
```diff
@@ -0,0 +1,60 @@
...
+        public int ResolveDepth { get; set; } = 10;
+
+        private Lazy<Dictionary<AnalysisValue, IAnalysisSet>> _cache = new Lazy<Dictionary<AnalysisValue, IAnalysisSet>>();
+        public Dictionary<AnalysisValue, IAnalysisSet> Cache => _cache.Value;
+
+        public bool Push() {
```
AlexanderSher: `ResolutionContext.Push` is called only in two places. Is that intentional?

Python/Product/Analysis/AnalysisValueSetExtensions.cs
```diff
@@ -259,6 +274,60 @@ public static class AnalysisValueSetExtensions {
...
+            // or changing anything.
+
+            if (!context.Push()) {
+                changed = false;
+                return self;
+            }
+            try {
+                List<AnalysisValue> removed = null;
+                IAnalysisSet added = null;
+                foreach (var ns in self) {
```
AlexanderSher: Why variable name is `ns`?

Python/Product/Analysis/AnalysisHashSet.cs
zooba: This doesn't take up any storage space (`AnalysisHashSet` is our most frequently created type)

Python/Product/Analysis/AnalysisHashSet.cs
zooba: Sure they can. In this class, we always copy it (under lock) and then copy it back (under lock) - it never gets updated directly.

Python/Product/Analysis/AnalysisHashSet.cs
zooba: Should be safe without locks for one writer, multiple readers.

Python/Product/Analysis/AnalysisHashSet.cs
zooba: Checking `.Capacity` or calling `Contains()` implicitly includes the null check, and the other places should silently pass on the nulls without impact. This one is special because we know we need to replace `null` with actual buckets so we can add the value, and the rest of this function assumes it is non-null.

Python/Product/Analysis/AnalysisHashSet.cs
zooba: I don't understand the question? We have a selection of 4-5 different comparers that affect how values are combined, so it's very easy for two different sets to have different comparers.

Python/Product/Analysis/AnalysisHashSet.cs
zooba: According to the comment two lines below (contributed by @DinoV), it's to make sure that the hashcode is stored before the key changes.

Python/Product/Analysis/AnalysisHashSet.cs
```diff
@@ -265,96 +323,93 @@ public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable, IEqualityComparer<
...
-                        oldBuckets[i].Key = _removed;
-                        newSize--;
-                        _count--;
-                    }
+            for (int i = 0; i < oldBuckets.Capacity; i++) {
+                var key = oldBuckets.Buckets[i].Key;
+                if (key != null && !key.IsAlive) {
+                    oldBuckets.Buckets[i].Key = _removed;
+                    newSize--;
+                    oldBuckets.Count--;
```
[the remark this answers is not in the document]
zooba: I'll make it a ref parameter like the rest

Python/Product/Analysis/ResolutionContext.cs
zooba: Removed. It's not useful for comparisons, so there's no value in having it.

Python/Product/Analysis/ResolutionContext.cs
zooba: Not the point of this class. All of these can be set in code before it gets passed to the Resolve function.

Python/Product/Analysis/ResolutionContext.cs
zooba: Yes. One of those places is the extension method that gets called from most calls to `Resolve`.

Python/Product/Analysis/AnalysisValueSetExtensions.cs
zooba: Legacy reasons mainly. This type used to be `Namespace` instead of `AnalysisValue` and `NamespaceSet` instead of `AnalysisSet`. I did a huge refactor a few years ago, but there was nothing gained by changing all the loop variables.

Python/Product/Analysis/AnalysisHashSet.cs
AlexanderSher: Nevermind, I got it.

Python/Product/Analysis/AnalysisHashSet.cs
AlexanderSher: Then `Thread.MemoryBarrier` must be called before every call to `HashCode` field, otherwise it is useless.

zooba: Presumably between reads of `Key` and reads of `HashCode`, yes? So that we ensure that we always read `Key` first and always write to `Key` last?

AlexanderSher:
We don't have time for the full review of the analysis logic, but one thing that worries me a lot is `AnalysisHashSet`. It heavily relies on mutable structs which is a bug prone event in single-threaded environment. `AddOne` uses an array which is shared with all other methods and accessed without locks. Rather inconsistent usage of `Thread.MemoryBarrier();`. And no multi-threaded stress testing.

@zooba , if you can give me more info about `AnalysisHashSet` and why it is such a critical type, that would be great. 

For now, let's merge the PR.

Python/Product/Analysis/AnalysisHashSet.cs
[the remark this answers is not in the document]
AlexanderSher: Sorry, I've written this comment before I've realized that `BucketSet` is a struct. `volatile` can't be used for it.

Python/Product/Analysis/AnalysisHashSet.cs
AlexanderSher: To ensure that we read `Key` first and always write to `Key` last we need several calls:

`AddOne`:
```cs
bucket.HashCode = hashCode;
Thread.MemoryBarrier();
bucket.Key = key;
Thread.MemoryBarrier();
```

`Contains`:
```cs
Thread.MemoryBarrier();
var key = bucket.Key;
if (key != null) {
    Thread.MemoryBarrier();
    var hashCode = buckets[index].HashCode;
    if (Object.ReferenceEquals(key, existingKey) ||
        (existingKey != _removed && hashCode == hc && _comparer.Equals(key, (AnalysisValue)existingKey))) {
        return index;
    }
}
```

And that is correct if we have these two methods.

AlexanderSher: But let's not change it right now.

zooba:
I still need to replace the `lock (this)` with volatile, but @DinoV will have to speak to the design of AnalysisHashSet as he's the one who implemented it.

AlexanderSher:
> I still need to replace the `lock (this)` with volatile

It won't compile. I've put that comment before I've found that `BucketSet` is a struct.

zooba:
Just found the same thing. Merging now

PR #3766, opened 2018-02-21
Fixes #3765 Create MSBuild files package

### LINKED ISSUE #3765: Create MSBuild files package
We need to create a package with enough of our files to satisfy MSBuild on our projects. We don't need to be fully functional or add any new build functionality

zooba:
Fixes #3765 Create MSBuild files package
Creates a new project and package with a minimal set of MSBuild tasks
We use simpler environment detection and embed more code files to avoid dependencies. The new assembly won't be installed on regular VS builds.


huguesv:
Is this solving a problem? If so, what is it?

zooba:
The problem is that when you install the Build Tools SKU, you can't build solutions that include Python projects. We'll add this package into that SKU so that it doesn't crash out completely.

zooba:
@huguesv Could you take another look with the new changes? I've been testing on a build tools install and it works fine.

PR #3768, opened 2018-02-21
LEGO: Pull request from lego/hb_839_20180221154044281 to 15.6rel

csigs:
LEGO: Pull request from lego/hb_839_20180221154044281 to 15.6rel with localized lcls


zooba:
This makes an incorrect change to the fwlink, so will not be merging it.

PR #3773, opened 2018-02-21
Fix #3761 & some more perf turning

### LINKED ISSUE #3761: Parser with Verbatim=false throws assert for function definition
Test:
```cs
[TestMethod]
public void FunctionDefTest() {
    var code = @"def func(p):\r\n    pass";
    var parser = Parser.CreateParser(new StringReader(code), PythonLanguageVersion.V36, new ParserOptions { Verbatim = false });
    var ast = parser.ParseFile();
}
```

Result:
```cs
   at Microsoft.PythonTools.Parsing.Parser.AddPreceedingWhiteSpace(Node ret, String whiteSpace) in D:\GitHub\PTVS\Python\Product\Analysis\Parsing\Parser.cs:line 5479
   at Microsoft.PythonTools.Parsing.Parser.AddPreceedingWhiteSpace(Node ret) in D:\GitHub\PTVS\Python\Product\Analysis\Parsing\Parser.cs:line 5451
   at Microsoft.PythonTools.Parsing.Parser.ParseVarArgsList(TokenKind terminator, Boolean allowAnnotations, List`1& commaWhiteSpace, Boolean& ateTerminator) in D:\GitHub\PTVS\Python\Product\Analysis\Parsing\Parser.cs:line 2040
   at Microsoft.PythonTools.Parsing.Parser.ParseFuncDef(Boolean isCoroutine) in D:\GitHub\PTVS\Python\Product\Analysis\Parsing\Parser.cs:line 1913
   at Microsoft.PythonTools.Parsing.Parser.ParseStmt() in D:\GitHub\PTVS\Python\Product\Analysis\Parsing\Parser.cs:line 460
   at Microsoft.PythonTools.Parsing.Parser.ParseFileWorker() in D:\GitHub\PTVS\Python\Product\Analysis\Parsing\Parser.cs:line 4610
   at Microsoft.PythonTools.Parsing.Parser.ParseFile() in D:\GitHub\PTVS\Python\Product\Analysis\Parsing\Parser.cs:line 152
```

AlexanderSher:
- Fix #3761: Parser with Verbatim=false throws assert for function definition
- Cache field value in Tokenizer.Peek (+3%)
- Optimize IsIdentifierChar for Latin1 case (+7%)

### DIFF
--- a/Python/Product/Analysis/Parsing/Tokenizer.cs
+++ b/Python/Product/Analysis/Parsing/Tokenizer.cs
@@ -24,6 +24,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Numerics;
+using System.Runtime.CompilerServices;
 using System.Text;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 
@@ -815,10 +816,21 @@ namespace Microsoft.PythonTools.Parsing {
         public static bool IsIdentifierStartChar(char ch) {
             // Identifiers determined according to PEP 3131
 
-            switch (ch) {
+            // ASCII case
+            if (ch <= 'z') {
                 // Underscore is explicitly allowed to start an identifier
-                case '_':
-                    return true;
+                return ch <= 'Z' ? ch >= 'A' : ch >= 'a' || ch == '_';
+            }
+
+            if (ch < 0xAA) {
+                return false;
+            }
+
+            return IsIdentifierStartCharNonAscii(ch);
+        }
+
+        private static bool IsIdentifierStartCharNonAscii(char ch) {
+            switch (ch) {
                 // Characters with the Other_ID_Start property
                 case '\x1885':
                 case '\x1886':
@@ -846,10 +858,14 @@ namespace Microsoft.PythonTools.Parsing {
 
         public static bool IsIdentifierChar(char ch) {
             // ASCII case
-            if (ch <= 7F) {
+            if (ch <= 'z') {
                 return ch <= 'Z'
                     ? ch >= 'A' || ch >= '0' && ch <= '9'
-                    : ch >= 'a' && ch <= 'z' || ch == '_';
+                    : ch >= 'a' || ch == '_';
+            }
+
+            if (ch < 0xAA) {
+                return false;
             }
 
             switch (ch) {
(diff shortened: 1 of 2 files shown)


Python/Product/Analysis/Parsing/Tokenizer.cs
```diff
@@ -844,8 +845,15 @@ public Tokenizer(PythonLanguageVersion version, ErrorSink errorSink = null, Toke
         }
 
         public static bool IsIdentifierChar(char ch) {
-            if (IsIdentifierStartChar(ch)) {
-                return true;
+            // Latin1 case
+            if (ch <= byte.MaxValue) {
+                return ch <= 'z'
+                    ? ch <= 'Z'
+                      ? ch >= 'A' || ch >= '0' && ch <= '9'
```
zooba: Identifiers can't start with a number.

This would be as much of a gain if it just checked ASCII letters, I suspect.

AlexanderSher: Yep, ASCII is even better.

Code below returns `true` for `UnicodeCategory.DecimalDigitNumber`:
```cs
            switch (char.GetUnicodeCategory(ch)) {
                // Supported categories for continuing an identifier
                case UnicodeCategory.NonSpacingMark:
                case UnicodeCategory.SpacingCombiningMark:
                case UnicodeCategory.DecimalDigitNumber:
                case UnicodeCategory.ConnectorPunctuation:
                    return true;
            }
```

Python/Product/Analysis/Parsing/Tokenizer.cs
```diff
@@ -844,8 +845,11 @@ public Tokenizer(PythonLanguageVersion version, ErrorSink errorSink = null, Toke
         }
 
         public static bool IsIdentifierChar(char ch) {
-            if (IsIdentifierStartChar(ch)) {
-                return true;
+            // ASCII case
+            if (ch <= 7F) {
+                return ch <= 'Z'
+                    ? ch >= 'A' || ch >= '0' && ch <= '9'
+                    : ch >= 'a' && ch <= 'z' || ch == '_';
```
zooba: (Lost the old comment)

Digits are correct here. Should we add the fast case (without digits) to `IsIdentifierStartChar` as well? Does that have an impact?

AlexanderSher: In mutate std lib tests impact is small (about 1% average), but there is no big deal to add it.

PR #3778, opened 2018-02-23
Analysis reliability improvements

zooba:
Fixes crash due to bad indexing
Ensures errors are properly logged and raised.
Attempts to reduce false positives when warning about undefined variables.


zooba:
I think this change is changing that file back to the correct style, but I'll merge and see.

PR #3794, opened 2018-02-28
Fixes #3793 Migrate hover to server

### LINKED ISSUE #3793: Migrate hover to server
Part of #3484, enables the Hover() message so we can return tool tip info.

We should also extend this to include specific type information, since this is the place where it is most available. This greatly increases the use of the LSP for general code analysis where you care more about canonical names than user-visible text.

zooba:
Fixes #3793 Migrate hover to server
Implements the Hover() command on server
Also disables verbatim parsing by default, as it was interfering with tests. Some work for lambda was required to restore the previous behaviour.

### DIFF
--- a/Python/Product/Analysis/LanguageServer/Server.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.cs
@@ -22,6 +22,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PythonTools.Analysis.Infrastructure;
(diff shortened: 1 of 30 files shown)


huguesv:
I started fixing this.

This is breaking go to definition for class instance methods (I'm comparing against 15.6 installed in a VM). F12 on `add` in the last line takes you to the definition of `c` instead of the definition of `add`.
```
from typing import Union

class Calculator:
    """docs for class"""
    def add(self, a: Union[int, float], b: int) -> int:
        '''docs for add'''
        return int(a + b)

c = Calculator()

r1 = c.add(8.0, 9)
```

zooba:
I doubt that's caused by this PR, but I'll take a look.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -508,6 +509,87 @@ public sealed class Server : ServerBase, IDisposable {
             }).Concat(extras).ToArray();
         }
 
+        public override async Task<Hover> Hover(TextDocumentPositionParams @params) {
```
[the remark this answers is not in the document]
zooba: Following the protocol naming here - none of the other methods have "Async" in their name, and they're all async.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -508,6 +509,87 @@ public sealed class Server : ServerBase, IDisposable {
...
+                // Return module information
+                return new Hover { contents = "{0} : module".FormatUI(w.ImportedName) };
+            }
+
+            Expression expr;
+            SourceSpan exprSpan;
+            Analyzer.InterpreterScope scope = null;
+
+            if (!string.IsNullOrEmpty(@params._expr)) {
+                TraceMessage($"Getting hover for {@params._expr}");
```
AlexanderSher: Since we have `TraceMessage(string)`, `TraceMessage(IFormattable)` isn't called. Need to rename one of overloads.

Python/Product/Analysis/Parsing/ParserOptions.cs
```diff
@@ -31,7 +31,7 @@ public sealed class ParserOptions {
 
         public Severity IndentationInconsistencySeverity { set; get; }
 
-        public bool Verbatim { get; set; } = true;
+        public bool Verbatim { get; set; } = false;
```
AlexanderSher: What is affected by this change?

Python/Product/Analysis/Values/BuiltinClassInfo.cs
```diff
@@ -81,6 +81,27 @@ public BuiltinClassInfo(IPythonType classObj, PythonAnalyzer projectState)
...
+                    if (_type.IsBuiltin) {
+                        return _type.Name;
+                    }
+                    return _type.DeclaringModule.Name + "." + _type.Name;
+                }
+                return null;
+            }
+        }
+
+        public KeyValuePair<string, string> FullyQualifiedNamePair {
```
AlexanderSher: I can't find usage of this property

zooba: Not yet, but it's the only way to preserve some of this information, and we have consumers of this API that will likely need it.

I can rewrite some of the `FullyQualifiedName` implementation by using this properly if you prefer? :)

Python/Product/Analysis/Parsing/ParserOptions.cs
zooba: EVERYTHING! This is what we suggested in response to your performance analysis finding the huge amount of time spent in GetNodeAttributes, as this should skip most of those.

The main thing it affects is anywhere we try to convert an AST branch back into code. Before, we had all the whitespace, etc. required to do it perfectly, but now we have to rely on formatting options (and we now have `CodeFormattingOptions.Traditional`, which is basically the normal style for Python code).

It was relevant here because we convert AST nodes into text for hover that we don't necessarily do elsewhere - default values for function parameters is the main one, IIRC. If we've parsed verbatim, then comments before or within a node will appear in the representation, which is not what we want. We only really want to do verbatim parsing explicitly when we know we are going to need to round-trip (i.e. formatting, extract method).

Python/Product/Analysis/LanguageServer/Server.cs
zooba: I'll just delete the `string` overload and add `$` prefix for all the strings.

zooba:
Go To Definition won't be completely fixed until we do #3803, but it's reasonably better now with a simple fix.

The underlying problem is we merge distinct messages between VS and the analyzer, and then split them back out for the language server, then recombine the responses. So we lose some context along the way. Once there are separate messages here (and eventually when we switch to the pure VS protocol) .

PR #3795, opened 2018-03-01
Small optimizations for conda discovery

### LINKED ISSUE #3540: Optimize code that determines which conda.exe to use
When more than one global conda.exe is found, we look at the version information `conda.exe -V`, and the latest version wins.

The problem is that it's slow to run `conda.exe -V`. And we do it in a few places. And when the global list of environments is rescanned, it has to be recomputed because a new conda.exe might show up (or an existing one deleted).

We need to minimize the number of times we do this. We should be able to cache the version per file path (maybe along with file time stamp in case it's upgraded).

Users of `CondaEnvironmentManager` are the ones that are computing this every time and need to be updated.

huguesv:
- Read from .version instead of calling conda -V when trying to determine which global conda.exe is the latest
- Do some parallel environment creation (instead of serially calling python.exe for each environment)

Fix #3540 

Timing of complete discovery of all conda environments varies quite a bit from run to run, but it seems to be slightly better on average (for me by about 1 sec)

As far as fixed cost of finding which global conda.exe to use, I don't know that it's worth doing a lot more than this (like caching results), especially if we plan on having a private conda at some point.

### DIFF
--- a/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
+++ b/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
@@ -292,7 +292,8 @@ namespace Microsoft.PythonTools.Interpreter {
                     .AsParallel()
                     .Where(folder => Directory.Exists(folder))
                     .Select(folder => CreateEnvironmentInfo(folder))
-                    .Where(env => env != null);
+                    .Where(env => env != null)
+                    .ToList();
             }
 
             return Enumerable.Empty<PythonInterpreterInformation>();
--- a/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
+++ b/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
@@ -285,25 +285,17 @@ namespace Microsoft.PythonTools.Interpreter {
             public string[] EnvironmentRootFolders = null;
         }
 
-        private List<PythonInterpreterInformation> FindCondaEnvironments(string condaPath) {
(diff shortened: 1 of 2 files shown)


Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -285,25 +285,17 @@ class CondaInfoResult {
...
-                    PythonInterpreterInformation env = CreateEnvironmentInfo(folder);
-                    if (env != null) {
-                        found.Add(env);
-                    }
-                }
+                return condaInfoResult.EnvironmentFolders
+                    .AsParallel()
+                    .Where(folder => Directory.Exists(folder))
+                    .Select(folder => CreateEnvironmentInfo(folder))
+                    .Where(env => env != null);
```
AlexanderSher: Since this method uses file system, it is better to call `ToList` at the end to avoid multiple enumerations

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -285,25 +285,18 @@ class CondaInfoResult {
             public string[] EnvironmentRootFolders = null;
         }
 
-        private List<PythonInterpreterInformation> FindCondaEnvironments(string condaPath) {
-            var found = new List<PythonInterpreterInformation>();
-            var watchFolders = new HashSet<string>();
-
+        private IEnumerable<PythonInterpreterInformation> FindCondaEnvironments(string condaPath) {
```
zooba: I see no reason not to keep returning a List here (or `IReadOnlyList`, if you prefer).

PR #3811, opened 2018-03-02
Profilermanager

afaruqui:
Enables ability to call external profiler, specifically Intel® VTune™ Amplifier.


Python/Product/ExternalProfilerDriver/DWJsonModel.cs
```diff
@@ -0,0 +1,165 @@
+using System;
```
zooba: We'll need a header comment with license information on each source file. Apache 2.0 is preferred.

Python/Product/ExternalProfilerDriver/DWJsonModel.cs
```diff
@@ -0,0 +1,165 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver
+{
+    // Most numerical integer data is represented in dwjson as a 64-bit unsigned integer
+    // Due to JSON limitations, this is represented as a 2D named tuple, where the fields are
+    // "l" (low) and "h" (high), each containing a 32-bit *signed* integer.
+    // In this way, the max value (0xffffffffffffffff) is represented as {"h" : -1, "l" : -1}
+    public class LongInt
+    {
```
zooba: Our preferred style puts most opening braces on the same line as the class/if/function/etc.

Python/Product/ExternalProfilerDriver/DWJsonModel.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver
+{
+    // Most numerical integer data is represented in dwjson as a 64-bit unsigned integer
+    // Due to JSON limitations, this is represented as a 2D named tuple, where the fields are
+    // "l" (low) and "h" (high), each containing a 32-bit *signed* integer.
+    // In this way, the max value (0xffffffffffffffff) is represented as {"h" : -1, "l" : -1}
+    public class LongInt
```
zooba: Avoid making anything public unless it needs to be. As this is an executable, in theory nothing should be public. (Doesn't have to be explicitly private - simply using `class LongInt` is fine)

Python/Product/ExternalProfilerDriver/DWJsonModel.cs
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    public static class TraceUtils
+    {
+        public static LongInt ToNanoseconds(double timepoint)
+        {
+            const ulong factor = 1000000000;
+            ulong timespec = (ulong)(timepoint * factor);
+            string hexrep = timespec.ToString("X");
```
zooba: Check out the [`BitConverter`](https://docs.microsoft.com/en-us/dotnet/api/system.bitconverter?view=netframework-4.7) class for this conversion.

Python/Product/ExternalProfilerDriver/ExternalProfilerDriver.csproj
```diff
@@ -0,0 +1,104 @@
...
+        <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+        <FileUpgradeFlags>
+        </FileUpgradeFlags>
+        <UpgradeBackupLocation>
+        </UpgradeBackupLocation>
+        <OldToolsVersion>14.0</OldToolsVersion>
+      </PropertyGroup>
+    </When>
+  </Choose>
+  <Import Project="..\ToolProjectBefore.settings" />
```
zooba: This is not actually a tool project, as the intent is to include it in a release. Just use `ProjectBefore` and after.

Python/Product/ExternalProfilerDriver/ExternalProfilerDriver.csproj
```diff
@@ -0,0 +1,104 @@
...
+    <ProjectReference Include="..\Analysis\Analysis.csproj">
+      <Project>{a85d479d-67a9-4bdb-904a-7d86daf68a6f}</Project>
+      <Name>Microsoft.PythonTools.Analysis</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\Common\Common.csproj">
+      <Project>{b3db0521-d9e3-4f48-9e2e-e5ecae886049}</Project>
+      <Name>Microsoft.PythonTools.Common</Name>
+    </ProjectReference>
+  </ItemGroup>
+  -->
```
zooba: You won't need the commented code, so feel free to remove it.

Python/Product/ExternalProfilerDriver/ProcessAsyncRunner.cs
```diff
@@ -0,0 +1,102 @@
...
+            };
+            psi.EnvironmentVariables["AMPLXE_EXPERIMENTAL"] = "time-cl";
+
+            Process process = new Process
+            {
+                StartInfo = psi,
+                EnableRaisingEvents = true
+            };
+
+            process.Exited += (sender, localEventArgs) =>
```
zooba: Since you're using redirected output, you should use its event when `e.Data == null` to detect when the process exits. 

The Exited event may be raised before the output is fully read.

Python/Product/ExternalProfilerDriver/ProcessAsyncRunner.cs
```diff
@@ -0,0 +1,102 @@
...
+                Task<Process> t = ProcessAsyncRunner.Run(exe, args, cts.Token, progress);
+
+                bool processRunning = true;
+                t.ContinueWith((p) => { processRunning = false; });
+
+                while (processRunning)
+                {
+                    Console.Write(".");
+                }
+                Console.WriteLine("\nDone with the task!");
```
zooba: Guessing there's some tidying up of the output left to do here? (Flushing after writing the `'.'`, or overwriting the current line with the progress event, and possibly rate limiting the dots.)

Python/Product/ExternalProfilerDriver/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+            string vtuneReportTimeArgs = reptimespec.FullCLI();
+
+            if (!DryRunRequested)
+            {
+                ProcessAsyncRunner.RunWrapper(vtuneExec, vtuneCollectArgs);
+                ProcessAsyncRunner.RunWrapper(vtuneExec, vtuneReportArgs);
+                ProcessAsyncRunner.RunWrapper(vtuneExec, vtuneReportTimeArgs);
+            }
+
+            string tempOutDir = Environment.GetEnvironmentVariable("USERPROFILE");
```
zooba: This is one of the worst places to write output to. If it's temporary, use `Path.GetTempPath()` (and probably create a randomly named subdirectory to avoid overwriting anything), or allow it to be overridden from the command line (which will make testing easier, as it can be redirected to a non-system wide location).

Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
```diff
@@ -0,0 +1,349 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+using Microsoft.DotNet.PlatformAbstractions;
+
+namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver
+{
+    public class VTuneInvoker
+    {
+        private static readonly string _vtuneCl = @"\bin32\amplxe-cl.exe";
```
zooba: `const string`

Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
```diff
@@ -0,0 +1,349 @@
...
+                {
+                    throw new VTuneNotInstalledException();
+                }
+
+            }
+            else
+            {
+                envvarval = "OS not supported"; // should this throw an exception?
+            }
+            if (File.Exists(envvarval + _vtuneCl)) // not exactly sure why Path.Combine doesn't work here
```
zooba: `Path.Combine` should work, unless you had the leading backslash in `_vtuneCl` at the time. Without the backslash, it should be okay.

Also note that the backslashes won't work on Linux, and this line looks like it will be run in both.

Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
```diff
@@ -0,0 +1,349 @@
...
+                 RuntimeEnvironment.OperatingSystemPlatform == Platform.Linux)
+            {
+                envvarval = Environment.GetEnvironmentVariable(_vtune17Envvar);
+                if (envvarval == null)
+                {
+                    envvarval = Environment.GetEnvironmentVariable(_vtune18Envvar);
+                }
+                if (envvarval == null)
+                {
+                    throw new VTuneNotInstalledException();
```
zooba: We do allow use of this syntax if you prefer it:

```
envvarval = Environment.GetEnvironmentVariable(_vtune17Envvar) ??
    Environment.GetEnvironmentVariable(_vtune18Envvar) ??
    throw new VTuneNotInstalledException();
```

Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
```diff
@@ -0,0 +1,349 @@
...
+
+namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver
+{
+    public class VTuneInvoker
+    {
+        private static readonly string _vtuneCl = @"\bin32\amplxe-cl.exe";
+        private static readonly string _vtune17Envvar = "VTUNE_AMPLIFIER_2017_DIR";
+        private static readonly string _vtune18Envvar = "VTUNE_AMPLIFIER_2018_DIR";
+
+        public static string VTunePath()
```
zooba: Looks like this can be private and only used when someone passes null/empty to the constructor?

Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
```diff
@@ -0,0 +1,349 @@
...
+        }
+    }
+
+    public class VTuneNotInstalledException : Exception
+    {
+        public override string Message
+        {
+            get
+            {
+                return "Only VTune 2017 or 2018 supported, see https://software.intel.com/en-us/intel-vtune-amplifier-xe";
```
zooba: Always avoid including `en-us` in URLs. This page will automatically redirect to the user's language if it's just "....com/intel-vtune..."

Python/Product/ExternalProfilerDriver/VTuneStackParser.cs
```diff
@@ -0,0 +1,159 @@
...
+            if (str.Length > 0)
+            {
+                if (str[str.Length - 1] == '"') { str = str.Substring(0, str.Length - 1); }
+            }
+            return str;
+        }
+
+        public static IEnumerable<SampleWithTrace> ParseFromStream(this IEnumerable<string> seq)
+        {
+            Regex startsWithComma = new Regex("^(,+)(.*)$");
```
zooba: May as well use a regex for the entire string. That will actually simplify the rest of this function fairly dramatically. (Or possibly compare the first character with `,` and then have two separate regexes.)

Python/Product/ExternalProfilerDriver/VTuneToDWJSON.cs
```diff
@@ -0,0 +1,261 @@
...
+        {
+            var samples = VTuneStackParser.ReadFromFile(filename)
+                                          .Skip(1)
+                                          .Select(s => String.Join(",", s.Split(',')
+                                          .Select(VTuneStackParser.RemovePrePosComma)))
+                                          .ParseFromStream();
+            return samples;
+        }
+
+        public static IEnumerable<int> SequenceGenerator(int start = 0)
```
zooba: `Enumerable.Range(0, int.MaxValue)` also does this

Python/Product/ExternalProfilerDriver/ProcessAsyncRunner.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver
+{
+    class ProcessAsyncRunner
+    {
+        static public Task<Process> Run(string execfname,
```
AlexanderSher: `public static`

Python/Product/ExternalProfilerDriver/ProcessAsyncRunner.cs
AlexanderSher: I believe we also need error stream to be drained as well.

PR #3814, opened 2018-03-02
Basic VS Code language server

MikhailArkhipov:
- .NET Standard project for Analysis
- Language Server facade to VSC as .NET Core App
- Project to publish .NET Core app
- some VSC supporting file (formatting setting, gitignore) - for for in VSC with C#


Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -16,8 +16,10 @@
 
 using System;
 using System.Collections.Generic;
+using Newtonsoft.Json;
 
 namespace Microsoft.PythonTools.Analysis.LanguageServer {
+    [JsonObject(MemberSerialization.OptOut)]
```
zooba: Isn't this the default?

Python/Product/Analysis/PythonAnalyzer.cs
```diff
@@ -784,9 +786,9 @@ class LazyModuleEnumerator {
             }
         }
 
-        #endregion
+#endregion
```
zooba: Why does automatic formatting have to be so difficult... can we keep these indented? (Just the `#region` tags)

Python/Product/NetStandard.settings
```diff
@@ -0,0 +1,48 @@
...
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <DebugType>portable</DebugType>
+    <NetStandard>true</NetStandard>
+    <DefineConstants>NETSTANDARD;$(DefineConstants)</DefineConstants>
+  </PropertyGroup>
+
+  <!-- Signing -->
+  <PropertyGroup>
+    <MicroBuild_DoNotStrongNameSign>true</MicroBuild_DoNotStrongNameSign>
```
zooba: Why do we need this if we're (a) strong name signing and (b) only strong name signing when building for release?

Python/Product/VSCode/.gitignore
```diff
@@ -0,0 +1,2 @@
+*/bin/
+*/obj/
```
zooba: Rather than ignore these, can we force the output directories to be under the paths set up by the common targets? (or at least based on the same logic)

Python/Product/VSCode/AnalysisVsc/Commands/Controller.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.DsTools.Core.Services;
```
zooba: Can we rename this namespace (wherever it came from)?

Python/Product/VSCode/AnalysisVsc/obj/AnalysisVsc.csproj.nuget.g.props
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Condition=" '$(ExcludeRestorePackageImports)' != 'true' ">
+    <RestoreSuccess Condition=" '$(RestoreSuccess)' == '' ">True</RestoreSuccess>
+    <RestoreTool Condition=" '$(RestoreTool)' == '' ">NuGet</RestoreTool>
+    <ProjectAssetsFile Condition=" '$(ProjectAssetsFile)' == '' ">F:\PTVS\Python\Product\VSCode\AnalysisVsc\obj\project.assets.json</ProjectAssetsFile>
+    <NuGetPackageRoot Condition=" '$(NuGetPackageRoot)' == '' ">$(UserProfile)\.nuget\packages\</NuGetPackageRoot>
+    <NuGetPackageFolders Condition=" '$(NuGetPackageFolders)' == '' ">C:\Users\mikhaila.REDMOND\.nuget\packages\;C:\Program Files\dotnet\sdk\NuGetFallbackFolder</NuGetPackageFolders>
```
zooba: You probably didn't mean to include this /obj directory

Python/Product/Analysis/LanguageServer/Structures.cs
MikhailArkhipov: They are probably not needed

Python/Product/NetStandard.settings
MikhailArkhipov: It is for .NET Core only. We need to tests this but we had to add in RTVS since MB auto-signs in .NET Core projects by default

Python/Product/VSCode/.gitignore
MikhailArkhipov: It is when building from VSC C# extension (i.e. `dotnet build`). I'll see if I can make it drop output properly.

Python/Product/VSCode/AnalysisVsc/Commands/Controller.cs
MikhailArkhipov: It used to be Microsoft.R.Common.Core. Basically various useful extensions etc. How do we want it named?

Python/Product/VSCode/AnalysisVsc/obj/AnalysisVsc.csproj.nuget.g.props
MikhailArkhipov: Yes, removed. Accidental.

PR #3815, opened 2018-03-03
Fixes #3786 OOM adding text to the repl

### LINKED ISSUE #3786: OOM adding text to the repl
We've seen `System.OutOfMemoryException` reported from `Microsoft.PythonTools.Repl.PythonCommonInteractiveEvaluator.AppendTextWithEscapes`.

Internal issue 503097

zooba:
Fixes #3786 OOM adding text to the repl
Ensures our search for color codes always makes forward progress.
Improves handling of debug assertions when in a context that doesn't handle dialogs well.

### DIFF
--- a/Common/Product/SharedProject/UIThread.cs
+++ b/Common/Product/SharedProject/UIThread.cs
@@ -15,11 +15,13 @@
 // permissions and limitations under the License.
 
 using System;
+using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


huguesv:
I've looked at this for a while and I still haven't figured out what input would work with the new loop but not work with the old loop. Any clues?

PR #3844, opened 2018-03-08
Make analyzer selection completely async

zooba:
* requires an async call to potentially create a new analyzer - this should handle a lot of deadlocks
* enable multiple shared analyzers - this will eventually let us better handle standalone files

Still going to keep testing this for a while, but want to wrap up the first step on its own before getting into the next phase.


Python/Product/Analysis/Projects/PythonProject.cs
```diff
@@ -58,9 +55,14 @@ public abstract class PythonProject {
         /// <summary>
         /// Gets the current analyzer for the project, or null if no analyzer is available.
         /// </summary>
-        public abstract ProjectAnalyzer Analyzer {
-            get;
-        }
+        [Obsolete("Use the async version if possible")]
+        public abstract ProjectAnalyzer Analyzer { get; }
```
AlexanderSher: Is this a public API?

zooba: Yes, and formally public as well (unlike some of the "accidentally" public ones). The whole point of the abstract base class is to make it available.

Python/Product/PythonTools/PythonTools/Commands/ImportCoverageCommand.cs
```diff
@@ -37,7 +40,7 @@ internal sealed class ImportCoverageCommand : Command {
             
         }
 
-        public override void DoCommand(object sender, EventArgs args) {
+        public override async void DoCommand(object sender, EventArgs args) {
```
AlexanderSher: Do we have async commands?

Python/Product/PythonTools/PythonTools/Commands/ImportCoverageCommand.cs
```diff
@@ -59,19 +62,29 @@ internal sealed class ImportCoverageCommand : Command {
...
             if (file != null) {
-                var outFilename = Path.ChangeExtension(file, ".coveragexml");
-
                 try {
-                    ConvertCoveragePy(file, outFilename, version);
-                } catch (IOException ioex) {
-                    MessageBox.Show(String.Format(Strings.FailedToConvertCoverageFile, ioex.Message));
+                    await DoConvert(file, version);
+                } catch (Exception ex) {
+                    Debug.Fail(ex.ToUnhandledExceptionMessage(GetType()));
```
AlexanderSher: What's the point of swallowing it release?

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -739,4 +744,157 @@ private Lazy<Tokenizer> GetTokenizerLazy()
...
+                    viewer.DifferenceBuffer.LeftBuffer.TryGetAnalysisEntry();
+                if (entry != null) {
+                    return entry;
+                }
+            }
+
+            return null;
+        }
+
+        public static async Task<ProjectAnalyzer> FindAnalyzerAsync(this IServiceProvider site, ITextView view) {
```
AlexanderSher: Do we need `CancellationToken` here?

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
```diff
@@ -739,4 +744,157 @@ private Lazy<Tokenizer> GetTokenizerLazy()
...
+            ProjectAnalyzer analyzer;
+
+            var bi = view.TextBuffer.TryGetInfo();
+            if (bi != null && (analyzer = await FindAnalyzerAsync(site, bi)) != null) {
+                return analyzer;
+            }
+
+            var diffViewer = site.GetComponentModel().GetService<IWpfDifferenceViewerFactoryService>();
```
AlexanderSher: Should we check for UI thread?

Python/Product/PythonTools/PythonTools/Repl/PythonCommonInteractiveEvaluator.cs
```diff
@@ -142,45 +142,45 @@ internal abstract class PythonCommonInteractiveEvaluator :
...
+            }
 
-                        return a;
-                    }, CancellationToken.None);
-                }
-                return _analyzer;
+            if (factory == null) {
+                _analyzer = await _serviceProvider.GetPythonToolsService().GetSharedAnalyzerAsync();
+            } else {
+                _analyzer = await _serviceProvider.GetUIThread().InvokeTask(async () => {
```
AlexanderSher: Is it possible that this code is called twice before `_analyzer` is assigned?

Python/Product/PythonTools/PythonTools/Commands/ImportCoverageCommand.cs
zooba: Not yet. Can we implement them?

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
zooba: There's exactly one case where this method won't be synchronous, and there's _a lot_ of plumbing required to get the cancellation somewhere useful. I don't think it's worth the churn.

Python/Product/PythonTools/PythonTools/Repl/PythonCommonInteractiveEvaluator.cs
zooba: Right now this function is not called... guess I should fix that first :)

Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
zooba: Because this is accessing via ITextView, I think it's okay to require the function be called from the UI thread. I'll add an assert.

huguesv:
I'm testing this. I switched my project from default environment (Python 3.6) to Python 2.7. I got this assert:
```
---------------------------
---------------------------
Changing from wrong analyzer



   at Microsoft.PythonTools.Navigation.PythonLibraryManager.AnalysisCompleteHandler.Project_ProjectAnalyzerChanging(Object sender, AnalyzerChangingEventArgs e) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Navigation\PythonLibraryManager.cs:line 145

   at Microsoft.PythonTools.Project.PythonProjectNode.<ReanalyzeProjectHelper>d__146.MoveNext() in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Project\PythonProjectNode.cs:line 1415

   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.InvokeMoveNext(Object stateMachine)

   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.Run()

   at ......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------
```

zooba:
I can't get the assert, but since you can now race VS by hand when changing analyzer (whereas before it simply froze), we have some new UI issues to deal with. Let's hold this for the next preview and revisit for the subsequent one.

huguesv:
I've updated to the latest code with this PR. I was able to easily get the "Buffer content mismatch", so we know that's not completely fixed.

I opened an existing project where several files were opened in last session. I closed all windows, and reopened the one that was active after project load. I hit delete on some blank lines near the top of the file and got the assert.

zooba:
@huguesv Try again - just fixed some event handler issues :)

Python/Product/PythonTools/PythonTools/Project/PythonFileNode.cs
```diff
@@ -190,7 +190,7 @@ internal PythonFileNode(CommonProjectNode root, ProjectElement e)
         }
 
         private VsProjectAnalyzer GetAnalyzer() {
-            return ((PythonProjectNode)ProjectMgr).GetAnalyzer();
+            return ((PythonProjectNode)ProjectMgr).TryGetAnalyzer();
```
huguesv: Now that `GetAnalyzer()` may return `null`, there's going to be an issue with the implementation of `GetAnalysisEntry` right below. It's not expecting to get a `null` analyzer. Are there any callers left for `GetAnalysisEntry` or did you remove them all?

huguesv: i would rename this function to `TryGetAnalyzer`.

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -1464,18 +1438,18 @@ private static T FindImmediateChild<T>(HierarchyNode parent)
...
                 log?.WriteLine($"Project includes files:{Environment.NewLine}    {string.Join(Environment.NewLine + "    ", files)}");
 
-                if (defAnalyzer != null) {
-                    foreach (var f in files) {
-                        var entry = defAnalyzer.GetAnalysisEntryFromPath(f);
-                        if (entry != null) {
-                            log?.WriteLine($"Unloading {f} from default analyzer");
-                            await defAnalyzer.UnloadFileAsync(entry);
-                            foreach (var b in entry.TryGetBufferParser().AllBuffers) {
+                foreach (var existing in _services.Python.GetActiveSharedAnalyzers().Select(kv => kv.Value).Where(v => !v.IsDisposed)) {
```
huguesv: Do you need to try/catch ObjectDisposedException? Can it become disposed while this loop is executing?

Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
```diff
@@ -166,7 +166,7 @@ internal class PythonDebugReplEvaluator :
                     var tid = _serviceProvider.GetDTE().Debugger.CurrentThread.ID;
                     var result = CustomDebugAdapterProtocolExtension.EvaluateReplRequest(text, tid);
                     CurrentWindow.Write(result);
-                }
+            }
```
huguesv: indentation

Python/Product/PythonTools/PythonTools/Commands/ImportCoverageCommand.cs
huguesv: I filed https://github.com/Microsoft/PTVS/issues/3942 for this.
For now you could just extract it in a Task returning method and call DoNotWait(). At least that way @AlexanderSher 's special `DoNotWait` injection for detecting non terminating task will work.

Issue #3687, opened 2018-02-01
Python Django cannot run anymore


Auvite:
Trying to debug before update worked django project and now vs cannot run it.
Gives following traceback.

```
Traceback (most recent call last):
  File "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\Common7\IDE\Extensions\Microsoft\Python\Core\ptvsd_launcher.py", line 34, in <module>
    import ptvsd.debugger as vspd
  File "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\Common7\IDE\Extensions\Microsoft\Python\Core\ptvsd\debugger.py", line 29, in <module>
    import ctypes
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\Lib\ctypes\__init__.py", line 7, in <module>
    from _ctypes import Union, Structure, Array
ImportError: DLL load failed: The specified procedure could not be found.

Internal error detected. Please copy the above traceback and report at
https://go.microsoft.com/fwlink/?LinkId=293415

Press Enter to close. . .
```

Auvite:
Okay so the vs disappeared the virtual environment for some reason. I added new virtual environment and got it back to situation as it was. 

Now i have another problem. 
The project runs fine with cmd->manage.py runserver
But with vs debugger i get just errors i don't understand why.<-This was the reason that i updated visual studio.
Django version 2
Python 3.6


The traceback log
```
Traceback (most recent call last):
  File "C:\Users\<user>\source\repos\customer\blank\blank\views.py", line 12, in <module>
    from blank.tokens import account_activation_token
  File "C:\Users\<user>\source\repos\customer\blank\blank\tokens.py", line 1, in <module>
    from django.contrib.auth.tokens import PasswordResetTokenGenerator
  File "C:\Users\<user>\source\repos\customer\blank\customer_env\lib\site-packages\django\contrib\auth\tokens.py", line 8, in <module>
    class PasswordResetTokenGenerator:
  File "C:\Users\<user>\source\repos\customer\blank\customer_env\lib\site-packages\django\contrib\auth\tokens.py", line 14, in PasswordResetTokenGenerator
    secret = settings.SECRET_KEY
  File "C:\Users\<user>\source\repos\customer\blank\customer_env\lib\site-packages\django\conf\__init__.py", line 56, in __getattr__
    self._setup(name)
[...]

PR #3856, opened 2018-03-09
Show a message box with details in case of ptvsd ImportError

### LINKED ISSUE #3852: Debugging without ptvsd installed error message is confusing
1. For a windows app, they see this message box (nothing in output window like the message suggest)
```
---------------------------
Microsoft Visual Studio
---------------------------
Failed to launch debug adapter.  Additional information may be available in the output window.



Cannot process request because the process (8080) has exited.
---------------------------
OK   
---------------------------
```

2. For a console app, they see this in the console but the console automatically disappears after 5 secs
```
  File "C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\Common7\IDE\Extensions\Microsoft\Python\Core\ptvsd_launcher.py", line 76, in <module>
    import ptvsd
ModuleNotFoundError: No module named 'ptvsd'

Internal error detected. Please copy the above traceback and report at
https://go.microsoft.com/fwlink/?LinkId=293415

Press Enter to close. . .

```

1) is pretty bad because they get no clue as to what is wrong.
2) is okay, but they only get 5 secs to read it (I guess they can F5 again until they've read it all)

We could give them some more info in both cases (please pip install ptvsd --pre in your Python environment in order to use the experimental debugger)

Later when it's bundled this may not be worth keeping, so not sure if it's 'too late' to do anything now.

karthiknadig:
Fixes #3852

### DIFF
--- a/Python/Product/Debugger/DebugAdapterProcess.cs
+++ b/Python/Product/Debugger/DebugAdapterProcess.cs
@@ -25,6 +25,7 @@ using System.Net.Sockets;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using System.Web;
+using System.Windows.Forms;
 using Microsoft.PythonTools.Debugger.DebugEngine;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.VisualStudio.Debugger.DebugAdapterHost.Interfaces;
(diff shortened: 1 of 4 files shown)


huguesv:
This code is common to both regular and experimental debugger, right? `pip install ptvsd --pre` won't help them if this ever happens with non-experimental debugger (that can happen in broken virtual environment like https://github.com/Microsoft/PTVS/issues/3687).

zooba:
Really not thrilled about requiring `ctypes` right at the start, either. We should import that as late as possible and have a fallback.

Better yet, some back-channel to VS would be better and then display the message box from there. Can we get to the exit code?

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -178,6 +179,24 @@ sealed class DebugAdapterProcess :
...
+                // 126 : ERROR_MOD_NOT_FOUND
+                // This error code is returned only for the experimental debugger. MessageBox must be
+                // bound to the VS Main window otherwise it can be hidden behind the main window and the 
+                // user may not see it.
+                MessageBox.Show(
+                    new VSWin32Window(Process.GetCurrentProcess().MainWindowHandle),
+                    Strings.ImportPtvsdFailedMessage,
+                    Strings.ImportPtvsdFailedTitle,
+                    MessageBoxButtons.OK,
+                    MessageBoxIcon.Error);
```
zooba: I'm okay with using a plain message box right now, as we intend to remove the need for it fairly soon (though we'll leave it here in case someone runs into a corrupt install).

A permanent error like this would be better as a TaskDialog that can trigger the action (in this case, install `--pre ptvsd` into the current environment).

Python/Product/Debugger/DebugAdapterProcess.cs
```diff
@@ -178,6 +179,24 @@ sealed class DebugAdapterProcess :
             if (_stream != null) {
                 _stream.Dispose();
             }
+
+            if (_process.ExitCode == 126) {
```
huguesv: What if debugging started successfully but the user's program exited with code 126? Can you do this only if debuggee didn't connect back?

karthiknadig: sure i can restrict it.

PR #3864, opened 2018-03-12
Fixes #3863 Crash getting list of packages with invalid environment

### LINKED ISSUE #3863: Crash getting list of packages with invalid environment
Sometimes we may crash when unable to launch a detected version of Python in order to list packages.

We should be handling all exceptions and failing quietly.

zooba:
Fixes #3863 Crash getting list of packages with invalid environment
Handles all exceptions when a process fails to start


huguesv:
What made the environment invalid? Missing executable, or executable but it can't load due to missing dlls, etc?

It looks easy to add a test for missing executable to `ProcessOutputTests`.

zooba:
`File.Exists()` is checked first, so if the file is simply inaccessible we don't get this far. I triggered it by removing the "Read+Execute" permission but allowing "Read". I believe missing DLLs will also cause it, so that may be an easier way.

zooba:
@huguesv Force me to do a test when I forward port to master.

PR #3870, opened 2018-03-13
Forward port fixes from 15.6

zooba:


Python/Product/Analysis/Infrastructure/StringExtensions.cs
```diff
@@ -103,20 +112,30 @@ static class StringExtensions {
...
+                // We do not want to return unquoted though
+                arg = arg.Substring(1, arg.Length - 2);
+            } else if (arg.IndexOf(' ') < 0 && !FindUnescapedChar(arg, '"').Any()) {
                 // Does not need quoting
                 return arg;
             }
 
-            if (arg.Last() == '\\') {
+            if (arg.Length > 1 && arg[arg.Length - 1] == '\\' &&
+                (arg.Length < 2 || arg[arg.Length - 2] != '\\')) {
```
huguesv: The check for `arg.Length < 2` can be removed, it will always be false

PR #3871, opened 2018-03-13
Fixes #3841 15.6 broke mixed mode debugging.

### LINKED ISSUE #3841: 15.6 broke mixed mode debugging.
I'm using hosted mixed mode environment (ie my hosting executable is not python.exe) when starting the process with F5 i now get a popup saying. 

```
Unable to start debugging. Check your debugger settings by opening 
project properties and navigating to 
'Configuration Properties-->Debugging'
```

the issue is, there's not a whole lot of settings there, i tried changing the target path from $(TargetPath) to the full path to my binary, but that didn't change anything.

proppages

Repro case: (this is going to be a 'tiny' bit painful, we tried making building as easy as possible, but blender is a big project)
1) Build blender from source with vs2017.6, there's a guide available [here ](https://wiki.blender.org/index.php/Dev:Doc/Building_Blender/Windows) to cut down on the build time, you could generate project files using `make.bat lite 2017 x64 nobuild` which cuts most of the optional components out of blender. 
2) Apply [D3039](https://developer.blender.org/D3039)
3) Change the debugger to 'Python/Native Debugging'
4) hit f5.

edit: Just to clarify, this scenario used to work on 15.5 , attaching to a running instance still works in 15.6, but it be real convenient if launching with f5 worked again.

zooba:
Fixes #3841 15.6 broke mixed mode debugging
Updates version of the VCDebugLauncher package to ensure the newer build is installed.

For the forward-port we should update all VSIX versions to 15.7

### DIFF
--- a/Python/Product/VCDebugLauncher/source.extension.vsixmanifest
+++ b/Python/Product/VCDebugLauncher/source.extension.vsixmanifest
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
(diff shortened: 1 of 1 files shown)


zooba:
Note that this fix does not require modifying the VS setup integration (because we fall within the `[3.0, 4.0)` range already specified). The correct fix will require a change there.

PR #3900, opened 2018-03-16
Fixes #3536 VS code debugger installation

### LINKED ISSUE #3536: VS code debugger installation
To be determined - amounts to F5 should 'just work'.

zooba:
Fixes #3536 VS code debugger installation
Embeds the latest release of ptvsd into the bundle.

This needs to be changed at some point to lock to a specific release, but as the only available release right now is not suitable for embedding.

### DIFF
--- a/Python/Product/PythonTools/ptvsd_launcher.py
+++ b/Python/Product/PythonTools/ptvsd_launcher.py
@@ -71,6 +71,9 @@ sys.path[0] = ''
 # Load the debugger package
 try:
     if bundled_ptvsd:
+        ptvs_lib_path = os.path.dirname(__file__)
+        sys.path.insert(0, ptvs_lib_path)
+    else:
         ptvs_lib_path = os.path.join(os.path.dirname(__file__), 'Packages')
         sys.path.append(ptvs_lib_path)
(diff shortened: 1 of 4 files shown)


Python/Product/PythonTools/ptvsd_launcher.py
```diff
@@ -71,8 +71,8 @@
 # Load the debugger package
 try:
     if bundled_ptvsd:
-        ptvs_lib_path = os.path.dirname(__file__)
-        sys.path.insert(0, ptvs_lib_path)
+        ptvs_lib_path = os.path.join(os.path.dirname(__file__), 'Packages')
+        sys.path.append(ptvs_lib_path)
```
huguesv: Is it still possible to use the old debugger?

zooba: Um... yeah... good point. I'll add another option to the launch script for that so we can still do it.

zooba:
@huguesv I've corrected the launch option so that we use the new bundled ptvsd as a fallback instead of a preference, but debugging is busted on my machine right now so I haven't been able to check it.

PR #3904, opened 2018-03-18
Improves scraping

zooba:
Improves scraping
Fixes buffer not resetting last parse correctly.

### DIFF
--- a/Python/Product/Analysis/Parsing/Parser.cs
+++ b/Python/Product/Analysis/Parsing/Parser.cs
@@ -957,7 +957,7 @@ namespace Microsoft.PythonTools.Parsing {
             Expression ret = ParseTestListAsExpr();
             bool hasAnnotation = false;
 
-            if (PeekToken(TokenKind.Colon) && (_stubFile || _langVersion >= PythonLanguageVersion.V36)) {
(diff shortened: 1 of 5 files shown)


Python/Product/Analysis/Parsing/Parser.cs
```diff
@@ -957,7 +957,7 @@ struct Name {
             Expression ret = ParseTestListAsExpr();
             bool hasAnnotation = false;
 
-            if (PeekToken(TokenKind.Colon) && (_stubFile || _langVersion >= PythonLanguageVersion.V36)) {
+            if (PeekToken(TokenKind.Colon) && (_stubFile || _langVersion.Is3x())) {
```
huguesv: Intentional? Was >= 3.6, Now >= 3.0

zooba: Nope... guess I misread it.

zooba:
Some tests were failing, but not horribly. These opportunities were found by manually comparing scraped functions (mostly from numpy and scipy) with their docstrings (which contained invalid Python code that looks like a function definition) - we were seeing the functions, but now we also see their parameters.

PR #3909, opened 2018-03-19
Analysis fixes

### LINKED ISSUE #3733: Intellisense squiggles are some times off
In this code there should not be any squiggles but they appear any way. They also appear in wrong position:
image

```python
def F(n):
  a = {0:1, 1:1}
  if n < 0:
    raise ValueError('Value is invalid: ' + str(n))
  if n in a:
    return a[n]
  x = F(n - 1) + F(n - 2)
  a[n] = x
  return x

for x in range(1, 300):
  print(F(x))

print('done')
```

zooba:
Fixes a range of analysis crashes and inconsistencies. These were mostly found by running with `FULL_VALIDATION` enabled against large open-source libraries.
Also includes updates to the analysis memory tester tool to make it easier to automatically load an entire package.
Also fixes #3733 Intellisense squiggles are sometimes off

I'm still finding more issues (up to libraries starting with "C" now :) ), but we can merge parts whenever we like.


zooba:
Just pushed two more changes that should fix (or significantly mitigate) some of the more annoying analysis-getting-out-of-sync issues.

@huguesv - probably best if you bash on this for a bit, since these are the ones that are hard to prove.

huguesv:
Okay, I'll test this.

huguesv:
In one session, I plenty of "Buffer content mismatch" asserts, it required a project reload (file reload wasn't enough).

In another session, I stopped getting completions altogether, nothing in the logs that was interesting, the analyzer process was at 0% cpu. I put a breakpoint in VsProjectAnalyzer.SendRequestAsync and it was never getting called. When I loaded another project it was immediately hit.

In another session, I got a bunch of these asserts:
```
---------------------------
---------------------------
Expected 14 lines; got 13 lines



   at Microsoft.PythonTools.Intellisense.LocationTracker.ValidateSnapshotLines(ITextSnapshot snapshot) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\LocationTracker.cs:line 293

   at Microsoft.PythonTools.Intellisense.LocationTracker.Translate(SourceSpan span, Int32 fromVersion, ITextSnapshot toSnapshot) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\LocationTracker.cs:line 350

   at Microsoft.PythonTools.PythonAnalysisClassifier.GetClassificationSpans(SnapshotSpan span) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\PythonAnalysisClassifier.cs:line 194

   at Microsoft.VisualStudio.Text.Classification.Implementation.ClassifierTagger.<GetTags>d__5.MoveNext()

   at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1.<GetTagsForBuffer>d__39.MoveNext()

   at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1.<InternalGetTags>d__40.MoveNext()

   at Microsoft.VisualStudio.Text.Classification.Implementation.ClassifierAggregator.InternalGetClassificationSpans(SnapshotSpan span, IEnumerable`1 tags)

   at Microsoft.VisualStudio.Text.Classification.Implementation.ClassifierAggregator.GetClassificationSpans(SnapshotSpan span)

   at Microsoft.VisualStudio.Text.Formatting.Implementation.NormalizedSpanGenerator.CreateClassifiedRuns(SnapshotSpan span, IAccurateClassifier classifier, IClassificationFormatMap classificationFormatMap, Nullable`1 cancel)

   at Microsoft.VisualStudio.Text.Formatting.Implementation.NormalizedSpanGenerator.Create(ITextAndAdornmentCollection elements, ITextSnapshot sourceSnapshot, IAccurateClassifier classifier, IClassificationFormatMap classificationFormatMap, Nullable`1 cancel)

   at Microsoft.VisualStudio.Text.Formatting.Implementation.FormattedLineSource.FormatLineInVisualBufferIfChanged(ITextSnapshotLine visualLine, IList`1 oldLines, Nullable`1 cancel)

   at Microsoft.VisualStudio.Text.Editor.Implementation.WpfTextView.FormatSnapshotLine(ITextSnapshotLine visualLine, Nullable`1 cancel)
```

which triggered an activity log error dialog with this in the activity log:
```
  <entry>
    <record>1443</record>
    <time>2018/03/20 01:04:43.740</time>
    <type>Error</type>
    <source>Editor or Editor Extension</source>
    <description>System.InvalidOperationException: IWpfTextView.TextViewLines is invalid.&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Editor.Implementation.WpfTextView.Microsoft.VisualStudio.Text.Editor.IWpfTextView.get_TextViewLines()&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Editor.Implementation.Outlining.OutliningMarginControls`2.Invalidate(SnapshotSpan span)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Editor.Implementation.Outlining.OutliningMargin.&lt;&gt;c__DisplayClass7_0.&lt;.ctor&gt;b__2(Object sender, RegionsChangedEventArgs args)&#x000D;&#x000A;   at System.EventHandler`1.Invoke(Object sender, TEventArgs e)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Outlining.OutliningManager.UpdateAfterChange(NormalizedSnapshotSpanCollection changedSpans)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Outlining.OutliningManager.OutliningRegionTagsChanged(Object sender, BatchedTagsChangedEventArgs e)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Utilities.GuardedOperations.RaiseEvent[TArgs](Object sender, EventHandler`1 eventHandlers, TArgs args)&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)</description>
  </entry>
  <entry>
    <record>1444</record>
    <time>2018/03/20 01:04:43.835</time>
    <type>Error</type>
    <source>Editor or Editor Extension</source>
    <description>System.InvalidOperationException: IWpfTextView has not completed its layout.&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Editor.Implementation.WpfTextView.PerformLayout(ITextSnapshot newSnapshot, ITextSnapshot newVisualSnapshot, SnapshotPoint anchorPosition, Double verticalDistance, ViewRelativePosition relativeTo, Double effectiveViewportWidth, Double effectiveViewportHeight, Boolean preserveViewportTop, Nullable`1 cancel)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Editor.Implementation.WpfTextView.PerformLayout(ITextSnapshot newSnapshot, ITextSnapshot newVisualSnapshot)&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Editor.Implementation.WpfTextView.&lt;QueueLayout&gt;b__260_0()&#x000D;&#x000A;   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)&#x000D;&#x000A;   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)&#x000D;&#x000A;   at System.Windows.Threading.DispatcherOperation.InvokeImpl()&#x000D;&#x000A;   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(Object state)&#x000D;&#x000A;   at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(Object obj)&#x000D;&#x000A;   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)&#x000D;&#x000A;   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)&#x000D;&#x000A;   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)&#x000D;&#x000A;   at MS.Internal.CulturePreservingExecutionContext.Run(CulturePreservingExecutionContext executionContext, ContextCallback callback, Object state)&#x000D;&#x000A;   at System.Windows.Threading.DispatcherOperation.Invoke()&#x000D;&#x000A;   at System.Windows.Threading.Dispatcher.ProcessQueue()&#x000D;&#x000A;   at System.Windows.Threading.Dispatcher.WndProcHook(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean&amp; handled)&#x000D;&#x000A;   at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean&amp; handled)&#x000D;&#x000A;   at MS.Win32.HwndSubclass.DispatcherCallbackOperation(Object o)&#x000D;&#x000A;   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)&#x000D;&#x000A;   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)&#x000D;&#x000A;   at System.Windows.Threading.Dispatcher.LegacyInvokeImpl(DispatcherPriority priority, TimeSpan timeout, Delegate method, Object args, Int32 numArgs)&#x000D;&#x000A;   at MS.Win32.HwndSubclass.SubclassWndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam)&#x000D;&#x000A;   at Microsoft.Win32.SafeNativeMethods.MessageBoxSystem(IntPtr hWnd, String text, String caption, Int32 type)&#x000D;&#x000A;   at Microsoft.Win32.SafeNativeMethods.MessageBox(IntPtr hWnd, String text, String caption, Int32 type)&#x000D;&#x000A;   at System.Diagnostics.AssertWrapper.ShowMessageBoxAssert(String stackTrace, String message, String detailMessage)&#x000D;&#x000A;   at System.Diagnostics.DefaultTraceListener.Fail(String message, String detailMessage)&#x000D;&#x000A;   at System.Diagnostics.DefaultTraceListener.Fail(String message)&#x000D;&#x000A;   at System.Diagnostics.TraceInternal.Fail(String message)&#x000D;&#x000A;   at System.Diagnostics.Debug.Fail(String message)&#x000D;&#x000A;   at Microsoft.PythonTools.Intellisense.LocationTracker.ValidateSnapshotLines(ITextSnapshot snapshot) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\LocationTracker.cs:line 293&#x000D;&#x000A;   at Microsoft.PythonTools.Intellisense.LocationTracker.Translate(SourceLocation loc, ITextSnapshot fromSnapshot, Int32 toVersion) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Intellisense\LocationTracker.cs:line 308&#x000D;&#x000A;   at Microsoft.PythonTools.PythonAnalysisClassifier.GetClassificationSpans(SnapshotSpan span) in C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\PythonAnalysisClassifier.cs:line 174&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Classification.Implementation.ClassifierTagger.&lt;GetTags&gt;d__5.MoveNext()&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1.&lt;GetTagsForBuffer&gt;d__39.MoveNext()&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)</description>
  </entry>
```

Fortunately, I haven't hit any of the newly added asserts in Connection.cs. I also haven't seen colorization get out of sync.

Python/Product/Analysis/Values/ClassInfo.cs
```diff
@@ -577,36 +577,58 @@ internal class ClassInfo : AnalysisValue, IReferenceableContainer, IHasRichDescr
...
+
+            if (ci1 == null && ci2 != null) {
+                return true;
+            } else if (ci1 != null && ci2 == null) {
+                return false;
+            } else if (ci1 != null && ci2 != null) {
+                return ci1.ClassDefinition.StartIndex > ci2.ClassDefinition.StartIndex;
+            }
+
+            return string.CompareOrdinal(ns1.Name, ns2.Name) > 0;
```
AlexanderSher: Why `>`?

zooba:
@AlexanderSher I can't get to your "why" comment anymore, but the answer is to basically ensure a stable comparison to minimize the likelihood of a changed result when the comparison is performed backwards. The more often the `UnionMergeTypes` functions return the same object, the better, and the MRO can change over time as the analysis changes but the class location and name are static.

Python/Product/Ipc.Json/Connection.cs
```diff
@@ -556,30 +556,67 @@ CancellationToken cancel
...
+                            break;
+                        }
+                    } finally {
+                        _writeLock.Release();
+                    }
+
+                    --retries;
+                    if (retries > 100) {
+                        // Fast looping
+                        await Task.Yield();
```
AlexanderSher: I don't like this approach. We have no explicit message pool, so with multiple events sent concurrently, we can end up sending these messages forever. Also, since there is no guarantee that this code is executed in background thread, this loop can run on UI thread.

On the other hand, `_sentSeq` doesn't change in normal situation outside of the lock, so it is either correct in the first loop iteration or it will never be correct for that message again unless there is another message with incorrect `seq`.

Why can't we just assing `ProtocolMessage.seq` in the `SendMessage` after `_writeLock` is acquired?

Python/Product/PythonTools/PythonTools/Intellisense/LocationTracker.cs
```diff
@@ -33,7 +33,7 @@ namespace Microsoft.PythonTools.Intellisense {
     /// version of the buffer.
     /// </summary>
     internal class LocationTracker {
-        private ITextSnapshot _snapshot;
+        private readonly Queue<ITextSnapshot> _snapshots;
```
AlexanderSher: What is this change for? `_snapshots` is never iterated.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -1018,7 +1018,7 @@ public sealed class Server : ServerBase, IDisposable {
             }
         }
 
-        private async void EnqueueItem(IDocument doc, AnalysisPriority priority = AnalysisPriority.Normal, bool enqueueForAnalysis = true) {
+        private async Task EnqueueItem(IDocument doc, AnalysisPriority priority = AnalysisPriority.Normal, bool enqueueForAnalysis = true) {
```
AlexanderSher: 👍

Python/Product/Ipc.Json/Connection.cs
zooba: Because this starts running on the thread that's sending the message, we can have multiple interleaving contexts in this loop at once. An alternative would be to always create a new thread for each connection and serialize messages through that (i.e. use a queue that the worker thread pulls from), and that may still turn out to be necessary. But this approach certainly works for the core case (multiple threads hit the lock, then get released in random order).

The messages themselves need to be in order when posted from the same thread, and by assigning seq synchronously we can ensure that. If we assign it later, then the messages are still out of order, but we've lost the only useful piece of information that can prove it.

Python/Product/PythonTools/PythonTools/Intellisense/LocationTracker.cs
zooba: It's a buffer to extend the lifetime of our cache. We limit it to five entries, and as the items leave the queue, their cache entries are cleaned up.

Previously, the queue was effectively one item long, and we were cleaning up cache entries too quickly. Taken to the other extreme, we would never clean up cache entries. This is a tunable compromise. A true solution would be to lock the cache before sending a request to the analyzer, and then unlock it once we've finished dealing with the request, but I didn't want to touch that much code.

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Note that it's still fire-and-forget, as we very rarely want to await the entire operation (perhaps in tests...), but Mikhail confirmed that without Task here we weren't getting the first step performed synchronously, which was certainly the intent.

AlexanderSher: It's still fire-and-forget in product, but with injection into `DoNotWait` tests can wait for those tasks to compete.

zooba:
@AlexanderSher Just waiting on your verdict here. This is a net improvement over the current code, but I'm going to have to spend some serious time tracking down a few weird interactions between multiple buffers into the same file. There's no point holding up this PR because of that (the issues existed before).

Python/Product/Ipc.Json/Connection.cs
AlexanderSher: Unless we create tasks in a single-threaded synchronization context, we have no guarantee that two code pieces, separated by `await` operator, will be executed in the same thread. Increasing `Connection._seq` twice inside asynchronous code flow doesn't guarantee that numbers will be sequential if there is another asynchronous code flow.

If we want to ensure that several events are sent without interruption, we should either provide access to some shared lock that can be acquired by event-raising, or collect events in a batch before sending. Current approach only ensures that events are sent in the same order they are created, but it doesn't guarantee that they are created in the right order.

This code may create problems when queue of events is too long. For instance, if there are 3 events that come to the lock in order `1,3,2`, third event will be waiting on the lock while the first is processed. If during this waiting 10 more events arrive, they all will be put on the tail of the lock and after first event is processed third one will end up in the tail of this queue. `SemaphoreSlim`, at least in its current implementation, execute continuations in the same order they are scheduled, but `Task.Yield` doesn't preserve any order, so on the second loop of every element in the queue they all will be shuffled, even if originally they were in the right order. As a result, we get a bubble sort with an expense of creating 2 tasks for every iteration of every message that isn't in the right place.

It may completely fail if event with expected `seq` never arrives. To avoid this in case of cancellation, `CancellationToken` isn't passed into `WaitAsync` anymore, which means that task can't be cancelled unless it is actually processed. But if, for instance, `JsonConvert.SerializeObject` can fail with exception, `_sentSeq` will become unsynced with `_seq`.

zooba: "In the same thread" doesn't matter here - we have a lock - and numbers only have to be increasing, not sequential (that is, if thread A sends message M1 and M2 and B sends M3, as long as M2 comes after M1 it doesn't matter if M3 is in the middle).

`Task.Yield` is only going to have one continuation, so I don't see how things get shuffled. The semaphore is where they all collect, and when it becomes available then one will be released (in order of arrival, as you said). If a continuation is executed, doesn't match the expected seq, and _does not_ yield, it will release and immediately reacquire the lock, repeating until it exhausts its retries (validated by testing).

Good point about not handling the exceptional case. I'll handle that.

AlexanderSher: > if thread A sends message M1 and M2 and B sends M3, as long as M2 comes after M1 it doesn't matter if M3 is in the middle

Ok, then I don't understand how is it possible that M2 comes before M1. There are only 4 methods that call `SendMessage`:  `SendRequestAsync`, `SendEventAsync`, `SendResponseAsync` and `WriteError`. For all of them, code flow piece between incrementing the `_seq` and calling `ConfigureAwait(false)` on a task returned from the `_writeLock` is 100% synchronous. And `SemaphoreSlim` is FIFO, it will preserve the order.

So as far as I understand it, current approach doesn't fix that M2 always comes after M1, cause it is happening already, it enforces that M3 comes after M2, while we actually doesn't care about it.

> so I don't see how things get shuffled. 

I've created a standalone simulation test. It sends 10 ints in order `0,2,1,3,4,5,6,7,8,9` to the following loop:

```cs
while (true) {
    var lockTask = _lock.WaitAsync();
    AddTaskToSequence(lockTask.Id, index);
    await lockTask;
    RemoveTaskFromSequence(lockTask.Id);
    try {
        if (Interlocked.CompareExchange(ref _index, index, index - 1) == index - 1) {
            TraceSequence(index);
            await Task.Delay(1000);
            return;
        }
    } finally {
        _lock.Release();
    }
    await Task.Yield();
}
```

I've also added another method that schedules short work items into thread pool to simulate real application environment.

Here is what I got:
- `0` is processing, queue is `2,1,3,4,5,6,7,8,9`
- `2` is skipped and put in the tail of the queue
- `1` is processing, queue is `3,4,5,6,7,8,9,2`
- All the queue is skipped until `2` is reached. At this point, every item in the queue (except `2`) reaches `Task.Yield()` and their continuations are executed in random order
- `2` is processing, queue is `3,5,4,8,6,7,9`
- `3` is processing, queue is `5,4,8,6,7,9`
- `5` is skipped and put in the tail of the queue
- `4` is processing, queue is `8,6,7,9,5`
- All the queue is skipped until `5` is reached.
- `5` is processing, queue is `8,6,9,7`

And so on.

zooba: That doesn't look like shuffling to me - it's just looping through the tasks until it finds the next one (insertion sort, perhaps?).

I'm not convinced that `SemaphoreSlim` is reliably FIFO, as it relies on `lock()` statements upon entry to its methods, but if it is then we should just get optimal performance in most cases. You haven't suggested an alternative approach yet (other than assuming that everything will be okay, which we've already seen doesn't work :) ), but if you have an idea I'd love to hear it.

AlexanderSher: > I'm not convinced that `SemaphoreSlim` is reliably FIFO, as it relies on `lock()` statements upon entry to its methods

Locks don't matter. If `_seq` increment for M2 can happen only after `SendRequestAsync`, `SendEventAsync`, `SendResponseAsync` or `WriteError` returns its task, then it is guarantee to happen after first synchronous part is completed. That synchronous part can wait on the lock, but it is a synchronous wait, so method won't return in this case. `SemaphoreSlim` is reliably FIFO unless someone changes that implementation. This is a simple experiment:

```cs
        private static SemaphoreSlim _ss = new SemaphoreSlim(1);
        private static int _value = -1;

        static void Main(string[] args) {
            var tasks = new Task[100];
            for (int i = 0; i < 100; i++) {
                tasks[i] = AddToSemaphore(i);
            }
            Console.WriteLine("Done scheduling");
            Task.WaitAll(tasks);
            Console.ReadKey();
        }

        private static async Task AddToSemaphore(int i) {
            await _ss.WaitAsync().ConfigureAwait(false);
            await Task.Delay(100);
            Console.Write($"{i},");
            if (Interlocked.Exchange(ref _value, i) != i - 1) {
                Console.WriteLine();
                Console.WriteLine($"Wrong order!");
            }
            _ss.Release();
        }
```

Note that if you add `await Task.Yield();` at the beginning of the `AddToSemaphore`, you'll get several errors.

> You haven't suggested an alternative approach yet (other than assuming that everything will be okay, which we've already seen doesn't work :) )

I suggest that we find the real cause of the issue. If it is guaranteed that message M2 won't be created until synchronous part of the method that created previous message M1 is completed, messages will be sent in order they are created. So the issue is somewhere else:
- Maybe some of them weren't sent because of the cancellation token, and receiving side doesn't expect that.
- Maybe they aren't created in the expected order.
- Maybe they are actually created in different threads.
- Maybe we have more than one instance of connection.

If you share what are the repro steps and symptoms of the issue, I can take a look.

zooba: I started writing out the exact sequence and proved to myself that you're right :) I'll revert this part of the change and reanalyze the likely cause.

The repro steps are basically to edit Python code for a period of time, making edits, and then the buffer gets out of sync. When that happens, there are typically two fileUpdate messages (in the connection logs - see the wiki for enabling these): one resets to version 0 and then the next applies a change to the version prior to the reset (i.e. the reset preceded the update). I'll try and track it down.

zooba:
@AlexanderSher FYI, I reverted the change to attempt to sequence messages and will address that separately. The rest of the fix seemed to be fine, so we're taking it now.

PR #3920, opened 2018-03-20
Fixes #3919 Imported modules do not allow go to definition

### LINKED ISSUE #3919: Imported modules do not allow go to definition
If you import code from outside your project, we cannot go to the definition.

This is because we map documentUri back to local files through the loaded file map. We should fall back directly where possible (and eventually that loaded file map will go away anyway).

zooba:
Fixes #3919 Imported modules do not allow go to definition
Attempt to read local path from documentUri if file path is missing
Return class/function and external values as definitions rather than values


zooba:
Indeed they are - more changes coming

zooba:
@huguesv I've fixed those tests as well. Feel free to merge this if you're happy, so we can get a full test run overnight.

PR #3926, opened 2018-03-21
Remote attach using experimental debugger.

karthiknadig:


Python/Product/Debugger/DebugAdapterRemoteProcess.cs
```diff
@@ -0,0 +1,94 @@
...
+            var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.IP);
+            EndPoint endpoint;
+            if (uri.IsLoopback) {
+                endpoint = new IPEndPoint(IPAddress.Loopback, uri.Port);
+            } else {
+                endpoint = new DnsEndPoint(uri.Host, uri.Port);
+            }
+            Debug.WriteLine("Connecting to remote debugger at {0}", uri.ToString());
+            var connection = Task.Factory.FromAsync(
+                (AsyncCallback callback, object state) => socket.BeginConnect(endpoint, callback, state),
```
int19h: Take a look at `SocketTaskExtensions`.

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
```diff
@@ -0,0 +1,94 @@
...
+                socket.EndConnect,
+                null);
+
+            try {
+                if (connection.Wait(_debuggerConnectionTimeout)) {
```
int19h: This should probably use whatever machinery PTVS has to do blocking waits on async tasks in a way that prevents deadlocks and similar, since we don't know which thread this is going to be called on.

Python/Product/Debugger/DebugAdapterUtils.cs
```diff
@@ -0,0 +1,81 @@
...
+            ProcessStartInfo psi = new ProcessStartInfo {
+                FileName = @"C:\Windows\System32\NETSTAT.EXE",
+                Arguments = "-a -n -o -p TCP",
+                UseShellExecute = false,
+                RedirectStandardOutput = true,
+                RedirectStandardInput = true,
+                RedirectStandardError = true
+            };
+
+            using (var process = Process.Start(psi)) {
```
int19h: This needs a check that `uri` is loopback, first, otherwise we might get an accidental match.

Also, `Process.Start` can throw a bunch of stuff; not sure about `WaitForExit`. But basically this needs to gracefully handle any errors during retrieval - we don't want to fail to show the process as available to attach just because we couldn't get info on it.

(If you feel like dropping this workaround instead, and just going with the hardcoded PID and name, that works too. Showing accurate information for remote processes is not a goal at this stage. The important part is that the endpoint gets validated, and can be connected to.

karthiknadig: I agree, will remove this workaround.

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
karthiknadig: That is available in 4.7. Should I switch the framework version?

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
karthiknadig: I will try with `TaskHelpers.RunSynchronouslyOnUIThread`. Let me know if this is not the right thing to use here.
```
TaskHelpers.RunSynchronouslyOnUIThread(ct => Task.WhenAny(
                    Task.Factory.FromAsync(socket.BeginConnect, socket.EndConnect, endpoint, null), 
                    Task.Delay(_debuggerConnectionTimeout), 
                    Task.FromCanceled(ct)));
```

int19h: That's overkill - we don't need the UI thread here really, we just need to wait safely. @AlexanderSher, what's the safe `Wait` equivalent that goes through VS joinable task factory machinery?

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
int19h: I think it's okay, since VS itself is 4.7. @zooba?

Python/Product/Debugger/DebugAdapterUtils.cs
```diff
@@ -0,0 +1,81 @@
...
+                pid = 55555;
+                processName = "python";
+            }
+        }
+
+        public static bool UseExperimentalDebugger() {
+            bool defaultValue = true;
+            if (!_useExperimental.HasValue) {
+                try {
+                    var experimentalKey = @"Software\Microsoft\PythonTools\Experimental";
```
int19h: This should probably be going through `PythonToolsService.ExperimentalOptions`.

int19h: Oh, right, it would actually be circular then. Also, the adapter might not even be in VS process (in msvsmon.exe scenarios, for example).

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
zooba: I don't know that's the case actually, pretty sure VS still only requires 4.6. Eventually it'll get onto 4.7, but not sure what the timeline is.

AlexanderSher: It is not 4.7 and I'm not sure it will be easy to use 4.7 with VS. I've tried couple months ago to switch RTVS, there were tons of assembly load issues.

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
AlexanderSher: `JoinableTaskFactory.Run`. Why can't you make `DebugAdapterRemoteProcess.Attach` method async?

Python/Product/Debugger/DebugAdapterUtils.cs
AlexanderSher: We had some good implementation of this in RTVS.

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
zooba: If you have a service provider, `_site.GetUIThread().InvokeTaskSync(...)`. If not, use `ThreadHelper.JoinableTaskFactory.Run()`. (Don't use the global service provider without also having a fallback to `ThreadHelper`, and in general we strongly avoid using the global service provider.)

Python/Product/Debugger/DebugAdapterUtils.cs
zooba: You can add a reference to `VSInterpreters`, which has the actual `ExperimentalOptions` class. PythonTools only has it so we can expose them to users - anything checking the options should be going through the lower-level DLL.

zooba: (And add an `InternalsVisibleTo` to `VSInterpreters` if you need it)

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteDebugPortSupplier.cs
```diff
@@ -75,13 +75,14 @@ public class PythonRemoteDebugPortSupplier : IDebugPortSupplier2, IDebugPortSupp
...
             var port = new PythonRemoteDebugPort(this, pRequest, uri, DebugLog);
 
-            // Validate connection early. Debugger automation (DTE) objects are not consistent in error checking from this
-            // point on, so errors reported from EnumProcesses and further calls may be ignored and treated as successes
-            // (with empty result). Reporting an error here works around that.
-            IEnumDebugProcesses2 processes;
-            int hr = port.EnumProcesses(out processes);
-            if (hr < 0) {
-                return hr;
+            if (DebugAdapterUtils.UseExperimentalDebugger()) {
```
huguesv: Is this reversed? You're changing behavior for old debugger.

Python/Product/Debugger/DebugAdapterUtils.cs
```diff
@@ -0,0 +1,81 @@
...
+                    using (var root = Registry.CurrentUser.OpenSubKey(experimentalKey, false)) {
+                        var value = root?.GetValue("UseVsCodeDebugger", 1);
+                        if (value == null) {
+                            _useExperimental = defaultValue;
+                        } else {
+                            int? asInt = value as int?;
+                            _useExperimental = asInt.HasValue ? (asInt.GetValueOrDefault() == 1) : defaultValue;
+                        }
+                    }
+                } catch(Exception) {
```
huguesv: `catch (Exception ex) when (!ex.IsCriticalException())`

Python/Product/Debugger/DebugAdapterUtils.cs
karthiknadig: I think it is risky to do this using netstat directly. There are other issues involving user permissions with using netstat. If needed i will implement this using the underlying API used by netstat to avoid process handle probing which requires additional privileges.

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
AlexanderSher: `_site.GetUIThread().InvokeTaskSync` will also use UI thread, which isn't required here.

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
```diff
@@ -0,0 +1,107 @@
...
+            Microsoft.VisualStudio.Shell.ThreadHelper.JoinableTaskFactory.Run(() => Task.WhenAny(
+                    Task.Factory.FromAsync(socket.BeginConnect, socket.EndConnect, endpoint, null),
+                    Task.Delay(_debuggerConnectionTimeout)));
+            try {
+                if (socket.Connected) {
+                    _sockets.Add(uri.Port, socket);
+                    _debuggerConnected = true;
+                    _stream = new DebugAdapterProcessStream(new NetworkStream(socket, ownsSocket: true));
+                } else {
+                    Debug.WriteLine("Timed out waiting for debuger to connect.", nameof(DebugAdapterRemoteProcess));
```
huguesv: debugger typo (below too)

Python/Product/Debugger/Debugger/DebugEngine/Remote/PythonRemoteDebugProgram.cs
```diff
@@ -85,7 +86,9 @@ internal class PythonRemoteDebugProgram : IDebugProgram2 {
         }
 
         public int GetEngineInfo(out string pbstrEngine, out Guid pguidEngine) {
-            pguidEngine = AD7Engine.DebugEngineGuid;
+            pguidEngine = ExperimentalOptions.GetUseVsCodeDebugger() ?
```
zooba: The property is preferred over the getter function. These options require restarting VS, so they aren't going to change after launch.

Python/Product/Debugger/DebugAdapterRemoteProcess.cs
```diff
@@ -0,0 +1,107 @@
...
+            _debuggerConnected = false;
+            var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.IP);
+            EndPoint endpoint;
+            if (uri.IsLoopback) {
+                endpoint = new IPEndPoint(IPAddress.Loopback, uri.Port);
+            } else {
+                endpoint = new DnsEndPoint(uri.Host, uri.Port);
+            }
+            Debug.WriteLine("Connecting to remote debugger at {0}", uri.ToString());
+            Microsoft.VisualStudio.Shell.ThreadHelper.JoinableTaskFactory.Run(() => Task.WhenAny(
```
AlexanderSher: @karthiknadig , why do you insist that `ConnectSocket` should be synchronous?

Issue #3905, opened 2018-03-19
Environments window is slow to open


huguesv:
UI freezes for a few seconds, it's longer if you don't already have Python packages already loaded. A bit longer still if conda support is enabled.

If it's the discovery that is slow, we should see if we can use a cached list, and/or make it load them incrementally.

First discovery is done on a worker thread, which shouldn't be blocking the UI. Not sure what's going on.

huguesv:
Most of the time is spent in `InterpreterListToolWindow.List_ViewCreated`, more specifically the loop that creates all the `PipExtensionProvider` instances.

To be continued...

PR #3949, opened 2018-03-23
Improve environments window startup time

huguesv:
Fix https://github.com/Microsoft/PTVS/issues/3905

Stop initializing all extensions for all environments on window startup. Extensions are now initialized as needed, when an environment is selected the first time.

There were 2 difficulties, which have been resolved.

A. Install package from solution explorer brings up the package manager extension and sets focus to the search box. Before when everything was initialized early, that worked fine but now that everything is delayed it was trying to focus on the text box before the package manager was 'ready' ie the textbox was disabled.

B. When pip extension was selected from drop down and you clicked a different environment, it was skipping the refresh packages because its state wasn't ready yet, so no packages appeared and it never attempted to refresh once it became ready.

I discovered another performance issue while testing B where we refresh packages of the previous environment AND refresh packages of the new environment when you change selection in environment list. There's no point in refreshing packages for the old environment, that only takes away CPU/IO from the refresh of the new environment. I'll file a bug for it. This PR doesn't change that behavior.

On my machine with:
2 anaconda,
5 conda envs
2 ironpython
15 cpython

I only tested debug build (NOT running under debugger). Timing of `ToolWindow.UpdateEnvironments` function. Window does not become responsive until that method has finished executing. That's pretty bad if the tool window is visible when VS starts.

Before: 5735ms
After: 147ms

There's a very slight delay when you select an environment in the list as it creates the extensions for that environment (only once per env of course), but it's much better than the initial delay where we created all of them!

### DIFF
--- a/Python/Product/EnvironmentsList/PipExtension.xaml.cs
+++ b/Python/Product/EnvironmentsList/PipExtension.xaml.cs
@@ -40,10 +40,12 @@ namespace Microsoft.PythonTools.EnvironmentsList {
         public static readonly ICommand InstallPip = new RoutedCommand();
 
         private readonly PipExtensionProvider _provider;
+        private readonly Timer _focusTimer;
 
         public PipExtension(PipExtensionProvider provider) {
             _provider = provider;
             DataContextChanged += PackageExtension_DataContextChanged;
+            _focusTimer = new Timer(FocusWaitExpired);
             InitializeComponent();
         }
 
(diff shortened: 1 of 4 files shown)


Python/Product/EnvironmentsList/PipExtension.xaml.cs
```diff
@@ -40,30 +40,61 @@ internal sealed partial class PipExtension : UserControl, ICanFocus {
...
+                Keyboard.Focus(SearchQueryText);
+                _focusTimer.Change(Timeout.Infinite, Timeout.Infinite);
+            }
+        }
+
+        private void SearchQueryText_IsEnabledChanged(object sender, DependencyPropertyChangedEventArgs e) {
+            SearchQueryText.IsEnabledChanged -= SearchQueryText_IsEnabledChanged;
+            if (SearchQueryText.IsVisible && SearchQueryText.IsEnabled) {
+                SearchQueryText.IsVisibleChanged -= SearchQueryText_IsVisibleChanged;
+                Keyboard.Focus(SearchQueryText);
```
AlexanderSher: So previously, UI has been freezing because of the `Keyboard.Focus(SearchQueryText);`?

huguesv: Not freezing, but it wouldn't set focus if it was disabled.
FYI the one place where this is used is when you right click "Install package" from solution explorer. It opens the environments tool window, selects the interpreter, select packages from the drop down combo box, and sets focus to the search box.

PR #3957, opened 2018-03-26
VSCode build of the analysis engine

MikhailArkhipov:
- Separated common build setting to `Common.Build.Core.settings`
- Ported Rest -> Markdown documentation converter from VSC extension (`RestTextConverter`) (will ports tests next)
- Turned some structs to classes for VSC
- Fixed some [handled] exceptions for easier debugging
- Made sure document text change processing finishes before handling `completion` request
- Fixed few Mac/Windows paths issues
- Added discovery of `so` and `dyLib` Python compiled modules
- Multi-platform build and packaging of the engine as nuget package for Windows, Mac OS and Linux


Python/Product/Analysis/AnalysisNetStandard.csproj
```diff
@@ -8,9 +8,21 @@
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
   <ItemGroup>
     <Compile Remove="AssemblyInfo.cs" />
+  </ItemGroup>
+     <ItemGroup>
+        <PackageReference Include="MicroBuild.Core" Version="0.2.0" />
+        <PackageReference Include="NewtonSoft.Json" Version="10.0.1" />
+        <PackageReference Include="Python2" version="2.7.14" Condition="'$(OS)' == 'Windows_NT'" />
+        <PackageReference Include="Python" version="3.6.4" Condition="'$(OS)' == 'Windows_NT'" />
+    </ItemGroup> <ItemGroup>
```
AlexanderSher: Formatting

Python/Product/Analysis/Infrastructure/ProcessHelper.cs
```diff
@@ -99,34 +99,34 @@ sealed class ProcessHelper : IDisposable {
...
-                return;
             }
-
-            OnOutputLine?.Invoke(e.Data.TrimEnd());
         }
 
         public void Kill() {
             try {
-                _process?.Kill();
+                if (!_process.HasExited) {
```
AlexanderSher: `_process?.Kill()` has null check. Do we need it here as well?

Python/Product/Analysis/Infrastructure/UriExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Net;
+
+namespace Microsoft.PythonTools.Analysis.Infrastructure {
+    public static class UriExtensions {
+        public static string ToAbsolutePath(this Uri uri) {
+            if(IsWindows()) {
+                // VS Code always sends /-based paths. Also, its file URL looks like
+                // file:///C:/foo which yields AbsolutePath as /C:/foo which is wrong.
+                // We need to clean this up.
+                var path = WebUtility.UrlDecode(uri.AbsolutePath).Replace('/', '\\');
```
zooba: We should use LocalPath to extract local paths, not AbsolutePath. (If there are path forms that do not work well with LocalPath, we can special case those. But the normal case works fine.)

Python/Product/Analysis/Interpreter/Ast/AstScrapedPythonModule.cs
```diff
@@ -157,8 +157,14 @@ class AstScrapedPythonModule : IPythonModule
...
 
                     proc.Start();
-                    var exitCode = proc.Wait(60000);
-                    
+                    int? exitCode = null;
+
+                    try {
+                        exitCode = proc.Wait(60000);
+                    } catch (Exception ex) {
+                        fact.Log(TraceLevel.Error, $"ScrapeException {ex.Message}", proc.FileName, proc.Arguments);
```
zooba: Put the exception message as its own field. It makes processing the log file easier.

Python/Product/Analysis/Interpreter/InterpreterConfiguration.cs
```diff
@@ -73,15 +67,12 @@ public sealed class InterpreterConfiguration : IEquatable<InterpreterConfigurati
             Architecture = InterpreterArchitecture.TryParse(Read(properties, nameof(Architecture)));
             try {
                 Version = Version.Parse(Read(properties, nameof(Version)));
-            } catch (ArgumentException) {
-                Version = new Version();
-            } catch (FormatException) {
+            } catch (Exception ex) when (ex is ArgumentException || ex is ArgumentNullException || ex is FormatException) {
```
zooba: Doesn't `ex is ArgumentException` imply that it is also `ArgumentNullException`?

Python/Product/Analysis/LanguageServer/RestTextConverter.cs
```diff
@@ -0,0 +1,269 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+// Based on https://github.com/Microsoft/vscode-python/blob/master/src/client/common/markdown/restTextConverter.ts
+
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
```
zooba: Let's bump this up to `Microsoft.PythonTools.Analysis`. It isn't directly related to the language server.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -37,22 +38,25 @@ public sealed class Server : ServerBase, IDisposable {
         internal readonly ParseQueue _parseQueue;
         private readonly Dictionary<IDocument, VolatileCounter> _pendingParse;
         private readonly VolatileCounter _pendingAnalysisEnqueue;
+        private readonly RestTextConverter _textConverter = new RestTextConverter();
```
zooba: If this is not stateful, can we have a static instance? If it is stateful, we should instantiate it when needed (it looks cheap enough).

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -82,8 +86,15 @@ public sealed class Server : ServerBase, IDisposable {
...
         #region Client message handling
 
-        public async override Task<InitializeResult> Initialize(InitializeParams @params) {
-            _analyzer = await CreateAnalyzer(@params.initializationOptions.interpreter);
+        public override async Task<InitializeResult> Initialize(InitializeParams @params) {
+            _testEnvironment = @params.initializationOptions.interpreter.properties.ContainsKey("TestEnvironment");
+            if (_testEnvironment) {
+                // Test environment needs predictable initialization.
+                // Tests can only proceed when analysis is fully done.
+                _analyzer = CreateAnalyzer(@params.initializationOptions.interpreter).Result;
```
zooba: Can't the tests `await Initialize` to get the same result?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -111,9 +122,16 @@ public sealed class Server : ServerBase, IDisposable {
...
                 capabilities = new ServerCapabilities {
-                    completionProvider = new CompletionOptions { resolveProvider = true },
                     textDocumentSync = new TextDocumentSyncOptions { openClose = true, change = TextDocumentSyncKind.Incremental },
-                    hoverProvider = true
+                    completionProvider = new CompletionOptions {
+                        triggerCharacters = new[] { "." },
+                        resolveProvider = true
+                    },
+                    hoverProvider = true,
+                    signatureHelpProvider = new SignatureHelpOptions { triggerCharacters = new[] { "(,)" } },
```
zooba: Why does a close parenthesis trigger signature help?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -111,9 +122,16 @@ public sealed class Server : ServerBase, IDisposable {
...
+                    // https://github.com/Microsoft/vscode-python/issues/1061
```
zooba: Is there an issue in the PTVS repo for this?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -260,14 +280,17 @@ public sealed class Server : ServerBase, IDisposable {
...
             await _analyzer.ReloadModulesAsync();
-
             // re-analyze all of the modules when we get a new set of modules loaded...
             foreach (var entry in _analyzer.ModulesByFilename) {
                 _queue.Enqueue(entry.Value.ProjectEntry, AnalysisPriority.Normal);
             }
         }
 
         public override Task<CompletionList> Completion(CompletionParams @params) {
+            IfTestWaitForAnalysisComplete();
```
zooba: This is much better done in the test with one of the "Wait" methods. Why can't we do that?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -260,14 +280,17 @@ public sealed class Server : ServerBase, IDisposable {
...
+            // Make sure document is enqueued for processing
+            _documentChangeProcessingComplete.Wait(200, CancellationToken);
```
zooba: Shouldn't we wait for this before waiting for analysis? And shouldn't we wait forever if this is necessary? (I assume most of the time there will be no wait actually happen here.)

zooba: We're also likely better off marking the entry as parse pending as soon as possible when a change comes in, and then the WaitForCurrentParse below will do the correct wait.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -374,6 +397,8 @@ public sealed class Server : ServerBase, IDisposable {
         }
 
         public override Task<SignatureHelp> SignatureHelp(TextDocumentPositionParams @params) {
+            IfTestWaitForAnalysisComplete();
```
zooba: Same question as above

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -434,24 +468,11 @@ public sealed class Server : ServerBase, IDisposable {
...
-                    if (vc.Versions.TryGetValue(GetPart(uri), out var bv)) {
-                        tree = bv.Ast ?? tree;
-                        if (bv.Version >= 0) {
-                            version = bv.Version;
-                        }
-                    }
-                }
-            }
-
+            tree = GetParseTree(entry, uri, out var version) ?? tree;
```
zooba: A couple places above you can use the same helper

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -1022,71 +988,79 @@ public sealed class Server : ServerBase, IDisposable {
...
+                }
 
-                    if (doc is IAnalyzable analyzable && enqueueForAnalysis) {
-                        TraceMessage($"Enqueing document {doc.DocumentUri} for analysis");
-                        _queue.Enqueue(analyzable, priority);
+                // The call must be fire and forget, but should not be yielding.
+                // It is called from DidChangeTextDocument which must fully finish
+                // since otherwise Complete() may come before the change is enqueued
+                // for processing and the completion list will be driven off the stale data.
+                cookieTask.ContinueWith(t => {
```
zooba: For the sake of indentation, can we extract this lambda to its own function ("FinishParse" or similar)? At least for the non-faulting case.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -1275,6 +1260,57 @@ string prefix
...
+            if (_testEnvironment) {
+                WaitForDirectoryScanAsync().Wait();
+                WaitForCompleteAnalysisAsync().Wait();
+            }
+        }
+
+        private PythonAst GetParseTree(IPythonProjectEntry entry, Uri documentUri, out int? version) {
+            version = null;
+            PythonAst tree = null;
+            var parse = entry.WaitForCurrentParse(_clientCaps?.python?.completionsTimeout ?? Timeout.Infinite, CancellationToken);
```
zooba: Completions may not be the only timeout we want here (e.g. we have a longer timeout for hover in VS today). This may need to be a parameter

Python/Product/Analysis/ModulePath.cs
```diff
@@ -143,9 +143,9 @@ public ModulePath(string fullname, string sourceFile, string libraryPath)
             RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
         private static readonly Regex PythonStubRegex = new Regex(@"^(?!\d)(?<name>(\w|_)+)\.pyi$",
             RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
-        private static readonly Regex PythonBinaryRegex = new Regex(@"^(?!\d)(?<name>(\w|_)+)\.((\w|_|-)+?\.)?pyd$",
+        private static readonly Regex PythonBinaryRegex = new Regex(@"^(?!\d)(?<name>(\w|_)+)\.((\w|_|-)+?\.)?(pyd|so|dyLib)$",
             RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
```
zooba: This (and below) regex ignore case - does that need to change for other OS?

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -247,7 +247,7 @@ private class ActivePythonParse : IPythonParse {
             // the children were not registered. To handle this possibility, scan analyzed packages for children of this
             // package (checked by module name first, then sanity-checked by path), and register any that match.
             if (ModulePath.IsInitPyFile(FilePath)) {
-                string pathPrefix = Path.GetDirectoryName(FilePath) + "\\";
+                string pathPrefix = Path.GetDirectoryName(FilePath) + Path.DirectorySeparatorChar;
```
zooba: EnsureEndSeparator?

Python/Product/VSCode/Publish/Publish.csproj
```diff
@@ -2,16 +2,118 @@
...
+        <!-- Override build output during dotnet publish so it goes into a separate folder and does not stomp over signed assemblies -->
+        <DotNetPublishPath Condition="'$(DotNetPublishPath)' == ''">$(BuildOutputRoot)publish\</DotNetPublishPath>
+        <!-- Nuget packages must be under the $(OutputPath) folder or signing task will fail. -->
+        <DotNetPublishNugetPath Condition="'$(DotNetPublishNugetPath)' == ''">$(BuildOutputRoot)\raw\nuget\</DotNetPublishNugetPath>
+        <OutputPath Condition="'$(DotNetPublish)' != ''">$(DotNetPublishPath)</OutputPath>
+    </PropertyGroup>
+    <!-- Nuget packaging -->
+    <PropertyGroup>
+        <NuGetPackageRoot Condition=" '$(NuGetPackageRoot)' == '' ">$(UserProfile)\.nuget\packages\</NuGetPackageRoot>
+        <NugetPackageVersion>0.1.0</NugetPackageVersion>
```
zooba: Is this something we should promote to a build property?

zooba:
And when we merge this, we should do a squash merge :)

Python/Product/Analysis/Interpreter/InterpreterConfiguration.cs
MikhailArkhipov: Yep

Python/Product/Analysis/LanguageServer/RestTextConverter.cs
MikhailArkhipov: Sure

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Technically it has states as it is converting. Not that it is called from multiple threads concurrently, but theoretically it could.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Done

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
```diff
@@ -228,5 +262,32 @@ public Task<TextEdit[]> DocumentOnTypeFormatting(JToken token)
...
+                proc.Exited += (s, e) => {
+                    Environment.Exit(0);
+                };
+            }
+            Task.Run(() => {
+                while (true) {
+                    if (SafeGetProcessById(pid) == null) {
+                        Environment.Exit(0);
+                    }
+                    Thread.Sleep(5000);
```
AlexanderSher: `await Task.Delay(5000)`
And you can pass `CancellationToken`

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
```diff
@@ -140,7 +174,7 @@ public Task DidOpenTextDocument(JToken token)
            => _server.DidOpenTextDocument(token.ToObject<DidOpenTextDocumentParams>());
 
         [JsonRpcMethod("textDocument/didChange")]
-        public Task DidChangeTextDocument(JToken token)
+        public void DidChangeTextDocument(JToken token)
```
AlexanderSher: Is it a public API?

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
```diff
@@ -69,51 +89,65 @@ struct PublishDiagnosticsParams {
...
 
         #region Lifetime
         [JsonRpcMethod("initialize")]
         public Task<InitializeResult> Initialize(JToken token) {
             var p = token.ToObject<InitializeParams>();
+            _ui.LogMessage($"VS Code process: {(p.processId.HasValue ? p.processId.Value : 0)}", MessageType.Info);
+
+#if WAIT_FOR_DEBUGGER
+            while (!System.Diagnostics.Debugger.IsAttached) {
+                Thread.Sleep(1000);
```
AlexanderSher: `await Task.Delay(1000)`

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
```diff
@@ -69,51 +89,65 @@ struct PublishDiagnosticsParams {
         private void OnLogMessage(object sender, LogMessageEventArgs e) => _ui.LogMessage(e.message, e.type);
         private void OnPublishDiagnostics(object sender, PublishDiagnosticsEventArgs e) {
             var parameters = new PublishDiagnosticsParams {
-                uri = e.uri,
+                uri = e.uri.ToString(),
                 diagnostics = e.diagnostics.ToArray()
             };
-            _rpc.InvokeAsync("textDocument/publishDiagnostics", parameters);
+            _vscode.NotifyWithParameterObjectAsync("textDocument/publishDiagnostics", parameters);
```
AlexanderSher: `DoNotWait()`

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -144,8 +144,8 @@ private class ActivePythonParse : IPythonParse {
             }
         }
 
-        public IPythonParse WaitForCurrentParse(int timeout = -1) {
-            if (!_pendingParse.Wait(timeout)) {
+        public IPythonParse WaitForCurrentParse(int timeout, CancellationToken token) {
```
AlexanderSher: Can we make this method async?

Python/Product/VSCode/Publish/Publish.csproj
MikhailArkhipov: Yes

Python/Product/Analysis/ModulePath.cs
MikhailArkhipov: I think so and dyLib cover OSX and Linux

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Usually nested calls as in `foo(bar()|` <- you may want signature help return to `foo`

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Those tests are in typescript...

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: https://github.com/Microsoft/PTVS/issues/3964

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Then can you move the _testEnvironment variable to the VSC wrapper, rather than embedding it in the server? I've tried really hard to keep this kind of state out of the analyzer this time around :)

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Link both from the comment, so we can resolve the correct issue when it gets fixed.

Python/Product/Analysis/ProjectEntry.cs
zooba: I'd rather not. This is only ever used in contexts where blocking briefly (and it's only a parse, so it's brief) is the right action.

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
MikhailArkhipov: It's current in Program.cs which is not async

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
MikhailArkhipov: Sort of. Call comes from VSC process over RPC. StreamJsonRPC handles is async or plain depending on the signature.

Python/Product/Analysis/PythonDb.Generator.targets
```diff
@@ -0,0 +1,74 @@
+﻿<Project>
```
MikhailArkhipov: Unused, but keeping it around in case we want to pre-scrape on Mac

Python/Product/Analysis/Infrastructure/UriExtensions.cs
MikhailArkhipov: OK. LocalPath still needs to be cleaned up. It looks like `/f:/VSCP/src/test/pythonFiles`

Python/Product/Analysis/LanguageServer/Messages.cs
```diff
@@ -63,7 +63,7 @@ public sealed class ShowMessageEventArgs : EventArgs {
     }
 
     [Serializable]
-    public struct ShowMessageRequestParams {
+    public class ShowMessageRequestParams {
```
AlexanderSher: You have converted _some_ of the structs into classes. Does it matter only for those several types?

AlexanderSher:
Some of the files have disappeared from the PR. Is it intentional?

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
AlexanderSher: I'm just thinking if this method can become async in future version of API

Python/Product/Analysis/ProjectEntry.cs
AlexanderSher: How many times `GetCompletions` can be called while we're waiting for a parse?

Python/Product/Analysis/Infrastructure/ProcessHelper.cs
zooba: @MikhailArkhipov Still want the null check here

Python/Product/Analysis/Infrastructure/TaskExtensions.cs
```diff
@@ -17,7 +17,7 @@
 using System.Threading.Tasks;
 
 namespace Microsoft.PythonTools.Analysis.Infrastructure {
-    static class TaskExtensions {
+    public static class TaskExtensions {
```
zooba: Does this need to be public? InternalsVisibleTo is preferable.

Python/Product/Analysis/Infrastructure/UriExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Net;
+
+namespace Microsoft.PythonTools.Analysis.Infrastructure {
+    public static class UriExtensions {
```
zooba: Internal is preferred.

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Okay, let's instantiate it for each conversion, just to be safe.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -37,22 +38,25 @@ public sealed class Server : ServerBase, IDisposable {
...
         private readonly VolatileCounter _pendingAnalysisEnqueue;
+        private readonly RestTextConverter _textConverter = new RestTextConverter();
 
         // Uri does not consider #fragment for equality
         private readonly ConcurrentDictionary<Uri, IProjectEntry> _projectFiles;
         private readonly ConcurrentDictionary<Uri, Dictionary<int, BufferVersion>> _lastReportedDiagnostics;
 
         // For pending changes, we use alternate comparer that checks #fragment
         private readonly ConcurrentDictionary<Uri, List<DidChangeTextDocumentParams>> _pendingChanges;
+        private ManualResetEventSlim _documentChangeProcessingComplete = new ManualResetEventSlim(true);
```
zooba: `readonly`?

Python/Product/Analysis/LanguageServer/Messages.cs
MikhailArkhipov: Those which are important are the JSON serialization. Others can remain structs.

MikhailArkhipov:
Branch was squished, had to restore some changes

MikhailArkhipov:
See also https://github.com/Microsoft/vscode-python/pull/1231

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Sure

Python/Product/Analysis/Infrastructure/TaskExtensions.cs
MikhailArkhipov: They are generally useful extensions but up to you. @AlexanderSher - your opinion?

zooba:
Generally useful does not mean we make them public supported API. This assembly is public/supported now, so we only want things to be public when it’s deliberate.

Python/Product/Analysis/Infrastructure/ProcessHelper.cs
MikhailArkhipov: Done

MikhailArkhipov:
OK, made internal

MikhailArkhipov:
`MonitorParentProcess` is mostly for Mac where `dotnet` does not exit when launching process dies.

Python/Product/Analysis/Infrastructure/UriExtensions.cs
MikhailArkhipov: Done

Python/Product/Analysis/Infrastructure/TaskExtensions.cs
MikhailArkhipov: Made internal

PR #3963, opened 2018-03-27
[WIP] Send event on unhandled exceptions

### LINKED ISSUE #3946: Add event for unexpected error
Our unexpected error handling that brings up a dialog prevents VS from crashing, which means no Watson crash report. We need a way to measure the impact of these errors (not everyone files an issue when they see the dialog).

huguesv:
Fix #3946

### DIFF
--- a/Python/Product/Cookiecutter/Shared/Infrastructure/VSTaskExtensions.cs
+++ b/Python/Product/Cookiecutter/Shared/Infrastructure/VSTaskExtensions.cs
@@ -41,7 +41,8 @@ namespace Microsoft.CookiecutterTools.Infrastructure {
             [CallerFilePath] string callerFile = null,
             [CallerLineNumber] int callerLineNumber = 0,
             [CallerMemberName] string callerName = null,
-            bool allowUI = true
+            bool allowUI = true,
+            string description = null
         ) {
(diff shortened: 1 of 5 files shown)


zooba:
We almost certainly want to use `PostFault` rather than `PostEvent`. Haven't looked into the rest of the PR yet though.

zooba:
I'd kind of like this to go through the existing logging infrastructure, so that the in-memory logger can also collect these events, and we can also keep the telemetry pieces all in the one spot. That may not work with cookiecutter though, as we don't have the references there for it... any reason you didn't put it with the existing support?

huguesv:
1. Cookiecutter would have required a different implementation (all the telemetry classes are different).
2. No need to log to in-memory logger since we already fetch from the Windows Event log when building diagnostics info.
3. This can work before our PythonToolsLogger is created.

zooba:
1 is valid. 3 is irrelevant (if we're throwing unhandled errors during load then we can't recover through our own dialog), and if we log through the logger then there's no need to fetch the event log anymore (which is one of the slowest parts of building the diagnostic info).

Not sure if 1 alone is a good enough reason, especially since it seems like it needs to be a different implementation anyway?

PR #3972, opened 2018-03-29
Rename registry key for New Style IntelliSense option

### LINKED ISSUE #3953: Reset new IntelliSense to be enabled by default
We recently changed the default to be enabled, but anyone that had previously changed that setting won't pick up the new default.

By changing the registry key name, it will reset for everyone.

huguesv:
Fix #3953 Reset new IntelliSense to be enabled by default


zooba:
How about we change it to something that implies "Use legacy IntelliSense"? And also update the UI to indicate this as well.

huguesv:
What bothers me is that it turns it into a "legacy" option instead of a "experimental" option and it looks weird in the Experimental option page, where all the others are about new features and this would be about an old feature.

zooba:
> What bothers me is that it turns it into a "legacy" option

Well, it basically is anyway. We've shipped the new one as the default, so disabling it really is reverting to legacy behaviour. But getting the change in is more important, so let's just go with this.

PR #3976, opened 2018-03-30
Support AsyncPackage

AlexanderSher:
There are several known issues with async package right now:
- Unless you use UIContext-based autoload, package is loaded on UI thread, but it is going to change in the future. There is no way to fully simulate that behavior for testing.
- While `IVsAsyncToolWindowFactory` is supposed to be the only entry point for async tool window creation, `AsyncPackage`, which uses this interface and at the same time provides default implementation for it, ignores `IVsAsyncToolWindowFactory.CreateToolWindow` and uses its own logic to create tool windows. Because of that, we, for instance, can't make async instantiation of python interactive window.
- While `AsyncPackage.GetServiceAsync` should be safe to call from any thread, it is actually not.

Because of those issues, I'm not going to merge this PR right now, but I still would like a CR on it.


Common/Product/SharedProject/UIThread.cs
```diff
@@ -32,25 +32,19 @@ class UIThread : UIThreadBase, IDisposable {
...
+        public UIThread(JoinableTaskFactory joinableTaskFactory) {
+            if (joinableTaskFactory != null) {
+                _factory = joinableTaskFactory;
+                _context = joinableTaskFactory.Context;
                 Trace.TraceInformation("Using TID {0}:{1} as UI thread", _context.MainThread.ManagedThreadId, _context.MainThread.Name ?? "(null)");
-            } catch (NullReferenceException) {
+            } else {
                 _needDispose = true;
                 _context = new JoinableTaskContext();
                 _factory = new JoinableTaskFactory(_context);
```
zooba: Do we not need to use the instance on `ThreadHelper` if it's already there? We need to share the context with the rest of VS, right?

Python/Product/Cookiecutter/CookiecutterPackage.cs
```diff
@@ -21,13 +21,18 @@
...
+using System.Windows;
 using Microsoft.CookiecutterTools.Infrastructure;
 using Microsoft.CookiecutterTools.Model;
 using Microsoft.CookiecutterTools.Telemetry;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.OLE.Interop;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Threading;
+using TplTask = System.Threading.Tasks.Task;
```
zooba: We just use this as `Task` elsewhere

Python/Product/Cookiecutter/CookiecutterPackage.cs
```diff
@@ -201,27 +208,18 @@ public sealed class CookiecutterPackage : Package, IOleCommandTarget {
...
-                }
-
-                if (focus) {
-                    var content = window.Content as System.Windows.UIElement;
-                    if (content != null) {
-                        content.Focus();
-                    }
-                }
+        internal async Task<T> ShowWindowPaneAsync<T>(bool focus) where T : ToolWindowPane {
+            await JoinableTaskFactory.SwitchToMainThreadAsync(DisposalToken);
```
zooba: This will probably break mock tests (where we load packages), FYI. That's why we stick to using the UIThread class.

Python/Product/Cookiecutter/Shared/Infrastructure/CancellationTokenUtilities.cs
```diff
@@ -0,0 +1,27 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.CookiecutterTools.Infrastructure {
+    public static class CancellationTokenUtilities {
```
zooba: Don't make it public if it doesn't need to be

Python/Product/Cookiecutter/Shared/Infrastructure/TaskCompletionSourceExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.CookiecutterTools.Infrastructure {
+    public static class TaskCompletionSourceExtensions {
```
zooba: Don't make it public if it doesn't need to be

Python/Product/PythonTools/PythonToolsPackage.cs
```diff
@@ -383,70 +386,61 @@ internal sealed class PythonToolsPackage : CommonPackage, IVsComponentSelectorPr
...
-                new ShowCppViewCommand(this),
-                new ShowNativePythonFrames(this),
-                new UsePythonStepping(this),
-                new AzureExplorerAttachDebuggerCommand(this),
-                new OpenWebUrlCommand(this, "https://go.microsoft.com/fwlink/?linkid=832525", PkgCmdIDList.cmdidWebPythonAtMicrosoft),
-                new OpenWebUrlCommand(this, Strings.IssueTrackerUrl, PkgCmdIDList.cmdidWebPTVSSupport),
-                new OpenWebUrlCommand(this, "https://go.microsoft.com/fwlink/?linkid=832517", PkgCmdIDList.cmdidWebDGProducts),
-            }, GuidList.guidPythonToolsCmdSet);
+            RegisterCommands(GuidList.guidPythonToolsCmdSet
+                , new OpenReplCommand(this, (int)PkgCmdIDList.cmdidReplWindow)
```
zooba: What's with the comma at the start of the line? We don't use this style anywhere else, and I don't like it.

Common/Product/SharedProject/UIThread.cs
AlexanderSher: Yes, recommendation is to avoid `ThreadHelper`.

Common/Product/SharedProject/CommonProjectPackage.cs
```diff
@@ -61,13 +62,15 @@ public abstract class CommonProjectPackage : ProjectPackage, IVsInstalledProduct
         public abstract string GetProductVersion();
 
         protected override void Initialize() {
-            UIThread.EnsureService(this);
+            if (GetService(typeof(UIThreadBase)) == null) {
+                ((IServiceContainer)this).AddService(typeof(UIThreadBase), new UIThread(null), true);
+            }
```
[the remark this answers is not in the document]
AlexanderSher: `UIThread.EnsureService` was called from three places. Code is different in all three places now:
- In `CommonProjectPackage`, it uses synchronous `GetService`
- In `CommonPackage`, it uses synchronous `GetServiceAsync`
- In `MockPythonToolsPackage`, we don't need to check for the service at all.

Python/Product/PythonTools/PythonToolsPackage.cs
AlexanderSher: Sorry, auto formatting on copy-paste

Python/Product/Cookiecutter/CookiecutterPackage.cs
AlexanderSher: Well, Cookiecutter tests pass 100%, but if we have mock tests that load packages, we may have problems with them anyway - `AsyncPackage` uses `JoinableTaskFactory` instance quite intensively

@huguesv , can you suggest a mock test that loads `PythonToolsPackage`?

Python/Product/Cookiecutter/Shared/Infrastructure/CancellationTokenUtilities.cs
AlexanderSher: Sure

Python/Product/Cookiecutter/Shared/Infrastructure/TaskCompletionSourceExtensions.cs
AlexanderSher: Sure

Python/Product/Cookiecutter/CookiecutterPackage.cs
huguesv: Not for cookiecutter, it doesn't use mock tests.

AlexanderSher: Ok, what about other packages?

huguesv: I ran a random sample of PythonToolsMockTests (at least one for each test class) and didn't see PythonToolsPackage.Initialize being called.

AlexanderSher: `PythonToolsMockTests` fail on my machine with MEF composition exception: https://github.com/Microsoft/PTVS/issues/3978, so I can't check it right now.

Common/Product/SharedProject/UIThread.cs
zooba: So everyone just creates their own new joinable context? How does that work?

Python/Product/Cookiecutter/CookiecutterPackage.cs
zooba: It should compose fine - the test machines seem okay, so you probably need to fix it in part of the PR.

Common/Product/SharedProject/UIThread.cs
AlexanderSher: Not context, but collection, factory, associated sync context and so on. If I recall correctly, recommendation for package-specific factories was there from the very beginning, but no one followed it because there was `ThreadHelper.JoinableTaskFactory`. So now they have implemented it by themselves. When package is disposed, it waits for all the tasks to complete:
```cs
        protected override void Dispose(bool disposing)
        {
            // Let any in-flight async set sites know they should just abort whatever they are doing.
            this.disposeCancellationTokenSource.Cancel();

            // Join on the tasks BEFORE we unregister/Dispose of the service instances, in case one of those services kicked off async work
            // that relies on said service. If we dispose services before waiting one of those tasks could encounter an ObjectDisposedException.
            try
            {
                ThreadHelper.JoinableTaskFactory.Run(this.JoinableTaskCollection.JoinTillEmptyAsync);
            }
```

Python/Product/Cookiecutter/CookiecutterPackage.cs
AlexanderSher: I can repro it on a clean master after erasing `BuildOutput`. Probably need to update my VS to the recent build.

huguesv:
Internal bug 611209 says that PythonUwpPackage is marked for synchronous auto load and needs to be updated. I don't see it in the current set of changes.

AlexanderSher:
Yes, it wasn't originally in the list.

huguesv:
I'm testing this PR applied on latest master, on 15.8p2

If I go to View->Other Windows->Python Performance Explorer, I get an error

---------------------------

---------------------------
Profiling support seems to be missing or corrupt. Try repairing your Visual Studio installation.
---------------------------
OK   
---------------------------

Before it, it broke in the debugger with an NRE:

image

I had not loaded profiling package prior to that.

I haven't tested without the PR yet.

AlexanderSher:
Yes, looks like async tool window can't use clsid anymore. I've asked VS about it, they haven't provided a workaround yet.

PR #3994, opened 2018-04-03
Send event on unhandled exceptions (alternate)

### LINKED ISSUE #3946: Add event for unexpected error

huguesv:
Fix #3946 Add event for unexpected error

This one goes through our wrappers, unlike https://github.com/Microsoft/PTVS/pull/3963

It also bypasses watson sampling as suggested by telemetry folks.

Added several missing `HandleAllExceptions()` calls in cookiecutter.


Python/Product/Cookiecutter/Telemetry/StringTelemetryRecorder.cs
```diff
@@ -47,6 +47,10 @@ internal sealed class StringTelemetryRecorder : ITelemetryRecorder, ITelemetryLo
                 }
             }
         }
+
+        public void RecordFault(string eventName, Exception ex, string description, bool dumpProcess) {
```
AlexanderSher: Why is it empty?

Python/Product/Cookiecutter/CookiecutterPackage.cs
```diff
@@ -89,6 +89,8 @@ public sealed class CookiecutterPackage : Package, IOleCommandTarget {
             _projectSystem = new ProjectSystemClient(DTE);
 
             CookiecutterTelemetry.Initialize();
+
+            VSTaskExtensions._telemetry = CookiecutterTelemetry.Current;
```
AlexanderSher: What's the value of `_telemetry`? `CookiecutterTelemetry.Current` is accessible from `VSTaskExtensions`

Python/Tests/CookiecutterTests/TestTelemetryRecorder.cs
```diff
@@ -41,6 +42,11 @@ internal sealed class TestTelemetryRecorder : ITelemetryRecorder, ITelemetryTest
                 }
             }
         }
+
+        public void RecordFault(string eventName, Exception ex, string description, bool dumpProcess) {
+            throw ex;
```
AlexanderSher: `ExceptionDispatchInfo.Capture(ex).Throw()`

Python/Product/VSCommon/Infrastructure/VSTaskExtensions.cs
```diff
@@ -79,20 +82,28 @@ static class VSTaskExtensions {
...
-                        if (site != null) {
-                            TaskDialog.ForException(site, ex, issueTrackerUrl: Strings.IssueTrackerUrl).ShowModal();
-                        } else if (!string.IsNullOrEmpty(logFile)) {
-                            MessageBox.Show(Strings.SeeActivityLog.FormatUI(logFile), Strings.ProductTitle);
-                        }
-                    }
+            bool alreadySeen = true;
+            lock (_displayedMessages) {
+                var key = "{0}:{1}:{2}".FormatInvariant(callerFile, callerLineNumber, ex.GetType().Name);
+                if (_displayedMessages.Add(key)) {
```
AlexanderSher: only `_displayedMessages.Add(key)` requires lock:
```cs
var key = "{0}:{1}:{2}".FormatInvariant(callerFile, callerLineNumber, ex.GetType().Name);
bool alreadySeen;
lock (_displayedMessages) {
    alreadySeen = !_displayedMessages.Add(key);
}
```

Python/Product/VSCommon/Infrastructure/VSTaskExtensions.cs
```diff
@@ -79,20 +82,28 @@ static class VSTaskExtensions {
...
+                    alreadySeen = false;
+                }
+            }
+
+            Debug.Assert(_logger != null);
```
AlexanderSher: Do we set this in tests?

Python/Product/Cookiecutter/View/CookiecutterSearchPage.xaml.cs
```diff
@@ -162,15 +162,15 @@ internal partial class CookiecutterSearchPage : Page {
                 return;
             }
 
-            ViewModel.LoadTemplateAsync().DoNotWait();
+            ViewModel.LoadTemplateAsync().HandleAllExceptions(null, GetType()).DoNotWait();
         }
 
         public void UpdateTemplate() {
-            ViewModel.UpdateTemplateAsync().DoNotWait();
+            ViewModel.UpdateTemplateAsync().HandleAllExceptions(null, GetType()).HandleAllExceptions(null, GetType()).DoNotWait();
```
AlexanderSher: `HandleAllExceptions` is called twice.

Python/Product/Cookiecutter/CookiecutterPackage.cs
huguesv: That was the reason, but might as well just use one global instead of two.

huguesv:
@zooba is this ok with you?

PR #4004, opened 2018-04-05
Fixes #3876 Missing completions for os.path

### LINKED ISSUE #3876: Missing completions for Os.Path.
When attempting to use the VisualStudio 2017 Python functionality I find that Intellisense does not supply completions for the 'os.path' library--such as 'os.path.isdir()' and so on. It does however provide completions for functions directly from 'os' such as 'os.listdir()'.

If I attempt to switch back to what is apparently the 'old' Intellisense DB completion method I get a crash notification every time I attempt to update the database. It hangs and then crashes at the 'Analysing Standard Library' step no matter how many times I retry this procedure.

Is there some way to include the 'os.path' functions--and perhaps other missing completions--to the 'new' system?

zooba:
Fixes #3876 Missing completions for os.path
Correctly handles importing of modules that are defined as members rather than modules.


huguesv:
I'm seeing new test failures in AnalysisTest class.

zooba:
So am I... :( Working on them, but the bugs in Test Explorer right now are really annoying me

zooba:
@huguesv All fixed now (I hope)

PR #4008, opened 2018-04-05
Make analysis sets immutable by default

zooba:
Make analysis sets immutable by default
Make AnalysisHashSet.Clone() thread safe
Add peak memory usage to memory tester output

In my testing so far, this is looking around 2x improvement in analysis time (for e.g. bokeh and conda sources) with minor improvement in memory usage (as opposed to the huge regression I was expecting).

Please test and see if you notice values disappearing from your own code (imported packages are not affected). I'm keen to take this, provided we can't see regressions in results. The tests still pass.

### DIFF
--- a/Python/Product/Analysis/AnalysisHashSet.cs
+++ b/Python/Product/Analysis/AnalysisHashSet.cs
@@ -81,11 +81,15 @@ namespace Microsoft.PythonTools.Analysis.AnalysisSetDetails {
 
         public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable, IEqualityComparer<AnalysisValue> comparer)
             : this(comparer) {
-            Union(enumerable);
+            using (var e = enumerable.GetEnumerator()) {
+                AddFromEnumerator(e);
+            }
         }
 
         public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable) : this() {
-            Union(enumerable);
+            using (var e = enumerable.GetEnumerator()) {
+                AddFromEnumerator(e);
+            }
         }
 
         private BucketSet Buckets {
@@ -111,7 +115,7 @@ namespace Microsoft.PythonTools.Analysis.AnalysisSetDetails {
             return DebugViewProxy.ToString(this);
         }
 
-        public IAnalysisSet Add(AnalysisValue item, bool canMutate = true) {
+        public IAnalysisSet Add(AnalysisValue item, bool canMutate = false) {
             if (!canMutate) {
                 if (Contains(item)) {
                     return this;
@@ -122,7 +126,7 @@ namespace Microsoft.PythonTools.Analysis.AnalysisSetDetails {
             return this;
         }
 
-        public IAnalysisSet Add(AnalysisValue item, out bool wasChanged, bool canMutate = true) {
+        public IAnalysisSet Add(AnalysisValue item, out bool wasChanged, bool canMutate = false) {
 #if FULL_VALIDATION
             var r = AddWorker(item, out wasChanged, canMutate);
             Validation.Assert(r.Count == r.GetTrueCount(), $"Set count is incorrect. Expected {r.GetTrueCount()}. Actual {r.Count}");
@@ -158,13 +162,13 @@ namespace Microsoft.PythonTools.Analysis.AnalysisSetDetails {
             return this;
         }
 
-        public IAnalysisSet Union(IEnumerable<AnalysisValue> items, bool canMutate = true) {
+        public IAnalysisSet Union(IEnumerable<AnalysisValue> items, bool canMutate = false) {
             bool wasChanged;
             return Union(items, out wasChanged, canMutate);
         }
 
 
-        public IAnalysisSet Union(IEnumerable<AnalysisValue> items, out bool wasChanged, bool canMutate = true) {
+        public IAnalysisSet Union(IEnumerable<AnalysisValue> items, out bool wasChanged, bool canMutate = false) {
 #if FULL_VALIDATION
             var r = UnionWorker(items, out wasChanged, canMutate);
             Validation.Assert(r.Count == r.GetTrueCount(), $"Set count is incorrect. Expected {r.GetTrueCount()}. Actual {r.Count}");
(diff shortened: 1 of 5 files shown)


huguesv:
I'm not noticing any obvious regression.

AlexanderSher:
I've run some benchmarks versus standard `ImmutableHashSet`.

Here is the code:
```cs
        [GlobalSetup]
        public void SetupTestData() {
            var rand = new Random();
            _input = new AnalysisValue[1001];
            for (var i = 0; i < _input.Length; i++) {
                var offset = rand.Next();
                _input[i] = new TestAnalysisValue {_name = "Q" + offset, Value = "q" + offset };
            }

            _immutableHashSetAdd = ImmutableHashSet<AnalysisValue>.Empty.WithComparer(UnionComparer.Instances[0])
                .Add(new TestAnalysisValue { _name = "A", Value = "a" })
                .Add(new TestAnalysisValue { _name = "B", Value = "b" })
                .Add(new TestAnalysisValue { _name = "C", Value = "c" });

            _analysisHashSetAdd = (AnalysisHashSet)AnalysisSet.CreateUnion(
                new[] {
                    new TestAnalysisValue { _name = "A", Value = "a" },
                    new TestAnalysisValue { _name = "B", Value = "b" },
                    new TestAnalysisValue { _name = "C", Value = "c" }
                },
                UnionComparer.Instances[0]);

            _immutableHashSetContains = _immutableHashSetAdd.Union(_input);
            _analysisHashSetContains = (AnalysisHashSet)_analysisHashSetAdd.Clone().Union(_input);
        }

        [Benchmark]
        public void ImmutableHashSet_Add() {
            _immutableHashSet = _immutableHashSetAdd;
            foreach (var t in _input) {
                _immutableHashSet = _immutableHashSet.Add(t);
            }
        }

        [Benchmark]
        public void AnalysisHashSet_Add() {
            _analysisHashSet = (AnalysisHashSet)_analysisHashSetAdd.Clone();
            foreach (var t in _input) {
                _analysisHashSet = (AnalysisHashSet)_analysisHashSet.Add(t);
            }
        }

        [Benchmark]
        public void ImmutableHashSet_Contains() {
            foreach (var t in _input) {
                _b = _immutableHashSetContains.Contains(t);
            }
        }

        [Benchmark]
        public void AnalysisHashSet_Contains() {
            foreach (var t in _input) {
                _b = _analysisHashSetContains.Contains(t);
            }
        }
```

Here are the results:
``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.309)
Intel Core i7-4810MQ CPU 2.80GHz (Haswell), 1 CPU, 8 logical cores and 4 physical cores
Frequency=2728057 Hz, Resolution=366.5613 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2633.0
  DefaultJob : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2633.0


```
|                    Method |         Mean |       Error |     Gen 0 |  Allocated |
|-------------------------- |-------------:|------------:|----------:|-----------:|
|      ImmutableHashSet_Add |    742.03 us |   5.5747 us |  104.4922 |   331709 B |
|       AnalysisHashSet_Add | 70,181.17 us | 277.1101 us | 3750.0000 | 11937407 B |
| ImmutableHashSet_Contains |    100.63 us |   0.5204 us |         - |        0 B |
|  AnalysisHashSet_Contains |     63.41 us |   0.6005 us |         - |        0 B |


As you can see, `AnalysisHashSet` is faster than `ImmutableHashSet` by 60% in searching, but is significantly worse in adding, especially in terms of memory.

AlexanderSher:
I've added a test for enumeration:

```cs
        [Benchmark]
        public void ImmutableHashSet_Iterate() {
            foreach (var t in _immutableHashSetContains) {
                _value = t;
            }
        }

        [Benchmark]
        public void AnalysisHashSet_Iterate() {
            foreach (var t in _analysisHashSetContains) {
                _value = t;
            }
        }
```


|                    Method |         Mean |       Error |     Gen 0 |  Allocated |
|-------------------------- |-------------:|------------:|----------:|-----------:|
|  ImmutableHashSet_Iterate |    204.85 us |   0.5041 us |         - |        0 B |
|   AnalysisHashSet_Iterate |     16.26 us |   0.0537 us |         - |       56 B |

So iterating `ImmutableHashSet` is 12 times slower.

AlexanderSher:
Here is the full result, compiled against x64 and including mutable option:

 ``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.309)
Intel Core i7-4810MQ CPU 2.80GHz (Haswell), 1 CPU, 8 logical cores and 4 physical cores
Frequency=2728057 Hz, Resolution=366.5613 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2633.0
  DefaultJob : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2633.0


```
|                     Method |         Mean |         Error |     Gen 0 |  Allocated |
|--------------------------- |-------------:|--------------:|----------:|-----------:|
|       ImmutableHashSet_Add |    937.85 us |    18.4960 us |  187.5000 |   590087 B |
|        AnalysisHashSet_Add | 72,876.41 us | 1,375.8347 us | 7562.5000 | 23862718 B |
| AnalysisHashSet_Add_Mutate |    167.15 us |     0.3410 us |   11.2305 |    35464 B |
|  ImmutableHashSet_Contains |    100.54 us |     1.1985 us |         - |        0 B |
|   AnalysisHashSet_Contains |     67.58 us |     0.2566 us |         - |        0 B |
|   ImmutableHashSet_Iterate |    205.10 us |     1.8505 us |         - |        0 B |
|    AnalysisHashSet_Iterate |     16.46 us |     0.1114 us |         - |       56 B |

@zooba , I'm not sure how to interpret these results considering your measurements. One possibility is that most of the memory taken by `AnalysisHashSet_Add` becomes garbage in real life scenario, but that will automatically mean more GC calls. Another possibility is that AnalysisHashSet.Add isn't called that often.

zooba:
@AlexanderSher I wasn't comparing AnalysisHashSet to ImmutableHashSet, so you've started doing brand new work here.

I was comparing the default `.Union(canMutate: true)` which allows adds to occur in-place when supported (which is only supported by AnalysisHashSet, and not the special cases for 0, 1 and 2 element sets), and the alternative `.Union(canMutate: false)` which always creates a new set when adding new elements.

My guess is the difference is at the macro level. By mutating shared instances of the set, values were showing up in other locations and causing analysis to take much longer than it should have. Without that, we complete much more quickly. I would hope that at the individual operation level it is slower with the change, but overall it is much faster.

If I don't hear any concerns with this PR in the next hour or so, I'm going to merge it.

AlexanderSher:
> I was comparing the default `.Union(canMutate: true)` which allows adds to occur in-place when supported (which is only supported by AnalysisHashSet, and not the special cases for 0, 1 and 2 element sets), and the alternative `.Union(canMutate: false)` which always creates a new set when adding new elements.

How many elements do you add to the `Union`?

PR #4024, opened 2018-04-09
Improved tooltip and member doc display and other fixes

### LINKED ISSUE #3849: Implement information on modules
VS Code extension feature parity

Hovering over module name only shows `os:module` (similarly, no info in the completion item documentation tooltip)

image

VS Code + Jedi shows module documentation

image

### LINKED ISSUE #3869: Signature display inconsistent with `range`
Type `range(c, 1, `

Hover over `range`
image

Place cursor after `c` and hit Ctrl+Shift_Space
image

Expected signature `range(start, stop, ...)`. Almost works in VSC language server, just picks short signature.

image

May relate to https://github.com/Microsoft/PTVS/issues/3868

### LINKED ISSUE #3882: Information is duplicated in completion item tooltips
15.7

image

15.6.1 seems to be better

image

### LINKED ISSUE #3977: Option not to trim tooltip documentation
VS Code supports scrollable tooltip so there is no need to limit or artificially break lines.

### LINKED ISSUE #4031: Concurrency issue accessing analysis dictionary
Occasional AF in `Debug.Assert(_count < buckets.Length);`

Two threads access unprotected objects

```
>	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.AnalysisDictionary<string, Microsoft.PythonTools.Analysis.VariableDef>.AddWorker(Microsoft.PythonTools.Analysis.AnalysisDictionary<string, Microsoft.PythonTools.Analysis.VariableDef>.Bucket[] buckets, string key, Microsoft.PythonTools.Analysis.VariableDef value, int hc, bool throwIfExists) Line 199	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.AnalysisDictionary<string, Microsoft.PythonTools.Analysis.VariableDef>.Add(Microsoft.PythonTools.Analysis.AnalysisDictionary<string, Microsoft.PythonTools.Analysis.VariableDef>.Bucket[] buckets, string key, Microsoft.PythonTools.Analysis.VariableDef value, bool throwIfExists) Line 189	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.AnalysisDictionary<string, Microsoft.PythonTools.Analysis.VariableDef>.AddOne(string key, Microsoft.PythonTools.Analysis.VariableDef value, bool throwIfExists) Line 112	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.AnalysisDictionary<string, Microsoft.PythonTools.Analysis.VariableDef>.this[string].set(string key, Microsoft.PythonTools.Analysis.VariableDef value) Line 549	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.Analyzer.InterpreterScope.AddVariable(string name, Microsoft.PythonTools.Analysis.VariableDef variable) Line 259	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.Analyzer.InterpreterScope.CreateVariable(Microsoft.PythonTools.Parsing.Ast.Node node, Microsoft.PythonTools.Analysis.AnalysisUnit unit, string name, bool addRef) Line 235	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.Analyzer.ModuleScope.SetModuleVariable(string name, Microsoft.PythonTools.Analysis.IAnalysisSet value) Line 33	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.Values.ModuleInfo.EnsureModuleVariables(Microsoft.PythonTools.Analysis.PythonAnalyzer state) Line 68	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.PythonAnalyzer.ReloadModulesAsync() Line 193	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.LanguageServer.Server.DidChangeConfiguration(Microsoft.PythonTools.Analysis.LanguageServer.DidChangeConfigurationParams params) Line 317	C#
 	[Resuming Async Method]	
 	[External Code]	
Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.LanguageServer.Server.Initialize.AnonymousMethod__0(System.Threading.Tasks.Task<Microsoft.PythonTools.Analysis.PythonAnalyzer> t) Line 108	C#
```
and
``` 	
>	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.AnalysisDictionary<string, Microsoft.PythonTools.Analysis.VariableDef>.AddOne(string key, Microsoft.PythonTools.Analysis.VariableDef value, bool throwIfExists) Line 112	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.AnalysisDictionary<string, Microsoft.PythonTools.Analysis.VariableDef>.this[string].set(string key, Microsoft.PythonTools.Analysis.VariableDef value) Line 549	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.Analyzer.InterpreterScope.AddVariable(string name, Microsoft.PythonTools.Analysis.VariableDef variable) Line 259	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.Analyzer.InterpreterScope.CreateVariable(Microsoft.PythonTools.Parsing.Ast.Node node, Microsoft.PythonTools.Analysis.AnalysisUnit unit, string name, bool addRef) Line 235	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.Analyzer.ModuleScope.SetModuleVariable(string name, Microsoft.PythonTools.Analysis.IAnalysisSet value) Line 33	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.Values.ModuleInfo.EnsureModuleVariables(Microsoft.PythonTools.Analysis.PythonAnalyzer state) Line 68	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.ProjectEntry.Parse(bool enqueueOnly, System.Threading.CancellationToken cancel) Line 228	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Analysis.ProjectEntry.Analyze(System.Threading.CancellationToken cancel, bool enqueueOnly) Line 175	C#
 	Microsoft.Python.Analysis.Engine.dll!Microsoft.PythonTools.Intellisense.AnalysisQueue.Worker(object threadStarted) Line 233	C#
```

MikhailArkhipov:
Fixes #3849
Fixes #3869
Fixes #3882
Fixes #3977
Fixes #4031

- Improves startup time in VS Code via async startup mode
- Option not to trim/wrap text in tooltips
- Remove line breaks in docs letting host to decide on flow/wrap
- Somewhat reworked tooltip and completion member display (less duplication, closer to PEP style)
- Added doc extraction for modules


Python/Product/Analysis/LanguageServer/DisplayTextBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            if (documentations.Count == 1 && !string.IsNullOrEmpty(firstLongDescription)) {
+                result.Clear();
+                result.Append(firstLongDescription);
+            }
+
+            var displayText = result.ToString();
+            if (displayOptions.trimDocumentationText && displayText.Length > displayOptions.maxDocumentationTextLength) {
+                displayText = displayText.Substring(0,
```
AlexanderSher: You can use `result.Remove` to avoid allocation of a `displayText`

Python/Product/Analysis/LanguageServer/DisplayTextBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            return _textConverter.ToMarkdown(result.ToString());
+        }
+
+        public string MakeModuleHoverText(ModuleReference modRef) {
+            // Return module information
+            var contents = "{0} : module".FormatUI(modRef.Name);
+            if (!string.IsNullOrEmpty(modRef.Module?.Documentation)) {
+                contents += $"{Environment.NewLine}{Environment.NewLine}{modRef.Module.Documentation}";
```
AlexanderSher: Do we need `FormatUI` here as well?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -87,13 +95,47 @@ public sealed class Server : ServerBase, IDisposable {
...
 
         public override async Task<InitializeResult> Initialize(InitializeParams @params) {
-            _testEnvironment = @params.initializationOptions.interpreter.properties.ContainsKey("TestEnvironment");
-            if (_testEnvironment) {
-                // Test environment needs predictable initialization.
-                // Tests can only proceed when analysis is fully done.
-                _analyzer = CreateAnalyzer(@params.initializationOptions.interpreter).Result;
+            _testEnvironment = @params.initializationOptions.testEnvironment;
+            // Test environment needs predictable initialization.
+            if (@params.initializationOptions.asyncStartup && !_testEnvironment) {
```
AlexanderSher: Do we need `testEnvironment` if we already have `asyncStartup`?

MikhailArkhipov: There are effectively 3 modes
1. Test mode, all sync
2. VS mode, as it used to work since I am not 100% confident VS part can handle all-async as I don't know all the entry points.
3. VSC mode to minimize delays. VSC only need synchronization at the LS entry points. This is what asyncStartup controls.

VSC passes true for asyncStartup  and true/false for tests. asyncStartup == false does not make it completely sync, VS still uses awaits in places. VSC creates analyzer async while VS does not.

if VS only uses LS entry points then we can just do (3). But bunch of methods are internal with visibility so I don't know.

Python/Product/Analysis/LanguageServer/DisplayTextBuilder.cs
MikhailArkhipov: It just does `CurrentCulture`... I am actually not sure that current culture is right.  Here we should have invariant I believe. I.e. if we have programming language docs then numbers must remain `5.0` and should not be displayed as `5,0` in a culture-specific manner.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -168,6 +196,8 @@ public sealed class Server : ServerBase, IDisposable {
         }
 
         public override void DidChangeTextDocument(DidChangeTextDocumentParams @params) {
+            _analyzerCreationTcs.Task.Wait();
```
AlexanderSher: Do we ever call this from UI thread?

MikhailArkhipov: It is out-of-proc so technically there is no UI thread. `OutOfProcProjectAnalyzer' calls it in UpdateContent. In VS, however, this task completes in initialization so only VS Code may hit the wait.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -144,6 +171,7 @@ public sealed class Server : ServerBase, IDisposable {
 
         public override async Task DidOpenTextDocument(DidOpenTextDocumentParams @params) {
             TraceMessage($"Opening document {@params.textDocument.uri}");
+            await _analyzerCreationTcs.Task;
```
AlexanderSher: If there was an exception CreateAnalyzer, `_analyzerCreationTcs` will hold that exception. Is it correct to raise it every time API method is called?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -87,13 +95,47 @@ public sealed class Server : ServerBase, IDisposable {
...
+                            _analyzerCreationTcs.TrySetResult(true);
+                        } catch (Exception ex) {
+                            _analyzerCreationTcs.TrySetException(ex);
+                        }
+                    }
+                }).DoNotWait();
             } else {
                 _analyzer = await CreateAnalyzer(@params.initializationOptions.interpreter);
+                OnAnalyzerCreated(@params);
+                _analyzerCreationTcs.TrySetResult(true);
```
AlexanderSher: Behavior here is different from when `@params.initializationOptions.asyncStartup && !_testEnvironment` is true. If `CreateAnalyzer` raises an exception, `_analyzerCreationTcs` will not complete. Is it by design?

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: If analyzer creation failed, then the whole thing is broken. VSC restarts server automatically up to 3 times I believe.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Let me try-catch that. Although if it throws, nothing is going to work, but at least exception may give client an idea that server may need to be restarted.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -272,24 +311,26 @@ public sealed class Server : ServerBase, IDisposable {
...
+
         public override async Task DidChangeConfiguration(DidChangeConfigurationParams @params) {
+            await _analyzerCreationTcs.Task;
             if (_analyzer == null) {
                 LogMessage(MessageType.Error, "change configuration notification sent to uninitialized server");
                 return;
             }
 
-            await _analyzer.ReloadModulesAsync();
+            _queue.Enqueue(new AnalysisQueueWorkItem(() => _analyzer.ReloadModulesAsync().WaitAndUnwrapExceptions()), AnalysisPriority.Normal);
```
AlexanderSher: This looks dangerous.
- `_queue.Enqueue` won't execute `ReloadModulesAsync` immediately. If there is any code that awaits `DidChangeConfiguration` to ensure that modules are reloaded, it will be broken
- `AnalysisQueue` will reorder project entries, but it won't reorder `AnalysisQueueWorkItem` cause they all are considered different. This is good in the way that `ProjectEntry.Analyze` will be called always after `ReloadModulesAsync`, but has a side effect that several concurrent calls of `ReloadModulesAsync` will be executed one by one instead of swallowing all but first and last call.
- `ReloadModulesAsync` may start new tasks and await on them while `AnalysisQueue` is thread-bound. `AnalysisSynchronizationContext` doesn't support reentrancy and `AnalysisQueue` thread is blocked until method is completed, which means that if anywhere on the code path we have missed `ConfigureAwait(false)` or provide any other way for `AnalysisSynchronizationContext` to be captured, we'll get a deadlock.

Also, `ReloadModulesAsync` should handle `CancellationToken` by itself, but that is a separate issue.

MikhailArkhipov: Well, I am open to suggestions. Currently it is broken b/c ReloadModulesAsync is called from two threads while
```
        /// This method should be called on the analysis thread and is usually invoked
        /// when the interpreter signals that it's modules have changed.
```
So basically change is to queue item before calling `_queue.Enqueue` in `DidChangeConfiguration`. Nothing awaits on DidChangeConfiguration in reality, it is out of proc call, we can make it void.

`ReloadModulesAsync` should never be called in parallel AFAIU

MikhailArkhipov: I think analyzer should probably handle reload internally in its worker thread rather than caller know what is permitted and what not. This will be a bigger change.

MikhailArkhipov: If it is queued, then `_reloadLock.WaitAsync()` prob not needed as there are no concurrent calls (apart of one at the creation at `CreateAsync`. I think this one can be queued too theoretically.

Now, we can also avoid doing anything in `DidChangeConfiguration`. Changing interpreter will actually restart the server.

AlexanderSher: Ok, here is a version of small fix: https://github.com/AlexanderSher/PTVS/commit/bf068bdc4b0b6be3eb34cf3753996f705796e58c
- `DidChangeConfiguration` will wait for last `ReloadModulesAsync` to be processed
- `AnalysisQueue` will properly reorder `ReloadModulesAsync` call if it is already scheduled
- `ReloadModulesAsync` will be started in thread-pool so reentrancy shouldn't be a problem.

huguesv:
Is there enough test coverage for the documentation changes? Aside from spaces removed around operators, there doesn't seem to be much changed.

MikhailArkhipov:
Yes, I updated tests. Need to add tests for the new functionality (tooltips and docs for modules, see https://github.com/Microsoft/PTVS/issues/4050).

Issue #4028, opened 2018-04-10
Adding a new project to an existing solution tries to make a new solution


huguesv:
Split from https://github.com/Microsoft/PTVS/issues/13 which was bundling several issues / enhancements.

1) Create a blank visual studio solution
2) Right click the solution node, choose Add New Project -> Python -> From Existing Python Code
3) Go through the steps.  

At the end of the wizard, your project will have been added to a new solution, not the original one.

PR #4029, opened 2018-04-10
Fix the Project From Existing Code wizard to handle adding to an existing solution

huguesv:
Fix https://github.com/Microsoft/PTVS/issues/4028


huguesv:
There's no UI level test as far as I know. The ImportWizardTests are lower level and still all pass.

Issue #4052, opened 2018-04-12
Start debugging times out on Anaconda


huguesv:

karthiknadig:
It looks like the 5 second time out for connection we have is too short especially so on slower machines.

zooba:
@karthiknadig I wonder if there are any other signals we could use to keep waiting? Can the OS tell us if the process is running even if Python isn't fully initialized yet? Can we read process times and see if CPU time is changing?

Probably most important is to collect signal on how long this step is taking for people. I already mentioned that on your PRs. But if we can detect a hung or silently-dead process distinct from a slow-but-working process, that might help tidy this up.

PR #4054, opened 2018-04-12
Increase connection timeout for experimental debugger.

karthiknadig:
#4052


zooba:
Do we get any sort of telemetry event when timeout occurs? Would be helpful to know what % of people encounter timeouts (or just generally how long it takes to establish the connection - most distributions we've seen in the past for this kind of thing make it pretty obvious where the timeout should be set).

PR #4055, opened 2018-04-12
15.7 Increase connection timeout for experimental debugger.

karthiknadig:
#4052


zooba:
See my comment on #4054 - adding telemetry for 15.7 is totally okay.

PR #4056, opened 2018-04-13
Test improvements

AlexanderSher:
- Fix CodeFormatterTests so that test waits until analyzer finish parsing
- Switch ExtractMethodTests to real editor (100% pass)


Python/Tests/Utilities.Python/EventTaskSources.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
```
huguesv: missing header

PR #4067, opened 2018-04-16
Update version to 15.8.

huguesv:
Fix https://github.com/Microsoft/PTVS/issues/4057

@zooba does this require changes in VS repository as well?


zooba:
@huguesv There shouldn't be - all of our version references should be able to handle this change.

PR #4073, opened 2018-04-17
Fixes #3798 Migrate special completions to server

### LINKED ISSUE #3798: Migrate special completions to server
As part of #3484 we need to handle the various special completions in the language server rather than the IDE.

zooba:
Fixes #3798 Migrate special completions to server
Removes old completion analysis classes
Adds CompletionAnalysis to language server to identify correct completions based on context.


zooba:
Still making tweaks to ensure behavior does not regress, but this change is mostly here.

zooba:
The tooltip trimming is still busted, since the algorithm changed by the tests weren't updated. I've got it much closer, but it's not that big a deal and I don't have the brainpower right now to analyze the new algorithm.

PR #4078, opened 2018-04-18
15.7 Show warning when ptvsd installed is 3.* or lower

karthiknadig:


huguesv:
Internal bug is 602976

PR #4082, opened 2018-04-19
15.7 Disable experimental debugger by default and show info bar

karthiknadig:
There is a pending issue that, after the user hits "Enable". There are currently one of two ways that the experimental debugger can be enabled. Either restart VS or Open Python experimental options window (this will update `PythonToolsService.ExperimentalOptions`).


zooba:
We have another options set for “don’t show again” messages. Probably better to use that if you can find it (or ask me when I’m in and I’ll track it down… think it was “Suppress”-something-something)

Python/Product/Debugger/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -536,6 +539,10 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
                 return VSConstants.E_NOTIMPL;
             }
 
+            if (!ExperimentalOptions.UseVsCodeDebugger && !ExperimentalOptions.DontShowUseVsCodeDebuggerInfoBar) {
+                AddInfoBar();
```
huguesv: I think you'll end up with multiple info bars stack on top of each other if you do multiple debugging sessions and the user doesn't dismiss them.

huguesv:
@zooba are you thinking of SuppressDialogOptions? It lives on PythonToolsService, which isn't accessible from Debugger project. The options could be refactored into their own service, defined in VSInterpreters so it's accessible from Debugger project. Not sure if that's worth it for a throwaway option.

I was hoping we could do the info bar within PythonTools project, but couldn't think of a way to hook into "we just stopped debugging" other than in AD7Engine.

Though now I realize we could use DTE events (DebuggerEventsClass.OnEnterDesignMode Event). We'd have to be careful to only catch Python debugging with legacy debugger (not mixed-mode or experimental), and we'd have to ideally catch project debugging, standalone script debugging and AnyCode debugging. There's more of a risk of getting it wrong and annoying the wrong users.

karthiknadig:
Consider telemetry for enable and don’t show?

PR #4087, opened 2018-04-20
Fix memory leaks

AlexanderSher:
Fixes two cases of memory leaks:
- `LocatedVariableDef` held a link to the outdated `PythonAst`. This could also lead to the invalid calculation of span. Now it is updated to the latest one. 
- `ProjectEntry` of a closed file was held by `ReferenceDict`, `ParseQueue` and `PythonAnalyzer` instances.

Before the fix, 10 closing/reopening of the `test_io.py` file with 10 atomic changes of the `import` statements for each reopening resulted in a 200MB leak.

Also, some dead code is cleaned up.


AlexanderSher:
Ping

PR #4094, opened 2018-04-20
Code split + better change request synchronization + completion experience improvements

MikhailArkhipov:
- Splits Server into partial classes
- Remove custom doc change/completion synchronization and instead use new StreamJsonRpc sync context feature
- Preserve previous analysis dictionaries while new analysis still running (improves experience in VSC)
- Add small wait to completion letting analysis to complete if it is close
- Add ability to hide `__` members for VSC


Python/Product/Analysis/Analyzer/InterpreterScope.cs
```diff
@@ -278,7 +246,7 @@ public InterpreterScope(AnalysisValue av, InterpreterScope outerScope)
         }
 
         internal virtual void ClearVariables() {
-            _variables.Clear();
+            _variables = new AnalysisDictionary<string, VariableDef>();
```
AlexanderSher: `_variables` is directly referenced in cloning constructor. Is it affected?

AlexanderSher: Same thing with `_nodeScopes`

Python/Product/Analysis/LanguageServer/OpenFiles.cs
```diff
@@ -0,0 +1,115 @@
...
+                    toVersion,
+                    changes.Select(c => new DocumentChange {
+                        ReplacedSpan = c.range.GetValueOrDefault(),
+                        WholeBuffer = !c.range.HasValue,
+                        InsertedText = c.text
+                    })
+                ));
+
+                DidChangeTextDocumentParams? next = null;
+                lock (PendingChanges) {
```
AlexanderSher: `lock` on publicly available object

Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            _projectFiles.GetAnalysis(@params.textDocument, @params.position, @params._version, out var entry, out var tree);
+            TraceMessage($"Completions in {uri} at {@params.position}");
+
+            tree = GetParseTree(entry, uri, CancellationToken, out _) ?? tree;
+            // Give analysis opportunity to complete
+            if (!entry.IsComplete) {
+                entry.OnNewAnalysis += OnNewAnalysis;
+                _analysisCts = new TaskCompletionSource<bool>();
+                _analysisCts.Task.Wait(300);
```
AlexanderSher: We have `EventTaskSource` in `TestUtilities` which allows you to properly await on events. You can move it down here. See usage at `TestUtilities.Python.EventTaskSources` and `PythonToolsTests.CodeFormatterTests.CodeFormattingTest`.

Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -0,0 +1,136 @@
...
+            if (context.HasValue) {
+                var c = context.Value;
+                if (c._intersection) {
+                    opts |= GetMemberOptions.IntersectMultipleResults;
+                }
+            }
+            return opts;
+        }
+
+        private CompletionItem ToCompletionItem(MemberResult m, GetMemberOptions opts) {
```
AlexanderSher: `static`?

Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -0,0 +1,136 @@
...
+                // Place regular items first, advanced entries last
+                sortText = Char.IsLetter(m.Completion[0]) ? "1" : "2",
+                kind = ToCompletionItemKind(m.MemberType),
+                _kind = m.MemberType.ToString().ToLowerInvariant()
+            };
+
+            return res;
+        }
+
+        private CompletionItemKind ToCompletionItemKind(PythonMemberType memberType) {
```
AlexanderSher: `static`?

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -0,0 +1,108 @@
...
+                        new SourceLocation(loc.StartLine, loc.StartColumn),
+                        new SourceLocation(loc.EndLine ?? loc.StartLine, loc.EndColumn ?? loc.StartColumn)
+                    )
+                };
+            }
+
+            return res;
+        }
+
+        private SymbolKind ToSymbolKind(PythonMemberType memberType) {
```
AlexanderSher: `static`?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -116,14 +110,13 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
             _queue.Dispose();
         }
 
-        private void TraceMessage(IFormattable message) {
+        public void TraceMessage(string message) {
```
AlexanderSher: The reason behind using `IFormattable` is to avoid string allocation when `_traceLogging` is false.

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: Yep

Python/Product/Analysis/LanguageServer/Server.Completion.cs
MikhailArkhipov: Actually this one is no longer needed

Python/Product/VSCode/AnalysisVsc/Program.cs
```diff
@@ -30,10 +34,15 @@ internal static class Program {
                 using (var cin = Console.OpenStandardInput())
                 using (var cout = Console.OpenStandardOutput())
                 using (var server = new LanguageServer())
-                using (var rpc = JsonRpc.Attach(cout, cin, server)) {
+                using (var rpc = new JsonRpc(cout, cin, server)) {
+                    var ui = new UIService(rpc);
```
AlexanderSher: You have two `UIService` instances here.

Python/Product/Analysis/LanguageServer/Server.Completion.cs
MikhailArkhipov: OK

Python/Product/Analysis/LanguageServer/OpenFiles.cs
MikhailArkhipov: That's the way it used to be. Create separate lock object?

AlexanderSher: Yep.

Python/Product/VSCode/AnalysisVsc/Program.cs
```diff
@@ -51,5 +60,44 @@ internal static class Program {
...
+                _cts.Cancel();
+            }
+
+            private void QueueWorker() {
+                while(true) {
+                    _workAvailable.Wait(_cts.Token);
+                    if(_cts.IsCancellationRequested) {
+                        break;
+                    }
+                    if(_queue.TryDequeue(out var t)) {
```
AlexanderSher: `while (_queue.TryDequeue(out var t))`

Python/Product/VSCode/AnalysisVsc/Program.cs
```diff
@@ -51,5 +60,44 @@ internal static class Program {
             }
 #endif
         }
+
+        private sealed class SingleThreadSynchronizationContext : SynchronizationContext, IDisposable {
```
AlexanderSher: From what I understand, `JsonRpc` uses sync context to schedule continuation, so it guarantees the order of method execution start, but there can be any kind of mix on next await.

Python/Product/Analysis/Analyzer/InterpreterScope.cs
zooba: This isn't a problem because we only modify the fields on the analysis thread. On any query threads, we shouldn't be relying on reference matches between the contents of these dictionaries.

Python/Product/Analysis/LanguageServer/OpenFiles.cs
zooba: Does it even need to be publicly visible? It should only be relevant within this handler.

Python/Product/Analysis/LanguageServer/ProjectFiles.cs
```diff
@@ -0,0 +1,73 @@
...
+            entry = GetEntry(document) as ProjectEntry;
+            if (entry == null) {
+                throw new LanguageServerException(LanguageServerException.UnsupportedDocumentType, "unsupported document");
+            }
+            var parse = entry.GetCurrentParse();
+            tree = parse?.Tree;
+            if (expectedVersion.HasValue && parse?.Cookie is VersionCookie vc) {
+                if (vc.Versions.TryGetValue(GetPart(document.uri), out var bv)) {
+                    if (bv.Version != expectedVersion.Value) {
+                        // throw new LanguageServerException(LanguageServerException.MismatchedVersion, $"document is at version {bv.Version}; expected {expectedVersion.Value}");
```
zooba: If we're not going to throw, we shouldn't be returning its `tree` either.

Python/Product/Analysis/LanguageServer/Server.Completion.cs
zooba: If we've never seen a complete analysis for the entry, we should just return a list marked incomplete (which may not even be empty if we've published the in-progress analysis correctly). Better to get an immediate response from the server than to be left hanging.

Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -0,0 +1,96 @@
...
+                TraceMessage($"No analysis found for {uri}");
+                return new CompletionList();
+            }
+
+            var opts = GetOptions(@params.context);
+            var ctxt = new CompletionAnalysis(analysis, tree, @params.position, opts, this);
+            var members = ctxt.GetCompletionsFromString(@params._expr) ?? ctxt.GetCompletions();
+
+            if (_settings.SuppressAdvancedMembers) {
+                members = members.Where(m => !m.label.StartsWith("__"));
```
zooba: This check should also include `EndsWithOrdinal("__")`. Member prefixed with `__` are protected against accidental overwriting by subclasses (via automatic name mangling), but they are not "advanced" members.

Python/Product/Analysis/LanguageServer/Server.SignatureHelp.cs
```diff
@@ -0,0 +1,153 @@
...
+                        }
+                    }
+                    break;
+            }
+
+            si._returnTypes = (overload as IOverloadResult2)?.ReturnType.OrderBy(k => k).ToArray();
+            return si;
+        }
+
+        private MarkupKind SelectBestMarkup(IEnumerable<MarkupKind> requested, params MarkupKind[] supported) {
```
zooba: Seems like this will be used from a few places - should we have a "Server.PrivateHelpers.cs" partial class as well?

Python/Product/Analysis/MemberResult.cs
```diff
@@ -175,6 +173,34 @@ public struct MemberResult {
             }
             return result.ToString().Trim();
         }
+
+        private static string FormatModuleDocumentation(string doc) {
+            // Module doc does not nave example/signature lines like a function
+            // so just make it flow nicely in the tooltip by removing like breaks.
+            // Preserve double breaks and breaks before the indented text
+            var lines = doc.Replace("\r", string.Empty).Split('\n');
```
zooba: At least use `StringReader.ReadLine()` here (though I suspect this may be a worthwhile place to optimize to reduce response time of completion lists - then again, resolving documentation lazily would also help significantly there)

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -181,12 +179,14 @@ private class ActivePythonParse : IPythonParse {
...
 
         internal void RaiseOnNewAnalysis() {
+            IsComplete = true;
             OnNewAnalysis?.Invoke(this, EventArgs.Empty);
         }
 
         public int AnalysisVersion { get; private set; }
 
         public bool IsAnalyzed => Analysis != null;
+        internal bool IsComplete { get; private set; }
```
zooba: As mentioned elsewhere, I don't think this is valuable. If there is an available Analysis, we should treat it as complete (because it's the previous complete one, or the best available)

Python/Product/VSCode/AnalysisVsc/Program.cs
zooba: That's correct, and it's what we are aiming for. When reviewing command handlers, please make sure that all critical state (locks/counters/etc.) are updated before the first `await`.

Python/Tests/Utilities.Python/EventTaskSources.cs
```diff
@@ -14,6 +14,8 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using Microsoft.PythonTools.Analysis.Infrastructure;
```
zooba: Seems like an unnecessary change?

Python/Product/VSCode/AnalysisVsc/Program.cs
MikhailArkhipov: Good catch

Python/Product/Analysis/LanguageServer/Server.Completion.cs
MikhailArkhipov: OK

Python/Product/VSCode/AnalysisVsc/Program.cs
AlexanderSher: No, I don't want to change the outer loop. I meant that we'd like to drain the queue and wait only after it is empty.

Python/Product/Analysis/ProjectEntry.cs
MikhailArkhipov: This basically to give chance to analysis to complete. In VSC there is no way to re-trigger completion session or augment the list later, so the list should be there on the first try.  None of this solves the problem completely, but improves the experience a bit.

Python/Tests/Utilities.Python/EventTaskSources.cs
MikhailArkhipov: Actually this may need to remain a copy so we don't reference out-of-proc stuff

Python/Product/VSCode/AnalysisVsc/Program.cs
MikhailArkhipov: OK

Python/Product/Analysis/LanguageServer/Server.Completion.cs
zooba: I'm thinking of the `isIncomplete` flag on the resulting list: https://microsoft.github.io/language-server-protocol/specification#textDocument_completion

Python/Product/Analysis/LanguageServer/ProjectFiles.cs
zooba: We should already have a tree at this point, but this is a better tree if a specific version is required.

Python/Product/Analysis/LanguageServer/OpenFiles.cs
MikhailArkhipov: Made private

Python/Product/Analysis/MemberResult.cs
```diff
@@ -22,6 +22,7 @@
 using Microsoft.PythonTools.Analysis.Values;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
+using System.IO;
```
zooba: Sorted in wrong place

Python/Product/Analysis/MemberResult.cs
```diff
@@ -178,24 +179,30 @@ public struct MemberResult {
...
-                    } else {
-                        if (lines[i + 1].Length > 0) {
-                            sb.Append(line);
-                            sb.Append(' ');
+            using (var sr = new StringReader(doc)) {
+                while (true) {
+                    var line = sr.ReadLine();
+                    if(line == null) {
+                        break;
+                    }
```
zooba: These five lines work nicely in a `for` loop:
```
for (var line = sr.ReadLine(); line != null; line = sr.ReadLine())
```

Python/Product/Analysis/Values/ClassInfo.cs
```diff
@@ -423,7 +423,9 @@ internal class ClassInfo : AnalysisValue, IReferenceableContainer, IHasRichDescr
             var result = new Dictionary<string, IAnalysisSet>(Scope.VariableCount);
 
             foreach (var v in Scope.AllVariables) {
-                v.Value.ClearOldValues();
+                if ((options & GetMemberOptions.ForEval) != GetMemberOptions.ForEval) {
```
zooba: We have extension methods for these checks already. (In particular, it's much easier to read `!options.ForEval()` than the mask and `!=` comparison and figure out whether the flag should be set or unset.)

Python/Product/Analysis/Analyzer/InterpreterScope.cs
```diff
@@ -246,7 +246,7 @@ internal bool RemoveVariable(string name, out VariableDef value)
         }
 
         internal virtual void ClearVariables() {
-            _variables = new AnalysisDictionary<string, VariableDef>();
+            _variables.Clear();
```
zooba: Not sure that we should be reverting back to this? Sure, it'll only affect module-level variables, but those are important too :)

zooba:
Don't merge this yet - we want to take a build without this so we can update the next preview.

Python/Product/Analysis/Values/ClassInfo.cs
MikhailArkhipov: OK, will update next PR. Didn't find those for `GetMembers`

PR #4110, opened 2018-04-24
Fixes #4096 Add test that default analyzer is used for non project file

### LINKED ISSUE #4096: Add test that default analyzer is used for non project file
No project
Change global default to 3.6
Open File, Python file with type hints
Check that there are no syntax errors

zooba:
Fixes #4096 Add test that default analyzer is used for non project file
Make default analyzer files follow environment changes
Ensures stub files are always parsed correctly.

### DIFF
--- a/Python/Tests/Core.UI/ErrorListTaskListTests.cs
+++ b/Python/Tests/Core.UI/ErrorListTaskListTests.cs
@@ -220,5 +220,18 @@ namespace PythonToolsUITests {
             app.WaitForTaskListItems(typeof(SVsErrorList), 0);
             app.WaitForTaskListItems(typeof(SVsTaskList), 0);
         }
+
+        /// <summary>
+        /// Make sure *.pyi files ignore the active Python version
+        /// </summary>
+        public void ErrorListEmptyForValidTypingFile(VisualStudioApp app) {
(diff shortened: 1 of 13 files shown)


Python/Tests/Core.UI/ErrorListTaskListTests.cs
```diff
@@ -220,5 +220,18 @@ internal struct TaskItemInfo {
             app.WaitForTaskListItems(typeof(SVsErrorList), 0);
             app.WaitForTaskListItems(typeof(SVsTaskList), 0);
         }
+
+        /// <summary>
+        /// Make sure *.pyi files ignore the active Python version
+        /// </summary>
+        public void ErrorListEmptyForValidTypingFile(VisualStudioApp app) {
+            var project = app.OpenProject(@"TestData\Typings.sln");
+            project.ProjectItems.Item("mymod.pyi").Open();
```
huguesv: how does it find the error in usermod.py if the only file opened is mymod.pyi? Do you need to also open usermod.py?

zooba: All `<Compile>` files in a project contribute to the error list, but right now stub files only do when they are open (they default to `<Content>` or `<None>`, IIRC).

Issue #4025, opened 2018-04-10
Editing django run/debug server arguments breaks launch


huguesv:
Create django project.
In project settings, debug, edit the debug server command arguments by adding a space before `--settings`.
Save project.
Note: Now the .pyproj file contains `PythonDebugWebServerCommandArguments`  property definition (it doesn't until you edit the project settings).
F5 to debug

Result: `ModuleNotFoundError`

Workaround: Edit the .pyproj and replace `$(DjangoSettingsModule)` in `<PythonDebugWebServerCommandArguments>` with its actual value.

Also reported at https://developercommunity.visualstudio.com/content/problem/224142/djangowebproject-runserver-command-fails.html

zooba:
That property should be in the template by default (and `$(MSBuildProjectName)` isn't necessarily safe - it needs to be importable)

karthiknadig:
I had missed this bit of info. The project XML seems to have `DjangoSettingsModule` defined but in a different property group below the one where `PythonDebugWebServerCommandArguments` gets defined.  May be we should change the set property to add it to the end of the property group that contains `DjangoSettingsModule`.

PR #4131, opened 2018-04-25
Fix for editing django debug/run breaks launch

karthiknadig:
#4025

### DIFF
--- a/Common/Product/SharedProject/ProjectNode.cs
+++ b/Common/Product/SharedProject/ProjectNode.cs
@@ -2044,7 +2044,6 @@ namespace Microsoft.VisualStudioTools.Project {
             Utilities.ArgumentNotNull("propertyName", propertyName);
             Utilities.ArgumentNotNull("afterProperty", afterProperty);
 
-            Utilities.ArgumentNotNull("propertyName", propertyName);
             Site.GetUIThread().MustBeCalledFromUIThread();
 
             var oldValue = GetUnevaluatedProperty(propertyName) ?? string.Empty;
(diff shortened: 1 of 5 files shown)


Python/Product/PythonTools/PythonTools/WebProject/PythonWebLauncherProvider.cs
```diff
@@ -104,7 +104,7 @@ string environmentProperty
                 if (string.IsNullOrEmpty(config.ScriptArguments)) {
                     config.ScriptArguments = args;
                 } else {
-                    config.ScriptArguments = config.ScriptArguments + " " + args;
+                    config.ScriptArguments = args + " " + config.ScriptArguments;
```
karthiknadig: This was needed to prevent django launch from breaking if script arguments were provided:
image

This does not fix the issue where we may have script arguments repeated twice. However, the user can remove `$(CommandLineArgs)` from the end of `DebugWebServerArguments` or `RunWebServerArguments` to fix that.

Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -2040,7 +2040,41 @@ internal enum EventTriggering {
             this.currentConfig = null;
         }
 
+        public virtual void SetOrAddPropertyAfter(string propertyName, string propertyValue, string afterProperty) {
+            Utilities.ArgumentNotNull("propertyName", propertyName);
+            Utilities.ArgumentNotNull("afterProperty", afterProperty);
+
+            Utilities.ArgumentNotNull("propertyName", propertyName);
```
huguesv: duplicate check

Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -2040,7 +2040,41 @@ internal enum EventTriggering {
...
+            if (!this.QueryEditProjectFile(false)) {
+                throw Marshal.GetExceptionForHR(VSConstants.OLE_E_PROMPTSAVECANCELLED);
+            }
+
+            var oldProp = this.buildProject.GetProperty(propertyName);
+            if (oldProp == null || oldProp.IsImported) {
+                var propertyGroups = from g in this.buildProject.Xml.PropertyGroups
+                                     where g.Children.Where(c => c.ElementName == afterProperty).Any()
+                                     select g;
+                var group = propertyGroups.Any() ? propertyGroups.First() : this.buildProject.Xml.PropertyGroups.First();
```
huguesv: If it's defined in multiple property groups, should we add after the last group to get the correct value? Also, what if there are no property groups at all, `group` will be `null` and it will crash.
How about this?
```
var group = propertyGroups.LastOrDefault();
if (group != null) {
    var newProp = group.SetProperty(propertyName, propertyValue);
} else {
    this.buildProject.SetProperty(propertyName, propertyValue);
}
```

karthiknadig: I agree. I will update this.

PR #4143, opened 2018-04-27
Update django app "new item templates" to include subfolder under templates dir

### LINKED ISSUE #4079: Django app template should put index.html in a subfolder of templates
Instead of this structure:
image

We should do this:
image

Note that `django startapp myapp` doesn't create a templates folder, so we are free to change it, as we are already different from command line tool here.

To understand why, see _Template namespacing_ note in https://docs.djangoproject.com/en/2.0/intro/tutorial03/#write-views-that-actually-do-something

Summary is that on multi apps, you need to namespace in order to guarantee loading the correct template.

huguesv:
Fix #4079

Update django project ui tests to ensure they use an environment with django installed.

### DIFF
--- a/Python/Tests/DjangoUITests/DjangoProjectUITests.cs
+++ b/Python/Tests/DjangoUITests/DjangoProjectUITests.cs
@@ -31,7 +31,7 @@ using PythonConstants = Microsoft.PythonTools.PythonConstants;
 
 namespace DjangoUITests {
     public class DjangoProjectUITests {
-        public void NewDjangoProject(VisualStudioApp app) {
+        public void NewDjangoProject(VisualStudioApp app, DjangoInterpreterSetter interpreterSetter) {
             var project = app.CreateProject(
                 PythonVisualStudioApp.TemplateLanguageName,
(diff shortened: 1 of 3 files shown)


Python/Tests/DjangoUITests/DjangoProjectUITests.cs
```diff
@@ -186,7 +191,7 @@ public class DjangoProjectUITests {
             Assert.IsNotNull(project.ProjectItems.Item("NewPage.html"));
         }
 
-        public void StartNewAppDuplicateName(VisualStudioApp app) {
+        public void StartNewAppDuplicateName(VisualStudioApp app, DjangoInterpreterSetter interpreterSetter) {
```
AlexanderSher: I don't see where this parameter is used.

huguesv: It's instantiated via reflection by HostedPythonToolsTestRunner:
https://github.com/Microsoft/PTVS/blob/master/Common/Tests/Utilities.UI/HostedPythonToolsTestRunner.cs#L120

PR #4152, opened 2018-05-01
Prevent crash when search path changes.

### LINKED ISSUE #4147: Crash when search paths change
Seeing this call stack in some crash reports:

```
System.ObjectDisposedException: Cannot access a disposed object.
   at System.Threading.TimerQueueTimer.Change(UInt32 dueTime, UInt32 period)
   at System.Threading.Timer.Change(Int32 dueTime, Int32 period)
   at Microsoft.PythonTools.SearchPathManager.Microsoft.VisualStudio.Shell.Interop.IVsFileChangeEvents.DirectoryChanged(String pszDirectory)
   at Microsoft.VisualStudio.Services.FileChange.ChangeNotificationBuffer.BufferingSink.<>c__DisplayClass12_0.<SendDirectoryNotifications>b__0()
```

huguesv:
Fix #4147

### DIFF
--- a/Python/Product/PythonTools/PythonTools/SearchPathManager.cs
+++ b/Python/Product/PythonTools/PythonTools/SearchPathManager.cs
@@ -16,7 +16,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Threading;
 using Microsoft.PythonTools.Infrastructure;
@@ -43,12 +42,13 @@ namespace Microsoft.PythonTools {
 
         protected virtual void Dispose(bool disposing) {
             if (disposing) {
+                uint[] cookies;
                 lock (_paths) {
-                    foreach (var p in _paths) {
-                        Unwatch(p.Cookie);
-                    }
+                    cookies = _paths.Select(p => p.Cookie).ToArray();
                     _paths.Clear();
                 }
+
+                Unwatch(cookies);
 
                 var timer = _notifyChangeTimer;
                 if (timer != null) {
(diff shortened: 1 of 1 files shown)


AlexanderSher:
This type has several other potential issues:
- `IVsFileChangeEx.AdviseDirChange` and `IVsFileChangeEx.UnadviseDirChange` are called under lock
- Type registers itself into finalizer queue while has no unmanaged resources. 
- `SearchPathManager.Clear` doesn't call `Unwatch`, though it isn't used right now

huguesv:
I fixed the last 2, I'll try to fix the first one, and I'll probably delete more functions that are unused so I don't have to fix them.

Python/Product/PythonTools/PythonTools/SearchPathManager.cs
```diff
@@ -41,10 +41,6 @@ class SearchPathManager : IVsFileChangeEvents, IDisposable {
             GC.SuppressFinalize(this);
```
AlexanderSher: You can make this class `sealed`, remove `GC.SuppressFinalize(this);` and remove `protected virtual void Dispose` overload.
This is pretty much everything you need:
```cs
public void Dispose() {
    lock (_paths) {
        foreach (var p in _paths) {
            Unwatch(p.Cookie);
        }
        _paths.Clear();
    }

    var timer = _notifyChangeTimer;
    if (timer != null) {
        timer.Dispose();
    }
}
```

huguesv: Last time you suggested changes to IDisposable pattern, @zooba was saying that our static code analysis rules were complaining about this.

How do I run static code analysis on my local machine? Enabling it on the project causes tons of errors "The rule X in PTVS\Build\xTVS.ruleset could not be found"

huguesv: I'll change it, and if we get warnings we can change it back.

PR #4155, opened 2018-05-02
Cleaned up contribution and included PTVS-side stubs

rgesteve:


Python/Product/ExternalProfilerDriver/VTuneToDWJSON.cs
```diff
@@ -0,0 +1,238 @@
...
+                    };
+                    fi.frameIPs.Insert(0, tr.Translate(s.TOSFrame.Module, s.TOSFrame.Function));
+                    chains.Add(fi);
+
+                    idx++;
+                }
+            }
+
+            ProcessSpec proc = new ProcessSpec {
+                name = "python36.dll",
```
zooba: Is there a way to avoid hard-coding this?

Python/Product/Profiling/Profiling/ProfilingTargetView.cs
```diff
@@ -79,6 +84,7 @@ public ProfilingTargetView(IServiceProvider serviceProvider, ProfilingTarget tem
                 Project = new ProjectTargetView(template.ProjectTarget);
                 IsStandaloneSelected = false;
                 IsProjectSelected = true;
+                UseVTune = false;              // TODO -- why not allow projects?
```
zooba: Good question! All we do with the project is use it to find all the settings we'd provide for the standalone target.

rgesteve:
4c1b5f7 fixes lingering #ifs and errant hyphen in packages.config.  Working on addressing the rest of the comments, of which, only the "standalone target" handling should take a little time (mostly because I haven't taken the time to understand it at all :) ).

rgesteve:
As far as I can tell, 02d638d addresses the feedback above, updating the helper command-line utility (still unsigned).  Only change that remains to be made is the hard-coding of the python process, which is part of the upcoming dwjson generation.

Python/Product/ExternalProfilerDriver/ExternalProfilerDriver.csproj
```diff
@@ -0,0 +1,72 @@
...
+        <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+        <FileUpgradeFlags>
+        </FileUpgradeFlags>
+        <UpgradeBackupLocation>
+        </UpgradeBackupLocation>
+        <OldToolsVersion>14.0</OldToolsVersion>
+      </PropertyGroup>
+    </When>
+  </Choose>
+  <Import Project="..\ToolProjectBefore.settings" />
```
zooba: Just use `ProjectBefore` and `ProjectAfter.settings`

PR #4168, opened 2018-05-04
Fixes #4115: Completions dropdown shouldn't appear after def and class

### LINKED ISSUE #4115: Completions dropdown shouldn't appear after def and class
Regression in 15.8 branch

New Python application
Type: `def ` or `class `
Result: after typing the space, completions drop down appears.

Example: type `def he(` and you'll get `def help()` instead of `def he()`, so you're forced to dismiss with ESC to get what you want.

zooba:
Fixes #4115: Completions dropdown shouldn't appear after def and class
Fixes handling of incomplete def and class statements
Fixes default commit of overrides
Fixes references to import names

### DIFF
--- a/Python/Product/Analysis/LanguageServer/Structures.cs
+++ b/Python/Product/Analysis/LanguageServer/Structures.cs
@@ -658,7 +658,8 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
         /// </summary>
         public string _expr;
         /// <summary>
-        /// When true, completions should commit by default.
+        /// When true, completions should commit by default. When false, completions
+        /// should not commit. If unspecified the client may decide.
         /// </summary>
         public bool? _commitByDefault;
     }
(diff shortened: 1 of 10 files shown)


Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -657,6 +657,10 @@ public struct CompletionList {
         /// The expression that members are being displayed for.
         /// </summary>
         public string _expr;
+        /// <summary>
+        /// When true, completions should commit by default.
```
AlexanderSher: What does `null` mean?

zooba: Up to the IDE in that case - this is kind-of an option in VS, but in some cases we want to force it to false (e.g. override completions).

I'll clarify the docs and make it null most of the time.

PR #4177, opened 2018-05-07
master to 15.8

huguesv:
We'll be releasing from the 15.8 branch, so this takes the latest from master.

There were no merge conflicts.


huguesv:
I'm not authorized to merge the PR? That's weird. It must be per-branch permission. I'm the one that created the branch in the first place!

PR #4162, opened 2018-05-03
Pythia integration

MikhailArkhipov:

PR #4182, opened 2018-05-08
Tooltip fixes

MikhailArkhipov:
Some more work for https://github.com/Microsoft/vscode-python/issues/1485
Effectively https://github.com/Microsoft/PTVS/pull/4162 sans Pythia


Python/Product/Analysis/LanguageServer/Server.SignatureHelp.cs
```diff
@@ -88,34 +88,66 @@ public sealed partial class Server {
...
+
+            si._returnTypes = (overload as IOverloadResult2)?.ReturnType.OrderBy(k => k).ToArray();
+
             if (_clientCaps?.textDocument?.signatureHelp?.signatureInformation?._shortLabel ?? false) {
                 si.label = overload.Name;
             } else {
+                var doc = overload.Documentation;
+                // Some function contain signature in the documentation. Example: print.
+                // We want to use that signature in VS Code since it contains all arguments.
+                if (si.parameters.Length == 0 && !string.IsNullOrEmpty(doc) && doc.StartsWithOrdinal($"{overload.Name}(")) {
```
AlexanderSher: Can we have space between `{overload.Name}` and `(`?

Python/Product/Analysis/LanguageServer/Server.SignatureHelp.cs
```diff
@@ -88,34 +88,66 @@ public sealed partial class Server {
...
+        private static SignatureInformation GetSignatureFromDoc(string doc) {
+            var si = new SignatureInformation();
+            var firstLineBreak = doc.IndexOfAny(new[] { '\r', '\n' });
+            si.label = firstLineBreak > 0 ? doc.Substring(0, firstLineBreak) : doc;
+            si.documentation = doc.Substring(si.label.Length).TrimStart();
+            si.parameters = GetParametersFromDoc(si.label);
+            return si;
+        }
+
+        private static ParameterInformation[] GetParametersFromDoc(string doc) {
```
AlexanderSher: This method produces lots of garbage strings. Do we care?

Python/Product/Analysis/LanguageServer/Server.SignatureHelp.cs
MikhailArkhipov: Normally no since Python is typically formatted to PEP.

Python/Product/Analysis/LanguageServer/Server.SignatureHelp.cs
MikhailArkhipov: It is called once to get signature when user types ( after the function name. Long argument lists are rare, maybe 4-5 max. Can be parsed in place technically, but with more complicated code.

Python/Product/Analysis/LanguageServer/Server.SignatureHelp.cs
```diff
@@ -88,34 +88,66 @@ public sealed partial class Server {
...
+            var openBrace = doc.IndexOf('(');
+            var closeBrace = doc.LastIndexOf(')');
+
+            if (openBrace > 0 && closeBrace > 0) {
```
huguesv: sanity check that `closeBrace>openBrace` too? otherwise you will have a negative length

PR #4206, opened 2018-05-11
Port to 15.8

huguesv:
I took latest master, there were no merge conflicts.


AlexanderSher:
Do you want to port async packages as well?

huguesv:
Yes, I want it in early previews so we have enough time to catch issues by 15.8 RTM.

Issue #4118, opened 2018-04-25
StackOverflowException while retrieving overloads from analysis


lostmsu:
Detailed repro steps:

1. Get [IronPython3](https://github.com/IronLanguages/ironpython3), commit d2df6fbe9f70c2d2b2b0487682731820d10e4704
2. Install VS 15.6.6 (latest release as of today) with Python 3.6 x64 workload
3. Ensure you have only one Python 3.6 environment in "Python Environments" tool window (might be crucial for I never specify interpreter details)
4. Run the following (replace IronPythonStdLib with a proper path to Src/StdLib folder of IronPython3 repo)
```csharp
var interpreterFactory = InterpreterFactoryCreator.CreateAnalysisInterpreterFactory(
                            PythonLanguageVersion.V36.ToVersion());
PythonAnalyzer analyzer = await PythonAnalyzer.CreateAsync(interpreterFactory).ConfigureAwait(false);
var modules = new SortedList<string, IPythonProjectEntry>();
var parseTasks = new List<Task<(IPythonParse, IPythonProjectEntry)>>();
SearchOption searchOption = true ? SearchOption.TopDirectoryOnly : SearchOption.AllDirectories;
const string IronPythonStdLib = "IP3/Src/StdLib";
foreach (string sourcePath in Directory.EnumerateFiles(IronPythonStdLib,
    searchPattern: "*.py", searchOption: searchOption)) {
    string moduleName = Path.GetFileNameWithoutExtension(sourcePath);
    if (Debugger.IsAttached && moduleName.StartsWith("test_"))
        continue;
    IPythonProjectEntry module = analyzer.AddModule(moduleName, sourcePath);
    modules.Add(sourcePath, module);
    var p = module.BeginParse();

    var parseTask = Task.Run(() => {
        var errorSink = new CollectingErrorSink();

        using (var reader = new StreamReader(sourcePath)) {
            Parser parser = Parser.CreateParser(reader, analyzer.LanguageVersion,
                new ParserOptions { BindReferences = true, ErrorSink = errorSink });
            p.Tree = parser.ParseFile();
        }

        foreach (ErrorResult error in errorSink.Errors)
            Console.Error.WriteLine(error.ToString());

        return (p, module);
    });

    parseTasks.Add(parseTask);
}

while (parseTasks.Count > 0) {
    var completedParse = await Task.WhenAny(parseTasks);
    var (parse, module) = completedParse.Result;
    parse.Complete();
[...]

PR #4228, opened 2018-05-16
added reentry guard to FunctionInfo.Overloads

lostmsu:
Added reentry guard to FunctionInfo.Overloads when enumerating over
__wrapped__ (bug #4118). This fixes the immediate issue. It is possible,
that __wrapped__ attribute of FunctionInfo must not contain self
by design. In that case a different fix is required.


huguesv:
Thanks! Someone should get around to review it next week. Many in our team are still at PyCon US.

Python/Product/Analysis/Values/FunctionInfo.cs
```diff
@@ -450,69 +450,75 @@ private class StringArrayComparer : IEqualityComparer<string[]> {
...
-        public override IEnumerable<OverloadResult> Overloads {
-            get {
-                if (_functionAttrs != null && _functionAttrs.TryGetValue("__wrapped__", out VariableDef wrapped)) {
-                    foreach (var o in wrapped.TypesNoCopy.SelectMany(n => n.Overloads)) {
-                        yield return o;
-                    }
+        public override IEnumerable<OverloadResult> Overloads => this.EnumerateOverloads(new HashSet<FunctionInfo>());
+
+        IEnumerable<OverloadResult> EnumerateOverloads(HashSet<FunctionInfo> beenThrough) {
+            if (!beenThrough.Add(this))
```
zooba: We have `Push()`/`Pop()` functions on each value type for this kind of protection. Any reason it isn't sufficient here? (I'm totally willing to believe it's not sufficient, but I'd rather not have multiple ways of doing the same thing throughout the codebase if it's fine.)

Python/Product/Analysis/Values/FunctionInfo.cs
```diff
@@ -450,69 +450,75 @@ private class StringArrayComparer : IEqualityComparer<string[]> {
             }
         }
 
-        public override IEnumerable<OverloadResult> Overloads {
-            get {
-                if (_functionAttrs != null && _functionAttrs.TryGetValue("__wrapped__", out VariableDef wrapped)) {
-                    foreach (var o in wrapped.TypesNoCopy.SelectMany(n => n.Overloads)) {
-                        yield return o;
-                    }
+        public override IEnumerable<OverloadResult> Overloads => this.EnumerateOverloads(new HashSet<FunctionInfo>());
```
zooba: Can we avoid the HashSet allocation in cases until we find a `__wrapped__` member?

Python/Product/Analysis/Values/FunctionInfo.cs
```diff
@@ -450,69 +450,75 @@ private class StringArrayComparer : IEqualityComparer<string[]> {
...
+
+        IEnumerable<OverloadResult> EnumerateOverloads(HashSet<FunctionInfo> beenThrough) {
+            if (!beenThrough.Add(this))
+                yield break;
```
zooba: We use braces on all multi-line blocks.

lostmsu:
@zooba , I did not know about Push/Pop. Are they intended to be used for internal purposes only (e.g. should an implementation of Overloads call them)? Its sort of weird, if a client must call one before accessing a public property.

Also, I looked at their code, and noticed, that they are implemented strangely. _processing field is only used inside these methods, and since they only add this, why does it even have to be a hashtable. Seem like `bool` would be sufficient - it either stores nothing, or self.

zooba:
They are thread static, so it only gets created once per thread. They should also be used internal, so `Overloads` should call it before potentially recursing, and return its default value if `Push` returns false.

lostmsu:
@zooba , updated to use Push/Pop. Seems to be enough.

zooba:
What did you think about the other suggestion of only protecting when we know we're going to recurse? (i.e. after we've gotten the `__wrapped__` value)

lostmsu:
@zooba is there any advantage of not protecting the other parts? They might get recursive later.

zooba:
Just performance. Most of the time there won't be a `__wrapped__` member, so we can save recursion checking (this code isn't _extremely_ performance sensitive, but it is in a fairly critical path for some scenarios, so we don't want to unnecessarily slow things down).

When we make the other parts recursive later we can also protect them later :)

lostmsu:
@zooba , fair enough. I've verified that still fixes my issue.

lostmsu:
Nitpicker in me wants an IDisposable, so that I could do something like using(this.ReentryGuard()){ ... }

PR #4241, opened 2018-05-21
Specify correct encoding in WebClient when calling into GitHub API.

### LINKED ISSUE #4240: Cookiecutter wrong encoding
Filed at https://developercommunity.visualstudio.com/content/problem/256666/cookiecutter中文乱码.html

Load the following template in cookiecutter:
https://github.com/chenyinxin/cookiecutter-bitadmin-core

And compare the description string that appears in the cookiecutter window with the description that appears at the top of the github page.

The description in the cookiecutter window is all garbled.

huguesv:
Fix #4240

### DIFF
--- a/Python/Product/Cookiecutter/Model/GitHubClient.cs
+++ b/Python/Product/Cookiecutter/Model/GitHubClient.cs
@@ -16,10 +16,10 @@
 
 using System;
 using System.Net;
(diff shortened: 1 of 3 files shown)


huguesv:
I've updated it to the same as PyPI.

PR #4247, opened 2018-05-23
pip fixes

### LINKED ISSUE #4194: Package management broken on pip v10
I created a new python application
I created a 2.7 32bit based virtualenv
Right-click Install Package, requests
After that failed I tried to upgrade pip twice
In the screenshot, you can see pip twice in the list of packages, that was there before I tried to upgrade it.

```

----- Installing 'requests==2.18.4' -----
  File "<string>", line 1, in <module>
AttributeError: 'module' object has no attribute 'main'
----- Failed to install 'requests==2.18.4' -----

----- Installing 'pip' -----
  File "<string>", line 1, in <module>
AttributeError: 'module' object has no attribute 'main'
----- Failed to install 'pip' -----

----- Installing 'pip' -----
  File "<string>", line 1, in <module>
AttributeError: 'module' object has no attribute 'main'
----- Failed to install 'pip' -----

```

image

### LINKED ISSUE #4246: Crash with InvalidOperationException when installing requirements on Python 3.3 venv
Create new python app
Create requirements.txt: `bottle`
Add virtual env, select Python 3.3 as base
Keep install packages option checked

You'll get the crash below.

Alternative: Uncheck install packages option, and when it's done creating, right click on venv and Install Requirements.txt.

I have implemented a fix for both, I'm filing so we can keep track of when this was fixed.

```
Unhandled exception in Microsoft.PythonTools.Project.AddVirtualEnvironmentOperation.Run (C:\Projects\GitHub\huguesv\PTVS\Python\Product\PythonTools\PythonTools\Project\AddVirtualEnvironmentOperation.cs:116)

System.InvalidOperationException: Operation cannot be started because the environment is not configured.

   at Microsoft.PythonTools.Interpreter.PipPackageManager.<AbortIfNotReady>d__26.MoveNext() in C:\Projects\GitHub\huguesv\PTVS\Python\Product\VSInterpreters\PackageManager\PipPackageManager.cs:line 108

   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()

   at Microsoft.PythonTools.Interpreter.PipPackageManager.<InstallAsync>d__37.MoveNext() in C:\Projects\GitHub\huguesv\PTVS\Python\Product\VSInterpreters\PackageManager\PipPackageM......
```

huguesv:
Fix #4194 Package management broken on pip v10
Fix #4246 Crash with InvalidOperationException when installing requirements on Python 3.3 venv

On 2.7 and up, use `python -m pip` instead of `python -c 'import pip;pip.main()'` because that's needed for pip v10.

Change pip_downloader so that it fetches the appropriate get_pip.py, since the main one is v10 and doesn't work on 2.6.

Also change pip_downloader to not force upgrade pip on IronPython, because those newer versions of pip don't work for it.

I verified that we still need to use `pip.main()` on 2.6, but that's ok since pip v10 is not supported on that anyway.

The InvalidOperationException crashes were due to pip not being installed, and no one caught the exception. It now ensures that the package manager is 'prepared', before trying to install.

Add an info bar in the package manager to let the user know that pip likely won't work for Python 2.6, 3.1, 3.2. Click learn more to go to https://go.microsoft.com/fwlink/?linkid=874576 which is the packages tab documentation. I'll email Kraig to add a note on this to the docs.

Updated screenshot:
image

### DIFF
--- a/Python/Product/EnvironmentsList/PipExtension.xaml.cs
+++ b/Python/Product/EnvironmentsList/PipExtension.xaml.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Runtime.ExceptionServices;
@@ -38,6 +39,7 @@ namespace Microsoft.PythonTools.EnvironmentsList {
         public static readonly ICommand UpgradePackage = new RoutedCommand();
         public static readonly ICommand UninstallPackage = new RoutedCommand();
         public static readonly ICommand InstallPip = new RoutedCommand();
+        public static readonly ICommand PipSecurityLearnMore = new RoutedCommand();
 
         private readonly PipExtensionProvider _provider;
         private readonly Timer _focusTimer;
@@ -182,6 +184,15 @@ namespace Microsoft.PythonTools.EnvironmentsList {
             }
         }
 
+        private void PipSecurityLearnMore_CanExecute(object sender, CanExecuteRoutedEventArgs e) {
+            e.CanExecute = true;
+            e.Handled = true;
+        }
+
+        private void PipSecurityLearnMore_Executed(object sender, ExecutedRoutedEventArgs e) {
+            Process.Start("https://go.microsoft.com/fwlink/?linkid=874576");
+        }
+
         private void InstallPip_CanExecute(object sender, CanExecuteRoutedEventArgs e) {
             e.CanExecute = _provider.CanExecute;
             e.Handled = true;
@@ -252,6 +263,10 @@ namespace Microsoft.PythonTools.EnvironmentsList {
             _provider.InstalledPackagesChanged += PipExtensionProvider_InstalledPackagesChanged;
 
             IsPipInstalled = _provider.IsPipInstalled ?? true;
+            ShowSecurityWarning =
+                provider._packageManager.UniqueKey == "pip" &&
+                view.Configuration.Version != new Version(2, 7) &&
+                view.Configuration.Version < new Version(3, 3);
 
             _installCommandView = new InstallPackageView(this);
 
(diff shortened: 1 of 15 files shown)


Python/Product/EnvironmentsList/PipExtension.xaml.cs
```diff
@@ -252,6 +263,10 @@ PipExtensionProvider provider
             _provider.InstalledPackagesChanged += PipExtensionProvider_InstalledPackagesChanged;
 
             IsPipInstalled = _provider.IsPipInstalled ?? true;
+            ShowSecurityWarning =
+                provider._packageManager.UniqueKey == "pip" &&
```
zooba: I think we can make this not pip specific.

Python/Product/EnvironmentsList/Resources.resx
```diff
@@ -422,6 +422,12 @@ The folder will be deleted from disk. This operation cannot be undone.</value>
   <data name="PipExtensionClearSearchToolTip" xml:space="preserve">
     <value>Clear search filter</value>
   </data>
+  <data name="PipExtensionSecurityWarning" xml:space="preserve">
+    <value>Due to security issues, pip may no longer work on this version of Python.</value>
```
zooba: Rather than "issues", let's say "Due to new security restrictions, installing from the internet may not work ..."

Python/Product/EnvironmentsList/PipExtension.xaml.cs
huguesv: I didn't want to to show for conda package management, because there it's not dependent on the python version of the environment, but rather on the version of the global conda.exe.

Using latest Anaconda, if I create a python=2.6 env, I can successfully install conda packages into it.

Python/Product/EnvironmentsList/Resources.resx
huguesv: Agreed I like that better.

PR #4256, opened 2018-05-24
Fixes IronPython intellisense

zooba:
Fixes IronPython intellisense
Fixes IronPython tests
Removes unnecessary IronPython dependencies from test projects
Makes Ast.PythonModuleLoader public API

This is a prerequisite for #2943, though it's not a fix for it. Feel free to make architectural suggestions - I've gone for minimum top-level public APIs, but if you really hate the nested classes/enums then let me know why and I'll reconsider :)


zooba:
For clarity, most of this change is:
* adding a fallback to import using the AST analyzer for pure-Python modules in IronPython
* removing unused code and unnecessary assembly references (that were breaking tests)

PR #4261, opened 2018-05-25
Allow annotated but not assigned variables

### LINKED ISSUE #3608: NamedTuple completions only show fields with default values
From https://docs.python.org/3/library/typing.html#classes-functions-and-decorators

```
from typing import NamedTuple

class Employee(NamedTuple):
    name: str
    id: int = 3

employee = Employee('Guido')
print(employee.name)
print(employee.id)
employee.
```

Mouse hover tooltip on `employee` correctly shows that the type is `Employee`.

Completions on `employee.` only include `id`, missing `name`.

MikhailArkhipov:
Fixes #3608


Python/Product/Analysis/Analyzer/DDG.cs
```diff
@@ -213,6 +213,23 @@ internal class DDG : PythonWalker {
 
         public override bool Walk(ExpressionStatement node) {
             _eval.Evaluate(node.Expression);
+
+            if (node.Expression is ExpressionWithAnnotation annoExpr && annoExpr.Annotation != null && annoExpr.Expression is NameExpression ne) {
```
zooba: No real reason to restrict `annoExpr.Expression is NameExpression` here. If the parser allowed something else, we should go ahead and assign to it normally.

PR #4264, opened 2018-05-25
Fixes #3437 Typeshed integration

### LINKED ISSUE #3437: Typeshed integration
Recognize the existence of the [`typeshed`](https://github.com/python/typeshed) package of type stub files and use it to supplement existing IntelliSense information.

This interacts closely with #82, which provides support for _generic_ type stub files. However, this particular task is complete when we detect and handle sub-paths for the _specific_ case of `typeshed`.

- [x] detect installed `typeshed` module
- [x] load `typeshed` module in place of regular analysis
- [ ] merge `typeshed` info with regular analysis

zooba:
Fixes #3437 Typeshed integration
Enables value merging from typeshed
Improves some handling of type annotations in AST imported files


zooba:
@MikhailArkhipov In case you want to watch my work in progress on this. The basics are in there, just need to beef up the tests and the handling of the typing module when doing AST imports.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -643,32 +660,65 @@ TryImportModuleContext context
...
-                return new AstScrapedPythonModule(mp.FullName, mp.SourceFile);
+            return module;
+        }
+
+        private async Task<IReadOnlyList<string>> GetTypeShedPaths() {
+            var typeShedPaths = _typeShedPaths;
+            if (typeShedPaths != null) {
+                return typeShedPaths;
+            }
+            lock (_searchPathsLock) {
```
MikhailArkhipov: Looks like duplicate?

Python/Product/Analysis/Infrastructure/PathUtils.cs
```diff
@@ -184,6 +184,39 @@ static class PathUtils {
...
+        /// Returns the path to the parent directory segment of a path. If the
+        /// last character of the path is a directory separator, the segment
+        /// prior to that character is removed. Otherwise, the segment following
+        /// the last directory separator is removed.
+        /// </summary>
+        /// <remarks>
+        /// This should be used in place of:
+        /// <c>Path.GetDirectoryName(CommonUtils.TrimEndSeparator(path)) + Path.DirectorySeparatorChar</c>
+        /// </remarks>
+        public static string GetParent(string path) {
```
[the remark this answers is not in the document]
zooba: That's much slower and far more pedantic about paths (it will actually validate the paths...)

zooba:
@MikhailArkhipov @huguesv This is ready for review

Python/Product/Analysis/Interpreter/Ast/ImportScanner.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
```
huguesv: missing header

zooba: Huh, this whole file is unused and not meant to be checked in yet... whoops

Python/Product/Analysis/Analyzer/InterpreterScope.cs
```diff
@@ -289,15 +289,15 @@ internal bool RemoveVariable(string name, out VariableDef value)
         }
 
         internal virtual void ClearNodeValues() {
-            _nodeValues.Clear();
+            _nodeValues = new AnalysisDictionary<Node, NodeValue>();
```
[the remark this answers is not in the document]
zooba: Yes. We only change them on the thread that accesses it, and we replace it specifically so that references obtained through the properties continue to point at the old dictionary.

Python/Product/Analysis/Interpreter/Ast/AstPythonFunction.cs
```diff
@@ -54,6 +54,17 @@ LocationInfo loc
             Locations = loc != null ? new[] { loc } : Array.Empty<LocationInfo>();
         }
 
+        internal AstPythonFunction(IPythonFunction original) {
+            DeclaringModule = original.DeclaringModule;
+            DeclaringType = original.DeclaringType;
+            Name = original.Name;
+            _doc = (original as AstPythonFunction)?._doc ?? original.Documentation;
```
AlexanderSher: `AstPythonFunction.Documentation` first checks for `_doc` as well, no need to check for it separately.

Python/Product/Analysis/Interpreter/Ast/AstPythonFunction.cs
```diff
@@ -54,6 +54,17 @@ LocationInfo loc
...
+            IsClassMethod = original.IsClassMethod;
+            IsStatic = original.IsStatic;
+            _overloads = original.Overloads.ToList();
+            Locations = (original as AstPythonFunction)?.Locations ?? (original as ILocatedMember)?.Locations ?? Array.Empty<LocationInfo>();
```
AlexanderSher: `AstPythonFunction` implements `ILocatedMember`, so no need to check `AstPythonFunction` by itself

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -643,32 +672,67 @@ TryImportModuleContext context
...
+                _log?.Log(TraceLevel.Verbose, "Import", mp.FullName, FastRelativePath(mp.SourceFile));
+                module = PythonModuleLoader.FromFile(context.Interpreter, mp.SourceFile, LanguageVersion, mp.FullName);
+            }
+
+            return module;
+        }
+
+        private IReadOnlyList<string> GetTypeShedPaths(int timeout) {
+            // First check
+            var typeShedPaths = _typeShedPaths;
```
AlexanderSher: `_typeShedPaths` is accessed outside of the lock

Python/Tests/Analysis/AstAnalysisTests.cs
```diff
@@ -749,5 +749,52 @@ class BankAccount(object):
...
         }
         #endregion
+
+        #region Type Shed tests
+
+        private static PythonVersion VersionWithTypeShed =>
+            PythonPaths.Versions.LastOrDefault(v => Directory.Exists(Path.Combine(v.PrefixPath, "Lib", "site-packages", "typeshed")));
+
+        [TestMethod, Priority(0)]
+        public void TypeShedElementTree() {
```
huguesv: I hacked this test to run under a version of python that doesn't have typeshed installed, and it still passes. I'm not sure what to make of that.

Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleModules.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            if (mmx != null && mmy == null) {
+                mmx.AddModule(y);
+                return mmx;
+            } else if (mmy != null && mmx == null) {
+                mmy.AddModule(x);
+                return mmy;
+            } else if (mmx != null && mmy != null) {
+                mmx.AddModules(mmy._modules);
+                return mmx;
```
AlexanderSher: Does it matter anywhere that we return `mmx` instead of `mmy`?

Python/Tests/Analysis/AstAnalysisTests.cs
zooba: :( It means our analysis is so good that typeshed doesn't actually help here. Which version did you run it on? The one I tried seemed to need it

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
zooba: Deliberately, so we can avoid locking for the search if the search has already been done (it'll only ever be done once - we never change `_typeShedPaths` after initialization - but it depends on context too much to make it `Lazy<T>`).

Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleModules.cs
zooba: No. This should only be being used while we're constructing objects and before the analyzer ever gets a hold of it (great question though)

Python/Tests/Analysis/AstAnalysisTests.cs
huguesv: 3.6

Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleMembers.cs
```diff
@@ -39,6 +39,13 @@ class AstPythonMultipleMembers : IPythonMultipleMembers, ILocatedMember {
             _checkForLazy = true;
         }
 
+        public IMember Trim() {
```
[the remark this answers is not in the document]
huguesv: Oh, calling `Members` might resize `_members` potentially going from size 1 to 0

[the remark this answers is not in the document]
zooba: Good point though, should really check that here.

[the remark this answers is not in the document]
zooba: Actually, decided not to change it. If we trim down to a single lazy member, we should probably leave it lazy even if it will eventually (depending on context) resolve down to nothing.

Python/Tests/Analysis/AstAnalysisTests.cs
zooba: Also, an easy way to disable type shed is to go into AstPythonInterpreter, search for `IncludeTypeShed` and set it to false (which I'm currently turning into one of our secret registry key settings)

huguesv:
With typeshed installed:
image

Without:
image

huguesv:
I was expecting to see the types for the urlopen parameters. Error list is from running mypy.
image
Are they removed intentionally? For my own method with type hints, the type appears:
image

zooba:
About to push a fix for the module members issue, and it seems we're currently not resolving parameter annotations into displayable types, so there's a bit more work to do there. The default values of `...` need to be ignored properly as well.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
AlexanderSher: Since `_searchPathsLock` is used only for locking, it will be a thin lock and the overhead of using it is close to nothing:
https://blogs.msdn.microsoft.com/seteplia/2017/09/06/managed-object-internals-part-2-object-header-layout-and-the-cost-of-locking/
http://firstclassthoughts.co.uk/Articles/Programming/DotNetLocksAreReallyFast.html

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -643,32 +680,67 @@ TryImportModuleContext context
...
+                _log?.Log(TraceLevel.Error, "GetTypeShedPaths", ex.ToString());
+            }
+            return null;
+        }
+
+        private async Task<IReadOnlyList<string>> GetTypeShedPathsAsync() {
+            // Assume caller has already done the quick checks
+
+            IReadOnlyList<string> typeShedPaths;
+            var typeshed = await FindModuleInSearchPathAsync("typeshed");
```
AlexanderSher: Is it ok that `FindModuleInSearchPathAsync` can be called several times concurrently before `_typeShedPaths` is finally assigned?

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
zooba: Not using it has no overhead, and it's only for initialization.

Python/Product/Analysis/Interpreter/Ast/AstPythonFunction.cs
zooba: I meant to copy the `null` from `_doc`, but this doesn't do that either :) So a slightly different change coming

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
zooba: It's also read outside the `async` to avoid any overhead there.

Python/Tests/Analysis/AstAnalysisTests.cs
zooba: In the latest code, you can disable it by setting: `HKEY_CURRENT_USER\Software\Microsoft\PythonTools\15.0\Analysis\Project@UseTypeStubPackages` to `0` (defaults to `1`). You can also set `UseTypeStubPackagesExclusively` to `1` to not merge with the original results.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
zooba: Yep. We could call it every time if we wanted, it doesn't change any state. But since it's results (rarely) change, better to cache it.

zooba:
Still need to fix that test before merging, but that's all I'm doing

Python/Product/Analysis/Interpreter/Ast/AstAnalysisFunctionWalker.cs
```diff
@@ -25,25 +25,20 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
...
-            _overload = new AstPythonFunctionOverload(
-                AstPythonFunction.MakeParameters(_scope.Ast, _target),
-                _scope.GetLocOfName(_target, _target.NameExpression),
-                _returnTypes
-            );
+            _overload = overload;
         }
 
-        public IList<IPythonType> ReturnTypes => _returnTypes;
+        public IList<IPythonType> ReturnTypes => _overload.ReturnTypes;
```
huguesv: according to the constructor, `_overload` can be `null`.

huguesv: it's used in a few other places without checks, so I suppose ctor should throw if null.

huguesv:
Just that one comment.

PR #4269, opened 2018-05-28
Fix compiled module check on Mac

MikhailArkhipov:
- On Mac modules can have more than one dot in the name. Changed to explicit extension check.
- VSC Tests cannot await, yielding causes VSC immediately return empty result to the test code.

### DIFF
--- a/Python/Product/Analysis/LanguageServer/Server.Completion.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.Completion.cs
@@ -82,7 +82,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
         }
 
         public override Task<CompletionItem> CompletionItemResolve(CompletionItem item) {
-             // TODO: Fill out missing values in item
+            // TODO: Fill out missing values in item
             return Task.FromResult(item);
         }
 
--- a/Python/Product/Analysis/LanguageServer/Server.Completion.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.Completion.cs
@@ -24,7 +24,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
     public sealed partial class Server {
         public override async Task<CompletionList> Completion(CompletionParams @params) {
             await _analyzerCreationTask;
-            IfTestWaitForAnalysisComplete();
+            await IfTestWaitForAnalysisCompleteAsync();
 
             var uri = @params.textDocument.uri;
             // Make sure document is enqueued for processing
--- a/Python/Product/Analysis/LanguageServer/Server.Completion.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.Completion.cs
@@ -15,6 +15,7 @@
 // permissions and limitations under the License.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.PythonTools.Parsing;
@@ -23,7 +24,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
     public sealed partial class Server {
         public override async Task<CompletionList> Completion(CompletionParams @params) {
             await _analyzerCreationTask;
-            await IfTestWaitForAnalysisCompleteAsync();
+            IfTestWaitForAnalysisComplete();
 
             var uri = @params.textDocument.uri;
             // Make sure document is enqueued for processing
@@ -57,7 +58,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
                 members = members.Where(m => m.kind == filterKind.Value);
             }
 
-            var res = new CompletionList {
+             var res = new CompletionList {
                 items = members.ToArray(),
                 _expr = ctxt.ParentExpression?.ToCodeString(tree, CodeFormattingOptions.Traditional),
                 _commitByDefault = ctxt.ShouldCommitByDefault
@@ -81,7 +82,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
         }
 
         public override Task<CompletionItem> CompletionItemResolve(CompletionItem item) {
-            // TODO: Fill out missing values in item
+             // TODO: Fill out missing values in item
             return Task.FromResult(item);
         }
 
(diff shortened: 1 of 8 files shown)


MikhailArkhipov:
No, everywhere, but only in tests. Basically VSC runs another VSC as extension host that then launches LS process. Test code invokes completion via
`await vscode.commands.executeCommand<vscode.CompletionList>('vscode.executeCompletionItemProvider', textDocument.uri, position)`

This lands in `textDocument/completion` and hence in the completion method. However, if the method starts awaiting, `vscode.commands.executeCommand` instantly returns with a null results. Not sure where  await falls off, in VSC, in node.js or elsewhere.

zooba:
I still really want those waits moved to the tests, rather than in the product. We already have test code that can wait for analysis to be complete - I don't see why we need to do it in the product?

Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -57,7 +58,7 @@ public sealed partial class Server {
                 members = members.Where(m => m.kind == filterKind.Value);
             }
 
-            var res = new CompletionList {
+             var res = new CompletionList {
```
zooba: Why the extra space?

Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -81,7 +82,7 @@ public sealed partial class Server {
         }
 
         public override Task<CompletionItem> CompletionItemResolve(CompletionItem item) {
-            // TODO: Fill out missing values in item
+             // TODO: Fill out missing values in item
```
zooba: Why the extra space?

Python/Product/Analysis/ModulePath.cs
```diff
@@ -145,8 +142,11 @@ public ModulePath(string fullname, string sourceFile, string libraryPath)
             RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
         private static readonly Regex PythonBinaryRegex = new Regex(@"^(?!\d)(?<name>(\w|_)+)\.((\w|_|-)+?\.)?(pyd|so|dyLib)$",
             RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
-        private static readonly Regex PythonCompiledRegex = new Regex(@"^(?!\d)(?<name>(\w|_)+)\.(((\w|_|-)+?\.)?pyd|py[co]|so|dyLib)$",
-            RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
+
+        private static bool IsCompiledModule(string name) {
+            var ext = Path.GetExtension(PathUtils.GetFileName(name)).ToLowerInvariant();
+            return CompiledModuleExtensions.Any(x => x == ext);
```
zooba: Why not use `.EndsWithOrdinal(x, ignoreCase: true)`?

Python/Product/Analysis/ModulePath.cs
```diff
@@ -137,6 +133,7 @@ public ModulePath(string fullname, string sourceFile, string libraryPath)
             LibraryPath = libraryPath;
         }
 
+        private static readonly string[] CompiledModuleExtensions = new[] { ".pyd", ".pyc", ".pyo", ".so", ".dyLib" };
```
zooba: Capital `L` in `dyLib` means this never matches your current test. If it works now and didn't before, does that mean we shouldn't be testing for `dylib`?

Python/Product/Analysis/LanguageServer/Server.Completion.cs
MikhailArkhipov: Accident. Most prob VS Code C# formatter, was working on Mac.

Python/Product/Analysis/ModulePath.cs
MikhailArkhipov: This is a good question. Mac FS is case-insensitive *usually* but one can have it in case-sensitive form. Linux is always case-sensitive. However, I compare with `ToLowerInvariant()`

Python/Product/Analysis/ModulePath.cs
MikhailArkhipov: OK

zooba:
```csharp
private static bool IsCompiledModule(string name) =>
    CompiledModuleExtensions.Any(ext => name.EndsWithOrdinal(ext, true))
```

zooba:
Which modules weren't being picked up on macOS? I'm not able to see anywhere that the suffix is treated differently by platform in the CPython sources, but it's possible we need to update the `PythonBinaryRegex` as well if the usual `<import name>.<optional ABI tag>.<ext>` isn't cross platform enough.

MikhailArkhipov:
the import was math on Python 3 on Mac. Looks like `math.cpython-36m-darwin.so`. Regex _looks_ OK but didn't match.

Do we need to use regex here? Does it matter what form the name takes? The choice is to scrape or not to scrape so basically, as long as it is a binary with known extension, we need to try and scrape.

Yes, I can change `PythonBinaryRegex` as well.

As for test-related code , we probably should expose custom notifications to typescript so tests can await on those.

zooba:
I made a slight tweak to the original regex that may also solve the problem:
```regex
^(?!\d)(?<name>(\w|_)+)\.((\w|_|-)+?\.)?(pyd|py[co]|so|dylib)$
```

Previously the optional ABI tag was grouped with `pyd` and not the other extensions. After this change, any of the compiled files are allowed an ABI tag (which is correct - even `.pyc` files can have them). That may allow most of this change to be reverted?

zooba:
> we probably should expose custom notifications to typescript so tests can await on those.

+1, let's do this, it'll match the `AnalysisCompletEvent` we have under the old protocol. We can also include performance information (timing, memory use, file count, etc.) that may help us detect problems at runtime.

MikhailArkhipov:
Up to you. That particular regex didn't use name group so it seemed that checking extension would be simpler.

zooba:
Doesn't use name group _yet_ - we're still evolving lots of things here and I can totally see it being useful, at which point we would need to reinvent an already working regex :)

MikhailArkhipov:
OK. I have https://github.com/Microsoft/vscode-python/issues/1591, opened https://github.com/Microsoft/PTVS/issues/4283

MikhailArkhipov:
Will reopen from another branch for simplicity

PR #4275, opened 2018-05-29
Fixes #4271 Failed to install Anaconda

### LINKED ISSUE #4271: Failed to install Anaconda
Anaconda may not install correctly if we do not explicitly wait for any running uninstalls to complete.

Add a helper process that will wait for previous uninstalls to complete and then launch the next one.

zooba:
Fixes #4271 Failed to install Anaconda
Adds installer launcher.


Python/Product/AnacondaInstallLauncher/Program.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        void WaitForProcess(DateTime stopAt, bool requireProcess) {
+            if (!KeepWaiting) {
+                return;
+            }
+
+            var procs = GetProcesses();
+            int retries = 30;
+            while (requireProcess && procs.Length == 0 && retries-- > 0) {
+                Thread.Sleep(1000);
```
heaths: Why force context switching on the CPU every second when you could just wait on the process handle(s) (via `Process.WaitForExit`) like in the original example I provided?

Python/Product/AnacondaInstallLauncher/Program.cs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        Process[] GetProcesses() {
+            return Process.GetProcesses()
+                .Where(p => p.ProcessName.StartsWith("Un_") &&
+                    p.MainModule.FileVersionInfo.FileDescription.IndexOf("Anaconda", StringComparison.OrdinalIgnoreCase) >= 0)
+                .ToArray();
+        }
+
+        void WaitForProcess(DateTime stopAt, bool requireProcess) {
```
heaths: Why timeout? If you use the process handle (via `Process.WaitForExit`), this is an intrinsic wait handle (in the underlying native implementation). It will signal when the process terminates. If you merely timeout, the original problem of the race condition between the async uninstaller and sync installer will happen as well.

If you really feel it necessary to timeout (and with how much machines vary in performance, especially under additional workload, I don't recommend it), at least fail method and forego running the installer. Return an exit code so other chainers (like Visual Studio) can tell the user something failed and to repair.

Python/Product/AnacondaInstallLauncher/Program.cs
zooba: Because we're also checking the file system, and file system watchers are notoriously unreliable so I'd rather not use one just to avoid looping. This is not going to be the most CPU intensive operation while it's running (Windows Defender takes that prize...)

Python/Product/AnacondaInstallLauncher/Program.cs
zooba: Yeah, fair enough. I'll lose the timeout.

Python/Product/AnacondaInstallLauncher/Program.cs
heaths: You're not checking the file system in this loop. Waiting on the process handles then checking the file system would be less intensive.

Python/Product/AnacondaInstallLauncher/Program.cs
```diff
@@ -0,0 +1,192 @@
...
+            return p;
+        }
+
+        int DoInstall() {
+            WaitForUninstall(false);
+            if (_cancel) {
+                return ERROR_INSTALL_CANCEL;
+            }
+
+            var p = Run(_installer, string.Format(CultureInfo.InvariantCulture, "/InstallationType=AllUsers /RegisterPython=0 /S /D={0}", _targetDir));
```
huguesv: _targetDir probably needs to be quoted?

zooba: No. Anaconda installer is weird here - it must be last on the command line and it must be unquoted.

Python/Product/AnacondaInstallLauncher/Program.cs
```diff
@@ -0,0 +1,192 @@
...
+            try {
+                WaitForProcess(requireWait);
+            } catch (Exception ex) {
+                Console.Error.WriteLine(ex.ToString());
+            }
+        }
+
+        bool IsDirectoryPresent() {
+            if (!Directory.Exists(_targetDir)) {
+                Console.Error.WriteLine("Target directory is not present");
```
huguesv: Might be useful to print `_targetDir` at the end of all these messages

Python/Product/AnacondaInstallLauncher/Program.cs
huguesv: Please put that as a comment

Python/Product/AnacondaInstallLauncher/Program.cs
zooba: It'll be followed up immediately by the log message for launching the installer, and it'll be in the log for starting this process too. I don't think it needs repeating much more than that.

Python/Product/AnacondaInstallLauncher/Program.cs
```diff
@@ -0,0 +1,190 @@
...
+                Console.Error.WriteLine(ex.ToString());
+                return false;
+            }
+
+            return true;
+        }
+
+        Process[] GetProcesses() {
+            return Process.GetProcesses()
+                .Where(p => p.ProcessName.StartsWith("Un_") &&
```
huguesv: According to the docs at https://msdn.microsoft.com/en-us/library/system.diagnostics.process.processname(v=vs.110).aspx, ProcessName property may throw `InvalidOperationException` if process has exited. Given that we're looking at all processes on the machine, it could happen that a process exits between the call to `GetProcesses` and `ProcessName`.

We print `ProcessName` in `WaitForProcess` as well.

MikhailArkhipov:
LGTM

PR #4284, opened 2018-05-30
Fix compiled module check on Mac

MikhailArkhipov:


AlexanderSher:
Do we have any tests for this?

MikhailArkhipov:
Nope. This is why we either need to port test to .NET Core or abstract the file system or both :-)

PR #4297, opened 2018-06-01
Fixes #4285 Expose custom extensions through LSP

### LINKED ISSUE #4285: Expose custom extensions through LSP
We already have the extensibility model in the language service, but we need to add:

- [x] Custom request to load extension
- [x] Custom messages to control extension

zooba:
Fixes #4285 Expose custom extensions through LSP
Adds message to load extension in server.
Adds single hook for Completion.


Python/Tests/Analysis/LanguageServerTests.cs
```diff
@@ -453,6 +454,37 @@ class MyClass:
             );
         }
 
+        public class TestCompletionHook : ILanguageServerExtension {
```
AlexanderSher: Is it required to be public?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -428,6 +324,108 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
+            }
+        }
+
+        private T ActivateObject<T>(string assemblyName, string typeName, Dictionary<string, object> properties) {
+            if (string.IsNullOrEmpty(assemblyName) || string.IsNullOrEmpty(typeName)) {
+                return default(T);
+            }
+            try {
+                var assembly = File.Exists(assemblyName) ? AssemblyName.GetAssemblyName(assemblyName) : new AssemblyName(assemblyName);
+                var type = Assembly.Load(assembly).GetType(typeName, true);
```
AlexanderSher: Is it going to be for the public API, or is it for internal purposes only?

Python/Tests/Analysis/LanguageServerTests.cs
zooba: Yes, this is public.

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Yes, public/stable API

AlexanderSher: Is design finalized or we can discuss it?

zooba: We can discuss it, but consistency with the existing public API is going to be a strong argument in favour of this design :)

Python/Product/Analysis/LanguageServer/Hooks.cs
```diff
@@ -0,0 +1,27 @@
...
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using Microsoft.PythonTools.Parsing.Ast;
+
+namespace Microsoft.PythonTools.Analysis.LanguageServer.Hooks {
```
MikhailArkhipov: `.Extensions` or `.Extensibility` instead of `Hooks`?

Python/Product/Analysis/LanguageServer/Hooks.cs
```diff
@@ -0,0 +1,27 @@
...
+    public sealed class CompletionEventArgs : EventArgs {
+        public ModuleAnalysis Analysis;
```
MikhailArkhipov: I'd make it with readonly get properties

Python/Product/Analysis/LanguageServer/ILanguageServerExtension.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// the extension cannot be sent messages and may be garbage collected
+        /// if it does not hook any events in its <see cref="Register(Server)"/>
+        /// method.
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Called when an extension command arrives for this extension.
+        /// </summary>
+        Dictionary<string, object> ExecuteCommand(string command, Dictionary<string, object> properties);
```
MikhailArkhipov: `IReadOnlyDictionary`?

Python/Product/Analysis/LanguageServer/Messages.cs
```diff
@@ -334,6 +335,25 @@ public struct RenameParams {
         public int? _version;
     }
 
+    [Serializable]
+    public class PythonAnalysisExtensionParams {
+        public string assembly;
```
MikhailArkhipov: readonly props?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -81,6 +81,8 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
         private bool _testEnvironment;
         private ReloadModulesQueueItem _reloadModulesQueueItem;
 
+        private readonly Dictionary<string, ILanguageServerExtension> _extensions;
```
MikhailArkhipov: Concurrent so we don't have to lock it?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -428,6 +352,108 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
 
+        #region Private Helpers
+
+        private IProjectEntry RemoveEntry(Uri documentUri) {
+            var entry = _projectFiles.RemoveEntry(documentUri);
+            _openFiles.Remove(documentUri);
+            return entry;
+        }
+
+        private async Task DoInitializeAsync(InitializeParams @params) {
```
MikhailArkhipov: We could move it to `Server.Initialize.cs` and `Server.Extensions.cs`

Python/Product/Analysis/LanguageServer/Messages.cs
zooba: None of these have readonly props - they're mostly for serialization.

Python/Product/Analysis/LanguageServer/Server.cs
zooba: If you think it's worth splitting it up at that fine-grained a level, sure. I'm personally not a fan, but happy to follow your direction here

Python/Product/Analysis/LanguageServer/ILanguageServerExtension.cs
zooba: Maybe... is Newtonsoft going to serialize it properly? Or will I need to copy it into a real dictionary?

Python/Product/Analysis/LanguageServer/Hooks.cs
zooba: I didn't want to write a constructor, and some of these need to be fields because they're value types.

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Is that easily available in .NET Core without adding more dependencies?

Python/Product/Analysis/LanguageServer/Hooks.cs
AlexanderSher: This instance is shared between all subscribers. One subscriber shouldn't be able to affect the next one by changing args unless it is intentionally expected.

Python/Product/Analysis/LanguageServer/ILanguageServerExtension.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System.Collections.Generic;
+using Microsoft.PythonTools.Projects;
+
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
+    public interface ILanguageServerExtension {
+        /// <summary>
+        /// Called when the extension is registered for a language server.
+        /// </summary>
+        void Register(Server server);
```
AlexanderSher: Can we separate entry point from the extension itself?
```cs
public interface ILanguageServerExtensionProvider {
    Task<ILanguageServerExtension[]> GetExtensionsAsync(Server server, CancellationToken cancellationToken);
}
```

Also, `Server` seems too much for an input. Can we only provide an interface with required parameters?

Python/Product/Analysis/LanguageServer/ILanguageServerExtension.cs
AlexanderSher: Can't we explicitly accept and return `JToken`?

Python/Product/Analysis/LanguageServer/Hooks.cs
AlexanderSher: Also, if we want to be sure that extension has decided if it needs this notification during registration, it is better to provide it with interface it can implement.

Python/Product/Analysis/LanguageServer/ILanguageServerExtension.cs
AlexanderSher: Or even better, a separate object instance.

Python/Product/Analysis/LanguageServer/ILanguageServerExtension.cs
MikhailArkhipov: OK, if it is simpler for JSON serialization.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: We are already using `ConcurrentDictionary` in `Analysis\Interpreter\Ast`

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: OK. Just making files reasonably small to scroll through :-)

Python/Product/Analysis/LanguageServer/Hooks.cs
zooba: I don't want to provide that many interfaces (one per overridable call). And your point is well taken on it being shared - I'll make the input parameters readonly properties and leave the return value as a mutable field.

Python/Product/Analysis/LanguageServer/ILanguageServerExtension.cs
zooba: It looks like Newtonsoft has support for `IReadOnlyDictionary` so I'll go with that.

Python/Product/Analysis/LanguageServer/ILanguageServerExtension.cs
zooba: We'd basically end up with one provider per extension, so there's nothing gained here, and we still need to be able to pass in parameters. The IDE will likely have something like this though (e.g. in VS we use MEF to discover extensions, and then the provider gives us the info we need to pass over JSON to load the extension).

As for providing `Server`, I would prefer to pass `ServerBase`, but ultimately these extensions are for the Python Language Server and so probably won't ever be used with another one. Having the base class helps keep _us_ honest about "what is in LSP vs. what have we added", which is why it's there.

MikhailArkhipov: +1 on the interface. Prevents extension from having recompile when class changes. Later we can put interfaces into a separate interop assembly. Also, allows extension to mock server for testing.

zooba: It doesn't depend on the structure of `Server`, as there is no inheritance involved. And `Server` is already public API, which means we can't change existing public members anyway (adding new members is fine).

If we want to provide `ServerBase` as the parameter and make extensions downcast, that's fine by me. Given the need to pass `PythonAst` instances around, there isn't really a world where we can build extensions without depending on `Microsoft.PythonTools.Analysis.dll` anyway, so we may as well make it as easy as possible to build the extension.

AlexanderSher: There are several reasons why extension instantiation should be separated from extension itself:
- Lifetime. Right now, it is kinda equal to the `Server`, but we may need to change this in the future (for instance, recreate extension if it throws too many exceptions), and extension writers may want to track instances without static collections that are always a problem for testing.
- Half-initialized object. It is always a question what can be done in ctor and what must be done in `Register`. Good example of how it ends is instantiation of tool windows and all the problems VS have with it right now. It is better to give extension writer freedom to return fully initialized object. This is also better for tests, cause instead of two calls from host test has to simulate only one.
- Read-only fields. Extension code will need far less null checks for fields that are assigned in ctor.
- Factory pattern and SRP

> We'd basically end up with one provider per extension

Is it a limitation of VSC?

> we may as well make it as easy as possible to build the extension.

That includes ability to test functionality without instantiation of heavy objects. That's why it is better to pass interface. You can take CPS with their 'SwitchToMainThreadAsync' that can't be mocked because of the struct awaiter. As a result, everyone create a wrapper, so all the benefits of that awaiter got swallowed by those wrappers.

`Server` is most long-living object which contains tons of public methods that aren't intended to be called from extension (including LSP handlers). Public object was designed for access from external host inside some third-party project. It is different from loading extension into our host and give it access inside Analysis workflow. Extension should have access to the parts that are designed for it and nothing more.

zooba: >>We'd basically end up with one provider per extension

>Is it a limitation of VSC?

We have one `Server` instance per process. Each extension can only be loaded once per `Server` instance. Hence, each provider is only ever going to be instantiated once and used once to provide one instance of the extension. So we may as well merge them and have the factory be `.ctor`.

MikhailArkhipov: `IServer` makes it easy to mock. Extension may want to have tests without actually running the process.

zooba: What would `IServer` have that `ServerBase` doesn't have?

MikhailArkhipov: Class cannot be mocked. Although it is possible to derive some sort of a test class from base, common practice is to write tests against interfaces using mocking frameworks like NSubstitute or TypeMoq. Also, interfaces eventually can go to a separate interop assembly and extension tests the won't need actual LS assembly.

IMO common practice is to expose interfaces unless class is trivial and is easy to create such as range/point/eventargs. Although I found that IRange is way more useful than Range/Span classes that VS and VSC use.

zooba: Okay, adding an `IServer` class with the same interface as `ServerBase` so it can be mocked.

zooba: I'll add a provider class as well. Shouldn't add too much code to every extension...

zooba: I'm treating the interop assembly as out of scope right now though. When we move everything out, we'll just have to do type forwarding for them all

Python/Product/Analysis/LanguageServer/ILanguageServerExtensionProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
+    /// <summary>
+    /// Implemented on a class that can create a language server extension.
+    /// This class must have a default constructor.
+    /// </summary>
+    public interface ILanguageServerExtensionProvider {
+        /// <summary>
+        /// Called when the extension is loaded for a language server.
+        /// </summary>
+        ILanguageServerExtension Create(IServer server, IReadOnlyDictionary<string, object> properties);
```
MikhailArkhipov: I think this one needs to be async with cancellation token so we can abandon extension creation (i.e. it won't  be able to hang us)

Python/Product/Analysis/LanguageServer/IServer.cs
```diff
@@ -0,0 +1,76 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
+    public interface IServer {
```
MikhailArkhipov: Do we need to expose everything? I'd keep interface to absolute minimum. Extension definitely should not be able to call Shutdown.... ;-) If we want extension to be able to intercept calls in the protocol then it should have appropriate hooks we will call (not events since extension will definitely forget to detach). Something like

```cs
interface ILanguageServerExtension {
Task OnIntitialize(CancellationToken t); // we'll call it after Init
Task OnShutdown(...) 
Task<CompletionList> OnCompletion(...)
...
}

We can look at specific list of notifications.

Python/Product/Analysis/LanguageServer/ILanguageServerExtensionProvider.cs
zooba: We're async all the way down, so may as well. But it's unlikely to be an issue - extensions at this stage shouldn't have to do more than object instantiation and event subscription.

Python/Product/Analysis/LanguageServer/IServer.cs
zooba: You are far too distrustful of extensions :) Even with the interface, they can just cast to `Server` and do whatever they want anyway (or you know, run arbitrary code to pwn your machine right there) - I just added the interface because you all said it would make it easier to write tests.

Detaching from events doesn't matter, as neither Server nor any extensions will significantly out-live each other.

The whole point of being integrated into the server here is to have full access to everything. Proxying is a much better option if you want a thin and protected layer - this is for deep integrations that can and do use everything available.

PR #4308, opened 2018-06-04
Improve support for relative imports + bug fixes

### LINKED ISSUE #949: Intellisense is broken in the presence of relative imports
If you type this:

`from . import foo`

PTVS doesn't seem to understand this syntax.  After you write the `.` (note the space) you get an intellisense popup suggesting module names, so as you continue typing `import` it selects `importlib` and you end up with 

`from . importlib foo`

The modules that its suggesting in this intellisense popup are not located in the specified directory either it just seems to be all the modules in the search paths.  So I think PTVS simply has no concept of relative imports.

If you actually go and type out a valid relative import, it _is_ able to resolve it.  So something like

`from ..foo import bar`

if it is a valid module, you will have working intellisense on bar and no squigglies.  The problem seems to be isolated only to the case of typing the import statement.

### LINKED ISSUE #4138: No completion in namespace packages
```
.
├── projecta
│   └── foo
│       ├── bar
│       │   └── __init__.py
│       └── __init__.py
└── projectb
    └── foo
        ├── baz
        │   └── __init__.py
        └── __init__.py
```

test.py:
```
import projectA.foo
import projectA.foo.bar
import projectB.foo
import projectB.foo.baz

projectA.
```
image

Jedi:
image

PyCharm:
image

### LINKED ISSUE #4299: Deleted variables linger in analysis
Declare `p=1`
Observe `p.` gives completion for `int`.
Delete line with `p=1`
Type `p.`

image

MikhailArkhipov:
Fixes #949
- Adds support for `from . import a, b, c`.
- Fixes issues with `from .a import b,c.`
- Hovering over `abc` in `from x import abc` gives information on the `abc`.


Python/Product/Analysis/ModulePath.cs
```diff
@@ -664,17 +660,21 @@ out errorParameter
         internal static bool FromBasePathAndName_NoThrow(
             string basePath,
             string moduleName,
+            PythonLanguageVersion? languageVersion,
```
zooba: I think this makes more sense as a bool that says `requiresInitPyFiles`. That avoids encoding too much version-specific logic right here and moves it out to the caller (especially relevant for when we hit particular cases where the requirement is interesting and the version is unknown/irrelevant - it's much more clear why `true` is selected than `V33`).

Python/Product/Analysis/ModuleResolver.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// such as 'from . import a, b, c' or 'from .a import b, c'
+        /// </summary>
+        /// <param name="entry">Project entry</param>
+        /// <param name="node">Import statement node</param>
+        /// <returns>Names of modules</returns>
+        internal static IEnumerable<string> ResolveRelativeFromImport(IPythonProjectEntry entry, FromImportStatement node)
+            => ResolveRelativeFromImport(entry.ModuleName, entry.FilePath, node);
+
+        /// <summary>
+        /// Resolves modules listed in the relative import statetement
```
zooba: "statement"

Python/Product/Analysis/ModuleResolver.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// </summary>
+        /// <param name="importingFromModuleName">Name of the importing module</param>
+        /// <param name="importingFromFilePath">Disk path to the importing module</param>
+        /// <param name="node">Import statement node</param>
+        /// <returns></returns>
+        internal static IEnumerable<string> ResolveRelativeFromImport(string importingFromModuleName, string importingFromFilePath, FromImportStatement node) {
+            var root = node.Root.MakeString();
+
+            if (!string.IsNullOrEmpty(root) && root.StartsWith(".")) {
+                var prefix = root.All(c => c == '.') ? root : $"{root}.";
```
zooba: `$` (culture-sensitive formatting) should not be used in this context - use `"{0}.".FormatInvariant(root)` instead

Python/Product/Analysis/ModuleResolver.cs
```diff
@@ -0,0 +1,206 @@
...
+        private static string GetModuleFullName(string originatingModule, string relativePath) {
+            // Check if it is indeed relative
+            if (string.IsNullOrEmpty(relativePath) || relativePath[0] != '.') {
+                return relativePath;
+            }
+
+            var up = relativePath.TakeWhile(ch => ch == '.').Count();
+            var bits = originatingModule.Split('.');
+            if (up > bits.Length) {
+                return relativePath; // too far up
```
zooba: We should probably return `null` here to indicate that there's nothing to import

Python/Product/Analysis/ModuleResolver.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// </summary>
+        /// <param name="entry">Project entry</param>
+        /// <param name="node">Import statement node</param>
+        /// <returns>Names of modules</returns>
+        internal static IEnumerable<string> GetModuleNamesFromImport(IPythonProjectEntry entry, FromImportStatement node)
+            => GetModuleNamesFromImport(entry.ModuleName, entry.FilePath, node);
+
+        /// <summary>
+        /// Extracts actual module names from the 'from import' statetement
+        /// such as [a, b, c] from 'from . import a, b, c' or [a] from 'from .a import b, c'
```
zooba: These seem like inconsistent results. Given the first example, I'd expect `a.b, a.c` from the second, and it's not clear how this relates to the module it's being imported from (really, I'd expect `importingFromModuleName.a, ...` in the results)

Python/Product/Analysis/ProjectEntry.cs
```diff
@@ -56,8 +56,11 @@ IAnalysisCookie cookie
             DocumentUri = documentUri ?? MakeDocumentUri(filePath);
             FilePath = filePath;
             Cookie = cookie;
+
             MyScope = new ModuleInfo(ModuleName, this, state.Interpreter.CreateModuleContext());
             _unit = new AnalysisUnit(null, MyScope.Scope);
+            Analysis = new ModuleAnalysis(_unit, ((ModuleScope)_unit.Scope));
```
zooba: Is this why you're ending up with the unit with null Ast? Aren't we checking that `Analysis` isn't null everywhere already?

Python/Product/Analysis/PythonAnalyzer.cs
```diff
@@ -344,133 +344,14 @@ public partial class PythonAnalyzer : IGroupableAnalysisProject, IDisposable {
...
-                    var adjacentModuleName = string.Join(".", prefix.Take(prefix.Length - 1)) + "." + relativeModuleName;
-                    yield return adjacentModuleName;
-                }
-            }
-
-            // Without absolute_import, we treat the name as complete last.
-            if (!absoluteImports) {
-                yield return relativeModuleName;
+            var unresolved = importFrom.GetModuleInfo().GetAllUnresolvedModules();
+            if(unresolved.Count == 0) {
```
zooba: Minor spacing issues all over the place - suggest Ctrl+K, Ctrl+D everywhere before merging (and check that it hasn't broken anything when you do that).

Python/Tests/Analysis/AnalysisTest.cs
```diff
@@ -6804,16 +6805,18 @@ class Y:
...
-                new { Code="from .deg import A", Index=9, Expected="abc.deg", Base="abc" },
-                new { Code="from .hij import A", Index=9, Expected="abc.deg.hij", Base="abc.deg" },
-                new { Code="from ..hij import A", Index=10, Expected="abc.hij", Base="abc.deg" },
-                new { Code="from ..hij import A", Index=10, Expected="abc.deg.hij", Base="abc.deg.HIJ" },
+                new { Code="from .deg import A", Index=9, Expected="deg.A", Base="abc" },
+                new { Code="from .hij import A", Index=9, Expected="abc.hij.A", Base="abc.deg" },
+                new { Code="from ..hij import A", Index=10, Expected="hij.A", Base="abc.deg" },
+                new { Code="from ..hij import A", Index=10, Expected="abc.hij.A", Base="abc.deg.HIJ" },
+                new { Code="from .. import deg", Index=10, Expected="deg", Base="abc.deg" },
+                new { Code="from . import A", Index=10, Expected="abc.A", Base="abc.deg" },
```
zooba: These tests are all incorrect now - you're dropping too many levels when processing the leading `.`.

I suggest creating these packages and testing everything to prove to yourself that Python treats `.` as "current package" in all cases. Use `print(__name__)` to see the order they get imported in.

zooba:
And just to plant the idea - please squash merge this PR. There are a lot of commits here, and they don't all seem to be related

Python/Product/Analysis/AnalysisUnit.cs
```diff
@@ -164,30 +164,26 @@ internal AnalysisUnit(ScopeStatement ast, InterpreterScope scope)
         internal virtual void AnalyzeWorker(DDG ddg, CancellationToken cancel) {
             DeclaringModule.Scope.ClearLinkedVariables();
 
+            // Remove stale variables
+            var toRemove = DeclaringModule.Scope.AllVariables
+                .Where(v => !v.Value.VariableStillExists && !v.Value.IsAssigned).ToList();
```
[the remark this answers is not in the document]
MikhailArkhipov: After the walk stale variable receives types and `VariableStillExists` becomes true again. So attempt here is to remember what was dead before the walk. However, I can try and chase why it receives types and prevent it there.

Python/Product/Analysis/Analyzer/DDG.cs
```diff
@@ -343,60 +345,76 @@ internal class DDG : PythonWalker {
                     }
                 } else {
                     userMod.Imported(_unit);
-                    fullImpName[fullImpName.Length - 1] = impName;
-                    AssignImportedMember(nameNode, userMod, fullImpName, newName ?? impName);
+                    var modName = node.Root.MakeString();
+                    if (modRef.Name != modName) {
```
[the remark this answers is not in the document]
MikhailArkhipov: I'll double check. The issue was with `from .a import x.y` where bits were not empty.

Python/Product/Analysis/Analyzer/DDG.cs
```diff
@@ -343,60 +345,76 @@ internal class DDG : PythonWalker {
...
+                            resolved(moduleRef, remainingParts);
+                         } else {
+                            resolved(moduleRef, null);
                         }
-                        return true;
-                    } else {
-                        break;
+                        continue;
                     }
+                    unresolved(name);
```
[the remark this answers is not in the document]
MikhailArkhipov: OK, makes sense

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -453,7 +453,9 @@ TryImportModuleContext context
                 return TryImportModuleResult.ModuleNotFound;
             }
 
-            Debug.Assert(!name.EndsWithOrdinal("."), $"{name} should not end with '.'");
+            if(name.EndsWithOrdinal(".")) { // Incomplete construct
```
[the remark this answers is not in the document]
MikhailArkhipov: Let me see if I can repro this on partial construct and check where is it coming from.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -119,10 +120,14 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
 
         #region Client message handling
         public override async Task<InitializeResult> Initialize(InitializeParams @params) {
+            //while (!System.Diagnostics.Debugger.IsAttached) {
+            //    System.Threading.Thread.Sleep(1000);
+            //}
+
```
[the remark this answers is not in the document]
MikhailArkhipov: OK. Comes useful occasionally.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -474,6 +479,8 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
                 _analyzer.AddModuleAlias(first, a);
             }
 
+            AddParentModules(first, path);
```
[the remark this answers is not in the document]
MikhailArkhipov: Yes, I can add comments. The issue and project structure is here https://github.com/Microsoft/PTVS/issues/4138. Folders up the tree don't have `__init__.py` and we miss them.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -495,6 +502,45 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
 
             return Task.FromResult(item);
         }
+        private void AddParentModules(string module, string path) {
+            if (ModulePath.PythonVersionRequiresInitPyFiles(_analyzer.LanguageVersion) ||
+                string.IsNullOrEmpty(_rootDir)) {
+                return;
+            }
+
+            var dir = Path.GetDirectoryName(path);
```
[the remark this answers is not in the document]
MikhailArkhipov: Do we care if path is not valid? I.e. is there a case path with invalid characters or form is useful still?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -495,6 +502,45 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
+                        _analyzer.Modules.TryImport(parentModName, out parentModRef);
+
+                        if (_analyzer.Modules.TryImport(module, out var modRef)) {
+                            parentModRef.AddReference(modRef.AnalysisModule as ModuleInfo);
+                            modRef.AddReference(parentModRef.AnalysisModule as ModuleInfo);
+
+                            entry.GetModuleInfo().AddModuleReference(modRef);
+
+                            var varName = module.Substring(parentModName.Length + 1);
+                            var v = entry.Analysis.Scope.AddVariable(varName, new LocatedVariableDef(entry, new EncodedLocation()));
```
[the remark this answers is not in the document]
MikhailArkhipov: OK. I thought it might be useful to know if the variable is actually fake

Python/Product/Analysis/ModulePath.cs
MikhailArkhipov: OK

Python/Product/Analysis/LanguageServer/Server.cs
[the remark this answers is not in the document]
zooba: We probably need a new property on VariableDef (or a new subclass) to tell if it's fake. `EphemeralVariableDef` might be the right one for this, but I think we also exclude those from results in certain places so perhaps not...

Python/Product/Analysis/LanguageServer/Server.cs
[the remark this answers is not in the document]
zooba: We care because it raises an exception that you aren't handling :) `GetParent` won't raise until we try and use the path for actual I/O, where we'll be handling those exceptions anyway.

Python/Product/Analysis/LanguageServer/Server.cs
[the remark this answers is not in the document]
zooba: Namespace packages should be resolved downwards from our search paths. We can't just resolve them upwards or we'll never know where to stop generating package names.

We can assume the root of the workspace is a search path (and we should already have that information available). Otherwise, the IDE should provide any other search paths explicitly.

Python/Product/Analysis/Analyzer/DDG.cs
```diff
@@ -343,60 +345,76 @@ internal class DDG : PythonWalker {
...
+
+                ModuleReference moduleRef = null;
                 foreach (var part in ModulePath.GetParents(name, includeFullName: true)) {
                     if (ProjectState.Modules.TryImport(part, out var mref)) {
+                        // First part is module
                         moduleRef = mref;
                         if (part.Length < name.Length) {
                             moduleRef.Module?.Imported(_unit);
+                        } else {
+                            resolved(moduleRef, null);
```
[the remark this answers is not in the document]
zooba: `resolved` doesn't resolve it, it assigns it. Python will only ever import one module, so we should resolve the best single module using the available information and assign that (e.g. we have Python runtime version, search paths, ability to read the disk/open ZIP files, etc.)

zooba:
Added a few replies, hope they aren't too hard to find in github's UI. Will take another full look tomorrow (between meetings...)

Python/Product/Analysis/ModulePath.cs
```diff
@@ -664,17 +659,21 @@ out errorParameter
...
             out ModulePath modulePath
         ) {
-            return FromBasePathAndName_NoThrow(basePath, moduleName, null, null, out modulePath, out _, out _, out _);
+            var isPackage = languageVersion.HasValue && PythonVersionRequiresInitPyFiles(languageVersion.Value) ? (Func<string, bool>)null : s => true;
```
zooba: For least surprise, we should assume that `__init__.py` files are required if we don't know the language version. (That's the recommended way to create packages, and it works in all versions, whereas implicit namespace packages only work in recent versions and are only recommended for certain cases.)

Python/Product/Analysis/ModuleResolver.cs
```diff
@@ -0,0 +1,208 @@
...
+
+                var names = node.Names.Where(n => !string.IsNullOrEmpty(n.Name)).Select(n => n.Name);
+                if (!names.Any()) {
+                    return Enumerable.Empty<string>();
+                }
+
+                var resolved = names.SelectMany(n => ResolvePotentialModuleNames(
+                        importingFromModuleName, importingFromFilePath,
+                        "{0}{1}".FormatInvariant(prefix, n), node.ForceAbsolute)).ToArray();
+                if (resolved.Length == 1 && resolved[0].Length > 2 && resolved[0].EndsWithOrdinal(".*")) {
```
zooba: Let's check for `import *` first, and then `ResolvePotentialModuleNames` can be more strict about only returning actual importable names (i.e. `my-module.name` is not importable, so it isn't a potential module name, so we could filter it out in that function - this would probably also help deal with that case where we're passing `..a.` as a "valid" name to `ImportModule`).

Python/Product/Analysis/ModuleResolver.cs
```diff
@@ -0,0 +1,208 @@
...
+            if (up > bits.Length) {
+                return relativePath; // too far up
+            }
+
+            var root = string.Join(".", bits.Take(bits.Length - up));
+            var subPath = relativePath.Trim('.');
+
+            return string.IsNullOrEmpty(root)
+                ? subPath
+                : string.IsNullOrEmpty(subPath) ? root : $"{root}.{subPath}";
```
zooba: Use concatentation or `FormatInvariant` here

Python/Product/Analysis/PythonAnalyzer.cs
```diff
@@ -356,133 +356,14 @@ public partial class PythonAnalyzer : IGroupableAnalysisProject, IDisposable {
...
-                if (prefix.Length > 1) {
-                    var adjacentModuleName = string.Join(".", prefix.Take(prefix.Length - 1)) + "." + relativeModuleName;
-                    yield return adjacentModuleName;
-                }
-            }
-
-            // Without absolute_import, we treat the name as complete last.
-            if (!absoluteImports) {
-                yield return relativeModuleName;
+            var unresolved = importFrom.GetModuleInfo().GetAllUnresolvedModules();
```
zooba: `GetModuleInfo()` may return `null`

Python/Product/Analysis/Values/FunctionInfo.cs
```diff
@@ -115,9 +115,11 @@ internal class FunctionInfo : AnalysisValue, IReferenceableContainer, IHasRichDe
                     calledUnit.Enqueue();
                 }
 
-                res.Split(v => v.IsResolvable(), out _, out var nonLazy);
-                res = DoCall(node, unit, calledUnit, callArgs);
-                res = res.Union(nonLazy);
+                if (calledUnit != null) {
```
zooba: If we don't have a specific called unit here, we should get res from `_analysisUnit.ReturnValue.Types` (which should have the union of all possible return types for this function).

zooba: It's probably easiest to make `DoCall` handle a null calledUnit. We can also assert `calledUnit != null || callingUnit.ForEval` in there, since we should never end up in a situation where we are still analyzing but get a null `calledUnit`.

Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
```diff
@@ -993,9 +993,11 @@ class ImportStatementWalker : PythonWalker {
...
 
             public override bool Walk(FromImportStatement node) {
-                var name = node.Root.MakeString();
-                if (!_analyzer.IsModuleResolved(_entry, name, node.ForceAbsolute)) {
-                    Imports.Add(MakeUnresolvedImport(name, node.Root));
+                var names = ModuleResolver.GetModuleNamesFromImport(_entry, node);
+                foreach (var n in names) {
+                    if (!_analyzer.IsModuleResolved(_entry, n, node.ForceAbsolute)) {
+                        Imports.Add(MakeUnresolvedImport(n, node.Root));
+                    }
```
zooba: We should only add the unresolved import if none of the names are resolved. Otherwise we will warn everywhere that we generate multiple names for the same module, even when the "obvious" one is there.

Python/Tests/Analysis/AnalysisTest.cs
```diff
@@ -6807,16 +6808,18 @@ class Y:
...
             }) {
                 var entry = ProcessTextV3(item.Code);
-                var walker = new ImportedModuleNameWalker(item.Base, item.Index);
+                var walker = new ImportedModuleNameWalker(item.Base, string.Empty, item.Index);
                 entry.Modules[entry.DefaultModule].Tree.Walk(walker);
 
-                Assert.AreEqual(item.Expected, walker.ImportedName);
+                Assert.AreEqual(item.Expected, walker.ImportedModules.First());
```
zooba: We should validate the rest of the suggested names too.

MikhailArkhipov: I added PEP328 test

Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
MikhailArkhipov: This is to handle `from . import a, b, c` where only `a` is actually resolved. However, it can be limited to `.` case since in other cases it may not be necessary.

Python/Product/Analysis/Values/FunctionInfo.cs
MikhailArkhipov: I got null ref here since condition above does not guarantee it to be non-null (i.e. .ForEval case)

Python/Product/Analysis/Analyzer/DDG.cs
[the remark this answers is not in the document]
MikhailArkhipov: Fixed

Python/Product/Analysis/Values/FunctionInfo.cs
MikhailArkhipov: Perhaps it is better to add `|| calledUnit == null` to the `if` above

Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
MikhailArkhipov: So I guess we can add all names if import is from `.` or similar and keep old approach for other cases?

MikhailArkhipov: Actually, implementation of `IsModuleResolved` has changed - it no longer does TryImport so it won't report 'unresolved' entries. It only reports those that were not found at all and reported via 'unresolved' in TryImportModules. So here it only adds those names that appears in `GetModuleInfo()?.GetAllUnresolvedModules()`

Python/Product/Analysis/ProjectEntry.cs
zooba: Right, but that `_unit` would normally never have ended up in `Analysis`, which means `FindScope` would not have been called on it. The change here is what causes that, so I'm wondering why we want this change if it breaks things and we're checking for null `Analysis` already?

Python/Product/Analyzer/Intellisense/OutOfProcProjectAnalyzer.cs
zooba: In any case, this is going to change fairly soon to report via diagnostics rather than being a request from VS.

Python/Tests/Analysis/AnalysisTest.cs
zooba: Ah, you're right. I was assuming (and testing) in the `__init__.py` files in the package, but those will always include `__init__` as the last part of the module name and so the logic still checks out.

Thanks for fixing my error here!

Python/Product/Analysis/AnalysisUnit.cs
```diff
@@ -162,30 +162,28 @@ internal AnalysisUnit(ScopeStatement ast, InterpreterScope scope)
         }
 
         internal virtual void AnalyzeWorker(DDG ddg, CancellationToken cancel) {
+            DeclaringModule.Scope.ClearLinkedVariables();
```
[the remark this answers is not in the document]
MikhailArkhipov: Looks like merge issue, took it out

Python/Product/Analysis/AnalysisUnit.cs
```diff
@@ -162,30 +162,28 @@ internal AnalysisUnit(ScopeStatement ast, InterpreterScope scope)
...
-            List<KeyValuePair<string, VariableDef>> toRemove = null;
-
             foreach (var variableInfo in DeclaringModule.Scope.AllVariables) {
                 variableInfo.Value.ClearOldValues(ProjectEntry);
-                if (variableInfo.Value._dependencies.Count == 0 &&
-                    !variableInfo.Value.HasTypes) {
-                    if (toRemove == null) {
-                        toRemove = new List<KeyValuePair<string, VariableDef>>();
-                    }
+                if (variableInfo.Value._dependencies.Count == 0 && !variableInfo.Value.HasTypes) {
```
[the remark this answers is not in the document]
MikhailArkhipov: OK

Python/Product/Analysis/ProjectEntry.cs
MikhailArkhipov: This was for #4138 which we took out.

zooba:
@MikhailArkhipov Remember to squash merge, but I'd like to get this in before building today

PR #4331, opened 2018-06-06
Fixes build and launch of VTune.

zooba:


Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -361,10 +310,71 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+                    } else {
+                        MessageBox.Show($"The Python interpreter in question is: [{interpreter}]");
+                    }
+                }
+            }
+#endif
+
+            var interpreter = config.GetInterpreterPath();
+            if (!File.Exists(interpreter)) {
+                MessageBox.Show("Can't find specified python interpreter.");
```
huguesv: Move to string resource

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -361,10 +310,71 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+                }),
+                WorkingDirectory = config.WorkingDirectory,
+            };
+
+            var proc = new Process { StartInfo = procInfo };
+            var dte = (EnvDTE.DTE)session._serviceProvider.GetService(typeof(EnvDTE.DTE));
+            proc.EnableRaisingEvents = true;
+            proc.Exited += (_, args) => {
+                if (!File.Exists(Path.Combine(outPathDir, "Sample.dwjson"))) {
+                    MessageBox.Show($"Something happened, cannot find output file");
```
huguesv: Also string resource

Python/Product/Profiling/PythonProfilingPackage.cs
huguesv: Also use Strings.ProductTitle

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -361,10 +310,71 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+#endif
+
+            var interpreter = config.GetInterpreterPath();
+            if (!File.Exists(interpreter)) {
+                MessageBox.Show("Can't find specified python interpreter.");
+                return;
+            }
+
+            string outPathDir = Path.GetTempPath();
+            string outPath = Path.Combine(outPathDir, "pythontrace.diagsession");
```
huguesv: The pythontrace.diagsession and sample.dwjson below will be the exact same path for all visual studio instances. I think it needs to use a random tmp subfolder.

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -361,10 +310,71 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+
+            var proc = new Process { StartInfo = procInfo };
+            var dte = (EnvDTE.DTE)session._serviceProvider.GetService(typeof(EnvDTE.DTE));
+            proc.EnableRaisingEvents = true;
+            proc.Exited += (_, args) => {
+                if (!File.Exists(Path.Combine(outPathDir, "Sample.dwjson"))) {
+                    MessageBox.Show($"Something happened, cannot find output file");
+                } else {
+                    PackageTrace(outPathDir);
+                    dte.ItemOperations.OpenFile(Path.Combine(outPathDir, "trace.diagsession"));
```
huguesv: If `PackageTrace` is the one creating this `trace.diagsession` file, then it would be better to have it return the path instead of hard coding the name in both places.

PR #4339, opened 2018-06-07
Fix colors used for info tip in package manager.

### LINKED ISSUE #4325: Text color in pip security warning incorrect on dark theme
image

huguesv:
Fix #4325

We already had a style definition for TextBlock which forced window text color, so I define a new style to use within info tip.

Also, it turned out that while ThemeResourceKey for InfoBackground was working fine, InfoText wasn't. I changed both to use VsBrushes/VsColors for consistency.

I checked all 4 themes + high contrast (black) + high contrast (white), and they're all fine.

### DIFF
--- a/Common/Product/SharedProject/Wpf/Controls.xaml
+++ b/Common/Product/SharedProject/Wpf/Controls.xaml
@@ -226,6 +226,10 @@
         <Setter Property="img:ImageThemingUtilities.ImageBackgroundColor" Value="{DynamicResource {x:Static wpf:Controls.InfoBackgroundColorKey}}" />
     </Style>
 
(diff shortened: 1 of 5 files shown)


Common/Product/SharedProject/Wpf/Controls.xaml
```diff
@@ -226,6 +226,10 @@
         <Setter Property="img:ImageThemingUtilities.ImageBackgroundColor" Value="{DynamicResource {x:Static wpf:Controls.InfoBackgroundColorKey}}" />
     </Style>
 
+    <Style x:Key="InfoText" TargetType="{x:Type TextBlock}" BasedOn="{StaticResource {x:Type TextBlock}}">
```
AlexanderSher: Do we use this style anywhere right now?

huguesv: Yes, it's used in this PR in pipextension.xaml

Issue #3803, opened 2018-03-01
Migrate find definition to server


zooba:
As part of #3484 we need to migrate the FindDefinition request.

huguesv:
See also Microsoft/vscode-python#1061

PR #4354, opened 2018-06-11
Implement Go To Definition and Symbol Info in the LS

### LINKED ISSUE #4306: Goto definition on "import modulename" no longer works
```
import os
```

place caret on `os` and press F12

result: Cannot go to definition of "os".
expected: os.py is opened

also does not work with modules defined in your project

failing test: NavigableTests.ModuleDefinition

MikhailArkhipov:
Fixes #4306 
#3803 partial - to make it work in VS Code. Current VS part remains for now.

Fixes `ModuleDefinition` and `ModuleImportDefinition` tests 
- Module name was not added as variable definition (no assignment)
- Locations were missing URI or path which prevented them from being added to definitions
- Note - tests are still a bit unstable b/c of the hardcoded 1s timeout calling `GetExpressionAtPointAsync`.

Also fixes random 'invalid document' errors when URI from VS Code did not match URI created by loading files from directory. .NET creates URI with `file:///c:/` while VS Code sends `file:///c%3A/`


MikhailArkhipov:
Looking into `ModuleImportDefinition` which appears to be broken b/c variable type is reference rather than definition.

Python/Tests/PythonToolsMockTests/NavigableTests.cs
```diff
@@ -56,7 +56,7 @@ public class NavigableTests {
 ";
             using (var helper = new NavigableHelper(code, Version)) {
                 // os
-                await helper.CheckDefinitionLocations(7, 2, ExternalLocation(1, 1, "os.py"), Location(1, 8));
```
MikhailArkhipov: `os.py` opens fine, but there is no second location.  Current `FindReferences` only finds location in the `ImportedModuleNameWalker`. May be related to the `ModuleImportDefinition` being broken.

Python/Product/Analysis/LocationInfo.cs
```diff
@@ -16,17 +16,14 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.PythonTools.Analysis.Infrastructure;
```
zooba: Unused?

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisEntry.cs
```diff
@@ -51,7 +51,7 @@ internal sealed class AnalysisEntry : IDisposable {
         ) {
             Analyzer = analyzer;
             Path = path;
-            DocumentUri = documentUri;
+            DocumentUri = documentUri ?? (!string.IsNullOrEmpty(path) ? new Uri(path) : null);
```
zooba: Since DocumentUri is the canonical name, I'd rather not infer it from another property (or else how are we going to be able to match it later). How about we throw if it's null?

MikhailArkhipov: OK. However, `ProjectEntry` does derive it from the path via `MakeDocumentUri`.

zooba: @MikhailArkhipov Only for legacy reasons. None of the language server related code ever passes `null` (though some of it does convert URI back to file system...), but some of our tests and the legacy DB system don't/won't use document URIs at all.

PR #4355, opened 2018-06-11
Typeshed improvements

zooba:
Adds an option for specifying where typeshed is cloned to (and initial support for more general type stub packages)
Adds support to tests for locating typeshed
Improves handling of sequence types in type stub files.


zooba:
@MikhailArkhipov I have one more set of changes coming for this, once I clean up test failures.

PR #4357, opened 2018-06-11
Fixes stuck analysis

zooba:


Python/Product/Analysis/AnalysisUnit.cs
```diff
@@ -177,12 +177,18 @@ internal AnalysisUnit(ScopeStatement ast, InterpreterScope scope)
...
 
             if (toRemove != null) {
-                foreach (var nameValue in toRemove) {
-                    DeclaringModule.Scope.RemoveVariable(nameValue.Key);
-                    // if anyone read this value it could now be gone (e.g. user 
-                    // deletes a class definition) so anyone dependent upon it
-                    // needs to be updated.
-                    nameValue.Value.EnqueueDependents();
+                // Do not allow variable removal to re-enqueue our unit
+                _suppressEnqueue = true;
```
AlexanderSher: Can this method be reentrant?

zooba: No. If it needs to be called again, it calls `Enqueue` and that schedules the next call of this function.

PR #4365, opened 2018-06-12
Add incompatibility check to ptvsd_launcher

### LINKED ISSUE #4345: Show task dialog with option to disable experimental debugger on incompatible python envs
In `ptvsd_launcher.py`, return error 687 ERROR_DLL_MIGHT_BE_INCOMPATIBLE when the python environment is not supported with new ptvsd (2.5, 2.6, 3.1, 3.2, IronPyton). Check for that error in DebugAdapterProcess and DebugAdapterRemoteProcess and show task dialog.

Also, change existing instructions for installing ptvsd when an older version is detected. Instead of printing a command to run, link to a web page that will have a command we can update out of band.

huguesv:
Fix #4345 

Show a task dialog with a link to the documentation where the instructions can be updated to match the latest published ptvsd. Include a task button to disable the experimental debugger when launched from an incompatible environment.

image

image

image

image


Python/Product/Debugger/PtvsdVersionHelper.cs
```diff
@@ -51,26 +51,79 @@ internal class PtvsdVersionHelper {
...
+
+                dlg.Buttons.Add(TaskDialogButton.OK);
+                dlg.Buttons.Insert(0, learnMore);
+                if (allowDisable) {
+                    dlg.Buttons.Insert(0, disable);
+                }
+
+                var selection = dlg.ShowModal();
+                if (selection == learnMore) {
+                    Process.Start("https://aka.ms/upgradeptvsd");
```
zooba: `.Dispose()`

Python/Product/Debugger/PtvsdVersionHelper.cs
```diff
@@ -51,26 +51,79 @@ internal class PtvsdVersionHelper {
...
+        public static void ShowPtvsdModuleNotFoundError() {
+            ShowPtvsdMessage(
+                Strings.ImportPtvsdModuleNotFoundTitle,
+                Strings.ImportPtvsdModuleNotFoundMessage,
+                allowDisable: false,
+                isError: true
+            );
+        }
+
+        private static void ShowPtvsdMessage(string main, string content, bool allowDisable, bool isError) {
```
zooba: I feel like we should have some exception handling around this function. Can't say specifically why, though.

PR #4366, opened 2018-06-13
Allow loading an extension from absolute dll path

vivlimmsft:


Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -373,8 +373,14 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
                 return default(T);
             }
             try {
-                var assembly = File.Exists(assemblyName) ? AssemblyName.GetAssemblyName(assemblyName) : new AssemblyName(assemblyName);
-                var type = Assembly.Load(assembly).GetType(typeName, true);
+                Type type;
+                if (Path.IsPathRooted(assemblyName)) {
+                    type = Assembly.LoadFrom(assemblyName).GetType(typeName, true);
```
AlexanderSher: `Assembly.LoadFrom` loads assembly in a different hive which may create problems in the future.
`AssemblyName.GetAssemblyName(assemblyName)` creates `AssemblyName` that references absolute file path.
What issue are you trying to fix?

vivlimmsft: My assembly is in an arbitrary location on disk, and getting the `AssemblyName` from the path works fine:
```
AssemblyName.GetAssemblyName(assemblyName);
{foo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=9f01fad7930233b0}
```
but when I actually try to load it, it cannot be found:
```
Assembly.Load(AssemblyName.GetAssemblyName(assemblyName))
'Assembly.Load(AssemblyName.GetAssemblyName(assemblyName))' threw an exception of type 'System.IO.FileNotFoundException'
```
If there's a way to add search paths to `Assembly.Load` at runtime I'd be open to that, but as far as I can tell that's what `Assembly.LoadFrom` is for.

MikhailArkhipov: In RTVS we do this
https://github.com/Microsoft/RTVS/blob/ea74954fd2165c30e1e84ff8452373f18d40ab9f/src/VsCode/LanguageServer/Impl/Services/ServiceContainer.cs

```cs

            try {
                var thisAssemblyName = thisAssembly.GetName();
                var name = Path.GetFileNameWithoutExtension(platformAssemblyName);
                var token = thisAssemblyName.GetPublicKeyToken();
                var tokenString = token != null && token.Length > 0 ? Encoding.ASCII.GetString(token) : "null";
                var asmName = new AssemblyName($"{name}, Version={thisAssemblyName.Version}, Culture=neutral, PublicKeyToken={tokenString}");
                assembly = Assembly.Load(asmName);
            } catch(IOException) {
                var thisAssemblyPath = thisAssembly.GetAssemblyPath();
                var assemblyLoc = Path.GetDirectoryName(thisAssemblyPath);
                var platformServicesAssemblyPath = Path.Combine(assemblyLoc, platformAssemblyName);
                assembly = AssemblyLoadContext.Default.LoadFromAssemblyPath(platformServicesAssemblyPath);
            }
```

zooba:
I haven't tried it, but I think a simpler version of Mikhail's snippet is the right approach:

```c#
try {
    var name = File.Exists(assemblyName) ? AssemblyName.GetAssemblyName(assemblyName) : new AssemblyName(assemblyName);
    assembly = Assembly.Load(name);
} catch (IOException) {
    assembly = Assembly.LoadFrom(assemblyName);
}
```

From my reading of the [Assembly.LoadFrom](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=netframework-4.7.1#System_Reflection_Assembly_LoadFrom_System_String_) remarks, this will best handle the cases where extensions are on the search path, already loaded, or have NGEN binaries, but also support falling back on loading directly from a file.

@DinoV - any comments?

zooba:
Since it was my code, I'd like a second approval from @AlexanderSher or @MikhailArkhipov before merging.

zooba:
Since there's no further comments, I'll merge.

PR #4367, opened 2018-06-13
[WIP] Move unresolved imports to the language server + related bug fixes

### LINKED ISSUE #4349: LastReportedDiagnostics isn't cleared after document is closed and reopened
- Create new module or open an existing one
- Make some good amount of changes in the file to increase the  buffer version
- Make an error that is handled by parser, save & close
- Reopen the file, fix an error or add a new one

**Actual** 
Squiggles don't change 

**Expected**
Squiggles match parsing errors

### LINKED ISSUE #4351: `OpenFile._pendingChanges` isn't cleared after document is closed
- Follow the steps from https://github.com/Microsoft/PTVS/issues/4350
- Save file, close it and reopen.
- Make changes in the file until the following exception happen:
```cs
System.InvalidOperationException: 'changes must be in reverse order of start location'

Microsoft.PythonTools.Analysis.DocumentBuffer.Update(Microsoft.PythonTools.Analysis.DocumentChangeSet changes) Line 70
Microsoft.PythonTools.Analysis.ProjectEntry.UpdateDocument(int part, Microsoft.PythonTools.Analysis.DocumentChangeSet changes) Line 429
Microsoft.PythonTools.Analysis.LanguageServer.OpenFile.DidChangeTextDocument(Microsoft.PythonTools.Analysis.LanguageServer.DidChangeTextDocumentParams params, System.Action<Microsoft.PythonTools.Analysis.IDocument> enqueueAction) Line 81
Microsoft.PythonTools.Analysis.LanguageServer.OpenFile.DidChangeTextDocument(Microsoft.PythonTools.Analysis.LanguageServer.DidChangeTextDocumentParams params, System.Action<Microsoft.PythonTools.Analysis.IDocument> enqueueAction) Line 105
Microsoft.PythonTools.Analysis.LanguageServer.Server.DidChangeTextDocument(Microsoft.PythonTools.Analysis.LanguageServer.DidChangeTextDocumentParams params) Line 227
Microsoft.PythonTools.VsCode.LanguageServer.DidChangeTextDocument(Newtonsoft.Json.Linq.JToken token) Line 196
```

### LINKED ISSUE #4363: Inconsistent behavior when showing errors in project items
- Create new *.pyproj
- Add couple of new python modules, add some code with syntax error into each
- Save and close modules
- Save project, close VS
- Open VS, open project

Observe that syntax errors from modules are shown in Error List

- Open and close any of the modules 

Observe that syntax errors from closed module have disappeared

AlexanderSher:
- Move unresolved imports to the analysis server
- Fix #4349 for VS Code: LastReportedDiagnostics isn't cleared after document is closed and reopened bug
- Fix #4351 for VS Code: OpenFile._pendingChanges isn't cleared after document is closed
- Fix #4363: Inconsistent behavior when showing errors in project items
- Fixed memory leaks in ErrorTaskProvider

Fixes for #4349 and #4351 for VS require better synchronization between `BufferParser.AddBuffer` and `BufferParser.RemoveBuffer` that has to be converted into `async` to send and await for `textDocument/didOpen` and `textDocument/didClose`.


Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -27,8 +27,6 @@ public sealed partial class Server {
             await IfTestWaitForAnalysisCompleteAsync();
 
             var uri = @params.textDocument.uri;
-            // Make sure document is enqueued for processing
-            var openFile = _openFiles.GetDocument(uri);
```
AlexanderSher: This doesn't ensure anything, so removed

Python/Product/Analysis/LanguageServer/OpenFiles.cs
```diff
@@ -49,7 +49,7 @@ sealed class OpenFile {
 
         public void DidChangeTextDocument(DidChangeTextDocumentParams @params, Action<IDocument> enqueueAction) {
             var changes = @params.contentChanges;
-            if (changes == null) {
+            if (changes == null || changes.Length == 0) {
```
AlexanderSher: VSC can send `textDocument/didChange` with no changes.

Python/Product/Analysis/Analyzer/ImportStatementWalker.cs
```diff
@@ -0,0 +1,99 @@
...
+                if (!_analyzer.IsModuleResolved(_entry, name, node.ForceAbsolute)) {
+                    Diagnostics.Add(MakeUnresolvedImport(name, nameNode));
+                }
+            }
+            return base.Walk(node);
+        }
+
+        private static bool IsImportError(Expression expr) {
+            if (expr is NameExpression name) {
+                return name.Name == "Exception" || name.Name == "BaseException" || name.Name == "ImportError";
```
zooba: Python 3.6 added `ModuleNotFoundError`, which we should probably also have here

Python/Product/Analysis/LanguageServer/OpenFiles.cs
zooba: 🙄

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -480,9 +483,7 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
                 return Task.FromResult(actualItem);
             }
 
-            if (_clientCaps?.python?.analysisUpdates ?? false) {
-                pyItem.OnNewAnalysis += ProjectEntry_OnNewAnalysis;
-            }
+            pyItem.OnNewAnalysis += (o, e) => OnPythonEntryNewAnalysis(pyItem);
```
zooba: Why change the event handler signature? We don't need the lambda and closure here, and for large workspaces it's a cost we may notice.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -337,10 +340,10 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
             _analyzer = await CreateAnalyzer(@params.initializationOptions.interpreter);
 
-            _clientCaps = @params.capabilities;
-            _settings.SetCompletionTimeout(_clientCaps?.python?.completionsTimeout);
-            _traceLogging = _clientCaps?.python?.traceLogging ?? false;
-            _analyzer.EnableDiagnostics = _clientCaps?.python?.liveLinting ?? false;
+            _clientCaps = new Capabilities(@params.capabilities);
+            _settings.SetCompletionTimeout(_clientCaps.Python.CompletionsTimeout);
+            _traceLogging = _clientCaps.Python.TraceLogging;
+            _analyzer.EnableDiagnostics = _clientCaps.Python.LiveLinting;
```
zooba: `LiveLinting` is currently false by default, since we have other messages behind that flag that we do not want to release. If unresolved imports are going there too, we might need an alternative filter somewhere in the analyzer to allow "good" diagnostics.

The `LiveLinting` flag is meant to enable experimental stuff, so maybe internally we can change it to `EnableExperimentalDiagnostics` and filter those somehow?

MikhailArkhipov: Yeah, unresolved imports should be default on (as warning). We can expose controls for severity like VS does so user can suppress warnings if they chose to.

MikhailArkhipov: Another alternative is create new structure with flags for assorted diagnostic controls and I can expose it in the VSC settings. We can then keep expanding it as needed

MikhailArkhipov:
LGTM apart from the options setting

Python/Product/Analysis/Analyzer/ImportStatementWalker.cs
AlexanderSher: Done

Python/Product/Analysis/LanguageServer/Server.cs
AlexanderSher: `_analyzer.EnableDiagnostics` doesn't affect unresolved imports, and yes, they are reported as warnings.
I like the idea of having granulated diagnostics controls on the client side, that we can switch from `false` to `true` as needed.

zooba:
Yep, but let’s make it “from true to false” on the client side. The language server should be able to introduce new warnings without a client update. But I’d still like to have a “give me experimental warnings” option that doesn’t refer to each warning code individually.

Python/Product/Analysis/LanguageServer/Server.cs
AlexanderSher: Since `OnNewAnalysis` is a public API, I can't change its signature and provide correct type in `eventArgs`. Type check inside old handler was misleading - there can be no other type, so this code is cleaner and easier to understand (no need to look for all existing implementations of `IPythonProjectTree`). Closure contains two reference type fields, so I can't imagine how we may notice it considering hundreds of kilobytes of garbage strings in G2 per every new tree parse.

AlexanderSher:
Matching PR in vscode: https://github.com/Microsoft/vscode-python/pull/2049

Python/Product/Analysis/Analysis.csproj
```diff
@@ -44,6 +44,10 @@
     <PlatformTarget>AnyCPU</PlatformTarget>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Newtonsoft.Json">
```
zooba: I want to be _very_ certain that we need dependencies for this assembly before we add them. Right now this assembly has no external dependencies, so let's be 110% sure we need this and it can't be done in some other way (even if it's a little less convenient for us).

Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -320,23 +324,33 @@ public struct ExecuteCommandCapabilities { public bool dynamicRegistration; }
     }
 
     [Serializable]
-    public class TextDocumentClientCapabilities {
+    public struct TextDocumentClientCapabilities {
```
zooba: I believe this will prevent us from changing the contents of this struct/class in the future. I'd rather it stay as a class (and probably should be sealed).

Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -320,23 +324,33 @@ public struct ExecuteCommandCapabilities { public bool dynamicRegistration; }
...
         [Serializable]
         public struct SynchronizationCapabilities {
+            [DefaultValue(false)]
```
zooba: This is implied - no need to specify it.

Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -479,45 +515,61 @@ public struct RenameCapabilities { public bool dynamicRegistration; }
...
+            TraceLogging = traceLogging;
+            LiveLinting = liveLinting;
+        }
+
         /// <summary>
         /// Client expects analysis progress updates, including notifications
         /// when analysis is complete for a particular document version.
         /// </summary>
-        public bool? analysisUpdates;
+        public readonly bool AnalysisUpdates;
```
zooba: Why change these from how the JSON is represented?

Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -479,45 +515,61 @@ public struct RenameCapabilities { public bool dynamicRegistration; }
...
-
+    
     [Serializable]
     public class ClientCapabilities {
-        public WorkspaceClientCapabilities workspace;
-        public TextDocumentClientCapabilities textDocument;
-        public object experimental;
-        public PythonClientCapabilities python;
+        public readonly WorkspaceClientCapabilities Workspace;
+        public readonly TextDocumentClientCapabilities TextDocument;
```
zooba: These no longer serialize correctly.

Python/Product/Analysis/LanguageServer/Structures.cs
zooba: (Similar for the other types in this file)

AlexanderSher: I agree that we need to declare clear rules on how types for serialization are declared, but I'm not sure how classes would help with versioning. To be detached from protocol changes, we need to remove all references to protocol data types from public part of API and/or inject a group of abstractions.

The benefits of structs are that they are not nullable, which saves us from endless null checks, enforces default values and makes `null` a real expected value instead of a sign that we haven't serialized something.

The benefit of classes is that they are safely mutable, but I'm against mutable protocol data types. Another benefit is that class loses its default constructor when non-default is presented, which will immediately give us deseriazliation error if default value isn't properly defined.

@MikhailArkhipov , what is your opinion?

Python/Product/Analysis/LanguageServer/Structures.cs
AlexanderSher: Because inside C# code we don't care about JSON (that's the whole point of these types) and everywhere else public fields and properties are UpperCamelCase (which is also default for C#). Names in constructor match names in JSON

zooba: As the author of these types, I say the whole point is to represent the JSON messages as closely as possible :) (See my comment below for my actual argument about not changing this)

Python/Product/Analysis/Analyzer/ImportStatementWalker.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
```
zooba: Needs a copyright header (did you add this or just merge it in from someone else's PR? regardless, copy/paste a header please)

PR #4372, opened 2018-06-14
Ability to exclude files from LS directory scan

MikhailArkhipov:
- Specific to VS Code. VSC now passes down include and exclude specifications
- Loading files moved to the VSC LS to avoid additional dependencies on globbing since it is VSC specific

### DIFF
--- a/Python/Product/Analysis/LanguageServer/Server.Completion.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.Completion.cs
@@ -22,9 +22,7 @@ using Microsoft.PythonTools.Parsing;
 
 namespace Microsoft.PythonTools.Analysis.LanguageServer {
     public sealed partial class Server {
-        public override async Task<CompletionList> Completion(CompletionParams @params) {
-            await IfTestWaitForAnalysisCompleteAsync();
-
+        public override Task<CompletionList> Completion(CompletionParams @params) {
             var uri = @params.textDocument.uri;
             // Make sure document is enqueued for processing
             var openFile = _openFiles.GetDocument(uri);
@@ -36,7 +34,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
             var analysis = entry?.Analysis;
             if (analysis == null) {
                 TraceMessage($"No analysis found for {uri}");
-                return new CompletionList();
+                return Task.FromResult(new CompletionList());
             }
 
             var opts = GetOptions(@params.context);
@@ -44,7 +42,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
             var members = ctxt.GetCompletionsFromString(@params._expr) ?? ctxt.GetCompletions();
             if (members == null) {
                 TraceMessage($"Do not trigger at {@params.position} in {uri}");
-                return new CompletionList();
+                return Task.FromResult(new CompletionList());
             }
 
             if (_settings.SuppressAdvancedMembers) {
@@ -92,7 +90,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
                 }
             }
 
-            return res;
+            return Task.FromResult(res);
         }
 
         public override Task<CompletionItem> CompletionItemResolve(CompletionItem item) {
--- a/Python/Product/Analysis/LanguageServer/Server.Completion.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.Completion.cs
@@ -23,7 +23,6 @@ using Microsoft.PythonTools.Parsing;
 namespace Microsoft.PythonTools.Analysis.LanguageServer {
     public sealed partial class Server {
         public override async Task<CompletionList> Completion(CompletionParams @params) {
-            await _analyzerCreationTask;
             await IfTestWaitForAnalysisCompleteAsync();
 
             var uri = @params.textDocument.uri;
--- a/Python/Product/Analysis/ModulePath.cs
+++ b/Python/Product/Analysis/ModulePath.cs
@@ -23,10 +23,9 @@ using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
-using Microsoft.PythonTools.Parsing;
 
 namespace Microsoft.PythonTools.Analysis {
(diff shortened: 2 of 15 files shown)


Python/Product/VSCode/AnalysisVsc/LanguageServer.Lifetime.cs
```diff
@@ -0,0 +1,134 @@
...
+        private async Task LoadFromDirectoryAsync(string rootDir, PatternMatchingResult matchResult) {
+            foreach (var file in matchResult.Files) {
+                if (_sessionTokenSource.IsCancellationRequested) {
+                    break;
+                }
+
+                var path = Path.Combine(rootDir, file.Path);
+                if (!ModulePath.IsPythonSourceFile(path)) {
+                    if (ModulePath.IsPythonFile(path, true, true, true)) {
+                        // TODO: Deal with scrapable files (if we need to do anything?)
```
zooba: I think the answer to this TODO is that when we resolve imports via search paths and the project root is passed as a search path then we'll import them just fine (which is not the default, but there's nothing stopping it - we could also just pass `rootDir` as a search path unconditionally, but I'm not sure that's always going to be right...).

Python/Product/VSCode/AnalysisVsc/LanguageServer.Lifetime.cs
```diff
@@ -0,0 +1,134 @@
...
+                    };
+                }
+            }
+
+            if (parentProcess != null) {
+                Task.Run(async () => {
+                    while (!_sessionTokenSource.IsCancellationRequested) {
+                        await Task.Delay(2000);
+                        if (parentProcess.HasExited) {
+                            _sessionTokenSource.Cancel();
```
zooba: This whole task can probably be simplified to `parentProcess.Exited += (s, e) => _sessionTokenSource.Cancel()`

Python/Product/Analysis/ModulePath.cs
```diff
@@ -23,10 +23,9 @@
 using System.Text.RegularExpressions;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
-using Microsoft.PythonTools.Parsing;
 
 namespace Microsoft.PythonTools.Analysis {
-    struct ModulePath {
+    internal struct ModulePath {
```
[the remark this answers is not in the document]
MikhailArkhipov: It is not accessible from VS language server assembly otherwise. `LoadFileAsync` calls it to determine if file is Python. However we could limit it to `.py` and then won't need `ModulePath`

Python/Product/Analysis/ModuleResolver.cs
```diff
@@ -111,7 +111,10 @@ bool absoluteImports
 
             // Handle relative module names
             if (relativeModuleName.FirstOrDefault() == '.') {
-                yield return GetModuleFullName(importingFrom, relativeModuleName);
+                var fullName = GetModuleFullName(importingFrom, relativeModuleName);
+                if (fullName.Any(c => c != '.')) {
```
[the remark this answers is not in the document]
MikhailArkhipov: `StartsWithOrdinal` will do. Will handle `.......mod` case.

Python/Product/VSCode/AnalysisVsc/AnalysisVsc.csproj
```diff
@@ -11,6 +11,11 @@
         <DebugType>portable</DebugType>
         <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     </PropertyGroup>
+    <ItemGroup>
+      <Compile Remove="obj\**" />
+      <EmbeddedResource Remove="obj\**" />
+      <None Remove="obj\**" />
```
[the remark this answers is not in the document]
MikhailArkhipov: This is .NET Core + NuGet oddity, we *do* redirect `obj` but nuget still creates its props and targets under the project.

Python/Product/Analysis/ModulePath.cs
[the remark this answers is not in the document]
zooba: Does .NET Core not use `internal` by default for structs?

Python/Product/VSCode/AnalysisVsc/LanguageServer.Lifetime.cs
zooba: I see you already had it there... so why the loop as well?

Python/Product/Analysis/ModuleResolver.cs
```diff
@@ -112,7 +112,7 @@ bool absoluteImports
             // Handle relative module names
             if (relativeModuleName.FirstOrDefault() == '.') {
                 var fullName = GetModuleFullName(importingFrom, relativeModuleName);
-                if (fullName.Any(c => c != '.')) {
+                if (fullName.StartsWithOrdinal(".") || fullName.EndsWithOrdinal(".")) {
```
zooba: Invert the sense of this check. We _don't_ want to return the name if it hasn't been resolved (still starts/ends with dots).

Python/Product/VSCode/AnalysisVsc/LanguageServer.Lifetime.cs
MikhailArkhipov: On Mac parentProcess.Exited never happens.

Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -22,10 +22,7 @@
 
 namespace Microsoft.PythonTools.Analysis.LanguageServer {
     public sealed partial class Server {
-        public override async Task<CompletionList> Completion(CompletionParams @params) {
-            await _analyzerCreationTask;
-            await IfTestWaitForAnalysisCompleteAsync();
-
+        public override Task<CompletionList> Completion(CompletionParams @params) {
```
zooba: For the record, I disagree with the warning that says not to use `async` when you aren't using `await` (and it's disabled explicitly). We've had 100% more bugs due to returning `null` instead of a Task than we've had issues with any performance overhead.

MikhailArkhipov: Yeah, it is annoying. I'll try to suppress it in .NET Core projects too then.

PR #4377, opened 2018-06-14
Fixes #4373 socket completions are not correct

### LINKED ISSUE #4373: socket completions are not correct
When we scrape the `_socket` module, it generates an `import socket` statement. This overrides `class socket` defined later in the file, and we end up with poor completions on socket objects.

zooba:
Fixes #4373 socket completions are not correct
Updates list of modules/types that lie about where they are defined
Adds test to verify standard library extension modules have no external dependencies
Adds ability to detect whether ModulePath represents a debug module
Removes typing members from type stubs
Fixes case sensitivity of generated cache paths.
Stops returning very generic overloads when more specific ones are available.

### DIFF
--- a/Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
@@ -194,6 +194,10 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
 
             var hash = SHA256.Create();
             var dir = Path.GetDirectoryName(filePath);
+            if (IsWindows()) {
+                dir = dir.ToLowerInvariant();
+            }
+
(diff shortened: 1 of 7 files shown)


Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -194,6 +194,10 @@ bool useDefaultDatabase
 
             var hash = SHA256.Create();
             var dir = Path.GetDirectoryName(filePath);
+            if (IsWindows()) {
```
huguesv: Whatever problem you are fixing will still exist on Linux/Mac?

zooba: Paths are case sensitive on Linux/Mac, so we won't resolve the imports if the casing doesn't match.

PR #4382, opened 2018-06-15
Fix build and LS tests

MikhailArkhipov:
Restore FileFound event

### DIFF
--- a/Python/Product/Analysis/LanguageServer/Server.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.cs
@@ -301,6 +301,12 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
         public void SetSearchPaths(IEnumerable<string> searchPaths) => _analyzer.SetSearchPaths(searchPaths.MaybeEnumerate());
         public void SetTypeStubSearchPaths(IEnumerable<string> typeStubSearchPaths) => _analyzer.SetTypeStubPaths(typeStubSearchPaths.MaybeEnumerate());
 
+        public event EventHandler<FileFoundEventArgs> OnFileFound;
+        private void FileFound(Uri uri) {
+            TraceMessage($"Found file to analyze {uri}");
+            OnFileFound?.Invoke(this, new FileFoundEventArgs { uri = uri });
+        }
+
         #endregion
 
         #region Private Helpers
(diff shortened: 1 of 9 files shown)


Python/Tests/Analysis/LanguageServerTests.cs
```diff
@@ -106,13 +106,24 @@ public class LanguageServerTests {
...
             }
 
             return s;
         }
 
+        private async Task LoadFromDirectoryAsync(Server s, string rootDir) {
+            foreach (var dir in Directory.EnumerateDirectories(rootDir, "*", SearchOption.AllDirectories)) {
+                await LoadFromDirectoryAsync(s, dir);
+            }
+            foreach (var file in Directory.EnumerateFiles(rootDir)) {
```
zooba: `PathUtils.EnumerateFiles()`

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -301,6 +301,12 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
         public void SetSearchPaths(IEnumerable<string> searchPaths) => _analyzer.SetSearchPaths(searchPaths.MaybeEnumerate());
         public void SetTypeStubSearchPaths(IEnumerable<string> typeStubSearchPaths) => _analyzer.SetTypeStubPaths(typeStubSearchPaths.MaybeEnumerate());
 
+        public event EventHandler<FileFoundEventArgs> OnFileFound;
```
zooba: Do we need this for tests? I'm not seeing anywhere it's essential (and I can't get the `ChildFileAnalyzed` event to trigger in master right now either, which probably means we don't care about it).

zooba:
The mock changes look good. Not sure that raising the FileFound event for explicitly added items is the right fix though - we never used to do that previously.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: It looks functional. `PythonAnalyzer` maintains `_projectFilesByUri`

MikhailArkhipov:
It was firing from
``
                var entry = await LoadFileAsync(new Uri(PathUtils.NormalizePath(file)));
                if (entry != null) {
                    FileFound(entry.DocumentUri);
                }
```
so it is now doing the same just from inside the `LoadFileAsync`. Looks like event comes to the project analyzer which uses it to update `_projectFilesByUri`. This, however, may never actually be called in VS case as it is not loading files from the directory. If you prefer, I can remove code from the `PythonAnalyzer`

File: Python/Product/Analysis/LanguageServer/Server.cs [hunk 3]
```diff
@@ -465,172 +464,172 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
 
-        private void RemoveDocumentParseCounter(Task t, IDocument doc, VolatileCounter counter) {
-            if (t.IsCompleted) {
-                lock (_pendingParse) {
-                    if (counter.IsZero) {
-                        if (_pendingParse.TryGetValue(doc, out var existing) && existing == counter) {
-                            _pendingParse.Remove(doc);
-                        }
-                        return;
+    private void RemoveDocumentParseCounter(Task t, IDocument doc, VolatileCounter counter) {
```
zooba: Whole lot of whitespace changes from here onwards... unindent too much?

PR #4396, opened 2018-06-18
Duplicate experimental debugger ui tests into separate 2.7 and 3.6 tests

huguesv:
Previous behavior for the tests was to use whatever happened to be the default on that machine.

Since we're seeing version specific bugs right now, it will be useful to see results for both 2.7 and 3.6.

I didn't duplicate the old debugger tests, not as much benefit there since the code isn't changing much / at all and that would increase the test run time.


huguesv:
Also updated ptvsd to 4.1.1a8, which has the "wait for exit" bug fix.

PR #4399, opened 2018-06-19
Add command handling for LS extensions

MikhailArkhipov:
- Add OnCommand event to the server (for handling `CompletionItem.command` in extension)
- Register LS as command provider and register command for post-completion
- Suppress 'async without await' warning in .NET Core build
- Fix Linux URI handling (must always use absolute)

Works in VSC. Looking at how to update VS part.

### DIFF
--- a/Python/Product/Analysis/AnalysisNetStandard.csproj
+++ b/Python/Product/Analysis/AnalysisNetStandard.csproj
@@ -4,6 +4,12 @@
     <RootNamespace>Microsoft.PythonTools.Analysis</RootNamespace>
     <AssemblyName>Microsoft.Python.Analysis.Engine</AssemblyName>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;1998</NoWarn>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <NoWarn>1701;1702;1998</NoWarn>
+  </PropertyGroup>
   <Import Project="..\NetStandard.settings" />
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
   <ItemGroup>
--- a/Python/Product/Analysis/AnalysisNetStandard.csproj
+++ b/Python/Product/Analysis/AnalysisNetStandard.csproj
@@ -5,6 +5,10 @@
     <AssemblyName>Microsoft.Python.Analysis.Engine</AssemblyName>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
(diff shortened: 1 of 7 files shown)


Python/Product/Analysis/AnalysisNetStandard.csproj
```diff
@@ -4,6 +4,12 @@
     <RootNamespace>Microsoft.PythonTools.Analysis</RootNamespace>
     <AssemblyName>Microsoft.Python.Analysis.Engine</AssemblyName>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;1998</NoWarn>
```
AlexanderSher: Can you add a comment what are those warnings?

MikhailArkhipov: OK

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -78,7 +78,7 @@ string prefix
             var loc = m.Locations.FirstOrDefault(l => !string.IsNullOrEmpty(l.FilePath));
             if (loc != null) {
                 res.location = new Location {
-                    uri = new Uri(PathUtils.NormalizePath(loc.FilePath), UriKind.RelativeOrAbsolute),
+                    uri = new Uri(PathUtils.NormalizePath(loc.FilePath), UriKind.Absolute),
```
zooba: Since we're converting from a file path here, we can check whether it's rooted before crashing when it's not.

zooba: Or better yet, just use the `DocumentUri` member of the `LocationInfo`.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -245,6 +252,14 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
                 _queue.Enqueue(entry.Value.ProjectEntry, AnalysisPriority.Normal);
             }
         }
+
+        public override Task<object> ExecuteCommand(ExecuteCommandParams @params) {
```
AlexanderSher: I'd really like to pass `async` to the `Command` method. Unfortunately, events don't work with async. Who is going to use `OnCommand`?

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: Good point.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Extension that needs to receive command upon completion. I.e. the `Command` field in the `CompletionItem`

PR #4400, opened 2018-06-20
Fixes #4364 Send to Interactive creates a new interactive session

### LINKED ISSUE #4364: Send to Interactive create a new Interactive session
From https://developercommunity.visualstudio.com/content/problem/211707/python-tools-for-visual-studio-send-to-interactive.html

When i try to send code to the interactive window from a Python file (.py) with CTRL+ENTER, it creates a new interactive window unless I keep CTRL+ENTER pressed or spam the key very quickly.
This is very annoying since i'm not able to use the send to interactive like it would be practical, line by line. 

I'm using a virtual env in Anaconda 5.0.0, IPython interactive mode is enabled . When it starts opening new window it does it for every new evaluation.

zooba:
Fixes #4364 Send to Interactive creates a new interactive session
Searches all open windows in MRU order for a suitable one when sending from a project

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
+++ b/Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
@@ -36,7 +36,10 @@ namespace Microsoft.PythonTools.Repl {
         private readonly Dictionary<int, IVsInteractiveWindow> _windows = new Dictionary<int, IVsInteractiveWindow>();
         private int _nextId = 1;
 
-        private readonly List<IVsInteractiveWindow> _mruWindows = new List<IVsInteractiveWindow>();
+        /// <summary>
+        /// A reverse-ordered list of recently used windows. Last item is most recently used.
+        /// </summary>
+        private readonly List<IVsInteractiveWindow> _lruWindows = new List<IVsInteractiveWindow>();
         private readonly Dictionary<string, int> _temporaryWindows = new Dictionary<string, int>();
 
         private readonly IServiceProvider _serviceProvider;
@@ -91,13 +94,27 @@ namespace Microsoft.PythonTools.Repl {
             return false;
         }
 
-        public IVsInteractiveWindow Open(string replId) {
+        public void OnWindowUsed(IVsInteractiveWindow window) {
+            if (window == null) {
+                throw new ArgumentNullException(nameof(window));
+            }
+
+            lock (_windows) {
+                _lruWindows.Remove(window);
+                _lruWindows.Add(window);
+            }
+        }
+
+        public IVsInteractiveWindow Open(string replId) => Open(replId, null);
+
+        public IVsInteractiveWindow Open(string replId, Func<IInteractiveEvaluator, bool> predicate) {
             EnsureInterpretersAvailable();
 
             lock (_windows) {
-                foreach(var window in _windows.Values) {
+                foreach(var window in _lruWindows.AsEnumerable().Reverse().Concat(_windows.Values)) {
                     var eval = window.InteractiveWindow?.Evaluator as SelectableReplEvaluator;
-                    if (eval?.CurrentEvaluator == replId) {
+                    if (eval?.CurrentEvaluator == replId && predicate?.Invoke(eval) != false) {
+                        OnWindowUsed(window);
                         window.Show(true);
                         return window;
                     }
@@ -107,14 +124,17 @@ namespace Microsoft.PythonTools.Repl {
             return null;
         }
 
-        public IVsInteractiveWindow OpenOrCreate(string replId) {
+        public IVsInteractiveWindow OpenOrCreate(string replId) => OpenOrCreate(replId, null);
+
+        public IVsInteractiveWindow OpenOrCreate(string replId, Func<IInteractiveEvaluator, bool> predicate) {
             EnsureInterpretersAvailable();
 
             IVsInteractiveWindow wnd;
             lock (_windows) {
-                foreach(var window in _windows.Values) {
+                foreach(var window in _lruWindows.AsEnumerable().Reverse().Concat(_windows.Values)) {
                     var eval = window.InteractiveWindow?.Evaluator as SelectableReplEvaluator;
-                    if (eval?.CurrentEvaluator == replId) {
+                    if (eval?.CurrentEvaluator == replId && predicate?.Invoke(eval) != false) {
+                        OnWindowUsed(window);
                         window.Show(true);
                         return window;
                     }
@@ -145,12 +165,14 @@ namespace Microsoft.PythonTools.Repl {
 
             lock (_windows) {
                 _windows[curId] = window;
+                _lruWindows.Add(window);
             }
 
             window.InteractiveWindow.TextView.Closed += (s, e) => {
                 lock (_windows) {
                     Debug.Assert(ReferenceEquals(_windows[curId], window));
                     _windows.Remove(curId);
+                    _lruWindows.Remove(window);
                 }
             };
 
@@ -202,6 +224,7 @@ namespace Microsoft.PythonTools.Repl {
             lock (_windows) {
                 _windows[curId] = window;
                 _temporaryWindows[replId] = curId;
+                _lruWindows.Add(window);
             }
 
             window.InteractiveWindow.TextView.Closed += (s, e) => {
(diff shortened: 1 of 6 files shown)


Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
```diff
@@ -107,14 +124,17 @@ class InteractiveWindowProvider {
             return null;
         }
 
-        public IVsInteractiveWindow OpenOrCreate(string replId) {
+        public IVsInteractiveWindow OpenOrCreate(string replId) => OpenOrCreate(replId, null);
+
+        public IVsInteractiveWindow OpenOrCreate(string replId, Func<IInteractiveEvaluator, bool> predicate) {
```
AlexanderSher: Looks like override with `predicate != null` isn't used.

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
```diff
@@ -91,13 +94,27 @@ class InteractiveWindowProvider {
...
 
-        public IVsInteractiveWindow Open(string replId) {
+        public void OnWindowUsed(IVsInteractiveWindow window) {
+            if (window == null) {
+                throw new ArgumentNullException(nameof(window));
+            }
+
+            lock (_windows) {
+                _lruWindows.Remove(window);
+                _lruWindows.Add(window);
```
AlexanderSher: What does `lru` mean?

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
zooba: Not yet, no. But the API should mirror the `Open()` function.

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
zooba: Least-recently used (opposite order of MRU)

zooba: (To make it MRU we would `Insert(0, window)` instead of `Add(window)`, which takes longer)

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
AlexanderSher: `OpenOrCreate` looks for specific `replId` that can be skipped by predicate and as a result, we will get several windows with identical id's. Is that expected?

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
```diff
@@ -91,13 +94,27 @@ class InteractiveWindowProvider {
...
+            }
+        }
+
+        public IVsInteractiveWindow Open(string replId) => Open(replId, null);
+
+        public IVsInteractiveWindow Open(string replId, Func<IInteractiveEvaluator, bool> predicate) {
```
AlexanderSher: This code becomes very complicated. It has two callbacks, both happen under lock, and may return null even if window is already open. Method requires implicit information about the caller and caller has to know this method implementation, which pretty much devalues any decoupling between `InteractiveWindowProvider` and `ExecuteInReplCommand.EnsureReplWindow`. 

`IVsInteractiveWindow.Show` is publicly available, so `InteractiveWindowProvider` can simply return window instance and leave window opening to the `EnsureReplWindow` itself. Or the whole `EnsureReplWindow` can be moved into `InteractiveWindowProvider`, so that it knows what is actually going on.

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
```diff
@@ -91,13 +94,27 @@ class InteractiveWindowProvider {
             return false;
         }
 
-        public IVsInteractiveWindow Open(string replId) {
+        public void OnWindowUsed(IVsInteractiveWindow window) {
```
AlexanderSher: Does it have to be public?

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
zooba: The lock is probably unnecessary, as you can't safely call this from off the UI thread anyway. I don't see the second callback (unless you mean `OnWindowUsed`? Which isn't actually a callback and could easily be inlined here if I wanted to duplicate the code twice).

All of the complexity already existed, BTW. By suggesting a refactoring for the sake of it you're increasing the complexity of this change.

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
zooba: Yes. We can have multiple windows with the same REPL id - that's why we need the predicate to tell them apart without interrupting the search (which is the underlying cause of this bug)

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
zooba: It's a private class, but the point is that it may be used from elsewhere when we "use" an interactive window.

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
AlexanderSher: I'm totally fine with slightly bigger change, technical debts should be resolved as soon as they can be resolved. The small size of the change is actually not that important if we consider the amount of things one has to know and remember to understand it:

1. Input argument for predicate is always a `SelectableReplEvaluator`, otherwise it won't be called.
1. If `SelectableReplEvaluator.Evaluator` property is `null` or not a `PythonCommonInteractiveEvaluator` for project-matching `replId`, window it will be reused. Before the change, it was the opposite. The type doesn't really matter, because with current implementation it can be only `PythonInteractiveEvaluator` which is derived from `PythonCommonInteractiveEvaluator`, but property itself can be `null` if `SelectableReplEvaluator.SetEvaluator` has failed to create evaluator. So change is important, but not for the bug fix.
1. If `SelectableReplEvaluator._evaluator` is set, it won't change until `SelectableReplEvaluator` instance is disposed.
1. `AssociatedProjectHasChanged == true` doesn't mean that active project has changed, it means that interpreter for that project has changed or file structure has changed (project is assigned by `PythonReplEvaluatorProvider` and is never changed after that) and no action has been done . It also doesn't mean that REPL is stopped, hanged, closed, not yet initialized, etc. But it is set to `false` only by `PythonInteractiveEvaluator.EnsureConnectedAsync`, which is called when something is executed in REPL but not when it is actually reset.

So the bug was actually caused by the REPL window that had the lowest id in `InteractiveWindowProvider._windows` and was never used after interpreter has changed. Significant part of the fix is `_lruWindows` list that checks for newly created window before the old one, and `predicate` doesn't actually matter. Is that right?

zooba: 1. This is not guaranteed to always be the case. From outside the class, you can't tell this or rely on it.
2. I don't think there was a change here in the default case. If you don't provide a predicate, `null != false` is always true and so the condition is irrelevant.
3. Huh? It changes through `SetEvaluator`, not through `OpenOrCreate`. Python supports arbitrary numbers of interactive windows and each one has its own selector for the associated environment.
4. That is all it is meant to mean. You're right that resetting the window only resets that flag because of other hooked events - I'll add it to the `ResetAsync` calls.

You're right that in this particular case the logic works out such that the most recently used window is also the one we want to use (if any), so the predicate isn't necessary and we could simply check the result of `Open(replId)`. However, as soon as we make any other change that could reorder the windows, that breaks, so I'd rather have the more robust change in now so that we are protected.

Python/Product/PythonTools/PythonTools/Repl/InteractiveWindowProvider.cs
zooba: Hypothetically, we could reorder windows every time you use one, or switch focus to it. Neither of these requires "getting" the window from the provider, but it does require the provider to know that it's been used.

PR #4403, opened 2018-06-20
Fixes #3842 Class decorators do not pass type to function

### LINKED ISSUE #3842: Class decorators do not pass type to function
```python
def Dec(X): pass

@Dec
class DecoratorClass: pass

@Dec
def DecoratedFunc(): pass
```

The type of `X` will include `DecoratedFunc`, but not the class. It should include both.

zooba:
Fixes #3842 Class decorators do not pass type to function
Adds decorator handling for class definitions

### DIFF
--- a/Python/Product/Analysis/AnalysisUnit.cs
+++ b/Python/Product/Analysis/AnalysisUnit.cs
@@ -364,12 +364,14 @@ namespace Microsoft.PythonTools.Analysis {
 
     class ClassAnalysisUnit : AnalysisUnit {
         private readonly AnalysisUnit _outerUnit;
+        private readonly Dictionary<Node, Expression> _decoratorCalls;
 
         public ClassAnalysisUnit(ClassDefinition node, InterpreterScope declScope, AnalysisUnit outerUnit)
             : base(node, new ClassScope(new ClassInfo(node, outerUnit), node, declScope)) {
 
             ((ClassScope)Scope).Class.SetAnalysisUnit(this);
             _outerUnit = outerUnit;
+            _decoratorCalls = new Dictionary<Node, Expression>();
 
             AnalysisLog.NewUnit(this);
         }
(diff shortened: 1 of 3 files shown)


Python/Product/Analysis/AnalysisUnit.cs
```diff
@@ -421,8 +423,51 @@ public ClassAnalysisUnit(ClassDefinition node, InterpreterScope declScope, Analy
...
+            var v = _outerUnit.Scope.AddLocatedVariable(Ast.Name, Ast.NameExpression, this);
+            v.AddTypes(this, ProcessClassDecorators(ddg, classInfo));
+        }
+
+        private IAnalysisSet ProcessClassDecorators(DDG ddg, ClassInfo classInfo) {
+            var types = classInfo.SelfSet;
+            if (Ast.Decorators != null) {
+                Expression expr = Ast.NameExpression;
+
+                foreach (var d in Ast.Decorators.DecoratorsInternal) {
```
MikhailArkhipov: `Ast.Decorators.DecoratorsInternal.Where`(x => x != null)?

zooba: Alex changed all of these to use the arrays directly for speed. Fairly sure adding Linq here will remove all of that benefit

Python/Product/Analysis/AnalysisUnit.cs
```diff
@@ -421,8 +423,51 @@ public ClassAnalysisUnit(ClassDefinition node, InterpreterScope declScope, Analy
...
+                        }
+                        expr = nextExpr;
+                        var decorated = AnalysisSet.Empty;
+                        bool anyResults = false;
+                        foreach (var ns in decorator) {
+                            var fd = ns as FunctionInfo;
+                            if (fd != null && Scope.EnumerateTowardsGlobal.Any(s => s.AnalysisValue == fd)) {
+                                continue;
+                            }
+                            decorated = decorated.Union(ns.Call(expr, this, new[] { types }, ExpressionEvaluator.EmptyNames));
```
huguesv: The use of `types` here is what propagates the result of the previous iteration? There doesn't seem to be any test for that  though?

PR #4404, opened 2018-06-20
Fixes #4222 Don't automatically bring up completions after 'for'

### LINKED ISSUE #4222: Don't automatically bring up completions after 'for'
This is almost never what you want, and leads to accidental completions.

In the screenshot below, completions appeared after I typed `for` and space.

image

If I type a comma at that point, it will give me `object`, instead of the `j` that I want.

zooba:
Fixes #4222 Don't automatically bring up completions after 'for'
Fixes handling of completions for For statement.

### DIFF
--- a/Python/Product/Analysis/LanguageServer/CompletionAnalysis.cs
+++ b/Python/Product/Analysis/LanguageServer/CompletionAnalysis.cs
@@ -440,9 +440,17 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
         private IEnumerable<CompletionItem> GetCompletionsInForStatement() {
             if (Statement is ForStatement fs) {
                 if (fs.Left != null) {
-                    if (Index < fs.Left.StartIndex) {
-                        return null;
-                    } else if (Index <= fs.Left.EndIndex) {
+                    if (fs.InIndex > fs.StartIndex) {
+                        if (Index > fs.InIndex + 2) {
+                            return null;
+                        } else if (Index >= fs.InIndex) {
+                            ApplicableSpan = new SourceSpan(Tree.IndexToLocation(fs.InIndex), Tree.IndexToLocation(fs.InIndex + 2));
+                            return Once(InKeywordCompletion);
+                        }
+                    }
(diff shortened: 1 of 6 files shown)


Python/Product/Analysis/LanguageServer/CompletionAnalysis.cs
```diff
@@ -440,9 +440,17 @@ Func<TextReader> openDocument
...
+                        if (Index > fs.InIndex + 2) {
+                            return null;
+                        } else if (Index >= fs.InIndex) {
+                            ApplicableSpan = new SourceSpan(Tree.IndexToLocation(fs.InIndex), Tree.IndexToLocation(fs.InIndex + 2));
+                            return Once(InKeywordCompletion);
+                        }
+                    }
+                    if (fs.Left.StartIndex > fs.StartIndex && fs.Left.EndIndex > fs.Left.StartIndex && Index > fs.Left.EndIndex) {
+                        return Once(InKeywordCompletion);
+                    } else if (fs.ForIndex >= fs.StartIndex && Index > fs.ForIndex + 3) {
```
AlexanderSher: `else` not needed

Python/Product/Analysis/LanguageServer/CompletionAnalysis.cs
```diff
@@ -543,10 +552,23 @@ Func<TextReader> openDocument
             }
             // Allow keywords when we are in another keyword
             if (statement is Statement s && index <= s.KeywordEndIndex) {
+                int keywordStart = s.KeywordEndIndex - s.KeywordLength;
+                if (index >= keywordStart) {
+                    span = new IndexSpan(keywordStart, s.KeywordLength);
+                } else if (((s as ForStatement)?.IsAsync ?? false) ||
```
AlexanderSher: Can we introduce an interface with `IsAsync` property to reduce it to one cast?

Python/Product/Analysis/LanguageServer/CompletionAnalysis.cs
zooba: But expresses the intent better. It's actively harmful to have an `else` on the previous one though (that bug took me a few minutes to find...)

huguesv:
It's working at global scope, but inside a function, completions still appear.
image

PR #4407, opened 2018-06-21
Handle top level . import in VS Code

MikhailArkhipov:
This appears to be VS Code specific.
`from . import x` does not work if both importing module and `x` are at the top of the workspace.

If current module is at the root of the workspace, there is no parent module to resolve . against. In VS there is always parent (project) but in VS Code there may be none. The fix is to treat `from . import test` as `import test` when `.` resolved to nothing.

### DIFF
--- a/Python/Product/Analysis/Analyzer/DDG.cs
+++ b/Python/Product/Analysis/Analyzer/DDG.cs
@@ -304,6 +304,20 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
         public override bool Walk(FromImportStatement node) {
             var modName = node.Root.MakeString();
 
+            var candidates = ModuleResolver.ResolvePotentialModuleNames(_unit.ProjectEntry, modName, node.ForceAbsolute).ToArray();
+            if (candidates.Length == 1 && string.IsNullOrEmpty(candidates[0])) {
+                // If current module is at the root of the workspace, there is no parent module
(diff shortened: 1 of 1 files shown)


Python/Product/Analysis/Analyzer/DDG.cs
```diff
@@ -304,6 +304,20 @@ internal class DDG : PythonWalker {
         public override bool Walk(FromImportStatement node) {
             var modName = node.Root.MakeString();
 
+            var candidates = ModuleResolver.ResolvePotentialModuleNames(_unit.ProjectEntry, modName, node.ForceAbsolute).ToArray();
+            if (candidates.Length == 1 && string.IsNullOrEmpty(candidates[0])) {
+                // If current module is at the root of the workspace, there is no parent module
+                // to resolve . against. In VS there is always parent (project) but in VS Code
+                // there may be none. Treat `from . import test` as `import test`
```
zooba: It doesn't behave like that, so we shouldn't treat it so. `from . import name` fails if you do not have a package (parent module): `ImportError: attempted relative import with no known parent package`

zooba: It shouldn't work in VS either, since we don't treat the project as a package.

MikhailArkhipov: You are right, doesn't work in VS either. OK, if it is not supposed to work, so be it.

PR #4408, opened 2018-06-21
Add preselect field for VSC 1.25+

MikhailArkhipov:


Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -702,6 +702,7 @@ public struct CompletionItem {
         public MarkupContent documentation;
         public string sortText;
         public string filterText;
+        public bool? preselect; // VS Code 1.25+
```
AlexanderSher: Is there a default value for it we want to have on server side?

zooba: The server is only provided CompletionItem values when resolving the content, so I wouldn't expect it to ever care what's in this value.

Also, I think this structure had better become a class in order to support adding more members to it in future. Otherwise any extensions that use it when post-processing completions will fail to load against an updated copy of our assembly.

MikhailArkhipov: OK, I can make it class

MikhailArkhipov: Also, the field is nullable in TypeScript so it is reflected here accordingly.

AlexanderSher: Let's discuss it first. Adding members will help only in the simplest case. Any complex case will fail. For example:
```cs
// Version 1:
public class RootV1 {
    public MemberV1[] Members { get; }
}

public class MemberV1 {
    public int Value { get; }
}

// Version 2:
public class RootV2 {
    public bool Flag { get; }
    public MemberV2[] Members { get; }
}

public class MemberV2 {
    public bool OtherFlag { get; }
    public int Value { get; }
}
```

You can't simply derive `RootV2` from `RootV1` and add a new `Flag` member cause arrays are not co-variant.

MikhailArkhipov: Some time later :-) perhaps due to time constraints

PR #4409, opened 2018-06-21
Fix tooltip for `from import`

### LINKED ISSUE #4358: Tooltip for `from import` shows wrong module
image

When I hover over `module1` in the `from ... import` statement, it should show me the value of `module1`, but actually shows me the package it belongs to.

If I hover over the module elsewhere in the file, it is correct.

MikhailArkhipov:
Fixes #4358

- Add splitting `from ...` statement into parts in the walker to we can fetch appropriate types
- Try regular values from expression first, then imported parts
- Store and return ranges for the hover
- Tests


Python/Product/Analysis/SourceSpan.cs
```diff
@@ -84,8 +75,20 @@ public SourceSpan(int startLine, int startColumn, int endLine, int endColumn)
...
+                Start = other.Start;
+            } else if(other.Start.Line == Start.Line && other.Start.Column < Start.Column) {
+                Start = other.Start;
+            }
+            if (other.End.Line > End.Line) {
+                End = other.End;
+            } else if (other.End.Line == End.Line && other.End.Column > End.Column) {
+                End = other.End;
+            }
+            return this;
```
AlexanderSher: Let's not make mutable structs. We can return new instance of `SourceSpan` here.

MikhailArkhipov: OK

Python/Product/Analysis/SourceSpan.cs
```diff
@@ -60,16 +55,12 @@ public SourceSpan(int startLine, int startColumn, int endLine, int endColumn)
...
-        }
+        public SourceLocation Start { get; private set; }
 
         /// <summary>
         /// The end location of the span. Location of the first character behind the span.
         /// </summary>
-        public SourceLocation End {
-            get { return _end; }
-        }
+        public SourceLocation End { get; private set; }
```
AlexanderSher: Same thing with mutable fields. You can use `SourceSpan(SourceLocation start, SourceLocation end)` constructor in `Union`.

Python/Product/Analysis/SourceSpan.cs
```diff
@@ -94,39 +96,33 @@ public SourceSpan(int startLine, int startColumn, int endLine, int endColumn)
...
-            SourceSpan other = (SourceSpan)obj;
-            return _start == other._start && _end == other._end;
+            var other = (SourceSpan)obj;
+            return Start == other.Start && End == other.End;
         }
 
-        public override string ToString() {
-            return _start.ToString() + " - " + _end.ToString();
-        }
+        public override string ToString() => Start.ToString() + " - " + End.ToString();
```
AlexanderSher: `"{0} - {1}".FormatInvariant`?

Python/Product/Analysis/LanguageServer/Server.Hover.cs
```diff
@@ -76,29 +57,71 @@ public sealed partial class Server {
...
-            string originalExpr;
-            if (expr is ConstantExpression || expr is ErrorExpression) {
-                originalExpr = null;
+            // First try values from expression. This works for 
+            // import statements most of the time as well.
+            var index = tree.LocationToIndex(@params.position);
+            var w = new ImportedModuleNameWalker(entry, index, tree);
+            tree.Walk(w);
+
+            IEnumerable<AnalysisValue> values = null;
```
AlexanderSher: Can you make it `IList<AnalysisValue> values`?

Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
```diff
@@ -18,41 +18,83 @@
 using System.Linq;
 using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Analysis.Infrastructure;
+using Microsoft.PythonTools.Analysis.LanguageServer;
 using Microsoft.PythonTools.Parsing.Ast;
 
 namespace Microsoft.PythonTools.Intellisense {
+    sealed class NamedLocation {
+        public string Name { get; set; }
+        public Range Range { get; set; }
```
zooba: Use `SourceSpan` while we're fully within the analyzer. `Range` is really only for LSP

Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
```diff
@@ -18,41 +18,83 @@
...
+            _ast = ast;
         }
 
-        public IEnumerable<string> ImportedModules { get; private set; } = Enumerable.Empty<string>();
+        public IEnumerable<NamedLocation> ImportedModules { get; private set; } = Enumerable.Empty<NamedLocation>();
+        public IEnumerable<NamedLocation> ImportedMembers { get; private set; } = Enumerable.Empty<NamedLocation>();
+        public NamedLocation ImportedType { get; private set; }
 
         public override bool Walk(FromImportStatement node) {
             if (node.StartIndex <= Location && Location <= node.EndIndex) {
```
zooba: Also change this to check against `node.Root.StartIndex` and EndIndex, so we only get the tooltip when hovering over the module name and not the other imported names.

Python/Product/Analysis/LanguageServer/Server.Hover.cs
```diff
@@ -76,29 +57,71 @@ public sealed partial class Server {
...
+            // Perhaps it is an assigned member like 'p' in 'from os import path as p'
+            if (!string.IsNullOrEmpty(w.ImportedType?.Name)) {
+                values = analysis.GetValues(w.ImportedType.Name, @params.position).ToList();
             } else {
-                originalExpr = @params._expr?.Trim();
-                if (string.IsNullOrEmpty(originalExpr)) {
-                    originalExpr = expr.ToCodeString(tree, CodeFormattingOptions.Traditional);
+                // Perhaps it is a member like 'path' in 'from os import path'
```
zooba: In this and the previous case, they are properly defined variables and our normal handling will be just fine. We only need special handling for the module name (which is not a variable in the current file) and the imported name when it gets reassigned (the `path` in `from os import path as p`) - in the latter case, we should be able to easily set `expr` to the "AsName" from the import statement and it'll show the right content (even better, it'll show the name the user _ought_ to be using to access it).

Python/Product/Analysis/LanguageServer/Server.Hover.cs
```diff
@@ -76,29 +57,71 @@ public sealed partial class Server {
...
+                    _typeNames = names
+                };
+                return Task.FromResult(res);
+            }
+
+            if (w.ImportedModules.Any()) {
+                var sb = new StringBuilder();
+                var range = default(Range);
+                foreach (var n in w.ImportedModules) {
+                    if (_analyzer.Modules.TryImport(n.Name, out var modRef)) {
```
zooba: This one should probably be `TryGetImportedModule` so we aren't modifying the analysis state or triggering background tasks from a hover request.

Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -61,6 +61,13 @@ public struct Range {
         public static implicit operator Range(SourceSpan span) => new Range { start = span.Start, end = span.End };
 
         public override string ToString() => ((SourceSpan)this).ToString();
+
+        public Range Union(Range other) {
```
zooba: I think changing the earlier struct to SourceSpan will make this easier, but I think it's better to keep operations like this on our own classes and only use the LSP structures for interop.

Python/Product/Analysis/SourceSpan.cs
zooba: Yeah, there shouldn't be any need for the setter on these.

Python/Tests/Analysis/AnalysisTest.cs
```diff
@@ -6860,10 +6860,10 @@ class Y:
                 new { Code="from ..hij import A", Index=10, Expected="abc.hij", Base="abc.deg.HIJ" },
             }) {
                 var entry = ProcessTextV3(item.Code);
-                var walker = new ImportedModuleNameWalker(item.Base, string.Empty, item.Index);
+                var walker = new ImportedModuleNameWalker(item.Base, string.Empty, item.Index, null);
                 entry.Modules[entry.DefaultModule].Tree.Walk(walker);
 
-                Assert.AreEqual(item.Expected, walker.ImportedModules.First());
+                Assert.AreEqual(item.Expected, walker.ImportedModules.First().Name);
```
zooba: `FirstOrDefault()?.Name` - better to fail the test with `name != null` than InvalidOperationException

Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
MikhailArkhipov: OK

Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
MikhailArkhipov: Sure

MikhailArkhipov: Shall we unify this at some point? Something like `ISourceLocation` . Comes handy when searching trees (provided nodes implement some `ITextRange`.

Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
MikhailArkhipov: Actually, it is an entry point for the entire node, then code determines where exactly it is - i.e. in `from os import path as P` it finds if hover is over `os`, `path` or `P`.

Python/Product/Analysis/SourceSpan.cs
MikhailArkhipov: This b/c I used `Range`. No need anymore.

Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
zooba: You'll want to merge from master and undo my other fix for this then. But do make sure that the actual Hover code uses the normal handling when the user hovers over an actual variable in the file (`P` in your example). We certainly shouldn't be displaying `os` in that case, and it'll be most consistent to let the normal handling get it.

Python/Product/Analysis/LanguageServer/Server.Hover.cs
MikhailArkhipov: `analysis.GetValues` does not return anything for `path`. But yeah, we can try `AsNames`.

zooba: Right, because `path` isn't a real variable in this file. So we need to handle that one specially, but not `p`.

PR #4412, opened 2018-06-21
Fixes #4406 PEP 561 support

### LINKED ISSUE #4406: PEP 561 support
[PEP 561](https://www.python.org/dev/peps/pep-0561/) defines how to search for and resolve type stub packages.

We should use this process to locate extra type information for imported modules.

zooba:
Fixes #4406 PEP 561 support
Adds support for locating type stub packages.
Makes AstPythonInterpreterFactory import process fully async and cancelable
Fixes minor bugs in import completion and module scraping

### DIFF
--- a/Common/Tests/Utilities/AssertUtil.cs
+++ b/Common/Tests/Utilities/AssertUtil.cs
@@ -199,14 +199,17 @@ namespace TestUtilities
 
         [System.Diagnostics.DebuggerStepThrough]
         public static void DoesntContain<T>(IEnumerable<T> source, IEnumerable<T> value) {
+            var contains = new List<T>();
             foreach (var v in source) {
                 foreach (var v2 in value) {
                     if (v.Equals(v2)) {
-                        Assert.Fail(String.Format("{0} contains {1}", MakeText(source), value));
+                        contains.Add(v2);
                     }
                 }
             }
-
+            if (contains.Any()) {
+                Assert.Fail(String.Format("{0} contains {1}", MakeText(source), MakeText(contains)));
+            }
         }
 
         [System.Diagnostics.DebuggerStepThrough]
--- a/Python/Product/Analysis/Interpreter/Ast/AstCachedPythonModule.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstCachedPythonModule.cs
@@ -16,20 +16,18 @@
 
 using System.Collections.Generic;
 using System.IO;
+using Microsoft.PythonTools.Analysis.Infrastructure;
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
     class AstCachedPythonModule : AstScrapedPythonModule {
-        private readonly string _cachedModuleName;
+        private readonly string _cachePath;
 
-        public AstCachedPythonModule(string name, string cachedModuleName) : base(name, null) {
-            _cachedModuleName = cachedModuleName + ".pyi";
+        public AstCachedPythonModule(string name, string cachePath) : base(name, null) {
+            _cachePath = cachePath;
         }
 
         protected override Stream LoadCachedCode(AstPythonInterpreter interpreter) {
-            if (interpreter.Factory is AstPythonInterpreterFactory factory) {
-                return factory.ReadCachedModule(_cachedModuleName);
-            }
-            return null;
+            return PathUtils.OpenWithRetry(_cachePath, FileMode.Open, FileAccess.Read, FileShare.Read);
         }
 
         protected override List<string> GetScrapeArguments(IPythonInterpreterFactory factory) {
--- a/Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
@@ -99,7 +99,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             return res;
         }
 
-        private async Task<IReadOnlyDictionary<string, string>> GetUserSearchPathPackagesAsync(CancellationToken cancel) {
+        private async Task<IReadOnlyDictionary<string, string>> GetUserSearchPathPackagesAsync(CancellationToken cancellationToken) {
             _log?.Log(TraceLevel.Verbose, "GetUserSearchPathPackagesAsync");
             var ussp = _userSearchPathPackages;
             if (ussp == null) {
@@ -113,7 +113,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
                 }
 
                 _log?.Log(TraceLevel.Verbose, "GetImportableModulesAsync");
-                ussp = await AstPythonInterpreterFactory.GetImportableModulesAsync(usp, cancel);
+                ussp = await AstPythonInterpreterFactory.GetImportableModulesAsync(usp, cancellationToken);
                 lock (_userSearchPathsLock) {
                     if (_userSearchPathPackages == null) {
                         _userSearchPathPackages = ussp;
--- a/Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
@@ -20,6 +20,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Analysis.Infrastructure;
@@ -98,7 +99,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             return res;
         }
 
-        private async Task<IReadOnlyDictionary<string, string>> GetUserSearchPathPackagesAsync() {
+        private async Task<IReadOnlyDictionary<string, string>> GetUserSearchPathPackagesAsync(CancellationToken cancel) {
             _log?.Log(TraceLevel.Verbose, "GetUserSearchPathPackagesAsync");
             var ussp = _userSearchPathPackages;
             if (ussp == null) {
@@ -112,7 +113,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
                 }
 
                 _log?.Log(TraceLevel.Verbose, "GetImportableModulesAsync");
-                ussp = await AstPythonInterpreterFactory.GetImportableModulesAsync(usp);
+                ussp = await AstPythonInterpreterFactory.GetImportableModulesAsync(usp, cancel);
                 lock (_userSearchPathsLock) {
                     if (_userSearchPathPackages == null) {
                         _userSearchPathPackages = ussp;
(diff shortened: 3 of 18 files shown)


Python/Product/Analysis/Interpreter/Ast/AstCachedPythonModule.cs
```diff
@@ -16,20 +16,18 @@
...
+        public AstCachedPythonModule(string name, string cachePath) : base(name, null) {
+            _cachePath = cachePath;
         }
 
         protected override Stream LoadCachedCode(AstPythonInterpreter interpreter) {
-            if (interpreter.Factory is AstPythonInterpreterFactory factory) {
-                return factory.ReadCachedModule(_cachedModuleName);
-            }
-            return null;
+            return PathUtils.OpenWithRetry(_cachePath, FileMode.Open, FileAccess.Read, FileShare.Read);
```
zooba: This returns null immediately if the file does not exist, which is the correct behavior here.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -360,7 +364,7 @@ bool useDefaultDatabase
             ImportableModulesChanged?.Invoke(this, EventArgs.Empty);
         }
 
-        protected virtual async Task<IReadOnlyList<string>> GetCurrentSearchPathsAsync() {
+        protected virtual async Task<IReadOnlyList<string>> GetCurrentSearchPathsAsync(CancellationToken cancel) {
```
zooba: This has only been public API in preview so far, so we can still change it. (Previously the whole class was internal)

Python/Product/Analysis/Interpreter/Ast/AstPythonModule.cs
```diff
@@ -76,6 +76,15 @@ string moduleFullName
             );
         }
 
+        public static IPythonModule FromTypeStub(
+            IPythonInterpreter interpreter,
+            string stubFile,
+            PythonLanguageVersion langVersion,
+            string moduleFullName
+        ) {
```
zooba: The signature intentionally matches the other creation functions for consistency and so we can support requiring those arguments in the future without changing the public API.

Common/Tests/Utilities/AssertUtil.cs
```diff
@@ -199,14 +199,17 @@ public static void DoublesEqual(double expected, double actual, string error)
 
         [System.Diagnostics.DebuggerStepThrough]
         public static void DoesntContain<T>(IEnumerable<T> source, IEnumerable<T> value) {
+            var contains = new List<T>();
```
MikhailArkhipov: value.Any(v => source.Any(s => s.Equals(v)) ? Up to you. Linq is hard to debug ;-)

zooba: That also doesn't give back the items that cause the test to fail. So `source.Intersect(value).ToArray()` is probably what's needed, but it's no big deal here. This is just tests

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -98,7 +99,7 @@ class AstPythonInterpreter : IPythonInterpreter, IModuleContext, ICanFindModuleM
             return res;
         }
 
-        private async Task<IReadOnlyDictionary<string, string>> GetUserSearchPathPackagesAsync() {
+        private async Task<IReadOnlyDictionary<string, string>> GetUserSearchPathPackagesAsync(CancellationToken cancel) {
```
AlexanderSher: I'd prefer to have full name for the parameter: `CancellationToken cancellationToken`

Python/Product/Analysis/Interpreter/SentinelModule.cs
```diff
@@ -34,38 +34,39 @@ sealed class SentinelModule : IPythonModule, IDisposable {
...
-                evt = Interlocked.CompareExchange(ref _event, evt, null) ?? evt;
+            var tcs = _tcs;
+            if (tcs == null) {
+                tcs = new TaskCompletionSource<IPythonModule>();
+                tcs = Interlocked.CompareExchange(ref _tcs, tcs, null) ?? tcs;
             }
 
-            if (!evt.Wait(millisecondsTimeout)) {
-                return _realModule;
+            if (cancel.CanBeCanceled) {
```
AlexanderSher: I don't understand this code piece. Why `CancellationToken` can't cancel `tcs` directly?

zooba: If we have multiple waiters and one cancels, that would prevents all the others from receiving the result.

AlexanderSher: Ok, I need to clarify the lifetime of this thing:
- All this code matters only if `importing == true`. When it is `false`, `_realModule == this` and we don't do anything.
- In this mode, `Completes` should be called for every instance of the `SentinelModule` only once inside `TryImportModule` (should we check for `null` there?). 
- `SentinelModule.Dispose` when `importing == true` is called only after `Complete`

If I'm correct, single instance of `SemaphoreSlim` should be enough:

```cs
private readonly SemaphoreSlim _semaphore;
```
```cs
        public SentinelModule(string name, bool importing) {
            _thread = Thread.CurrentThread;
            Name = name;
            if (importing) {
                _semaphore = new SemaphoreSlim(0, 1);
            } else {
                _realModule = this;
            }
        }
```
```cs
        public Task<IPythonModule> WaitForImport(CancellationToken cancellationToken) {
            var mod = _realModule;
            if (mod != null) {
                return Task.FromResult(mod);
            }
            if (_thread == Thread.CurrentThread) {
                return Task.FromResult(this);
            }

            var task = _semaphore.WaitAsync(cancellationToken);
            return _realModule != null ? Task.FromResult(_realModule): 
        }
```
```cs
public void Complete(IPythonModule module) {
    _realModule = module;
    while (sm.CurrentCount == 0) {
        sm.Release();     
    }
}
```

zooba: Sure, that'll work too. It's basically doing the same thing :)

Python/Product/Analysis/Interpreter/SentinelModule.cs
```diff
@@ -34,38 +34,39 @@ sealed class SentinelModule : IPythonModule, IDisposable {
...
-            _event?.Set();
+            _tcs?.TrySetResult(module);
         }
 
         public void Dispose() {
-            var evt = Interlocked.Exchange(ref _event, null);
-            if (evt != null) {
-                evt.Dispose();
-            }
+            Interlocked.Exchange(ref _tcs, null)?.TrySetCanceled();
```
AlexanderSher: Looks like `Dispose` is not needed at all.

Python/Product/Analysis/LanguageServer/CompletionAnalysis.cs
```diff
@@ -245,10 +245,8 @@ Func<TextReader> openDocument
                 }
                 foreach (var t in ZipLongest(imp.Names, imp.AsNames).Reverse()) {
                     if (t.Item2 != null) {
-                        if (Index > t.Item2.EndIndex && t.Item2.EndIndex > t.Item2.StartIndex) {
+                        if (Index >= t.Item2.StartIndex && t.Item2.EndIndex > t.Item2.StartIndex) {
```
AlexanderSher: Does it actually matter that `t.Item2.EndIndex > t.Item2.StartIndex`?

zooba: Yes. We can get empty names in this part of the ast for partially typed statements.

zooba: Oh wait, I see what you mean (had other changes here in my head). Yeah, this time it doesn't matter, but generally when we're comparing against `EndIndex` it does.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -441,30 +446,53 @@ bool useDefaultDatabase
...
+
+        public enum TryImportModuleResultCode {
             Success,
             ModuleNotFound,
             NeedRetry,
             NotSupported,
             Timeout
         }
 
+        public struct TryImportModuleResult {
```
AlexanderSher: Is it part of a protocol?

zooba: You mean language server protocol? No. It's public API for anyone deriving their own interpreter factory (which is basically nobody but us).

zooba: And it's intentionally scoped to `AstPythonInterpreterFactory`, not to interpreter factories in general. The protocol is `IPythonInterpreter.ImportModule`, which returns `null` if the module can't be imported (and we really need a new async version, but until it can be plumbed through everywhere there's basically no point)

AlexanderSher: Ok, so it's a general purpose struct. Let's make it immutable then.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -264,8 +297,8 @@ class AstPythonInterpreter : IPythonInterpreter, IModuleContext, ICanFindModuleM
         }
 
         public IEnumerable<string> GetModulesNamed(string name) {
-            var usp = GetUserSearchPathPackagesAsync().WaitAndUnwrapExceptions();
-            var ssp = _factory.GetImportableModulesAsync().WaitAndUnwrapExceptions();
+            var usp = GetUserSearchPathPackagesAsync(CancellationToken.None).WaitAndUnwrapExceptions();
+            var ssp = _factory.GetImportableModulesAsync(CancellationToken.None).WaitAndUnwrapExceptions();
```
AlexanderSher: `GetUserSearchPathPackagesAsync` calls `GetImportableModulesAsync`. Should we change the order?

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -316,17 +319,18 @@ bool useDefaultDatabase
...
                 IReadOnlyCollection<string> packages = null;
                 if (File.Exists(searchPath)) {
-                    packages = GetPackagesFromZipFile(searchPath);
+                    packages = GetPackagesFromZipFile(searchPath, cancel);
                 } else if (Directory.Exists(searchPath)) {
-                    packages = await Task.Run(() => GetPackagesFromDirectory(searchPath)).ConfigureAwait(false);
+                    packages = await Task.Run(() => GetPackagesFromDirectory(searchPath, cancel)).ConfigureAwait(false);
                 }
                 foreach (var package in packages.MaybeEnumerate()) {
+                    cancel.ThrowIfCancellationRequested();
```
AlexanderSher: Why? `packageDict` is simply a dictionary, adding values to it is very fast. If we made it this far, there is no reason to throw away results that are already acquired.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
AlexanderSher: Should we add `CancellationToken` to other methods of this class as well?

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -441,30 +446,53 @@ bool useDefaultDatabase
...
+            public static TryImportModuleResult Timeout => new TryImportModuleResult { Status = TryImportModuleResultCode.Timeout };
+        }
+
         public sealed class TryImportModuleContext {
             public IPythonInterpreter Interpreter { get; set; }
             public ConcurrentDictionary<string, IPythonModule> ModuleCache { get; set; }
-            public int Timeout { get; set; } = 5000;
             public IPythonModule BuiltinModule { get; set; }
-            public Func<string, Task<ModulePath?>> FindModuleInUserSearchPathAsync { get; set; }
+            public Func<string, CancellationToken, Task<ModulePath?>> FindModuleInUserSearchPathAsync { get; set; }
```
AlexanderSher: I don't understand the idea of this delegate. Referenced method itself is called only from `AstPythonInterpreterFactory`, why it is in the `AstPythonInterpreter`?

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
zooba: It calls it with different arguments. The `AstPythonInterpreter` implementation of `GetUserSearchPathPackagesAsync` uses the static `GetImportableModulesAsync` to search its own search paths, while the `AstPythonInterpreterFactory.GetImportableModulesAsync` non-static version uses the factory's search paths.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
zooba: It uses private properties and private state of `AstPythonInterpreter`, and may be implemented differently for different interpreters.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
AlexanderSher: Ok

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
AlexanderSher: It is very hard to read. Ok, it is out of this PR scope.

zooba: The only alternative is defining another public interface, and since it's meant to be private to this class, that doesn't really make sense. This gets worse if other callers (i.e. us) decide to start partially (and incorrectly) resolving methods through a public interface that is meant for one precise purpose.

zooba:
@AlexanderSher Any further comments?

PR #4418, opened 2018-06-22
Fixes #4417 from completions incorrectly autocomplete

### LINKED ISSUE #4417: from completions incorrectly autocomplete
with django installed, type:

```
from django.for
```

1. Completions appear after you type `django.`, but as you type `for`, they don't track in the combo box, ie 'focus' stays on first item. This is different behavior compared to `import django.for`.
image

2. Doesn't matter if completions combo box is still up or not, if you press control+space in the hopes of getting `forms`, you'll get `forimport` instead.
image

zooba:
Fixes #4417 from completions incorrectly autocomplete
Fixes selection of applicable span for from names.

### DIFF
--- a/Python/Product/Analysis/LanguageServer/CompletionAnalysis.cs
+++ b/Python/Product/Analysis/LanguageServer/CompletionAnalysis.cs
@@ -274,6 +274,15 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
             return Empty;
         }
 
+        private void SetApplicableSpanToLastToken(Node containingNode) {
+            if (containingNode != null && Index >= containingNode.EndIndex) {
+                var token = Tokens.LastOrDefault();
+                if (token.Key.End >= Index) {
+                    ApplicableSpan = GetTokenSpan(token.Key);
+                }
+            }
+        }
+
         private IEnumerable<CompletionItem> GetCompletionsInImport(ref GetMemberOptions opts, ref List<CompletionItem> additional) {
             if (Statement is ImportStatement imp) {
                 if (imp.Names == null || imp.Names.Count == 0) {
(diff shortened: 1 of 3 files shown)


huguesv:
That works better, but I noticed this:
```
from django.forms im
```

put caret at the end and press ctrl+space and you'll get:
```
from django.forms imimport
```

huguesv:
Completions don't automatically appear after a comma, and a ctrl+space doesn't bring them up either.
image

zooba:
I was seeing that latter one as well - are you on the latest internal VS build? I think it's flakey... the completions are definitely being returned, but the session is being dismissed instead of started.

huguesv:
I'm on Preview 3 [27818], I'll update to 27820 later.

I can file that bug separately, I don't want to pile on bugs and make this PR never ending, but wanted to give you the chance to get the fix in if it's quick/easy.

zooba:
The incomplete "import" case is looking kinda tough, just because of the way we build the AST in that situation. I think I have an easy workaround for it, but if not then it'll just have to stick around for a bit.

huguesv:
The fix for `imimport` looks ok.

It's somewhat ridiculous to be completing this because you might as well type `s`:
```from django.forms import BaseForm a```
after ctrl+space becomes:
```from django.forms import BaseForm aas```

huguesv:
Sometimes ctrl+space after this:
```from django.forms import BaseForm, BaseFormS```
becomes
```from django.forms import BaseForm, BaseFormSas```
instead of getting the completions drop down.

zooba:
The "as" completion matters because it should appear after typing the space. So you can do `from<space>dj<space><tab><space>Base<space><tab><space>AsNameForBaseForm`. But it's hard to get perfectly right with incomplete statements. The same check as I added for import should also work for as though.

PR #4419, opened 2018-06-22
Prevent task not completed failures in debugger ui tests

### LINKED ISSUE #4415: Task not completed errors in debugger UI tests
On a local test run, I get at least 10% of the debugger UI tests failing due to that.

huguesv:
Fix #4415 
Avoid reanalyzing the project at the end of each test by not restoring the global default interpreter while the project is still open.

The tests now pass much more reliably.

### DIFF
--- a/Python/Tests/DebuggerUITests/DebugProjectUITests.cs
+++ b/Python/Tests/DebuggerUITests/DebugProjectUITests.cs
@@ -1044,13 +1044,44 @@ namespace DebuggerUITests {
             }
         }
 
+        class AggregateDisposable : IDisposable {
+            private readonly IDisposable[] _disposables;
+
+            public AggregateDisposable(params IDisposable[] disposables) {
+                _disposables = disposables;
+            }
+
+            public void Dispose() {
+                foreach (var disposable in _disposables) {
+                    disposable.Dispose();
+                }
+            }
+        }
+
+        class CloseAllOnDispose : IDisposable {
+            private readonly VisualStudioApp _app;
+
+            public CloseAllOnDispose(VisualStudioApp app) {
+                _app = app;
+            }
+
+            public void Dispose() {
+                _app.CloseAll();
+            }
+        }
+
         private static IDisposable SelectDefaultInterpreter(PythonVisualStudioApp app, string pythonVersion) {
             if (string.IsNullOrEmpty(pythonVersion)) {
                 // Test wants to use the existing global default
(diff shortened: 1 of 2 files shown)


Python/Tests/DebuggerUITests/DebugProjectUITests.cs
```diff
@@ -1044,13 +1044,44 @@ class EmptyDisposable : IDisposable {
             }
         }
 
+        class AggregateDisposable : IDisposable {
```
AlexanderSher: You can port this one: https://github.com/Microsoft/RTVS/blob/774e5c746bf0bb99d45247a4fdf082199abd2461/src/Common/Core/Impl/Disposables/DisposableBag.cs
But it's up to you.

zooba: `VisualStudioApp` already has `OnDispose(Action)` - any reason why that won't work?

zooba: Maybe we just need to reorder `VisualStudioApp.Dispose` to do the close all first?

huguesv: I'm just chaining the "close all" to run before the end of the existing using statement that's there to restore the global default interpreter.
```
            using (SelectDefaultInterpreter(app, interpreter))
            using (new PythonExperimentalGeneralOptionsSetter(pyService, useVsCodeDebugger: useVsCodeDebugger)) {
                var proj = OpenDebuggerProject(app);
                // do stuff
                
                // close all here before the "default interpreter" is disposed (causes interpreter change, which project subscribes to and causes reanalysis)
            }
```

Close all is already done in app.Dispose, and that's too late for me.

DisposableBag could be useful in general, but isn't saving me any code here.

zooba: I'm not so keen on having many ways to do multiple disposals... what if we changed `SelectDefaultInterpreter` to not be disposable itself, but it becomes a method on `app` and hooks up via `OnDispose()`, and we put that after closing the project?

Given what we use `OnDispose()` for (mostly options), it would seem to make sense to reset those after stopping debugging/closing dialogs/solution anyway.

zooba:
See #4443 for an alternate fix for this that should be even more reliable.

huguesv:
👍 for the alternate fix.

Issue #4279, opened 2018-05-29
Unknown appears in method signature for dict.get


huguesv:
`Unknown()` looks suspicious in this method signature. That's not a real Python thing, is it?

```
d = {'a': 1 }
d.get(
```
![image](https://user-images.githubusercontent.com/1696845/40690106-2d95a2b6-635a-11e8-84e5-715175b28b77.png)

zooba:
Not in this case. We normally filter them out before they ever reach the user, but there have been recent changes here that may have caused it to leak through our module scraper as a different type.

PR #4424, opened 2018-06-22
Fixes #3276 Cross-path intellisense

### LINKED ISSUE #3276: [NoDatabaseFactory] Cross-path Intellisense can't detect information about functions and variables
Branching this off from #3248.  The issue described at #3248 involves the default installation configuration.  In that scenario, when you import something that isn't in the current project and is not from a system library, there is no IntelliSense.

Upon enabling `NoDatabaseFactory`, this is better but not quite solved.  See this screenshot.  Note the active file is from the `llvm` project, and has a search path pointing to the `lit` project.

screen2

Notice here that now `FindTool` and `ToolSubst` are colored (compare to screenshot in #3248), and it properly detects that these are classes and I get Intellisense from them.  But as you can see, `llvm_config` finds no completions.

If I do this same thing from a file that is inside the `lit` project, everything works.  See here:
screen1

Here we are inside of `config.py` which is included in the same project, so I get completion.

zooba:
Fixes #3276 Cross-path intellisense
Improves cross-path intellisense

### DIFF
--- a/Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
+++ b/Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
@@ -36,7 +36,7 @@ namespace Microsoft.PythonTools.Intellisense {
         public IEnumerable<string> ImportedModules { get; private set; } = Enumerable.Empty<string>();
 
         public override bool Walk(FromImportStatement node) {
-            if (node.StartIndex <= Location && Location <= node.EndIndex) {
+            if (node.Root != null && node.Root.StartIndex <= Location && Location <= node.Root.EndIndex) {
                 // Determine if location is over imported parts such as 
                 // over 'a' in 'from . import a, b, c' or over 'x' in 'from a import x'
                 // and store module names and imported parts
--- a/Python/Product/Analysis/OverloadResult.cs
+++ b/Python/Product/Analysis/OverloadResult.cs
@@ -253,18 +253,27 @@ namespace Microsoft.PythonTools.Analysis {
             _extraParameters = extraParams;
             _removedParams = removedParams;
             _projectState = state;
-
-            if (!overload.ReturnType.Any()) {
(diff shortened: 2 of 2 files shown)


Python/Product/Analysis/Intellisense/ImportedModuleNameWalker.cs
```diff
@@ -36,7 +36,7 @@ class ImportedModuleNameWalker : PythonWalkerWithLocation {
         public IEnumerable<string> ImportedModules { get; private set; } = Enumerable.Empty<string>();
 
         public override bool Walk(FromImportStatement node) {
-            if (node.StartIndex <= Location && Location <= node.EndIndex) {
+            if (node.Root != null && node.Root.StartIndex <= Location && Location <= node.Root.EndIndex) {
```
zooba: @MikhailArkhipov I just suggested this change to you on your PR. If you take it there as well, it should merge just fine and I'm pretty sure it doesn't affect the other changes you made either.

Python/Product/Analysis/OverloadResult.cs
```diff
@@ -267,6 +262,20 @@ internal BuiltinFunctionOverloadResult(PythonAnalyzer state, string name, IPytho
             : this(state, name, overload, removedParams, null, extraParams) {
         }
 
+        private static IEnumerable<string> GetInstanceDescriptions(PythonAnalyzer state, IEnumerable<IPythonType> type) {
+            foreach (var t in type) {
+                var av = state.GetAnalysisValueFromObjects(t);
+                var inst = av?.GetInstanceType();
+                if (inst.IsUnknown()) {
```
huguesv: Does this take care of https://github.com/Microsoft/PTVS/issues/4279 ?

zooba: No, this only applies to the return types. But it'll likely be a similar change in the same class for that one.

PR #4430, opened 2018-06-25
Fixes #4429 Incorrect refactor rename of parameter with default value

### LINKED ISSUE #4429: Incorrect refactor rename of parameter with default value
Refactor rename `y`:
```
def test(x, y=2):
    return x * y / (x + y)
```
image

Setting new name to `eee` results in:
```
def test(x, eee):
    return x * y / (x + y)
```

Reported at https://developercommunity.visualstudio.com/content/problem/279682/wrong-renaming-of-python-function-argument-with-de.html

zooba:
Fixes #4429 Incorrect refactor rename of parameter with default value
Get correct location of parameter name.

### DIFF
--- a/Python/Product/Analysis/AnalysisUnit.cs
+++ b/Python/Product/Analysis/AnalysisUnit.cs
@@ -346,13 +346,9 @@ namespace Microsoft.PythonTools.Analysis {
 
         public LocationInfo ResolveLocation(object location) {
             Node node = (Node)location;
-            MemberExpression me = node as MemberExpression;
-            SourceSpan span;
-            if (me != null) {
-                span = me.GetNameSpan(Tree);
-            } else {
-                span = node.GetSpan(Tree);
-            }
+            var span = (node as MemberExpression)?.GetNameSpan(Tree)
+                ?? (node as Parameter)?.NameExpression?.GetSpan(Tree)
+                ?? node.GetSpan(Tree);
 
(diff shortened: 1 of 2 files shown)


Python/Product/Analysis/AnalysisUnit.cs
```diff
@@ -346,13 +346,9 @@ internal AnalysisUnit(ScopeStatement ast, InterpreterScope scope)
...
-            MemberExpression me = node as MemberExpression;
-            SourceSpan span;
-            if (me != null) {
-                span = me.GetNameSpan(Tree);
-            } else {
-                span = node.GetSpan(Tree);
-            }
+            var span = (node as MemberExpression)?.GetNameSpan(Tree)
+                ?? (node as Parameter)?.NameExpression?.GetSpan(Tree)
+                ?? node.GetSpan(Tree);
```
AlexanderSher: Is it possible that `Parameter.NameExpression` is `null`?

zooba: Very theoretically. No harm in making sure here though.

AlexanderSher: I'm asking because semantics depends on it. There is a difference between

- If node is `Parameter`, then call `Parameter.NameExpression.GetSpan`
- If it is not, call `node.GetSpan` 

and

 - If node is `Parameter` and `NameExpression != null` then call `GetSpan`, otherwise fallback to `node.GetSpan(Tree)`

PR #4443, opened 2018-06-26
Test fixes

### LINKED ISSUE #4119: Task List always empty in PTVS
I recently restart using PTVS for the first time in a couple of years, and I've noticed that none of the comments which used to show up in the Task List are appearing.  

e.g.
```
# TODO: an item to be done
# HACK: a hack to be rewritten
```

The Task List window is always empty. 

I'm running Visual Studio 2017, build 15.6.6 and Python 15.6.18072.2.

The Task List feature is working normally in my C++ projects, so I don't think it's a problem with my Visual Studio configuration.  I haven't made any changes to the defaults in Tools->Environment->Task List.

Is this a known problem?  If not, any suggestions on how to debug it or workaround it?

### LINKED ISSUE #4415: Task not completed errors in debugger UI tests

### LINKED ISSUE #4440: Invalid cross-thread call in CustomCommands tests
`PythonToolsUITestsRunner.BuildTasksUITests.CustomCommands*` are failing with this error. 

```
Test method PythonToolsUITestsRunner.BuildTasksUITests.CustomCommandsErrorList_27 threw exception: 

TestRunnerInterop.TestFailedException: AssertFailedException: Assert.Fail failed. Invalid cross-thread call from thread 26
```
```
Assert.Fail failed. Invalid cross-thread call from thread 26

   at void TestUtilities.AssertListener.Fail(string message, string detailMessage) in E:\A\_work\27\s\Common\Tests\Utilities\AssertListener.cs:line 140

   at void System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()

   at void TestUtilities.AssertListener.ThrowUnhandled() in E:\A\_work\27\s\Common\Tests\Utilities\AssertListener.cs:line 67

   at void TestUtilities.UI.VisualStudioApp.Dispose(bool disposing) in E:\A\_work\27\s\Common\Tests\Utilities.UI\UI\VisualStudioApp.cs:line 136

   at void TestUtilities.UI.Python.PythonVisualStudioApp.Dispose(bool disposing) in E:\A\_work\27\s\Python\Tests\Utilities.Python\PythonVisualStudioApp.cs:line 116

   at void TestUtilities.UI.VisualStudioApp.Dispose() in E:\A\_work\27\s\Common\Tests\Utilities.UI\UI\VisualStudioApp.cs:line 141

   at IVsHostedPythonToolsTestResult TestUtilities.UI.HostedPythonToolsTestRunner.InvokeTest(Type type, MethodInfo method, object[] arguments) in E:\A\_work\27\s\Common\Tests\Utilities.UI\HostedPythonToolsTestRunner.cs:line 148
```

zooba:
Fixes #4440 Invalid cross-thread call in CustomCommands tests
Fixes #4415 Fixes tests that reset default analyzer failing due to the analyzer not closing
Fixes #4119 Task List always empty

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
@@ -1047,8 +1047,11 @@ namespace Microsoft.PythonTools.Intellisense {
             // remove our sources.
             buffer.Services.MaybeErrorTaskProvider?.RemoveBufferForErrorSource(buffer.Filename, ParserTaskMoniker, buffer.Buffer);
             buffer.Services.MaybeErrorTaskProvider?.RemoveBufferForErrorSource(buffer.Filename, UnresolvedImportMoniker, buffer.Buffer);
+            buffer.Services.MaybeErrorTaskProvider?.RemoveBufferForErrorSource(buffer.Filename, InvalidEncodingMoniker, buffer.Buffer);
             buffer.Services.MaybeCommentTaskProvider?.RemoveBufferForErrorSource(buffer.Filename, ParserTaskMoniker, buffer.Buffer);
+
             buffer.Services.MaybeUnresolvedImportSquiggleProvider?.RemoveBuffer(buffer);
+            buffer.Services.MaybeInvalidEncodingSquiggleProvider?.RemoveBuffer(buffer);
         }
 
         internal void OnAnalysisStarted() {
@@ -1779,6 +1782,7 @@ namespace Microsoft.PythonTools.Intellisense {
             } else {
                 _services.MaybeErrorTaskProvider?.ClearErrorSource(entry.Path, ParserTaskMoniker);
                 _services.MaybeErrorTaskProvider?.ClearErrorSource(entry.Path, UnresolvedImportMoniker);
+                _services.MaybeErrorTaskProvider?.ClearErrorSource(entry.Path, InvalidEncodingMoniker);
                 _services.MaybeCommentTaskProvider?.ClearErrorSource(entry.Path, ParserTaskMoniker);
             }
             if (entry?.Path != null) {
--- a/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
@@ -237,7 +237,7 @@ namespace Microsoft.PythonTools.Intellisense {
                 Trace.TraceInformation($"Connection log: {_conn.LogFilename}");
             }
 
-            _processingTask = Task.Run(() => _conn.ProcessMessages().WaitAndHandleAllExceptions(_services.Site, allowUI: false));
+            _processingTask = Task.Run(() => _conn.ProcessMessages().HandleAllExceptions(_services.Site, allowUI: false));
 
             _toString = $"<{GetType().Name}:{_interpreterFactory.Configuration.Id}:{_analysisProcess}:{comment.IfNullOrEmpty(DefaultComment)}>";
             _userCount = 1;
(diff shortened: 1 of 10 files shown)


Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -236,7 +236,7 @@ IPythonInterpreterFactory factory
                 Trace.TraceInformation($"Connection log: {_conn.LogFilename}");
             }
 
-            Task.Run(() => _conn.ProcessMessages().HandleAllExceptions(_services.Site, allowUI: false)).DoNotWait();
+            Task.Run(() => _conn.ProcessMessages().HandleAllExceptions(_services.Site, allowUI: false)).DoNotWait(waitInTests: false);
```
AlexanderSher: I don't think we should trick the test. Have you figured out why `ProcessMessages` doesn't finish?

zooba: Because the default analyzer doesn't shut down at the end of the test.

zooba: This task would be better represented as a complete thread that we join on when disposing the analyzer (after it's terminated through another mechanism). It really shouldn't be going through `DoNotWait` at all.

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -1767,7 +1773,14 @@ class ExprWalker : PythonWalker {
...
             _analysisComplete = false;
 
-            entry?.TryGetBufferParser()?.ClearBuffers();
+            var bp = entry?.TryGetBufferParser();
+            if (bp != null) {
+                bp.ClearBuffers();
+            } else {
+                _services.MaybeErrorTaskProvider?.ClearErrorSource(entry.Path, ParserTaskMoniker);
+                _services.MaybeErrorTaskProvider?.ClearErrorSource(entry.Path, UnresolvedImportMoniker);
+                _services.MaybeCommentTaskProvider?.ClearErrorSource(entry.Path, ParserTaskMoniker);
```
AlexanderSher: Do you need `_services.MaybeCommentTaskProvider?.ClearErrorSource(entry.Path, InvalidEncodingMoniker);` here as well?

zooba: Probably... that one is fairly inconsistent right now...

Ideally it would be detected and reported as a regular error by the language server, and then we wouldn't need a separate moniker for it.

Python/Product/PythonTools/PythonTools/Intellisense/ProjectAnalyzer.cs
```diff
@@ -236,7 +237,7 @@ IPythonInterpreterFactory factory
                 Trace.TraceInformation($"Connection log: {_conn.LogFilename}");
             }
 
-            Task.Run(() => _conn.ProcessMessages().HandleAllExceptions(_services.Site, allowUI: false)).DoNotWait();
+            _processingTask = Task.Run(() => _conn.ProcessMessages().WaitAndHandleAllExceptions(_services.Site, allowUI: false));
```
AlexanderSher: Synchronously waiting for task inside a `Task.Run` isn't a good idea. It may end up in a deadlock even in SyncContext-free environment. 
`_processingTask = Task.Run(() => _conn.ProcessMessages()).HandleAllExceptions(_services.Site, allowUI: false)` should be enough.

But this doesn't answer the question what was the problem during test run. `TaskObserver` waits several seconds before it throws. Is it a problem only with default analyzer?

zooba: Yes. The default analyzer stays running for the VS instance, and we reuse the VS instance between successful tests, so the analyzer is running still. So the problem is that our test infrastructure is waiting on a task that isn't meant to complete.

I _think_ you're right that we don't need to wait on this task, but let me check it before changing.

PR #4444, opened 2018-06-26
Fixes #4192 Quick Info tooltip follows the mouse as it moves

### LINKED ISSUE #4192: Quick Info tooltip follows the mouse as it moves
Comment from the editor team:

> I noticed that python’s quick info likes to ‘follow the mouse’ as it moves. Quick info is intended to stay in one place until the mouse cursor moves outside of the token’s span, so this feels a bit weird when compared with Roslyn or C++ behavior.

zooba:
Fixes #4192 Quick Info tooltip follows the mouse as it moves
Only dismisses quick info if we have moved outside the applicable span.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
@@ -102,47 +102,67 @@ namespace Microsoft.PythonTools.Intellisense {
                 .DoNotWait();
         }
 
+        private static async Task DismissQuickInfo(IAsyncQuickInfoSession session) {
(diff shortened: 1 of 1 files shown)


huguesv:
I notice that you can go left to right on an expression, but not right to left.
Example:
`HOST = os.environ.get()`
If you put the mouse on `os`, wait for quick info then move right, it will update when you get to `environ`. It doesn't update if you do it in reverse.

zooba:
That requires changing how we calculate the applicable span here, so that the span doesn't match the expression we're evaluating.

It's actually in a totally separate part of the code from this, so we can do it separately and there's no overlap.

PR #4452, opened 2018-06-27
Fixes #4450 No type information on sys.exc_info() return type

### LINKED ISSUE #4450: No type information on sys.exc_info() return type, with typeshed
In the editor, no quick info or completions for:
```
import sys
e1, e2, e3 = sys.exc_info()
e2.
```
image

This worked some time ago.

zooba:
Fixes #4450 No type information on sys.exc_info() return type
Replaces default import of typing module with fully faked module based on a custom pyi file
Makes AstPythonMultipleMembers transparently provide the most specific available interface.
Fixes missing comparisons that were causing types to disappear
Fixes incorrect classification of words from type stubs.


Python/Product/Analysis/Analysis.csproj
```diff
@@ -470,6 +469,12 @@
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
   </ItemGroup>
+  <ItemGroup>
+    <Content Include="typing-stub.pyi">
```
MikhailArkhipov: Should I package it in VSC?

Python/Product/Analysis/Interpreter/Ast/AstAnalysisWalker.cs
```diff
@@ -216,39 +230,94 @@ private IMember Clone(IMember member)
...
                         }
                         (mem as IPythonModule)?.Imported(Scope.Context);
-                    } else {
-                        mem = new AstNestedPythonModuleMember(name.Key, mod, Scope.Context, GetLoc(name.Value));
-                    }
-                    Scope.SetInScope(name.Value.Name, mem);
-                    if (isTyping) {
-                        _typingMembers.Add(name.Value.Name);
                     }
+                    Scope.SetInScope(name.Value.Name, mem, scope: scope);
```
MikhailArkhipov: OK for `mem` to be null?

Python/Product/Analysis/Interpreter/Ast/AstAnalysisWalker.cs
```diff
@@ -216,39 +230,94 @@ private IMember Clone(IMember member)
...
+                            (int)((te.Items.ElementAtOrDefault(0) as ConstantExpression)?.Value ?? 0),
+                            (int)((te.Items.ElementAtOrDefault(1) as ConstantExpression)?.Value ?? 0)
+                        );
+                    } catch (ArgumentException) {
+                        // Unsupported comparison, so walk all children
+                        return true;
+                    }
+
+                    switch (cmp.Operator) {
+                        case PythonOperator.LessThan:
```
MikhailArkhipov: All cases appear to be the same? Collapse into one?

Python/Product/Analysis/Analysis.csproj
zooba: Yep, you'll need it.

Python/Product/Analysis/Interpreter/Ast/AstAnalysisWalker.cs
zooba: It shouldn't ever be null, but yes it's okay (it means to remove the value from the current scope).

Python/Product/Analysis/Interpreter/Ast/AstAnalysisWalker.cs
zooba: The comparison should be different between each, but I can simplify it a bit with a temporary variable.

huguesv:
Class instances are being colorized the same as classes. On another machine without this PR, I'm not seeing that.
image

huguesv:
Same machine without PR is ok, so it's a regression in this PR.

PR #4453, opened 2018-06-27
Implement error/warnings control settings + diagnostics in open files only

MikhailArkhipov:
- Add support for reporting diagnostics in open files only
- Pass typeshed paths from VS Code
- Add appropriate LS settings to control error/warnings/disable assignments
- Handle settings change (updating diagnostics)
- Rename OpenFile(s) => EditorFile(s) since they are no necessarily 'open'
- Consolidate diagnostic publishing in the EditorFile(s)
- Add severity translation

Will try and add tests soon


Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,257 @@
...
+            _server = server;
+        }
+
+        public EditorFile GetDocument(Uri uri) => _files.GetOrAdd(uri, _ => new EditorFile(_server));
+        public void Remove(Uri uri) => _files.TryRemove(uri, out _);
+        public void Open(Uri uri) => GetDocument(uri).Open(uri);
+        public void Close(Uri uri) => GetDocument(uri).Close(uri);
+
+        public void UpdateDiagnostics() {
+            foreach (var entry in _server.ProjectFiles.All) {
```
AlexanderSher: `_server.ProjectFiles.All` returns values from `ConcurrentDictionary`, so it can be modified while enumerator is working. Do we care about that?

AlexanderSher:
> diagnostics in open files only

Is it to match JS/TS behavior or to improve performance?

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
```diff
@@ -108,18 +109,32 @@ private void OnUnregisterCapability(object sender, UnregisterCapabilityEventArgs
...
 
-            var settings = pythonSection.ToObject<LanguageServerSettings>();
-            return _server.DidChangeConfiguration(new DidChangeConfigurationParams { settings = settings });
+            var autoComplete = pythonSection["autoComplete"];
+            settings.completion.showAdvancedMembers = GetSetting(autoComplete, "showAdvancedMembers", true);
+
+            var analysis = pythonSection["analysis"];
+            settings.analysis.openFilesOnly = GetSetting(analysis, "openFilesOnly", false);
+
+            settings.analysis.errors = GetSetting(analysis, "errors", Array.Empty<string>());
```
AlexanderSher: Are there any problem with converting `JToken` into `LanguageServerSettings` directly?

Python/Product/Analysis/LanguageServer/Settings.cs
```diff
@@ -15,24 +15,62 @@
...
+            }
 
-        [Serializable]
-        internal class DiagnosticsSettings {
-            public DiagnosticSeverity unresolvedImports;
+            public bool Show(DiagnosticSeverity severity) => severity != DiagnosticSeverity.Unspecified;
+
+            private void Init() {
+                if (errors != null && errors.Length > 0) {
+                    _errors = new HashSet<object>(errors);
```
AlexanderSher: I'm not sure I understand this code. Why do we need `errors` field?

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,257 @@
...
+            if (ShouldHideDiagnostics) {
+                ClearDiagnostics(documentUri);
+                return true;
+            }
+            return false;
+        }
+
+        private bool ShouldHideDiagnostics => !IsOpen && _server.Settings.analysis.openFilesOnly;
+
+        class PublishDiagComparer : IEqualityComparer<PublishDiagnosticsEventArgs> {
```
zooba: Use the full name

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,257 @@
...
+                    diags = _parseDiagnostics.Values.Except(new[] { documentParseDiags }).Concat(new[] { combined }).ToArray();
+                } else {
+                    diags = _parseDiagnostics.Values.Concat(new[] { _analysisDiagnostics }).ToArray();
+                }
+            }
+
+            if (_lastPublishedDiagnostics == null || !_lastPublishedDiagnostics.SequenceEqual(diags, new PublishDiagComparer())) {
+                _lastPublishedDiagnostics = diags;
+                foreach (var d in diags) {
+                    _server.PublishDiagnostics(d);
```
zooba: Don't we need to do these as a single PublishDiagnostics call?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -613,49 +603,44 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
+            }
+
+            if (newSettings.analysis.openFilesOnly != oldSettings.analysis.openFilesOnly) {
+                _editorFiles.UpdateDiagnostics();
+            }
+
+            if (!newSettings.analysis.errors.SequenceEqual(oldSettings.analysis.errors) ||
+                !newSettings.analysis.warnings.SequenceEqual(oldSettings.analysis.warnings) ||
+                !newSettings.analysis.information.SequenceEqual(oldSettings.analysis.information) ||
+                !newSettings.analysis.disabled.SequenceEqual(oldSettings.analysis.disabled)) {
```
zooba: We should use set equality here, not sequence equality. (`errors.Except(oldErrors).Any()`)

Python/Product/Analysis/LanguageServer/Settings.cs
```diff
@@ -15,24 +15,62 @@
...
-    internal class LanguageServerSettings {
-        public AutoCompleteSettings autoComplete;
-        public DiagnosticsSettings diagnostics;
+    public sealed class LanguageServerSettings {
+        public class PythonAnalysisOptions {
+            public bool openFilesOnly;
+            public object[] errors = Array.Empty<object>();
+            public object[] warnings = Array.Empty<object>();
+            public object[] information = Array.Empty<object>();
+            public object[] disabled = Array.Empty<object>();
```
zooba: I'm happy to require these all be strings. LSP allows them to be numbers or strings, but I think we get to dictate which one they'll be.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,257 @@
...
+            for (var i = 0; i < diags.Count; i++) {
+                diags[i].severity = _server.Settings.analysis.GetEffectiveSeverity(diags[i].code, diags[i].severity);
+            }
+
+            var severity = _server.Settings.analysis.GetEffectiveSeverity(ErrorMessages.UnresolvedImportCode, DiagnosticSeverity.Warning);
+            var pythonProjectEntry = projectEntry as IPythonProjectEntry;
+            var parse = pythonProjectEntry?.GetCurrentParse();
+
+            if (parse != null && _server.Settings.analysis.Show(severity)) {
+                var walker = new ImportStatementWalker(parse.Tree, pythonProjectEntry, _server.Analyzer, severity);
```
zooba: If this walker is solely used to find missing imports, put that in its name.

If it's used during analysis, collect the diagnostics there instead of running it afterwards.

Python/Product/Analysis/LanguageServer/Settings.cs
```diff
@@ -15,24 +15,62 @@
...
 
-        [Serializable]
-        internal class AutoCompleteSettings {
-            public bool showAdvancedMembers;
-        }
+            private HashSet<object> _errors;
+            private HashSet<object> _warnings;
+            private HashSet<object> _information;
+            private HashSet<object> _disabled;
```
zooba: Initialize these in the constructor?

zooba: Or better yet, remap them to a `Dictionary<code, severity>` so that getting the overridden severity is a simple `TryGetValue`. (But leave the interop schema as the four arrays.)

Python/Product/Analysis/LanguageServer/Settings.cs
```diff
@@ -15,24 +15,62 @@
...
+                if (_information != null && _information.Contains(code)) {
+                    return DiagnosticSeverity.Information;
+                }
+                return defaultSeverity;
+            }
 
-        [Serializable]
-        internal class DiagnosticsSettings {
-            public DiagnosticSeverity unresolvedImports;
+            public bool Show(DiagnosticSeverity severity) => severity != DiagnosticSeverity.Unspecified;
```
zooba: Is this really useful? Being a non-static method implies that some calculation is going on, but that's really not the case, so we may as well just check it where we are decided to hide the entire message.

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
zooba: `LanguageServerSettings` isn't an LSP structure (and I've already suggested changing it so that it doesn't look like a JSON object either).

Also, we think there's a 99% change that the settings sent from the editor won't match how we want to use it, at least once PM creates a design for how users will set the settings... so may as well have the code here so we can easily adapt it later.

AlexanderSher: > LanguageServerSettings isn't an LSP structure 

Ok, but why we can't use JSON.Net to feel properties of the object instead of doing it manually? JSON.Net compiles all converters, so there is no performance issue.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
MikhailArkhipov: If file is gone, we do want to skip it so there is no harm. If file was added and diagnostic is available, then good too.

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
MikhailArkhipov: This is sort of like TS side does it - i.e. fetching settings one by one in most cases (there are couple of cases of direct conversion from interface to object). The code translates extension json settings to the structure we have internally. Fields and their types may or may not match extension json.

Python/Product/Analysis/LanguageServer/Settings.cs
MikhailArkhipov: errors is a string array that comes from VSC (i.e. user stuff). The setting is used to map error code like `unresolved-import` to severity - i.e. we need to know which array the code belongs to (see my PR in vscode-python.
```json
                "python.analysis.errors": {
                    "type": "array",
                    "default": [],
                    "items": {
                        "type": "string"
                    },
                    "description": "List of diagnostics messages to be shown as errors.",
                    "scope": "resource"
                },
```

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: What is old is longer than the new one? Would it make new.Except(old).Any empty?

MikhailArkhipov: `old.Intersect(new).Any()` I guess

Python/Product/Analysis/LanguageServer/Settings.cs
MikhailArkhipov: Sure. I just thought they may be empty most of the time so avoided creation of hashsets

Python/Product/Analysis/LanguageServer/Settings.cs
MikhailArkhipov: This is just for the analyzer convenience since `AddDiagnostic` et al use `object code`

MikhailArkhipov: I can change that to strings if you prefer

zooba: Those can change to strings as well

Python/Product/Analysis/LanguageServer/Server.cs
zooba: Yeah, you're right. Turn them into sets and do SetEquals() I guess

Python/Product/Analysis/LanguageServer/Settings.cs
zooba: They can still be initialized to null if they'd be empty. Doing it lazily doesn't really achieve anything here apart from keeping the original arrays alive for longer.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        public void UpdateParseDiagnostics(VersionCookie vc, Uri documentUri) {
+            lock (_lock) {
+                var last = _parseBufferDiagnostics;
+                foreach (var kv in vc.GetAllParts(documentUri)) {
+                    var part = _server.ProjectFiles.GetPart(kv.Key);
+                    if (!last.TryGetValue(part, out var lastVersion) || lastVersion.Version < kv.Value.Version || _ignoreDiagnosticsVersion) {
+                        last[part] = kv.Value;
+
+                        PublishDiagnostics(new PublishDiagnosticsEventArgs {
```
AlexanderSher: `PublishDiagnostics` raises events. Let's move it out of the lock.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,218 @@
...
+                            _version = kv.Value.Version
+                        });
+                    }
+                }
+                _ignoreDiagnosticsVersion = false;
+            }
+        }
+
+        public void UpdateAnalysisDiagnostics(IProjectEntry projectEntry, int version) {
+            lock (_lock) {
```
AlexanderSher: Whole method seems too much for a lock. Do we actually need it for anything except `_parseBufferDiagnostics`?

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,218 @@
...
+                var diags = _server.Analyzer.GetDiagnostics(projectEntry);
+                for (var i = 0; i < diags.Count; i++) {
+                    diags[i].severity = _server.Settings.analysis.GetEffectiveSeverity(diags[i].code, diags[i].severity);
```
AlexanderSher: I think we can cache ` _server.Settings.analysis` at the beginning of the method

Python/Product/Analysis/LanguageServer/EditorFiles.cs
AlexanderSher: - `_server.Analyzer` is shared and has its own thread safety mechanisms
- `_server.Settings` is also shared and can be changed from code that is outside of the `EditorFile` scope

Python/Product/Analysis/LanguageServer/Settings.cs
MikhailArkhipov: This stems from parser and tokenizer use of `ErrorCodes` which are integers and `DiagnosticsErrorSink`. We can change `Diagnostic` to hold a string and do something like

```cs
            var d = new Diagnostic {
                code = "E{0}".FormatInvariant(errorCode),
                message = message,
                source = _source,
                severity = GetSeverity(severity),
                range = span
            };
```

Python/Product/Analysis/LanguageServer/EditorFiles.cs
MikhailArkhipov: Ah, but there is a catch. Parser and analysis run on different threads. So out of the lock you can get events in a random order while they should always come as parse 1 => analysis 1 => parse 2 => analysis 2. I may have to create a separate task/queue for this then.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
MikhailArkhipov: We need definitive order of even firing, see above.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -151,15 +151,17 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
             };
         }
         public override Task Shutdown() {
-            Interlocked.Exchange(ref _analyzer, null)?.Dispose();
-            _projectFiles.Clear();
+            Analyzer?.Dispose();
```
AlexanderSher: Is `Analyzer` protected against double dispose?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -214,46 +216,46 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
+                // Make sure reload modules is executed on the analyzer thread.
+                var task = _reloadModulesQueueItem.Task;
+                _queue.Enqueue(_reloadModulesQueueItem, AnalysisPriority.Normal);
+                await task;
 
-            // re-analyze all of the modules when we get a new set of modules loaded...
-            foreach (var entry in _analyzer.ModulesByFilename) {
-                _queue.Enqueue(entry.Value.ProjectEntry, AnalysisPriority.Normal);
+                // re-analyze all of the modules when we get a new set of modules loaded...
+                foreach (var entry in Analyzer.ModulesByFilename) {
```
AlexanderSher: `Analyzer` may become null here because of the `Shutdown` concurrent call. We should either check it for null again or cache the value of the field at the beginning

AlexanderSher: Or we can catch NRE and throw `OperationCanceledException` instead

zooba: It shouldn't ever be concurrent, but I vote for either caching it or just not setting it to null and making it do ObjectDisposedException properly.

zooba: Actually, looking at the broader scope, that may happen here.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -622,49 +611,45 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
+
+            if (newSettings.analysis.openFilesOnly != oldSettings.analysis.openFilesOnly) {
+                _editorFiles.UpdateDiagnostics();
+                return false;
+            }
+
+            if (!newSettings.analysis.errors.Intersect(oldSettings.analysis.errors).Any() ||
+                !newSettings.analysis.warnings.Intersect(oldSettings.analysis.warnings).Any() ||
+                !newSettings.analysis.information.Intersect(oldSettings.analysis.information).Any() ||
+                !newSettings.analysis.disabled.Intersect(oldSettings.analysis.disabled).Any()) {
```
zooba: This isn't the right check either - will skip updating when changing from `['a', 'b']` to `['b', 'c']`.

I wish Linq had a `.SetEqual` extension method like `.SequenceEqual` (feel free to add it to Analysis.Infrastructure if you want), but it really does have to look like:

```
if (!new HashSet<string>(newSetting.analysis.errors).SetEquals(oldSettings.analysis.errors)) { ... }
```

Python/Product/Analysis/LanguageServer/Settings.cs
```diff
@@ -15,24 +15,49 @@
...
+            public string[] disabled = Array.Empty<string>();
 
-        [Serializable]
-        internal class AutoCompleteSettings {
-            public bool showAdvancedMembers;
-        }
+            private Dictionary<string, DiagnosticSeverity> _map;
+
+            public DiagnosticSeverity GetEffectiveSeverity(string code, DiagnosticSeverity defaultSeverity) {
+                Init();
```
zooba: Pretty sure this can race (e.g. multiple parses completing on separate threads).

Python/Product/Analysis/LanguageServer/SingleThreadSynchronizationContext.cs
```diff
@@ -0,0 +1,60 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.PythonTools.Analysis.LanguageServer {
```
zooba: Since this isn't really language server specific, it could go into Infrastructure.

Python/Product/Analysis/LanguageServer/SingleThreadSynchronizationContext.cs
```diff
@@ -0,0 +1,60 @@
...
+            while (true) {
+                _workAvailable.Wait(_cts.Token);
+                if (_cts.IsCancellationRequested) {
+                    break;
+                }
+                while (_queue.TryDequeue(out var t)) {
+                    try {
+                        t.Item1(t.Item2);
+                    } catch (Exception ex) {
+                        _logger?.TraceMessage($"Exception processing request: {ex.Message}");
```
zooba: Do we really want this? If errors fall out this far, we really want to know about it.

At a minimum, `Debug.Fail` as well. Though I'd be just as happy to let it crash, since then we'll hear about real problems instead of simply hiding them and wondering why certain functionality doesn't work.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Analyzer is pretty big to track disposed state atm... Ditto for `AnalysisQueue`. We may need to port `DisposableBag` here. I'll just check it here. The condition is not particularly critical since if server is shutdown, the worst thing is that it continues consuming some CPU for few seconds. Also, we are in a single thread context here so I am not sure Shutdown can come while `DidChangeConfiguration` is still running.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Nope

Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -697,14 +688,13 @@ public class CompletionList {
     }
 
     [Serializable]
-    public class CompletionItem {
+    public struct CompletionItem {
```
zooba: For future versioning, I'd rather this stayed as `class`.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: However, this is single thread context

Python/Product/Analysis/LanguageServer/Structures.cs
```diff
@@ -239,16 +240,18 @@ public struct Interpreter {
...
-        public string[] searchPaths;
+        public string[] searchPaths = Array.Empty<string>();
+
         /// <summary>
         /// Secondary paths to search when resolving modules. Not supported by all
         /// factories. In generaly, only source files will be discovered, and their
         /// contents will be merged with the initial module.
         /// </summary>
-        public string[] typeStubSearchPaths;
+        public string[] typeStubSearchPaths = Array.Empty<string>();
```
zooba: Do you think we should move `searchPaths` and `typeStubSearchPaths` into `LanguageServerSettings` so we can update them without restarting the process? (Not for this PR, unless it's easy)

Python/Product/Analysis/LanguageServer/SingleThreadSynchronizationContext.cs
zooba: Or it can stay in the VSCode LS host?

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: OK

Python/Product/Analysis/LanguageServer/SingleThreadSynchronizationContext.cs
MikhailArkhipov: This is to use Post and have PublishDiagnostic to be callable under the lock since it fires events.

Python/Product/Analysis/LanguageServer/SingleThreadSynchronizationContext.cs
MikhailArkhipov: I am usually on a fence here. If exception is trivial such as canceled operation, we can keep user running. Or if it was an extension exception. On the other hand, we do want to know...

Python/Product/Analysis/LanguageServer/Structures.cs
MikhailArkhipov: Eventually, yes. Just kept them here for simplicity for now. They don't change that often.

Python/Product/Analysis/LanguageServer/Server.cs
AlexanderSher: `ObjectDisposedException` should be thrown only if we want calling code to be able to restart/reinitialize disposed object. If we simply want to cancel the execution, it is better to throw `OperationCanceledException`. We have another disposable primitive, `DisposeToken` (https://github.com/Microsoft/RTVS/blob/master/src/Common/Core/Impl/Disposables/DisposeToken.cs), which can be used in async methods.

MikhailArkhipov: I'd do OperationCanceledException since it is generally normal condition.

Python/Product/Analysis/LanguageServer/Settings.cs
AlexanderSher: It can race, but I don't see any issue with this dictionary filled multiple times. We can use `ConcurrentDictionary` or simply assign the instance to the field at the end of the method to avoid concurrent access before dictionary is filled. The `EditorFile._lock` is widely used object, but it is still scoped to one file, while `PythonAnalysisOptions` is shared for the whole server.

Python/Product/Analysis/LanguageServer/Server.cs
zooba: We previously awaited on a task running in the analysis thread, so I think we could end up having Shutdown come in between.

IMO, ObjectDisposedException is a perfectly good signal that we should abandon the object. In this context, that means exit the process, since we only have one server per process.

zooba: OperationCanceledException may already be used from this class to indicate that you can send another request, so it becomes overloaded on context if we use it here because the server was shut down.

Python/Product/Analysis/LanguageServer/Settings.cs
zooba: We're going to end up filling the dictionary in probably 99.9% of instances, so we may as well do it as soon as we have the new values.

The problem is with a partially filled dictionary, which is either going to randomly NRE, or show diagnostics to the user that they explicitly disabled.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Would it cause unnecessary Watsons? I.e. analysis still running when host exits?

AlexanderSher: But you don't abandon the object. You abandon the code flow. In rare case where you care why it is abandoned you need to ask the object directly, cause `ObjectDisposedException` doesn't show you what object is disposed (it can be server, it can be stream, it can be some local scope). But in most cases you just want to stop execution.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,240 @@
...
+                var last = _parseBufferDiagnostics;
+                var lastPublishedAnalysisVersion = _lastReportedAnalysisDiagnostics?.FirstOrDefault()._version;
+
+                foreach (var kv in vc.GetAllParts(documentUri)) {
+                    var part = _server.ProjectFiles.GetPart(kv.Key);
+                    if (!last.TryGetValue(part, out var lastVersion) || lastVersion.Version < kv.Value.Version || _ignoreDiagnosticsVersion) {
+                        last[part] = kv.Value;
+                        // Verify that this version is newer than the last published one so
+                        // parse diagnostics does not overwrite analysis of the same version.
+                        if (!lastPublishedAnalysisVersion.HasValue || lastPublishedAnalysisVersion.Value < kv.Value.Version) {
```
zooba: `Diagnostic` has a source property on it. How much of this merging logic is unnecessary if we publish the two sets of diagnostics independently?

Python/Product/Analysis/LanguageServer/Server.cs
AlexanderSher: I don't want to add try/catch blocks to every method that uses server. This is pretty much the same trash code that handles error codes.

zooba: Except the reason this error occurs is because `Shutdown` was called and the server should be abandoned.

If we were handling request cancellation here, then yeah, we just want to cancel the operation. But when it's because there is no point using this server ever again I think ObjectDisposedException is totally appropriate (even if it's being thrown somewhat asynchronously).

@MikhailArkhipov If the call that's awaiting on this is handling whatever exception we throw, then there shouldn't be any crash reports. The callback shouldn't go away, though it likely will fail to send a response at that point (though I don't think it has a response to send here anyway).

Python/Product/Analysis/LanguageServer/Settings.cs
AlexanderSher: Setting value to the `PythonAnalysisOptions._map` field after dictionary is fully filled solves the problem with partially filled dictionary.

Python/Product/Analysis/LanguageServer/SingleThreadSynchronizationContext.cs
zooba: Is it worth having this on Server itself then? Then we can route all events through it automatically

Python/Product/Analysis/LanguageServer/Settings.cs
zooba: Initializing it once in the constructor solves that and the multiple-creation problem introduced by doing it lazily without a lock.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
MikhailArkhipov: Just this check I think. Analysis includes parsing diagnostic so it is OK for it to overwrite parser but not the other way around. We could have 'Python parser' and 'Python analysis' sources, but it might be confusing to the user, esp with combination with third party linter.

Python/Product/Analysis/LanguageServer/Server.cs
AlexanderSher: > Except the reason this error occurs is because `Shutdown` was called and the server should be abandoned.

How this case is different? Shutdown is called only through JsonRPC. There will be no more new requests after that, only those that are currently unfinished. Client code that triggered them doesn't care why they aren't completed and it shouldn't handle server disposed state. Its the job of the code that called Shutdown or the code that is going to send new request (spin the new instance of the process first).

zooba: Then we shouldn't throw any exception from here at all and should silently exit.

Though I'd rather do that by either checking a flag on the server, on the analyzer, or handling an ObjectDisposedException here and using that to silently stop (as we do throughout the debugger, for example).

Python/Product/Analysis/LanguageServer/EditorFiles.cs
zooba: The "confusing to user" might be worth testing, if it's going to let us have simpler code here.

We could label them "Python" (parser) and "Python static analysis" (analysis) instead?

Python/Product/Analysis/Intellisense/AnalysisQueue.cs
```diff
@@ -97,19 +98,20 @@ sealed class WaitAnalyzableTask : IAnalyzable {
         public event EventHandler<UnhandledExceptionEventArgs> UnhandledException;
 
         public void Enqueue(IAnalyzable item, AnalysisPriority priority) {
-            int iPri = (int)priority;
+            ThrowIfDisposed();
```
AlexanderSher: This is a good example of what I'm trying to explain. `Enqueue` is called from `OnDocumentChangeProcessingComplete`, which is called from a thread-pool scheduled task: https://github.com/Microsoft/PTVS/blob/8470db3f81016111613d569f117c331a2f11ee6a/Python/Product/Analysis/LanguageServer/Server.cs#L556.
If `OCE` is thrown, `DoNotWait` will perfectly handle it. If `ODE` is thrown, we will get a Watson. And adding something like `IgnoreObjectDisposedException` isn't a solution, because we need to crash on `ObjectDisposedException` from other disposed objects.

Python/Product/Analysis/LanguageServer/Server.cs
AlexanderSher: > Then we shouldn't throw any exception from here at all and should silently exit

Exactly, so that is what `OperationCanceledException` does. C# doesn't have any language-specific mechanism that allows us to abort some code flow, it uses OCE instead. 

`OperationDisposedException` is very old, it was added to the FCL when C# code was totally thread-bound, and `IDisposable` intended to serve completely different purpose (from which we still have that absurd FxCop requirement about full `IDisposable` implementation).

Python/Product/Analysis/LanguageServer/Settings.cs
MikhailArkhipov: The map should not be changing inside the loop or user may get some codes translated to old settings and some to new.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Soooo, `OperationCanceledException` it is?

Python/Product/Analysis/LanguageServer/Settings.cs
MikhailArkhipov: Added explicit initialization. It shouldn't need lock since settings are replaced wholesale and then are only read.

Python/Product/Analysis/LanguageServer/SingleThreadSynchronizationContext.cs
MikhailArkhipov: I can move it to the server base then

MikhailArkhipov: On the other hand, it is not strictly needed there. Editor file uses Post to execute sequence of code or events. Routing events at the server level still may cause mix ups. Currently loop is executed fully at once, 
```cs
            _syncContext.Post(_ => {
                foreach (var a in args) {
```
but if Post becomes part of `_server.PublishDiagnostics` it will no longer hold true.

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Apparently, this discussion is mostly academical :-) in VS Code we get 'Shutdown' which, per LSP spec should NOT exit, but never get Exit since right after the Shutdown, without any waits, the process is terminated. Happens when either VSC exits or workspace is closed. 

So, the process is insta-gone, it is pretty low probability it will be able to do much after the shutdown request.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
MikhailArkhipov: OK, done

Python/Product/Analysis/Intellisense/AnalysisQueue.cs
MikhailArkhipov: We should get Watson then. I tested with VSC, we get 'shutdown' and then process is insta-terminated. No exit comes. So it is unlikely process will be able to do much before it is killed.

AlexanderSher: Then why do we care about calling `ThrowIfDisposed` at all? There is no value in this code.

MikhailArkhipov: Might come useful for VS when it switches to actual LS from the current OOP analyzer.

Python/Product/Analysis/Infrastructure/EnumerableExtensions.cs
```diff
@@ -48,5 +48,11 @@ public static bool IsNullOrEmpty<T>(this IEnumerable<T> source)
         public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T> source) where T : class {
             return source.Where(NotNull);
         }
+
+        public static bool SetEquals<T>(this IEnumerable<T> source, IEnumerable<T> other) where T : class {
+            var set1 = new HashSet<T>(source);
+            var set2 = new HashSet<T>(other);
+            return set1.Count == set2.Count && set1.All(x => set2.Contains(x));
```
zooba: What's wrong with the SetEquals method of HashSet?

Also, I predict it will take about 5 minutes before we want to pass a comparer into this method (if only for case-insensitive string equality). Worth adding now (with a default to `null`)?

Python/Product/Analysis/Intellisense/AnalysisQueue.cs
zooba: We should definitely handle this where it comes up and Watsons (ideally before we ship). But explicitly checking `ODE` is important, whereas in many cases quietly stopping work on `OCE` is fine because there's no need to abandon the overall process.

Semantically, `ODE` is the right choice here. But we will need to do some work to make it safe. Watch the PTVS test results after merging to see if a lot of things start failing as a result.

Python/Product/Analysis/Intellisense/AnalysisQueue.cs
```diff
@@ -165,24 +167,20 @@ sealed class WaitAnalyzableTask : IAnalyzable {
...
-        }
+        public int AnalysisPending => _analysisPending;
 
         #region IDisposable Members
-
         public void Dispose() {
             Stop();
             _workEvent.Dispose();
             _cancel.Dispose();
+            _disposed = true;
```
zooba: Should we set the flag first so that items stop being queued while we are stopping?

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,237 @@
...
+            List<PublishDiagnosticsEventArgs> diags = null;
+
+            lock (_lock) {
+                var last = _parseBufferDiagnostics;
+
+                foreach (var kv in vc.GetAllParts(documentUri)) {
+                    var part = _server.ProjectFiles.GetPart(kv.Key);
+                    if (!last.TryGetValue(part, out var lastVersion) || lastVersion.Version < kv.Value.Version || _ignoreDiagnosticsVersion) {
+                        last[part] = kv.Value;
+                            diags = diags ?? new List<PublishDiagnosticsEventArgs>();
```
zooba: Is the extra indentation here actually in the code? Or just showing up on GH for some reason?

Python/Product/Analysis/LanguageServer/EditorFiles.cs
```diff
@@ -0,0 +1,237 @@
...
+
+                for (var i = 0; i < diags.Count; i++) {
+                    diags[i].severity = settings.analysis.GetEffectiveSeverity(diags[i].code, diags[i].severity);
+                }
+
+                var severity = settings.analysis.GetEffectiveSeverity(ErrorMessages.UnresolvedImportCode, DiagnosticSeverity.Warning);
+                var pythonProjectEntry = projectEntry as IPythonProjectEntry;
+                var parse = pythonProjectEntry?.GetCurrentParse();
+
+                if (parse != null && severity != DiagnosticSeverity.Unspecified) {
```
zooba: Can you add a `TODO` comment to move this scan into the normal analysis process?

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -69,22 +68,17 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
-        // Uri does not consider #fragment for equality
-        private readonly ProjectFiles _projectFiles = new ProjectFiles();
-        private readonly OpenFiles _openFiles;
-
-        internal PythonAnalyzer _analyzer;
-        private ClientCapabilities _clientCaps;
-        private LanguageServerSettings _settings = new LanguageServerSettings();
+        internal ClientCapabilities _clientCaps;
 
+        // Uri does not consider #fragment for equality
```
zooba: This comment doesn't make sense now that the comparer isn't visible here.

Python/Product/Analysis/Intellisense/AnalysisQueue.cs
zooba: Probably the best choice is to catch the `ODE` and throw an `OCE` with the `ODE` as the inner exception. That properly captures the cause of the cancellation in a way that simply throwing the cancellation exception does not.

Python/Product/Analysis/LanguageServer/EditorFiles.cs
MikhailArkhipov: Sure

Python/Product/Analysis/Infrastructure/EnumerableExtensions.cs
MikhailArkhipov: Done

PR #4454, opened 2018-06-27
Fixes #4434 No imports without Python installed

### LINKED ISSUE #3802: TestDiscovererTests failures
- DecoratedTests
- TestCaseRunTests
- TestCaseRunTestsWithTest
- TestCaseSubclasses

I tested all 4 scenarios manually and they work fine in the product, so it's a matter of fixing the tests so we stop thinking the product is broken.

### LINKED ISSUE #4434: No imports without Python installed
We include default databases with the install for when you don't have Python installed, but we aren't using them.

zooba:
Fixes #4434 No imports without Python installed
Fixes #3802 TestDiscovererTests failures
Replace NoInterpretersFactory with analysis-only factory


huguesv:
What are your results for `TestDiscovererTests`? Only `TestDiscover` passes for me. All others get 0 results. I'm not sure that `TestDiscover` even does any parsing/analysis of code to find tests, it looks it's just validating the parsing of the run settings.

zooba:
It needs the fix from #4452 as well, I think. I rebased onto a new branch for the PR.

zooba:
Or you need to force a full rebuild to regenerate `DefaultDB`. It doesn't automatically refresh like the regular cache does.

huguesv:
I got both PRs and did a full rebuild and now all TestDiscovererTests pass.

PR #4459, opened 2018-06-28
Fix analysis crash

zooba:
Prevents recursion occurring while resolving lazy values in multiple members container
Adds IPythonMultipleMembers2 interface to return immutable collection of members.

### DIFF
--- a/Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleMembers.cs
+++ b/Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleMembers.cs
@@ -21,15 +21,15 @@ using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
-    class AstPythonMultipleMembers : IPythonMultipleMembers2, ILocatedMember {
-        private readonly IReadOnlyList<IMember> _members;
-        private IReadOnlyList<IMember> _resolvedMembers;
+    class AstPythonMultipleMembers : IPythonMultipleMembers, ILocatedMember {
+        private readonly IMember[] _members;
+        private IList<IMember> _resolvedMembers;
 
         public AstPythonMultipleMembers() {
             _members = Array.Empty<IMember>();
         }
 
-        private AstPythonMultipleMembers(IReadOnlyList<IMember> members) {
+        private AstPythonMultipleMembers(IMember[] members) {
             _members = members;
         }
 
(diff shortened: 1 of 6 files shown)


Python/Product/Analysis/Interpreter/Ast/AstTypingModule.cs
```diff
@@ -33,10 +33,10 @@ class AstTypingModule : AstCachedPythonModule {
             if (type is IPythonType pyType) {
                 return pyType.DeclaringModule is AstTypingModule;
             }
-            if (type is IPythonMultipleMembers mm) {
-                return mm.Members.Any(IsTypingType);
-            }
-            return false;
+
+            return (type as IPythonMultipleMembers2)?.MembersNoCopy.Any(IsTypingType) ??
```
AlexanderSher: Let's use pattern matching instead of null checks here as well

Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleMembers.cs
```diff
@@ -21,17 +21,42 @@
...
             _members = Array.Empty<IMember>();
         }
 
         private AstPythonMultipleMembers(IReadOnlyList<IMember> members) {
             _members = members;
-            _resolvedMembers = new Lazy<IList<IMember>>(() => _members.Select(m => (m as ILazyMember)?.Get() ?? m).Where(m => m != this).ToArray());
+        }
+
+        private void EnsureMembers() {
+            if (_resolvedMembers != null) {
```
AlexanderSher: Can it be called from multiple threads?

Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleMembers.cs
```diff
@@ -21,17 +21,42 @@
 using Microsoft.PythonTools.Analysis.Infrastructure;
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
-    class AstPythonMultipleMembers : IPythonMultipleMembers, ILocatedMember {
+    class AstPythonMultipleMembers : IPythonMultipleMembers2, ILocatedMember {
```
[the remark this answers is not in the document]
zooba: `IReadOnlyList<T>` does not implement `IList<T>`, though we're only using that interface internally, so I guess we can change it to something more concrete.

Python/Product/Analysis/Interpreter/Ast/AstPythonMultipleMembers.cs
zooba: It shouldn't be - this is only used in analysis.

zooba: Actually, maybe through `Locations`, yeah.

PR #4465, opened 2018-06-28
Bundles Typeshed

zooba:
Bundles Typeshed
Improves options for customising or disabling Typeshed


zooba:
@MikhailArkhipov You'll need to update the VSC build to both include the entire Typeshed directory (so we get any modifications we need to make) and detect and provide the path in the initialization options. The analyzer deliberately doesn't pick it up automatically - it has to be specified as a type stub directory.

zooba:
Note that #4460 is required for changes to these options to be noticed without reopening the project.

huguesv:
Ideally I would see us both (PTVS+VSC) use a submodule that points to our own fork of typeshed.

So that if we want to make patches to it and not wait for typeshed to accept the changes, we can do it in one place and have both products pick it up. Also both products would have more control on when we take updates, and we'll reduce user reports of IntelliSense working for some things in PTVS but not VSC or vice-versa when it just amounts to different typeshed.

Is adding all the files to this repo just a temporary thing until we set up our fork?

huguesv:
To be clear, my main objection is to PTVS and VSC not bundling the same thing.

zooba:
I’m not even that keen to fork the repository, I’d rather just take a copy. We don’t need yet another repo floating around the Microsoft org for this.

I’m also opposed to submodules for the same reason. It’s in its own folder already, so it’s easy enough to copy changes back to a personal fork to send a PR if we need to. But I think we'll need to carry our own changes here, and since our model is richer than MyPy (we support simple code in these files) we may not merge back anyway.

MikhailArkhipov:
VSC takes the latest but later may list specific commit to sync to for the submodule since typeshed releases are spotty. @brettcannon and @DonJayamanne can comment more.

huguesv:
You won't have the opportunity for any customization, or be able to pick up any changes to the files copied in this repo, if you submodule from typeshed repo directly.

Do you have any way of embedding the files from the ptvs repo, without making another copy that needs to be kept in sync?

zooba:
That's the problem with submodules - you can't customize. I doubt we'll ever need to pin to an _older_ release of typeshed (unless they start adding incompatible features to MyPy, which... okay... that seems pretty likely at some point), but we _already_ want to replace parts of it with our own versions.

To do that we either need to put it in the language server repository (currently a directory in PTVS, but will be split out eventually) or in our own fork and then pull it into the language server repository. My vote is obviously to put it into the language server.

It should build and be part of the language server just as our other pregenerated files are (which we'll be keeping because they include docs, and typeshed does not).

PR #4467, opened 2018-06-29
Fix #4350: VSC can send `textDocument/didChange` incrementing version by any number.

### LINKED ISSUE #4350: VSC can send `textDocument/didChange` incrementing version by any number.
- Create new module, type some text
- Quickly press `Backspace` twice to erase two symblos, then press `Ctrl+Z` (undo). Both symbols should be restored at once
- Attach debugger to `dotnet` process, put breakpoint at the first line of `Server.DidChangeTextDocument`, add `@params.textDocument.version`  to watch
- Press `Ctrl+Y` (redo), then `Ctrl+Z` (undo), observe that VS Code sends JSON with with version incremented by 2:

```cs
@params.textDocument.version: 22
@params.textDocument.version: 24
@params.textDocument.version: 26
```

Increment can be by any number, depending on the amount of changes VS Code merges together during undo/redo operations. `OpenFile.DidChangeTextDocument` can't handle changes like that and keeps putting incoming changes into `_pendingChanges`.

```
if (fromVersion > docVersion && @params.contentChanges?.Any(c => c.range == null) != true) {
    // Expected from version hasn't been seen yet, and there are no resets in this
    // change, so enqueue it for later.
    _log.TraceMessage($"Deferring changes for {uri} until version {fromVersion} is seen");
    lock (_pendingChanges) {
        _pendingChanges.Add(@params);
    }
    return;
}
```

AlexanderSher:
Requires #4453 to finish


zooba:
Looks good so far. Make sure to ping us when you remove the WIP marker.

MikhailArkhipov:
@AlexanderSher said it can be merged

PR #4472, opened 2018-07-02
Fixes #2943 XAML has no Intellisense

### LINKED ISSUE #2943: XAML has no Intellisense

zooba:
Fixes #2943 XAML has no Intellisense
Fixes Resolved property
Reduce impact of .NET build targets with properties instead of excluding the entire target
Adds mscorlib reference automatically if missing
Adds upgrade test
Improves warning for unhandled types in analysis


Python/Templates/IronPython/ProjectTemplates/Python/WpfProject/WpfApp.pyproj
```diff
@@ -29,6 +29,10 @@
     </Content>
   </ItemGroup>
   <ItemGroup>
+    <Reference Include="mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
```
huguesv: b77a5c561934e089

Python/Product/Analysis/Interpreter/Ast/AstPythonModule.cs
```diff
@@ -130,6 +135,8 @@ sealed class AstPythonModule : IPythonModule, IProjectEntry, ILocatedMember {
             var walker = new AstAnalysisWalker(interpreter, ast, this, filePath, DocumentUri, _members, true, true);
             ast.Walk(walker);
             walker.Complete();
+
+            ParseErrors = parseErrors.ToArray();
```
huguesv: `parseErrors` can be null, causing ANE on this line

Repro with test `UnassignedClassMembers` and `OsPathMembers`

Python/Product/Analysis/Parsing/Parser.cs
```diff
@@ -4197,8 +4211,25 @@ struct Name {
 
         private bool PeekTokenForOrAsyncFor {
             get {
-                return PeekToken(Tokens.KeywordForToken) ||
-                    (AllowAsyncAwaitSyntax && PeekToken(Tokens.KeywordAsyncToken) && PeekToken2()?.Kind == TokenKind.KeywordFor);
+                if (PeekToken(Tokens.KeywordForToken)) {
+                    return true;
+                }
+                if (AllowAsyncAwaitSyntax || (_langVersion >= PythonLanguageVersion.V37 && _inGeneratorExpression)) {
```
MikhailArkhipov: Should `_langVersion >= PythonLanguageVersion.V37` be simply part of the `AllowAsyncAwaitSyntax` property?

Python/Product/Analysis/Parsing/Parser.cs
```diff
@@ -4197,8 +4211,25 @@ struct Name {
 
         private bool PeekTokenForOrAsyncFor {
```
MikhailArkhipov: Alternative is change it to `PeekTokenForOrAsyncFor (bool inGeneratorExpression)`. Up to you.

zooba: I started going that way and didn't like itp

Python/Product/Analysis/Parsing/Parser.cs
zooba: No, because it's still not allowed everywhere, but only in generator expressions.

PR #4483, opened 2018-07-05
Add analysis progress reporting to LS

MikhailArkhipov:
PVSC: https://github.com/Microsoft/vscode-python/pull/2099

- Move analysis and tracing options to PythonInitialization so they can be easily used as part of LanguageClientOptions in VSC.
- Add AnalysisQueued event
- Add progress reporting service to LS
- Add task to monitor analysis events


Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -345,9 +356,10 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
 
             ThrowIfDisposed();
             _clientCaps = @params.capabilities;
-            _traceLogging = _clientCaps?.python?.traceLogging ?? false;
-            Analyzer.EnableDiagnostics = _clientCaps?.python?.liveLinting ?? false;
+            _traceLogging = @params.initializationOptions.traceLogging;
+            _analysisUpdates = @params.initializationOptions.analysisUpdates;
 
+            Analyzer.EnableDiagnostics = _clientCaps?.python?.liveLinting ?? false;
```
zooba: This property should go away in favour of suppressing the individual messages (probably by default, at least for the ones currently hidden by this flag)

Python/Product/VSCode/AnalysisVsc/Program.cs
```diff
@@ -14,7 +14,7 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-// #define WAIT_FOR_DEBUGGER
+#define WAIT_FOR_DEBUGGER
```
zooba: Guessing you don't mean to check this in?

Python/Product/Analysis/LanguageServer/Server.cs
MikhailArkhipov: Yep. We should discuss messages, error codes, etc.

PR #4485, opened 2018-07-05
Change legacy/experimental debugger messages

### LINKED ISSUE #4479: Update experimental debugger warnings now that it is the only debugger
For this PR: https://github.com/Microsoft/PTVS/pull/4365

The debugger is now on by default, we should now talk about the previous version as the legacy debugger.

Title: Debugger does not support this Python environment
Text: To debug this Python environment use the legacy debugger instead. It is likely using an older version of Python, see the documentation for a list of supported Python versions and environment types.

Enable the legacy debugger
 - Future debugging sessions will use the legacy debugger. You can change this setting back in Tools > Options > Python > Experimental page.

Learn more
- View documentation on supported Python environments and how to upgrade or uninstall the ptvsd debugger package.

Additional note:
- Can we move the setting to Tools > Options > Python > Debugging instead?

huguesv:
Fix #4479 

Use legacy instead of experimental when mentioning the ptvsd debuggers.
Move the option from experimental page to debugger page.
Add Python 3.3 as an incompatible environment.

I kept the storage of the option in VSInterpreters, we still need to access it from the Debugger project which doesn't have access to DebuggerOptions.

### DIFF
--- a/Python/Product/Common/Strings.resx
+++ b/Python/Product/Common/Strings.resx
@@ -2257,10 +2257,10 @@ Expand selection to the full expression?</value>
     <value>Debugger package could not be loaded</value>
   </data>
   <data name="PtvsdDisableCaption" xml:space="preserve">
-    <value>Enable the legacy debugger</value>
+    <value>Use the legacy debugger</value>
   </data>
   <data name="PtvsdDisableSubtext" xml:space="preserve">
-    <value>Future debugging sessions will use the legacy debugger. You can change this setting back in Tools &gt; Options &gt; Python &gt; Debugging page.</value>
+    <value>Future debugging sessions will use the legacy debugger. You can also change this setting in Tools, Options, Python, Debugging page by changing the 'Use legacy debugger' option.</value>
   </data>
   <data name="PtvsdLearnMoreCaption" xml:space="preserve">
     <value>Learn more</value>
@@ -2272,10 +2272,7 @@ Expand selection to the full expression?</value>
     <value>Debugger does not support this Python environment</value>
   </data>
   <data name="PtvsdIncompatibleEnvMessage" xml:space="preserve">
-    <value>To debug this Python environment use the legacy debugger instead. It is likely using an older version of Python, see the documentation for a list of supported Python versions and environment types.</value>
-  </data>
-  <data name="DisableExperimentalDebugger" xml:space="preserve">
-    <value>Enable the legacy debugger</value>
+    <value>To debug this Python environment, please use the legacy debugger. Older versions of Python are not supported. See our &lt;a href="https://aka.ms/upgradeptvsd"&gt;documentation&lt;/a&gt; for a list of supported Python versions and environment types.</value>
   </data>
   <data name="DebugReplFeatureNotSupportedWithExperimentalDebugger" xml:space="preserve">
     <value>This debug REPL feature is only supported with the legacy debugger.</value>
--- a/Python/Product/Common/Strings.resx
+++ b/Python/Product/Common/Strings.resx
@@ -2257,28 +2257,28 @@ Expand selection to the full expression?</value>
     <value>Debugger package could not be loaded</value>
   </data>
   <data name="PtvsdDisableCaption" xml:space="preserve">
-    <value>Disable the experimental debugger</value>
+    <value>Enable the legacy debugger</value>
   </data>
   <data name="PtvsdDisableSubtext" xml:space="preserve">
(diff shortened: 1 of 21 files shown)


Python/Product/Common/Strings.resx
```diff
@@ -2257,28 +2257,28 @@ Expand selection to the full expression?</value>
     <value>Debugger package could not be loaded</value>
   </data>
   <data name="PtvsdDisableCaption" xml:space="preserve">
-    <value>Disable the experimental debugger</value>
+    <value>Enable the legacy debugger</value>
   </data>
   <data name="PtvsdDisableSubtext" xml:space="preserve">
-    <value>Future debugging sessions will not use the experimental debugger. You can also change this setting in the Tools/Options/Python/Experimental page.</value>
+    <value>Future debugging sessions will use the legacy debugger. You can change this setting back in Tools &gt; Options &gt; Python &gt; Debugging page.</value>
```
zooba: I normally use commas here and then break the list where the UI changes. e.g. "in Tools, Options, Python, Debugging and ensure that 'Use the legacy debugger' is checked" (and I carefully worded the "ensure ... is checked" part to avoid using control names)

Python/Product/Common/Strings.resx
```diff
@@ -2257,28 +2257,28 @@ Expand selection to the full expression?</value>
...
-    <value>View documentation in your web browser on supported Python environments and how to upgrade or uninstall the ptvsd debugger package.</value>
+    <value>View documentation on supported Python environments and how to upgrade or uninstall the ptvsd debugger package.</value>
   </data>
   <data name="PtvsdIncompatibleEnvTitle" xml:space="preserve">
-    <value>Debugger is incompatible with this environment</value>
+    <value>Debugger does not support this Python environment</value>
   </data>
   <data name="PtvsdIncompatibleEnvMessage" xml:space="preserve">
-    <value>The experimental debugger is incompatible with this Python environment. See the documentation for a list of supported Python environments.</value>
+    <value>To debug this Python environment use the legacy debugger instead. It is likely using an older version of Python, see the documentation for a list of supported Python versions and environment types.</value>
```
zooba: "To debug this Python environment, please use the legacy debugger. Older versions of Python are not supported. See our documentation ..."

(Also, if this is going as the content in a task dialog, we can use a hyperlink in the message - `<a href="...">the documentation</a>`

Python/Product/Common/Strings.resx
```diff
@@ -2257,28 +2257,28 @@ Expand selection to the full expression?</value>
...
   </data>
   <data name="DisableExperimentalDebugger" xml:space="preserve">
-    <value>Disable the experimental debugger</value>
+    <value>Enable the legacy debugger</value>
```
zooba: "Use" makes more sense than "Enable" alongside a checkbox.

zooba: Oh wait, this is going in the task dialog too, right?

Python/Product/Common/Strings.resx
```diff
@@ -2257,28 +2257,28 @@ Expand selection to the full expression?</value>
...
   </data>
   <data name="DebugReplFeatureNotSupportedWithExperimentalDebugger" xml:space="preserve">
-    <value>This debug REPL feature is not supported with the experimental debugger.</value>
+    <value>This debug REPL feature is only supported with the legacy debugger.</value>
```
zooba: We have a bug to fix this, right? @karthiknadig?

Python/Product/PythonTools/PythonTools/Options/IPythonOptions.cs
```diff
@@ -55,6 +55,11 @@ public interface IPythonOptions {
             get;
             set;
         }
+
+        bool UseLegacyDebugger {
+            get;
+            set;
+        }
```
zooba: This is a COM interface, so we need to add a new interface to expose the new option.

Python/Product/Common/Strings.resx
huguesv: Right, it's a task dialog button, not a check box. Also it's not immediate, the current debugging session that was just started is not affected.

huguesv: Oh, there's two of those identical strings, and this one is not used, will delete.

Python/Product/Common/Strings.resx
huguesv: I created https://github.com/Microsoft/PTVS/issues/4487

PR #4492, opened 2018-07-06
Fixes #4033 Remote attach with source reference crashes VS

### LINKED ISSUE #4033: Remote attach with source reference crashes VS
image

Repro steps:
on a remote machine:
1. Create a py file with following content `one.py`:
```py
import time


for x in range(10000):
    time.sleep(1)
    print(x)
```
2. start ptvsd for remote debugging:
```
python3 -m ptvsd --server --port 8765 --file one.py
```

On a machine with VS:
1. Go to Debug > Attach to process ...
2. Select Connection type : python remote, Connection target: tcp://<remote name>:8765
3. Hit pause after attaching

zooba:
Fixes #4033 Remote attach with source reference crashes VS
Handles non-absolute paths by generating a false URI

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
+++ b/Python/Product/PythonTools/PythonTools/Editor/PythonTextBufferInfo.cs
@@ -191,11 +191,18 @@ namespace Microsoft.PythonTools.Editor {
             var path = Filename;
             if (!string.IsNullOrEmpty(path)) {
                 try {
-                    return new Uri(path);
+                    if (Path.IsPathRooted(path)) {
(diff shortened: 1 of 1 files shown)


zooba:
I wasn't able to test this fully because for some reason I can't remote debug right now (and local-remote debugging doesn't show it, because the file is available). But all the pieces work individually :)

zooba:
(We use opaque DocumentUri throughout our tests with the `python://` scheme - it means we can find/use the analysis for it, but it won't be transformed back into a local path.)

karthiknadig:
I have verified the fix, it works.

PR #4496, opened 2018-07-06
Resolve deadlock on recursive imports

MikhailArkhipov:
When importing `celery` we time out because there is code that re-imports the module. Since importing is async, the second import comes back on a different thread and sentinel does not detect it.

Added check for the interpreter instance.


Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -498,6 +498,7 @@ public sealed class TryImportModuleContext {
             CancellationToken cancellationToken
         ) {
             IPythonModule module = null;
+            bool isCelery = name == "celery";
```
AlexanderSher: Is it case sensitive?

Python/Product/Analysis/Interpreter/SentinelModule.cs
```diff
@@ -36,12 +39,12 @@ sealed class SentinelModule : IPythonModule {
...
         }
 
-        public async Task<IPythonModule> WaitForImportAsync(CancellationToken cancellationToken) {
+        public async Task<IPythonModule> WaitForImportAsync(IPythonInterpreter interpreter, CancellationToken cancellationToken) {
             var mod = _realModule;
             if (mod != null) {
                 return mod;
             }
-            if (_thread == Thread.CurrentThread) {
+            if (_thread == Thread.CurrentThread || _interpreter == interpreter) {
```
AlexanderSher: I don't understand how it works:
- Thread checking doesn't work in async environment, it requires `AsyncLocal`.
- `IPythonInterpreter` instances are shared between requests, how do they help against reentrancy?

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
zooba: It is, but this is also unused and should be deleted :)

Python/Product/Analysis/Interpreter/SentinelModule.cs
zooba: Yeah, now I think about it, I'm not sure `IPythonInterpreter` is the right check here. I'd need to think through what is, it certainly isn't obvious. (The aim is to ensure parallel tasks will wait for some time for an import to complete, but not if it is recursing and will never make progress. `AsyncLocal` is probably correct, though I haven't convinced myself of that yet.)

MikhailArkhipov: I left the current thread check for legacy reasons. If it is safe to remove, I can nix it. The intent here is to determine that code is reentering on import (i.e. circular import). So sentinel remember which interpreter it relates to.

AlexanderSher: `_thread == Thread.CurrentThread` may become false positive, i.e. sentinel was created on thread A, then some importing code was started on thread B but async was continued on thread A.

In `AsyncLocal` we can store the modules that requested this import. But I don't fully understand how the sentinel works in case of reentrancy. If `WaitForImportAsync` has returned the sentinel itself, what happens next?

MikhailArkhipov: Yeah, I know about the thread issue, this is what we are trying to solve. `WaitForImportAsync` is called if some code requests the import and module table contains sentinel (i.e. module is being imported).

zooba: `SynchronizationContext.Current` might be the right comparison. We should always reach this point on the "main" analysis thread, since imports should be synchronous if they are already cached, and we're aren't doing locking here as much as we are detecting same-thread vs. other-thread.

That said, I can see it being a maintenance issue in the future. What if we store the current synchronization context on this object, but check it from the caller? That way "everyone" who comes into this function will get a valid result (or wait), and we can handle the `ImportModule` recursion in `ImportModule` (or whichever helper function we're calling this from).

AlexanderSher: `_interpreter.ImportModule` which calls `TryImportModuleAsync` on sync code path is called from `Task.Run(() => _interpreter.ImportModule(name))` - there is no sync context in it.

zooba: Hmm... you're right. But that also means that AsyncLocal wouldn't work either, right?

AlexanderSher: No, `AsyncLocal` has nothing to do with `SynchronizationContext`, it works through `ExecutionContext`.

MikhailArkhipov: `AsyncLocal` does not work. Actually, the interpreter instance appears to be a good choice. Import context (`TryImportModuleContext`) consists of interpreter and some of its members (such as module table). So the importing task is started in `ImportModule` for a given name and context which is, in fact, the interpreter. 

Similar effect can be achieved by tracking which import tasks already started in the `ImportModule` and then avoiding starting new ones for the same name/context pair. But it looks like same thing as checking the interpreter instance since context is built around it.

So, if we are importing into module table that is per interpreter instance, then checking said instance should be OK. No?

AlexanderSher:
> AsyncLocal does not work.

That's weird. Here is all the changes I've made:

```cs
 public SentinelModule(string name, bool importing) {
    Name = name;
    if (importing) {
        _asyncLocal = new AsyncLocal<SentinelModule> {
            Value = this
        };
        _semaphore = new SemaphoreSlim(0, 1000);
    } else {
        _realModule = this;
    }
} 
```  

```cs
public async Task<IPythonModule> WaitForImportAsync(CancellationToken cancellationToken) {
    var mod = _realModule;
    if (mod != null) {
        return mod;
    }

    if (_asyncLocal?.Value == this) {
        return this;
    }
```

Result:
image
 
Are there any different repro steps?

MikhailArkhipov:
It didn't work on my example, specifically on importing 'os'. With Python 3

`celery.py`:
```python
from __future__ import absolute_import # for python 2

import celery
celery.Celery('tasks')
```

project:
```
__init__.py
celery.py
```

MikhailArkhipov:
OK, Alex version works.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -195,7 +195,7 @@ class AstPythonInterpreter : IPythonInterpreter, IModuleContext, ICanFindModuleM
             return names.MaybeEnumerate().ToArray();
         }
 
-        public IPythonModule ImportModule(string name) {
+        public async Task<IPythonModule> ImportModuleAsync(string name) {
```
AlexanderSher: Let's add `CancellationToken` on this level as well and pass it to the `TryImportModuleAsync`

AlexanderSher: Nevermind, we use `CancellationTokenSource` below, mixing it with another `CancellationToken` requires too much code changes.

Python/Product/Analysis/Interpreter/AnalysisOnlyInterpreterFactory.cs
```diff
@@ -65,7 +66,7 @@ internal AnalysisOnlyInterpreterFactory(Dictionary<string, object> properties)
         }
 
         public void NotifyImportNamesChanged() {
-            
+
```
AlexanderSher: ?

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -261,6 +256,13 @@ class AstPythonInterpreter : IPythonInterpreter, IModuleContext, ICanFindModuleM
             return null;
         }
 
+        public IPythonModule ImportModule(string name) {
+            Task<IPythonModule> impTask = ImportModuleAsync(name);
+            if (!impTask.Wait(10000)) {
+                return null;
```
AlexanderSher: Do we need to call `Debug.Fail` here?

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
AlexanderSher: On the other hand, it is public API, so we may want to do it right.

Python/Product/Analysis/ModuleTable.cs
```diff
@@ -93,12 +93,12 @@ class ModuleTable : IEnumerable<KeyValuePair<string, ModuleLoadState>> {
             ModuleReference res;
             bool firstImport = false;
             if (!_modules.TryGetValue(name, out res) || res == null) {
-                var mod = await Task.Run(() => _interpreter.ImportModule(name)).ConfigureAwait(false);
+                var mod = await _interpreter.ImportModuleAsync(name).ConfigureAwait(false);
```
AlexanderSher: 👍

Python/Product/Analysis/Interpreter/AnalysisOnlyInterpreterFactory.cs
MikhailArkhipov: C# automatic formatting on save

AlexanderSher:
> Revert "Use AsyncLocal" 

So, did it work or not?

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
MikhailArkhipov: `ImportModuleAsync` is doing it

AlexanderSher: Ok

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
MikhailArkhipov: Added, change propagated all over...

MikhailArkhipov:
It did (in the specific example I tried) but so did sync context. Since we are really just checking that same instance of the analyzer is not reentering and the analyzer has associated sync context the check appears sufficient.

AlexanderSher:
> So I think capturing and then checking the synchronization context later should be fine for detecting recursive imports. If the context is the same, we return nothing immediately. If the import is currently happening on a different context then we wait and let it keep running (i.e. on the other analyzer, which is not a common case for us with the out-of-proc analyzer, though this can be disabled so this becomes relevant again).

`SynchronizationContext` won't work because of the several reasons, including:
- We have only one instance of `` per `AnalysisQueue`, so it is shared between all requests that go through the queue:
```cs
private void Worker(object threadStarted) {
    try {
        SynchronizationContext.SetSynchronizationContext(new AnalysisSynchronizationContext(this));
    } finally {
        ((AutoResetEvent)threadStarted).Set();
    }
```
- It is wiped out by any `Task.Run` or may be wiped out by `ConfigureAwait(false)`, so `WaitForImportAsync` may be called without `AnalysisSynchronizationContext` instance.
- There are code paths that call `ImportModule` have no `AnalysisQueue` in them, so they will have no sync context as well.

> AsyncLocal doesn't work because it doesn't actually run in a true async context - only one call would go through the Task.Run path, and that's on initialization. All the rest run in the context of a single analyzer.

Not exactly. `ProjectEntry`, which triggers `GroupAnalysis.Analyze`, is added to the `AnalysisQueue` in `OnDocumentChangeProcessingComplete`, which is called as continuation to the `TaskCompletionSource<IAnalysisCookie>.Task`:
```cs
                cookieTask.ContinueWith(t => {
                    if (t.IsFaulted) {
                        // Happens when file got deleted before processing
                        p.Dispose();
                        LogMessage(MessageType.Error, t.Exception.Message);
                        return;
                    }
                    OnDocumentChangeProcessingComplete(doc, t.Result as VersionCookie, enqueueForAnalysis, priority, p);
                }).DoNotWait();
```

But since the call goes through `AnalysisSynchronizationContext`, it is shared between all items in `AnalysisQueue`.

AlexanderSher:
But if both `AsyncLocal` and `SynchronizationContext` checks have fixed the symptoms, I wonder if that is a proper fix at all.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
```diff
@@ -261,6 +248,14 @@ class AstPythonInterpreter : IPythonInterpreter, IModuleContext, ICanFindModuleM
             return null;
         }
 
+        public IPythonModule ImportModule(string name) {
+            var token = Debugger.IsAttached ? CancellationToken.None : new CancellationTokenSource(5000).Token;
+            var impTask = ImportModuleAsync(name, token);
+            if (!impTask.Wait(10000)) {
```
zooba: I'd rather not time out here while debugging either, and ideally we should only have the IsAttached check in debug builds.

Python/Product/Analysis/Interpreter/IPythonInterpreter.cs
```diff
@@ -59,11 +61,17 @@ public interface IPythonInterpreter : IDisposable {
...
+        /// Synchronous variant of <see ref="ImportModuleAsync">.
+        /// Waits for the async import completion using timeouts.
         /// </summary>
         IPythonModule ImportModule(string name);
 
+         /// <summary>
+        /// Returns an IPythonModule for a given module name. Returns null if
+        /// the module does not exist. The import is performed asynchronously.
+        /// </summary>
+        Task<IPythonModule> ImportModuleAsync(string name, CancellationToken token);
```
zooba: This needs to go onto a separate interface.

Python/Product/Analysis/PythonAnalyzer.cs
```diff
@@ -71,11 +71,12 @@ public partial class PythonAnalyzer : IGroupableAnalysisProject, IDisposable {
         /// </summary>
         public static async Task<PythonAnalyzer> CreateAsync(
             IPythonInterpreterFactory factory,
+            CancellationToken token,
             IPythonInterpreter interpreter = null
```
zooba: Public API here - put the new parameter last with a sensible default.

Python/Product/Analysis/PythonAnalyzer.cs
```diff
@@ -166,7 +167,7 @@ public partial class PythonAnalyzer : IGroupableAnalysisProject, IDisposable {
         /// This method should be called on the analysis thread and is usually invoked
         /// when the interpreter signals that it's modules have changed.
         /// </summary>
-        public async Task ReloadModulesAsync() {
+        public async Task ReloadModulesAsync(CancellationToken token) {
```
zooba: Public API - need a new overload.

Python/Tests/Analysis/AnalysisTest.Perf.cs
```diff
@@ -224,7 +224,7 @@ import System
 
             var fact = InterpreterFactoryCreator.CreateAnalysisInterpreterFactory(version.ToVersion());
             var projectState = new PythonAnalyzer(fact, fact.CreateInterpreter());
-            projectState.ReloadModulesAsync().WaitAndUnwrapExceptions();
+            projectState.ReloadModulesAsync(cancel.HasValue ? cancel.Value : CancellationToken.None).WaitAndUnwrapExceptions();
```
zooba: `cancel ?? CancellationToken.None`

zooba:
@AlexanderSher The async local changes were never necessary. What we're trying to do here is determine whether to wait before assuming that the import will never complete. When running in the same analyzer, the answer is always "no, it will never complete" because the imports are single threaded. 

This only becomes relevant when we have multiple analyzers in the same process (tests, some VS scenarios, not ever VSC), when two analyzers are sharing the same module cache in the interpreter. In that case, if one analyzer tries to import a module that the other is currently importing, it can wait, because the other will continue to make progress. If it doesn't appear to make progress (i.e. times out), then we assume the module cannot be imported and return nothing (the same scenario as when a single analyzer tries to recursively import a module, except that time we don't use the timeout).

So this is nothing to do with async or threads, and everything to do with "are we recursively importing this module". Another fix besides using the analyzer sync context would be to use some other recursion protection attached to the interpreter instance. So when the sentinel object is found in the table, we can then check whether it's because we're partway through importing it ourselves or because another analyzer is doing it.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreter.cs
MikhailArkhipov: Sometimes I had to debug release builds but sure :-)

Python/Product/Analysis/Interpreter/IPythonInterpreter.cs
MikhailArkhipov: `IPythonInterpreter2` ?

AlexanderSher:
@zooba Yes, whole `GroupAnalysis.Analyze` is single-threaded and thread-bound, and no `async` call (including `WaitForImportAsync`) is awaited in runtime, otherwise it would've hanged because of the `AnalysisSynchronizationContext`. And for this specific code path, original thread check would've been enough.

But we have a bug that thread-checking has failed, so there should be other code paths. For example, `ModuleTable.TryImport` is called from `PythonAnalyzer.AddModuleAlias` 🡐 `Server.AddFileAsync` 🡐 `Server.DidOpenTextDocument` or from `ModuleAnalysis.ResolveModule` 🡐 `ModuleAnalysis.GetMembers` 🡐 `CompletionAnalysis.GetCompletionsFromMembers`. These call stacks don't include `AnalysisQueue` and can be executed from any thread. And in these cases, `AsyncLocal` will work as expected, separating one request from another.

What worries me is the real cause of the bug. IMO, if `WaitForImportAsync` was hanged because of the thread check failure, it means that import was started outside of the `GroupAnalysis.Analyze` call stack and in this case it will always hang if `_semaphore.WaitAsync` is awaited.

zooba:
@AlexanderSher I would argue that all of those are bugs, and we need to not do that. The rule is "nothing outside of the analyzer thread should mutate any state within the analyzer thread", and importing modules breaks that (this is why `ModuleTable.TryGetImportedModule` exists).

Anywhere we have an `AnalysisUnit`, its `ForEval` member should be `true` if we don't want to mutate state. It may not always be checked though...

Python/Tests/Utilities.Python.Analysis/MockPythonInterpreter.cs
```diff
@@ -23,7 +23,7 @@
 using Microsoft.PythonTools.Interpreter;
 
 namespace TestUtilities.Python {
-    public class MockPythonInterpreter : IPythonInterpreter {
+    public class MockPythonInterpreter : IPythonInterpreter2 {
```
zooba: I expected `AstPythonInterpreter` to implement this too, yes?

Python/Product/Analysis/Values/BuiltinNamespace.cs
```diff
@@ -125,7 +125,7 @@ class BuiltinNamespace<MemberContainerType> : AnalysisValue where MemberContaine
 
         public override bool Equals(object obj) {
             if (obj is BuiltinNamespace<MemberContainerType> bn && GetType() == bn.GetType()) {
-                return _type.Equals(bn._type);
+                return _type != null &&_type.Equals(bn._type);
```
zooba: If `_type` is null and so is `bn._type`, we need to return `true`

Python/Tests/Utilities.Python.Analysis/MockPythonInterpreter.cs
MikhailArkhipov: Yes, I guess we are not calling it from outside just yet.

Python/Product/Analysis/LanguageServer/Server.cs
```diff
@@ -126,7 +126,7 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
         #region Client message handling
         public override async Task<InitializeResult> Initialize(InitializeParams @params) {
             ThrowIfDisposed();
-            await DoInitializeAsync(@params);
+            await DoInitializeAsync(@params, CancellationToken);
```
zooba: `CancellationToken` only has a value when it's used within a `using (AllowRequestCancellation)` block.

Python/Product/Analyzer/Interpreter/LegacyDB/CPythonInterpreter.cs
```diff
@@ -98,6 +97,8 @@ class CPythonInterpreter : IPythonInterpreter {
             return fromDb;
         }
 
+        public Task<IPythonModule> ImportModuleAsync(string name, CancellationToken token) => Task.FromResult(ImportModule(name));
```
zooba: Unused (since the interface is not added here, and I'm quite happy to not have the interface on this one)

Python/Product/IronPython/Interpreter/IronPythonInterpreter.cs
```diff
@@ -405,6 +405,9 @@ class AssemblyResolver {
             return _modules[name];
         }
 
+        public Task<IPythonModule> ImportModuleAsync(string name, CancellationToken token)
+            => Task.FromResult(ImportModule(name));
```
zooba: Also unused

Python/Product/Uwp.Interpreter/PythonUwpInterpreter.cs
```diff
@@ -78,6 +78,9 @@ class PythonUwpInterpreter : IPythonInterpreter, IPythonInterpreterWithProjectRe
             return new List<string>(_typeDb.GetModuleNames());
         }
 
+        public Task<IPythonModule> ImportModuleAsync(string name, CancellationToken token)
+            => Task.FromResult(ImportModule(name));
```
zooba: Also unused

Python/Product/VSInterpreters/Interpreter/NotFoundInterpreterFactory.cs
```diff
@@ -26,6 +28,8 @@ sealed class NotFoundInterpreter : IPythonInterpreter {
         public IPythonType GetBuiltinType(BuiltinTypeId id) { throw new KeyNotFoundException(); }
         public IList<string> GetModuleNames() { return new string[0]; }
         public event EventHandler ModuleNamesChanged { add { } remove { } }
+        public Task<IPythonModule> ImportModuleAsync(string name, CancellationToken token)
+            => Task.FromResult(ImportModule(name));
```
zooba: Also unused

PR #4514, opened 2018-07-10
Fixes #4509 Nested imports not resolving correctly

### LINKED ISSUE #4509: AstTypeStubPaths_* test failures
```
AstTypeStubPaths_MergeStubsPath
AstTypeStubPaths_NoStubs
AstTypeStubPaths_MergeStubsPath
AstTypeStubPaths_ExclusiveStubs
```

They all fail like this:
```
System.AggregateException: One or more errors occurred. ---> Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Assert.Fail failed. Failed to get module we just imported
```

zooba:
Fixes #4509 Nested imports not resolving correctly
Corrects assigning submodules to analysis objects
Fixes references to unresolved module names
Fixes comparison of builtin functions

### DIFF
--- a/Python/Product/Analysis/Analyzer/DDG.cs
+++ b/Python/Product/Analysis/Analyzer/DDG.cs
@@ -430,30 +430,28 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
 
             var candidates = ModuleResolver.ResolvePotentialModuleNames(_unit.ProjectEntry, modName, forceAbsolute).ToArray();
             foreach (var name in candidates) {
-                if (ProjectState.Modules.TryImport(name, out moduleRef)) {
-                    _unit.DeclaringModule.AddModuleReference(moduleRef);
-                    return true;
-                }
-            }
+                ModuleReference modRef;
 
-            foreach (var name in candidates) {
-                moduleRef = null;
-                foreach (var part in ModulePath.GetParents(name, includeFullName: true)) {
-                    if (ProjectState.Modules.TryImport(part, out var mref)) {
-                        moduleRef = mref;
-                        if (part.Length < name.Length) {
-                            moduleRef.Module?.Imported(_unit);
-                        }
-                    } else if (moduleRef != null) {
-                        Debug.Assert(moduleRef.Name.Length + 1 < name.Length, $"Expected {name} to be a child of {moduleRef.Name}");
-                        if (moduleRef.Name.Length + 1 < name.Length) {
-                            remainingParts = name.Substring(moduleRef.Name.Length + 1).Split('.');
-                        }
-                        _unit.DeclaringModule.AddModuleReference(moduleRef);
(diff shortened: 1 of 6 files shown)


Python/Product/Analysis/Analyzer/DDG.cs
```diff
@@ -430,30 +430,28 @@ internal class DDG : PythonWalker {
...
+                bool gotAllParents = true;
+                AnalysisValue lastParent = null;
+                remainingParts = name.Split('.');
+                foreach (var part in ModulePath.GetParents(name, includeFullName: false)) {
+                    if (!ProjectState.Modules.TryImport(part, out modRef)) {
+                        gotAllParents = false;
                         break;
                     }
+                    moduleRef = modRef;
+                    (lastParent as BuiltinModule)?.AddChildModule(remainingParts[0], moduleRef.AnalysisModule);
```
zooba: This (and the call below) are the critical part of the fix. We may have been resolving `import package.module` into two separate modules (via type stubs) where `package` didn't know about its `.module` member. So now we explicitly associate it when importing.

zooba:
@huguesv Fixes a regression in 15.8 - I'd like to backport it.

Python/Product/Analysis/Analyzer/DDG.cs
```diff
@@ -430,30 +430,28 @@ internal class DDG : PythonWalker {
...
-                            remainingParts = name.Substring(moduleRef.Name.Length + 1).Split('.');
-                        }
-                        _unit.DeclaringModule.AddModuleReference(moduleRef);
-                        return true;
-                    } else {
+                ModuleReference modRef;
+
+                bool gotAllParents = true;
+                AnalysisValue lastParent = null;
+                remainingParts = name.Split('.');
```
MikhailArkhipov: What happens if there no dots?

zooba: We skip the `foreach` completely and just import it directly.

Python/Product/Analysis/Values/BuiltinModule.cs
```diff
@@ -33,9 +36,23 @@ public BuiltinModule(IPythonModule module, PythonAnalyzer projectState)
 
         public IPythonModule InterpreterModule => _interpreterModule;
 
+        public void AddChildModule(string memberName, AnalysisValue module) {
+            if (_childModules == null) {
+                _childModules = new Dictionary<string, IAnalysisSet>();
+            }
+            if (_childModules.TryGetValue(memberName, out var existing)) {
```
MikhailArkhipov: Sequence is getting popular... Might we worth writing extension to dictionary like `Add<T>(T value, Func<T, T> newValueConstructor);` where the new value constructor is called to make the actual value when existing is there.

PR #4519, opened 2018-07-10
Rename Language Server assemblies

MikhailArkhipov:
Includes https://github.com/Microsoft/PTVS/pull/4496, will updated when it is merged. Changes namespace but only in the LS.


AlexanderSher:
Do we need to do anything specific for the client-side update?

zooba:
I think I looked at the relevant commits, but I'll review properly once the other changes are not in this PR any more - they're far too distracting.

MikhailArkhipov:
Client side went in here https://github.com/Microsoft/vscode-python/pull/2122

PR #4520, opened 2018-07-10
Fixes #1865 Unit tests refresh too often

### LINKED ISSUE #1865: Unit tests refresh too often
When editing tests, the test window refreshes frequently, and the output window is often cleared and new text added (typically warnings about Nuget, for some reason).

This is annoying and distracting. We should avoid doing this.

zooba:
Fixes #1865 Unit tests refresh too often
Makes TestCaseInfo use value equality.

### DIFF
--- a/Python/Product/TestAdapter.Analysis/TestCaseInfo.cs
+++ b/Python/Product/TestAdapter.Analysis/TestCaseInfo.cs
@@ -57,6 +57,12 @@ namespace Microsoft.PythonTools.TestAdapter {
                 { TestAnalyzer.Serialize.Kind, Kind.ToString() },
             };
         }
+
+        // Compare based on Python name, so that editing the file doesn't replace the object
+        private object CompareKey => new { Filename, ClassName, MethodName, StartLine };
+
+        public override bool Equals(object obj) => CompareKey.Equals((obj as TestCaseInfo)?.CompareKey);
+        public override int GetHashCode() => CompareKey.GetHashCode();
     }
 
     internal enum TestCaseKind {


huguesv:
I didn't notice much of a difference on a small project. It is still too frequent, and we lose test results because of it.

Here I typed the B2 test after run all, which made C disappear for a while, until B2 and C appeared, and C lost its result.

image

C# doesn't have that issue:
image

If we only do 1 refresh, we also don't have that issue. Instead of adding B2, try removing B in one edit. C results remain (and its location updated). Then undo the delete, B appears (no results which is ok), and C is again updated and doesn't lose its result.

I think we are firing too many updates and that's why test explorer can't keep track of association between test case and test results.

In the C# example above, there was only one refresh, after I had stopped typing for a couple of seconds.

huguesv:
The problem with the test disappearing and then losing results is because we refresh the list while the code is not valid. If we could increase the delay after typing to at least 1 sec, the user should have enough time to define the new function and have valid code.

Anyway, as for this change, I'll do more testing on some code that has more tests in it, to see if this reduces the number of refreshes when the edits should not be affecting the existing test cases.

huguesv:
Okay, back to this fix: it's a great improvement when you're editing code that does not affect the set of tests.

When adding the `for m` loop, without fix:
image

When adding the `for l` loop, with the fix:
image

zooba:
Yeah, the problem I found was that we need to regenerate the tests to update the line numbers, and if we don't then double-clicking won't go to the right place. So every time the location moves we have to refresh.

If we could figure out mapping on double-click then we could remove most of the rest. But this is a step in the right direction.

PR #4530, opened 2018-07-11
Completed model for source line mapping

rgesteve:
Beginnings of PR with functionality for source-line mapping viz (needs 15.8p2 or better)

### DIFF
--- a/Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
+++ b/Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
@@ -150,6 +150,7 @@ namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver {
     public abstract class VTuneCollectSpec : VTuneSpec {
         public abstract string AnalysisName { get; }
         public string WorkloadSpec { get; set; }
+        public string SymbolPath { get; set; }
         public string AnalysisCLI {
             get { return "-collect" + " " + AnalysisName; }
         }
@@ -168,6 +169,11 @@ namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver {
             sb.Append(AnalysisCLI);
             sb.Append(" " + UserDataDirCLI());
             sb.Append(" " + ResultDirCLI);
+
+            if (this.SymbolPath != String.Empty) {
+                sb.Append($" -search-dir {this.SymbolPath}");
+            }
+
             sb.Append(WorkloadCLI);
             return sb.ToString();
         }
--- a/Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
+++ b/Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
@@ -25,6 +25,27 @@ using Microsoft.DotNet.PlatformAbstractions;
 
 namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver {
 
+    // Very simple Maybe "monad" to handle file number definitions
+    public struct Maybe<T>
+    {
+        public readonly static Maybe<T> None = new Maybe<T>();
+        public T Value;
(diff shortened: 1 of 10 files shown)


zooba:
@rgesteve I'm happy to merge this, but it looks incomplete? Or do those items just need to be there to be loaded?

rgesteve:
Yes, sorry, just added the second of third parts (support to collect symbol information from VTune, including reporting function name/line information from pdbs).  The third part I should get in this evening.  One question: I see that the debugger includes "Python/Product/Debugger.Concord/Microsoft.Dia.Interop.dll"... is this a wrapper around msdia*.dll?  I ask because the function/line number included here uses such a wrapper and I'd love it if I didn't have to include the corresponding dll.  Thanks!

rgesteve:
The second of *three* parts, not "third" :(

Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
```diff
@@ -25,6 +25,27 @@
 
 namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver {
 
+    // Very simple Maybe "monad" to handle file number definitions
+    public struct Maybe<T>
```
zooba: Seems unused? In any case, it's identical to `Nullable<T>`, which is what you seem to be using already.

Python/Product/ExternalProfilerDriver/VTuneInvoker.cs
```diff
@@ -168,6 +190,11 @@ public class VTuneCollectHotspotsSpec : VTuneCollectSpec {
             sb.Append(AnalysisCLI);
             sb.Append(" " + UserDataDirCLI());
             sb.Append(" " + ResultDirCLI);
+
+            if (this.SymbolPath != String.Empty) {
```
zooba: Use `string.IsNullOrEmpty(SymbolPath)` instead, to handle both cases.

Python/Product/ExternalProfilerDriver/VTuneUtils.cs
```diff
@@ -0,0 +1,145 @@
...
+            }
+            process.WaitForExit();
+        }
+    }
+
+    /// <summary>
+    /// A `Maybe` monad implementation based largely on the one(s) in the
+    /// Azure IoT Edge project.
+    /// </summary>
+    public struct Option<T> : IEquatable< Option<T> >
```
zooba: Also the same as `Nullable<T>`, and apparently unused.

Python/Product/ExternalProfilerDriver/VTuneToDWJSON.cs
```diff
@@ -209,6 +250,106 @@ class VTuneToDWJSON {
...
+                    name = mf.Module
+                };
+                List<FunctionSpec> funcsInCurrentModule = new List<FunctionSpec>();
+                foreach (var ff in mf.Fnames) {
+                    //FunctionSpec currrent = new FunctionSpec (ff.FunctionName, ff.Base, ff.Size);
+                    FunctionSpec current = new FunctionSpec (ff.FunctionName, 0,0);
+
+                    // look up information on function
+                    if (mfilesDWJSON.ContainsKey(mf.Module)) {
+                        int idx = mfilesDWJSON[mf.Module].FindIndex(fi => fi.file == ff.SourceFile);
```
zooba: Since you're just using this to get the file id anyway, try 
```
var idForFun = mfilesDWJSON[mf.Module].FirstOrDefault(fi => fi.file == ff.SourceFile)?.id;
if (idForFun.HasValue) { ... }
```

huguesv:
@rgesteve will this be resubmitted against master later on? Should we close it?

PR #4535, opened 2018-07-11
[15.8] Fixes #4493 Class name not colorized in import ... as

### LINKED ISSUE #4493: Class name not colorized in import ... as
image

zooba:
Fixes #4493 Class name not colorized in import ... as
Adds imported 'as' name as resolvable name for classifications.
Also fixes member with no types being treated as a function.


zooba:
@huguesv Feel free to close if you don't want this for 15.8, but I think it's a noticeable regression that I'd rather not have around.

PR #4549, opened 2018-07-13
Fix tooltips in VSC 1.25

MikhailArkhipov:
Change MarkupKind to actual strings since VSC now returns `unsupported markup kind: 1`. 
Includes https://github.com/Microsoft/PTVS/pull/4543


Python/Product/Analysis/LanguageServer/Enums.cs
```diff
@@ -34,13 +34,6 @@ public enum TraceLevel {
         Verbose
     }
 
-    public enum MarkupKind {
```
AlexanderSher: Is it a protocol change?

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
```diff
@@ -0,0 +1,105 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using Microsoft.DsTools.Core.Disposables;
+using Microsoft.PythonTools.Analysis.Infrastructure;
+
+namespace Microsoft.Python.LanguageServer {
+    internal sealed class PathsWatcher : IDisposable {
```
AlexanderSher: I thought  that VSC is a source of truth in terms of files.

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
```diff
@@ -365,6 +376,10 @@ public Task LoadExtension(JToken token)
 
         #endregion
 
+        #region Custom
+        [JsonRpcMethod("python/reloadModules")]
+        public Task ReloadModules() => _server.ReloadModulesAsync();
```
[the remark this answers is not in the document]
MikhailArkhipov: No, it is not needed as it is default behavior for LS. VS is not affected as it is does not use that layer.

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
MikhailArkhipov: TS file watcher is much less reliable as I found. Also, this is generally useful behavior. If we are to do it at TS level, we'd need custom `python/reloadModules` notification.

Python/Product/Analysis/LanguageServer/ServerBase.cs
```diff
@@ -124,9 +124,10 @@ private sealed class RequestLock : IDisposable {
 
         public virtual Task<WorkspaceEdit> Rename(RenameParams @params) => throw new NotImplementedException();
 
+        public virtual Task ReloadModulesAsync() => Task.CompletedTask;
```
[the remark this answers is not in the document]
MikhailArkhipov: OK

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
AlexanderSher: Ok, so how do we resolve conflicts with `workspace/didChangeWatchedFiles`?

MikhailArkhipov: It never come AFAICT b/c we don't have any watched files registered through LSP

Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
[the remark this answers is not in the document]
MikhailArkhipov: Wait, I thought I removed this

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
```diff
@@ -0,0 +1,105 @@
...
+                    .Add(() => fsw.Created -= OnChanged)
+                    .Add(() => fsw.Deleted -= OnChanged)
+                    .Add(fsw);
+            }
+        }
+
+        private void OnChanged(object sender, FileSystemEventArgs e) {
+            // Throttle calls so we don't get flooded with requests
+            // if there is massive change to the file structure.
+            lock (_lock) {
```
AlexanderSher: Do we care about the change content at all?

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
AlexanderSher: Is it possible that we end up with set of files and folders that is different from what user sees in VSC?

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
MikhailArkhipov: You mean what changed? Not really. We just want to filter out multiple file copy so we reload when install actually completes. Lock is for the timer that fires on a different thread.

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
MikhailArkhipov: Yes. Python folders are different from user code. I.e. we are watching Python search (i.e libraries) while VSC sees user code. In RTVS we watched `install.packages` while here we watch for folder changes.

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
```diff
@@ -0,0 +1,105 @@
...
+                _changedSinceLastTick = true;
+                _throttleTimer = _throttleTimer ?? new Timer(TimerProc, null, 500, 500);
+            }
+        }
+
+        private void TimerProc(object o) {
+            lock (_lock) {
+                if (!_changedSinceLastTick && _throttleTimer != null) {
+                    _syncContext.Post(s => _onChanged(), null);
```
AlexanderSher: In our case, `_onChanged = _server.ReloadModulesAsync(CancellationToken.None).DoNotWait()`.

`ReloadModulesAsync` schedules several items into the queue but since we use `DoNotWait`, it is possible that two `ReloadModulesAsync` calls will run concurrently. Not sure if it is dangerous, but with current throttling timeout, I don't think there is a benefit of using `_syncContext` over simply calling `_onChanged`.

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
AlexanderSher: Ok, so we don't actually care about client side but rather interesting in real state of file system.

Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
AlexanderSher: If we don't care about the order, I think `ThreadPool.QueueUserWorkItem(_ => _onChanged())` should be enough.

PR #4551, opened 2018-07-13
Update to build with VS "16.0"

zooba:
VS 16.0 doesn't exist yet, so right now this is mostly just making it possible to build with nuget references and a `VSTarget` variable that doesn't match the current version of VS (but I'm assuming that there'll be backwards-compatibility allowances so we can reference VS 15.0 assemblies and continue to work in the next version)


Build/15.0/packages.config
```diff
@@ -1,30 +1,52 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.VSSDK.BuildTools" version="15.5.100" />
+  <package id="Microsoft.VSSDK.BuildTools" version="15.7.109" />
   <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net45" />
   <package id="MicroBuild.Core" version="0.2.0" />
   <package id="Microsoft.VisualStudio.Debugger.InteropA" version="9.0.21022" targetFramework="net46" />
-  <package id="Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime" version="14.3.25408" />
+  <package id="Microsoft.VisualStudio.Debugger.Interop.10.0" version="10.0.30320" />
```
AlexanderSher: Do we need those changes for 15.0?

Python/Product/AnacondaInstallLauncher/AnacondaInstallLauncher.csproj
```diff
@@ -11,14 +11,14 @@
         <OldToolsVersion>14.0</OldToolsVersion>
       </PropertyGroup>
     </When>
-    <When Condition=" '$(VisualStudioVersion)'=='14.0'  Or '$(TargetVisualStudioVersion)'=='VS140' ">
+    <When Condition=" '$(VisualStudioVersion)'=='16.0'  Or '$(TargetVisualStudioVersion)'=='VS160' ">
```
AlexanderSher: Don't we have a common props file where we can store these conditions?

Build/Common.Build.Core.settings
```diff
@@ -32,12 +34,12 @@
     The version number that will identify a specific build. This appears in
     Programs and Features, Extension Manager, and Help About.
     -->
-    <FileVersionPrefix Condition="$(FileVersionPrefix) == ''">15.8</FileVersionPrefix>
+    <FileVersionPrefix Condition="$(FileVersionPrefix) == ''">$(VSTarget)</FileVersionPrefix>
```
huguesv: I've found having the minor version to be very useful (15.8 instead of 15.0). Why do you want to get rid of it?

Common/Tests/MockVsTests/MockVsTests.csproj
```diff
@@ -86,15 +77,59 @@
...
-    <When Condition="$(VSTarget) == '15.0'">
+    <When Condition="$(VSTarget) == '10.0'">
       <ItemGroup>
-        <Reference Include="Microsoft.VisualStudio.Shell.Framework, Version=$(VSTarget).0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
+        <Reference Include="Microsoft.VisualStudio.Shell.10.0" />
+        <Reference Include="Microsoft.VisualStudio.Shell.Immutable.10.0, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
       </ItemGroup>
     </When>
-    <Otherwise>
+    <When Condition="'$(VSTarget)' == '11.0'">
```
AlexanderSher: Do we really need `'$(VSTarget)' == '11.0'`?

Build/Common.Build.Core.settings
zooba: Ah, good point. I had a change to make it 16.0, and forgot that it previously had the minor version.

Common/Tests/MockVsTests/MockVsTests.csproj
zooba: We've never officially dropped old versions from the shared code ("Common" folder). So until we do that, yes. (And [PHP Tools](https://www.devsense.com/en/download) is still supporting VS 2010 using this code, so we should run it past them before making any significant change here.)

Python/Product/AnacondaInstallLauncher/AnacondaInstallLauncher.csproj
zooba: Yes, but unfortunately these ones need to stay in the .csproj or VS can't open the file properly. Maybe they've fixed these bugs in VS by now? But last time I checked they hadn't.

zooba: (The bug is that they only look in the current file to find these properties and don't resolve through Imports)

Build/15.0/packages.config
zooba: Not strictly, but they're generally good anyway. Right now the 16.0 and 15.0 sets of packages are identical, which is convenient for making it more isolated from the VS install, but I expect them to diverge. Still, we need to keep the builds working with 15.0 until people outside of MS have moved up to 16.0 or else they won't be able to build and use our newest bits against 2017.

huguesv:
Version of ProjectWizards assembly in .vstemplate files need to be updated

Example from WebRoleBottle.vstemplate

```
  <WizardExtension>
    <Assembly>Microsoft.PythonTools.ProjectWizards, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</Assembly>
    <FullClassName>Microsoft.PythonTools.ProjectWizards.InstallRequirementsWizard</FullClassName>
  </WizardExtension>
```

PR #4554, opened 2018-07-13
Allow non-IAnalyzable code to be executed in AnalysisQueue schedule

AlexanderSher:
This infrastructure change is a follow-up of the #4496 PR, related to the problem that we have code parts that change shared analysis data outside of the `AnalysisQueue`. 

**Current issues**
1. Changes may be implicit, triggered by requests to the data that is lazily initialized.
1. We have code that makes such changes, doesn't implement `IAnalyzable` and is quite async.
1. `AnalysisQueue` is single-threaded and thread-bound. Because of that, any code that is executed inside the queue must run on the same thread, so `ConfigureAwait(false)` can't be used.
1. It doesn't support reentrancy and can't execute async code that awaits on `AnalysisQueue` thread. 
1. `AnalysisSynchronizationContext` is presented but is used only as a scope token: any explicit scheduling goes through `AnalysisQueue` directly, implicit scheduling through `async/await` results in a deadlock and can't be used.
1. Any exception inside the queue kills it.

**Requirements**
- Any code that isn't a part of `IAnalyzable` should be able to execute and async delegate.
- If this delegate isn't hanging by itself, it shouldn't hang the queue 
- Result of the execution, as well as exceptions in it should be propagated to the caller and shouldn't affect the queue
- While delegate is executed, queue waits for it.
- Code should be able to check if it is executed inside the queue.
- Code that is executed inside the queue should be able to add new tasks to the queue.

**Changes**
- Introduced generic `PriorityProducerConsumer` type that implements producer/consumer pattern in the async way.
- `AnalysisQueue` is made free threaded and doesn't need `SynchronizationContext` anymore. Code that is executed inside the queue can run on any threads.
- No dedicated thread is required. This has also improved unit-test performance:
image
image
- Instance of the `AnalysisQueue` available through `AnalysisQueue.Current` is used as a scope token instead of `SynchronizationContext`. It can be used to schedule new tasks as well. 
- Some dead code is cleaned up. 

No changes in `Analysis` test results.


zooba:
Overall I can see a lot of benefits to this, but I'm wondering how it solves the problem you set out to address?

Specifically, code that is currently modifying analysis data from outside the analysis thread actually _should not_ be modifying analysis data. In my opinion, the bigger problem is that queries from outside analysis can change how we are interpreting user code (for example, by inserting function calls that don't really exist).

I think we probably dealt with the worst of these back when we made `IAnalysisSet` instances immutable by default. We detected that because of the race conditions. We then detected more bugs by testing `Deque` modifications were happening on the right thread. Would this change make it impossible to detect other violations in the same way?

Threading is nice here because it is side-channel context - the "official" context to determine whether we're doing the wrong thing is `AnalysisUnit.ForEval`, but since the bugs are mostly that we aren't setting this correctly when calling in, we need a second way to detect issues. Because even if we fix the race conditions, the `ForEval` bugs are still there.

Ideally, I think we should make sure that the mutation bugs are fixed before moving to free threading (if it is indeed going to make it harder to detect the mutation bugs).

zooba:
(Make sure you check individual test performance changes, too. There are so many factors impacting how long tests take that a simple aggregate over 30 tests showing 193ms of improvement each doesn't really mean much...)

AlexanderSher:
This change is not to make bugs harder to detect. The purpose is to make it easier to fix them. 

For example, `CompletionAnalysis.GetCompletionsInImport` indirectly calls `_interpreter.ImportModule`. Instead of wrapping part of the call stack into `IAnalyzable`, we can simply ensure that `_interpreter.ImportModule` is called inside the queue.

Or we have `ReloadModulesQueueItem`, which calls `_analyzer.ReloadModulesAsync(cancel).WaitAndUnwrapExceptions()` that doesn't hand now only by a fluke.

Scope checking, instead of being bound to the thread, is bound to the call stack. So in sync single-threaded case, there is no difference, but in async case it allows to run any code inside the queue.

Test performance wasn't a goal of the change, it is more a side effect. `AnalysisTests` took 4.53 seconds before change (887 tests), 4.03 seconds after (898 tests).

zooba:
`IPythonInterpreter.ImportModule` shouldn't need to be called within the queue, and it also shouldn't inject anything into the analyzer (into the module table is fine, since the analyzer should only ever pull from that).

And do we really want to block analysis because a query is trying to read from a different module? I don't think we need to. Getting results for "from x import ..." shouldn't have to wait for the analysis queue to get around to it, and shouldn't have to hold up the running analysis to get the `IPythonModule` object. The single-threadedness is there to protect `AnalysisValue` subclasses.

Maybe once we ship and have moved this out to its own repo with its own test suite we can take the time to look at major internal refactoring. I don't think it's a high priority this week.

AlexanderSher:
> `IPythonInterpreter.ImportModule` shouldn't need to be called within the queue

For that, code that runs in queue should be able to wait if code from outside has called `IPythonInterpreter.ImportModule` for requested module and sentinel is already created. With current implementation, it will lead to a deadlock.

> Maybe once we ship and have moved this out to its own repo with its own test suite we can take the time to look at major internal refactoring

This isn't a major change because there is only one new class and one changed. `DisposeToken`, `TaskCompletionSourceExtensions` and `TaskCompletionSourceEx` are taken from RTVS.
And it isn't a refactoring because it changes the way analysis queue works when it has to wait on async call.

zooba:
> code that runs in queue should be able to wait if code from outside has called `IPythonInterpreter.ImportModule` for requested module and sentinel is already created

Yep, that's the idea. Sorry I didn't explain this more clearly early on, but you've got it now :) And I don't think it will necessarily lead to a deadlock (it hasn't in the past, until more recent changes, and we've been using this model for quite a while), so we should fix that if so.

> This isn't a major change ...

Not yet, but you said in the original description that this is to enable making bigger changes. I'm all in favour of that, but I also prefer to apply [YAGNI](http://wiki.c2.com/?YouArentGonnaNeedIt) when we're this close to a release. If there's no other fix for the deadlock, then we can evaluate it in that context, but that isn't the reason you've been giving for it :)

huguesv:
@zooba which release are you worried about? A VSCode release? Because this isn't intended to go into 15.8 right? The PR is for master branch.

AlexanderSher:
> And I don't think it will necessarily lead to a deadlock (it hasn't in the past, until more recent changes, and we've been using this model for quite a while), so we should fix that if so.

Well, here is the test that reproduces the issue:

```cs
[TestMethod, Priority(0)]
public async Task AnalysisQueue_WaitOnTryImportModuleAsyncTest() {
    var opts = new InterpreterFactoryCreationOptions {
        DatabasePath = TestData.GetTempPath("AstAnalysisCache"),
        UseExistingCache = false
    };

    var factory = new AstPythonInterpreterFactory(
        Latest.Configuration,
        opts
    );

    var tcs = new TaskCompletionSource<bool>();
    var tcsModulePath = new TaskCompletionSource<ModulePath?>();
    var context = new AstPythonInterpreterFactory.TryImportModuleContext {
        ModuleCache = new ConcurrentDictionary<string, IPythonModule>(),
        FindModuleInUserSearchPathAsync = (s, ct) => {
            tcs.SetResult(true);
            return tcsModulePath.Task;
        }
    };

    // Simulate a call from ModuleTable.ImportModuleAsync
    var outsideTask = Task.Run(() => factory.TryImportModuleAsync("some.pyc", context, CancellationToken.None));

    // Ensure that sentinel is created
    await tcs.Task;

    var queue = new AnalysisQueue();
    var project = new AnalysisProject(factory, context);
    queue.Enqueue(new Analyzable(project), AnalysisPriority.Normal);

    // Wait for IGroupableAnalysisProject.AnalyzeQueuedEntries to finish its sync part
    await project.Task;

    // Release the sentinel
    tcsModulePath.SetResult(new ModulePath("some.pyc", "c:\\some.pyc", "Abc"));

    await outsideTask;

    // 10 seconds should be enough to finish the task
    await Task.WhenAny(Task.Delay(10000), queue.WaitForCompleteAsync());

    Assert.AreEqual(0, queue.Count);
}

private class Analyzable : IGroupableAnalysisProjectEntry, IAnalyzable {
    public Analyzable(IGroupableAnalysisProject analysisGroup) {
        AnalysisGroup = analysisGroup;
    }

    public void Analyze(CancellationToken cancel, bool enqueueOnly) {}
    public IGroupableAnalysisProject AnalysisGroup { get; }
    public void Analyze(CancellationToken cancel) {}
}

private class AnalysisProject : IGroupableAnalysisProject {
    private readonly AstPythonInterpreterFactory _factory;
    private readonly AstPythonInterpreterFactory.TryImportModuleContext _context;
    private readonly TaskCompletionSource<bool> _tcs;

    public Task Task => _tcs.Task;

    public AnalysisProject(AstPythonInterpreterFactory factory, AstPythonInterpreterFactory.TryImportModuleContext context) {
        _factory = factory;
        _context = context;
        _tcs = new TaskCompletionSource<bool>();
    }

    public void AnalyzeQueuedEntries(CancellationToken cancel) {
        // Get a task to ensure that sync part of the call stack is completed
        var task = _factory.TryImportModuleAsync("some.pyc", _context, cancel);
        // Notify test that 
        _tcs.SetResult(true);
        // Simulate a call from `AstPythonInterpreter.ImportModule`
        task.Wait(); 
    }
}
```

Now, in real scenario it doesn't hang completely because `AstPythonInterpreter.ImportModule` has a timeout:
```cs
        public IPythonModule ImportModule(string name) {
            var token = new CancellationTokenSource(5000).Token;
#if DEBUG
            token = Debugger.IsAttached ? CancellationToken.None : token;
#endif
            try {
                var impTask = ImportModuleAsync(name, token);
                if (impTask.Wait(10000)) {
                    return impTask.Result;
                }
            } catch (AggregateException ex) {
                throw ex.InnerException ?? ex;
            }
            return null;
```
So instead it can freeze the queue for 10 seconds and then return null. But that is still a bug.
 
With current `AnalysisQueue` implementation, this issue can't be fixed in async way - you'll have to switch back to the thread-blocking primitives. With the new implementation, this bug disappears, because continuation can be started on any thread-pool thread.

zooba:
I'd be more convinced by seeing assertions in `ImportModule` when it times out, or by a test that explicitly creates two threads and runs each part of it in those.

There are far too many async complexities in that test right now - how can we be sure that it isn't hanging trying to get back to the test thread? Or that the SetResult call isn't calling the continuation on a different thread than it would in the real scenario?

AlexanderSher:
Here is a `Thread.Sleep` version of the test. The only async is for `FindModuleInUserSearchPathAsync` to simulate its real implementation. Test fails on last assert, showing one item in the analysis queue.

```cs
[TestMethod, Priority(0)]
public void AnalysisQueue_WaitOnTryImportModuleAsyncTest() {
    var opts = new InterpreterFactoryCreationOptions {
        DatabasePath = TestData.GetTempPath("AstAnalysisCache"),
        UseExistingCache = false
    };

    var factory = new AstPythonInterpreterFactory(
        Latest.Configuration,
        opts
    );

    var context = new AstPythonInterpreterFactory.TryImportModuleContext {
        ModuleCache = new ConcurrentDictionary<string, IPythonModule>(),
        FindModuleInUserSearchPathAsync = async (s, ct) => {
            await Task.Delay(200, ct);
            return new ModulePath("some.pyc", "c:\\some.pyc", "Abc");
        }
    };

    // Simulate a call from ModuleTable.ImportModuleAsync
    var outsideTask = Task.Run(() => factory.TryImportModuleAsync("some.pyc", context, CancellationToken.None));

    Thread.Sleep(100);

    var queue = new AnalysisQueue();
    var project = new AnalysisProject(factory, context);
    queue.Enqueue(new Analyzable(project), AnalysisPriority.Normal);

    Thread.Sleep(200);

    Assert.AreEqual(outsideTask.Status, TaskStatus.RanToCompletion);

    Thread.Sleep(5000);

    Assert.AreEqual(0, queue.Count);
}

private class Analyzable : IGroupableAnalysisProjectEntry, IAnalyzable {
    public Analyzable(IGroupableAnalysisProject analysisGroup) {
        AnalysisGroup = analysisGroup;
    }

    public void Analyze(CancellationToken cancel, bool enqueueOnly) {}
    public IGroupableAnalysisProject AnalysisGroup { get; }
    public void Analyze(CancellationToken cancel) {}
}

private class AnalysisProject : IGroupableAnalysisProject {
    private readonly AstPythonInterpreterFactory _factory;
    private readonly AstPythonInterpreterFactory.TryImportModuleContext _context;

    public AnalysisProject(AstPythonInterpreterFactory factory, AstPythonInterpreterFactory.TryImportModuleContext context) {
        _factory = factory;
        _context = context;
    }

    public void AnalyzeQueuedEntries(CancellationToken cancel) {
        // Simulate a call from `AstPythonInterpreter.ImportModule`
        _factory.TryImportModuleAsync("some.pyc", _context, cancel).Wait();
    }
}
```

PR #4559, opened 2018-07-16
Streamline Language Server release builds

zooba:


Python/Product/Analysis/AnalysisNetStandard.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project>
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
AlexanderSher: There is no `netstandard2.1` target, only `netstandard2.0`

Python/Product/Analysis/AnalysisNetStandard.csproj
```diff
@@ -34,6 +29,9 @@
     <Content Include="typing-stub.pyi">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="Typeshed\**\*">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
MikhailArkhipov: VS Code extension includes typeshed as submodule to the extension rather than fixed version with the LS. Packaging it with LS may yield two typesheds installed.

zooba: Brett made an issue on the PVSC repo to remove the submodule and rely on this copy. (Importantly, I think we may need to modify typeshed's `sys.pyi` to make it work properly - it currently imports itself, which may be the cause of some typeshed issues.)

PR #4561, opened 2018-07-16
Fix REPL completions

### LINKED ISSUE #4178: Live completions in REPL not working
In REPL window:
```
from datetime import datetime
n = datetime.now()
n.
```
Gives no completions.
Using default REPL options.

image

zooba:
Fix #4178 Live completions in REPL not working

Fix REPL completions not respecting the "live completions only" option and not synchronizing the buffer.


huguesv:
It's a lot more reliable :)

Duplicated items in the results from analysis/live completions, but I'll take that over the previous unreliability.

image

zooba:
I can fix that - the comparison is just using display name, but we should really check both display name and completion for equality.

zooba:
image

Though in this case, it's not actually correct, since the current value is the `datetime` module which does not have that member. But that problem won't go away quite as easily (we'd need to carefully break analysis between REPL entries)

PR #4564, opened 2018-07-17
[15.8] Fix REPL completions

zooba:
Fix REPL completions not respecting the "live completions only" option and not synchronizing the buffer.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/CompletionComparer.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/CompletionComparer.cs
@@ -92,20 +92,31 @@ namespace Microsoft.PythonTools.Intellisense {
         /// <param name="y"></param>
         /// <returns></returns>
         public int Compare(CompletionResult x, CompletionResult y) {
-            return Compare(x.Name, y.Name);
+            int i = Compare(x.Completion, y.Completion);
+            return i == 0 ? 0 : Compare(x.Name, y.Name);
         }
 
         /// <summary>
         /// Compares two <see cref="MemberResult"/> structures for equality.
         /// </summary>
         public bool Equals(CompletionResult x, CompletionResult y) {
-            return x.Name.Equals(y.Name);
+            return x.Name.Equals(y.Name) || x.Completion.Equals(y.Completion);
         }
 
         /// <summary>
         /// Gets the hash code for a <see cref="MemberResult"/> structure.
         /// </summary>
         public int GetHashCode(CompletionResult obj) {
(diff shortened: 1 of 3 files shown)


Python/Product/PythonTools/PythonTools/Intellisense/CompletionComparer.cs
```diff
@@ -92,20 +92,31 @@ class CompletionComparer : IEqualityComparer<CompletionResult>, IComparer<Comple
...
+            int i = Compare(x.Completion, y.Completion);
+            return i == 0 ? 0 : Compare(x.Name, y.Name);
         }
 
         /// <summary>
         /// Compares two <see cref="MemberResult"/> structures for equality.
         /// </summary>
         public bool Equals(CompletionResult x, CompletionResult y) {
-            return x.Name.Equals(y.Name);
+            return x.Name.Equals(y.Name) || x.Completion.Equals(y.Completion);
```
AlexanderSher: This implementation of `Equals` breaks transitive relation:

|      | Name | Completion |
| --- | ------ | -------- |
| `x` | "ab" | "bc" |
| `y` | "ab" | "cd" |
| `z` | "de" | "cd" |

`completionComparer.Equals(x, y) == true`, `completionComparer.Equals(y, z) == true`, but `completionComparer.Equals(x, z) == false`,

Python/Product/PythonTools/PythonTools/Intellisense/CompletionComparer.cs
```diff
@@ -92,20 +92,31 @@ class CompletionComparer : IEqualityComparer<CompletionResult>, IComparer<Comple
...
         /// </summary>
         public int GetHashCode(CompletionResult obj) {
+            if (obj.Name.Length < obj.Completion.Length && obj.Completion.StartsWithOrdinal(obj.Name)) {
+                return obj.Name.GetHashCode();
+            } else if (obj.Name.Length > obj.Completion.Length && obj.Name.StartsWith(obj.Completion)) {
+                return obj.Completion.GetHashCode();
+            }
+            for (int i = 0; i < obj.Name.Length && i < obj.Completion.Length; ++i) {
+                if (obj.Name[i] != obj.Completion[i]) {
+                    return obj.Name.Substring(0, i).GetHashCode();
```
AlexanderSher: `GetHashCode` shouldn't take additional memory in heap

Python/Product/PythonTools/PythonTools/Intellisense/CompletionComparer.cs
zooba: You're right, we're going to take a different approach. Hugues is working on it

Issue #4547, opened 2018-07-13
Typeshed completions missing after switching interpreter


huguesv:
Create new Python application
Paste this code:
```
import sys

e1, e2, e3 = sys.exc_info()
```
`e1` and `e2` should have type info, and `e1` should be colorized as a class

In solution explorer, change the active interpreter (ex from 3.6 to 2.7)

Completions on `e1` and `e2` are missing, and colorization of `e1` is missing.

Change active interpreter again. Notice that for a brief moment (less than half a sec), `e1` is colorized correctly, but the color goes away immediately.

Close and reopen the project, and everything works again.

Maybe related to https://github.com/Microsoft/PTVS/issues/4533 ?

PR #4573, opened 2018-07-18
Fixes error squiggles not clearing when warnings or tasks remain.

zooba:
Fixes error squiggles not clearing when warnings or tasks remain.
Fixes unnecessary recursive self-imports.


zooba:
Found these while working on #4547. I think the squiggles one is worth taking for 15.8 (to repro, import a missing module and add a syntax error, then fix either), but the self-recursive import is very safe.

zooba:
I think it depends which one you fix, but fair enough.

PR #4585, opened 2018-07-21
Remove legacy db

huguesv:
Fix https://github.com/Microsoft/PTVS/issues/4570

Removes the option to enable legacy db, all UI related to it, all core code under LegacyDB namespace, tests and test data.

Maybe more to delete:
- AnalysisBrowser?

UWP changes compile but are untested.


zooba:
AnalysisBrowser can go away entirely. As proud as I am of that tool, it serves no purpose anymore :)

huguesv:
Removed the analysis browser. As discussed in person, I ran the tests locally and we have some time to handle test regressions, if any, before the next milestone.

PR #4597, opened 2018-07-26
Add option to disable language server

### LINKED ISSUE #4587: Option to disable all IntelliSense services
Don't start the analyzer when option is disabled. A lot of editor features will be unavailable, but giving the user the choice can be useful, if they have a particular need for lower resource usage.

Note that C++ has several IntelliSense related options, including "Disable all IntelliSense features".

https://github.com/Microsoft/PTVS/issues/4582 is a crash where a user attempted to rename the analyzer executable to prevent it from running. That didn't work, it just crashed VS when we tried to launch it.

huguesv:
Fix #4587 

Most code could already handle not getting analysis entry or getting null from an analyzer request, so this is mostly just adding the option, and not starting analysis.

Not entirely satisfied with the option name.

I'm convinced we need this as an escape hatch for problematic projects (I've seen users kill the analyzer process and try to rename it to prevent it from starting), and having the choice between feature and resource usage is good to have.

I'll leave it open until others are back from vacation so we can discuss.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
+++ b/Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
@@ -241,7 +241,9 @@ namespace Microsoft.PythonTools.Intellisense {
                 }
 
                 if (entry == null) {
-                    Debug.Fail($"Failed to analyze {buffer.DocumentUri}");
+                    if (!_services.Python.LanguageServerOptions.ServerDisabled) {
+                        Debug.Fail($"Failed to analyze {buffer.DocumentUri}");
+                    }
                     return;
                 }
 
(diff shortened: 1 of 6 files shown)


zooba:
I seem to recall another language having a similar option, that was basically just "Disable IntelliSense".

Having the option sounds good.

huguesv:
We can name the checkbox "Disable all IntelliSense features" like in the C++ option description. I was a bit worried it may not be obvious that this also disables things like refactoring, go to definition, etc. but it's probably ok.

image

Python/Product/PythonTools/PythonTools/Intellisense/IntellisenseController.cs
```diff
@@ -241,7 +241,9 @@ internal sealed class IntellisenseController : IIntellisenseController, IOleComm
                 }
 
                 if (entry == null) {
-                    Debug.Fail($"Failed to analyze {buffer.DocumentUri}");
+                    if (!_services.Python.LanguageServerOptions.ServerDisabled) {
```
zooba: Should we just avoid attaching the IntellisenseController completely when disabled?

zooba:
Do we need to add checks to some commands to disable refactoring, etc? Perhaps suppressing the IntellisenseController (and EditFilter?) would do that automatically?

MinKPark:
@huguesv is this PR still needed? if not, how about discarding it?

huguesv:
It's desirable, in the weeks at least 2 people asked for it.

We'll need to see if we can maintain that option when we use Language Server Client (probably).

We can probably bypass the IntelliSense controller as suggested. We'll do another PR if/when we schedule this work.

PR #4599, opened 2018-07-27
 Fix #4560: IntelliSense completions don't reliably appear

### LINKED ISSUE #4560: IntelliSense completions don't reliably appear
When using the PTVS IntelliSense in VS or VS Code (as a Language Server), completions don't reliably appear.

To reproduce, type any variable/symbol, then repeatedly type "." and backspace. Completions sometimes will not appear at all, the severity of the issue seems to vary greatly depending on the code and where IntelliSense is being used.

AlexanderSher:
Create a `TaskCompletionSource` at the beginning of parsing to be able to wait for the corresponding analysis.

### DIFF
--- a/Python/Product/Analysis/LanguageServer/Server.Completion.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.Completion.cs
@@ -25,26 +25,27 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
     public sealed partial class Server {
         public override Task<CompletionList> Completion(CompletionParams @params) => Completion(@params, CancellationToken.None);
 
-        internal Task<CompletionList> Completion(CompletionParams @params, CancellationToken cancellationToken) {
+        internal async Task<CompletionList> Completion(CompletionParams @params, CancellationToken cancellationToken) {
             var uri = @params.textDocument.uri;
  
             ProjectFiles.GetAnalysis(@params.textDocument, @params.position, @params._version, out var entry, out var tree);
             TraceMessage($"Completions in {uri} at {@params.position}");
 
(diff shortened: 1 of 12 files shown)


Python/Product/Analysis/LanguageServer/Server.Completion.cs
```diff
@@ -25,26 +25,27 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
...
-        internal Task<CompletionList> Completion(CompletionParams @params, CancellationToken cancellationToken) {
+        internal async Task<CompletionList> Completion(CompletionParams @params, CancellationToken cancellationToken) {
             var uri = @params.textDocument.uri;
  
             ProjectFiles.GetAnalysis(@params.textDocument, @params.position, @params._version, out var entry, out var tree);
             TraceMessage($"Completions in {uri} at {@params.position}");
 
             tree = GetParseTree(entry, uri, cancellationToken, out var version) ?? tree;
-            var analysis = entry?.Analysis;
+            var analysis = entry != null ? await entry.GetAnalysisAsync(50, cancellationToken) : null;
```
MikhailArkhipov: I actually tried this approach a while ago. I used 100 ms I think. But then resolved problem differently. 
The analysis should not be clearing old stuff until new is available. I.e. if class declares variables they should remain until analyzer figures out new set.

AlexanderSher: Well, there is definitely a moment when there are no variables, cleaning doesn't happen as atomic operation. And there is still another concurrency issue when `entry?.Analysis` is accessed before parsing is actually started.

PR #4601, opened 2018-07-27
[WIP] Enable running a module as a script

huguesv:
Relax the startup file check to allow not specifying a startup file.
Fixes interpreter arguments appearing twice in the process arguments.

TODO: Need to look at the web project types

Note that https://github.com/Microsoft/ptvsd/pull/655 is needed for the tests to pass (for the value of sys.argv to be correct). Legacy debugger is already correct.


huguesv:
F5 works on flask project:

Startup File: (empty)
Script Arguments: `-m flask run`
Environment: `FLASK_APP=app.py` (or `runserver.py` depending on the template)

One issue is that the browser does not start up.

Using the custom commands for start server and start debug server fail with this error:
```
---------------------------
---------------------------
An error occurred while running Start server.



Failed to get command PythonRunWebServerCommand from project
---------------------------
OK   
---------------------------
```

huguesv:
Also need to look at the Open Folder scenario. I'll close this and revisit later.

PR #4609, opened 2018-08-06
Fix #2017  Language service fails when using Python 2.7 interpreter

### LINKED ISSUE #2017: Feature Request: Django: add bootstrap to intellisense
At the moment, I don't see a way to add css classes of bootstrap to intellisense. There is no Nuget package avaliable. A much more helpful feature would be to scan through the html for js, css on a CDN, automatically download them and add to intellisense

MikhailArkhipov:
Fixes https://github.com/Microsoft/vscode-python/issues/2017
Make sure interpreter is initialized before attempting to import types


MikhailArkhipov:
Thanks to @patrys for debugging this down

PR #4611, opened 2018-08-06
Relative import fails in Python 2.7

### LINKED ISSUE #4610: Relative import fails in Python 2.7
See https://github.com/Microsoft/vscode-python/issues/2320

image

Manually executing code in Python 2.7 console works. With Python 3.6 import works fine, no squiggly.

Execution in VS REPL fails. But from console running from the app folder, code works.

image

MikhailArkhipov:
Fixes #4610

The issue was that file watcher ctor throws when folder in search paths does not exist.

I was able to repro the issue in VS but it was much harder, the reason is different. Most probably adding 'hello.py' first leads to creation of unresolved 'imported' and later, when 'imported' is actually added the module table is not updated.

### DIFF
--- a/Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
+++ b/Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
@@ -20,17 +20,20 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using Microsoft.DsTools.Core.Disposables;
+using ILogger = Microsoft.PythonTools.Analysis.LanguageServer.ILogger;
 
 namespace Microsoft.Python.LanguageServer {
     internal sealed class PathsWatcher : IDisposable {
         private readonly DisposableBag _disposableBag = new DisposableBag(nameof(PathsWatcher));
         private readonly Action _onChanged;
         private readonly object _lock = new object();
+        private readonly ILogger _log;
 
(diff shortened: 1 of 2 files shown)


Python/Product/VSCode/AnalysisVsc/PathsWatcher.cs
```diff
@@ -40,19 +43,26 @@ internal sealed class PathsWatcher : IDisposable {
             var list = new List<FileSystemWatcher>();
             var reduced = ReduceToCommonRoots(paths);
             foreach (var p in reduced) {
-                var fsw = new FileSystemWatcher(p);
-                fsw.IncludeSubdirectories = true;
-                fsw.EnableRaisingEvents = true;
+                if(!Directory.Exists(p)) {
```
AlexanderSher: Let's just catch `IOException` here

MikhailArkhipov: Ctor actually throws `ArgumentException`

AlexanderSher: Ok. Do we care about `PathTooLongException`?

MikhailArkhipov: Oh, I see

PR #4612, opened 2018-08-07
Introduce LS setting to delay diagnostic publishing to idle

MikhailArkhipov:
Addresses https://github.com/Microsoft/vscode-python/issues/2270

### DIFF
--- a/Python/Product/Analysis/LanguageServer/Settings.cs
+++ b/Python/Product/Analysis/LanguageServer/Settings.cs
@@ -28,6 +28,8 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
             public string[] information { get; } = Array.Empty<string>();
             public string[] disabled { get; } = Array.Empty<string>();
 
+            public int diagnosticPublishDelay = 1000;
+
             public DiagnosticSeverity GetEffectiveSeverity(string code, DiagnosticSeverity defaultSeverity)
                 => _map.TryGetValue(code, out var severity) ? severity : defaultSeverity;
 
--- a/Python/Product/Analysis/LanguageServer/Server.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.cs
@@ -104,7 +104,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
         }
 
         internal PythonAnalyzer Analyzer { get; private set; }
-        internal LanguageServerSettings Settings { get; private set; } = new LanguageServerSettings();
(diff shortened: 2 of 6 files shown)


Python/Product/Analysis/LanguageServer/Settings.cs
```diff
@@ -28,6 +28,8 @@ public class PythonAnalysisOptions {
             public string[] information { get; } = Array.Empty<string>();
             public string[] disabled { get; } = Array.Empty<string>();
 
+            public int diagnosticPublishDelay = 1000;
```
zooba: Not sure how I feel about having a setting here that isn't actually used by Server itself. Wouldn't this be better at the actual point of publishing anyway? (And maybe default to "no delay" but then make the VSC configuration default to 1s?)

MikhailArkhipov: These are `LanguageServerSettings` they are only used in the LS scenario. They happen to be defined at the Server level since it was simpler. I can try pushing them to the LS instead.

MikhailArkhipov: Generally speaking the setting may come useful, the complaint was not VSC specific. IT was more about "I don't want to see squiggles as I type, I know the construct is incomplete". Applies to VS too. `openFilesOnly` is already there too.

Now, handling is in VSC LS side and does not affect anything else. BTW, don't we merge everything into LS at some point anyway?

zooba: VS already throttles on the editor side and on the error list (though the latter may go away soon when we move to the proper table API), which is why we don't see this complaint so often. We don't really need to also throttle on sending the diagnostics, but it may be confusing later on if the setting doesn't appear to do what it claims to.

(And eventually we'll use LSP for both, but I expect it'll be a different adapter for VS that uses the service host support, rather than being a standalone executable. So the DLLs will be shared, but not necessarily the EXE.)

MikhailArkhipov: To summarize, throttling is happening at the LS side, not on the Server side. LanguageServerSettings declared in Server, but it does contain fields specific to LS as it was simpler to implement conditions in the Server code. openFiilesOnly comes to mind. I've spent some time trying to move LanguageServerSettings to LS and just use Server methods, but it gets complicated pretty soon.

MikhailArkhipov: It is possible to split class into ServerSettings and LanguageServerSettings and derive one from another so Server will see less.

PR #4619, opened 2018-08-08
Make tests run without a Visual Studio instance.

zooba:
Many tests are failing on our v16 builds because this crashes. We can fall back on the environment variable in most cases.

### DIFF
--- a/Common/Tests/Utilities/VisualStudioPath.cs
+++ b/Common/Tests/Utilities/VisualStudioPath.cs
@@ -38,9 +38,9 @@ namespace TestUtilities {
                 var path = current.ResolvePath(current.GetProductPath());
                 return Path.GetDirectoryName(path);
             } catch (COMException) {
-                var path = Environment.GetEnvironmentVariable($"VisualStudio_{AssemblyVersionInfo.VSVersion}");
+                var path = Environment.GetEnvironmentVariable($"VisualStudio_IDE_{AssemblyVersionInfo.VSVersion}");
                 if (string.IsNullOrEmpty(path)) {
-                    path = Environment.GetEnvironmentVariable("VisualStudio");
+                    path = Environment.GetEnvironmentVariable("VisualStudio_IDE");
                 }
                 return path;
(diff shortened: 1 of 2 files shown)


Common/Tests/Utilities/VisualStudioPath.cs
```diff
@@ -16,25 +16,34 @@
...
-            var current = (ISetupInstance2)configuration.GetInstanceForCurrentProcess();
-            var path = current.ResolvePath(current.GetProductPath());
-            return Path.GetDirectoryName(path);
+            try {
+                var configuration = (ISetupConfiguration2)new SetupConfiguration();
+                var current = (ISetupInstance2)configuration.GetInstanceForCurrentProcess();
+                var path = current.ResolvePath(current.GetProductPath());
+                return Path.GetDirectoryName(path);
+            } catch (COMException) {
+                var path = Environment.GetEnvironmentVariable($"VisualStudio_IDE_{AssemblyVersionInfo.VSVersion}");
```
AlexanderSher: Is it "class not registered" COM exception for `new SetupConfiguration()`?

zooba: Yeah. I suspect your work to re-split up the tests will help more, but this change is good enough to get our pass rate back up to a reasonable value.

AlexanderSher: There is one more thing that we can fallback (before or after looking for the variable):

https://github.com/dotnet/roslyn/blob/a3f904c792fbe6b22fad1254b299b8c178340247/src/Workspaces/MSBuildTest/Interop.cs

zooba: I think we're still not going to find the instance from the current process anyway, so falling back to the environment variable is the best option for our test runs.

We probably need to move the non-UI tests that depend on VS binaries to run in-proc. As I say, I'm sure you'll get this sorted out just fine in the bigger change :)

PR #4622, opened 2018-08-09
Return appropriate derived class when method returns self

### LINKED ISSUE #4614: IOBase is missing write (and some other methods)
`IOBase` in scraped io misses `write` but it does exist in typeshed

image

But it does work in regular assignment
image

MikhailArkhipov:
Fixes #4614 

Will add test next

### DIFF
--- a/Python/Product/Analysis/Values/BoundBuiltinMethodInfo.cs
+++ b/Python/Product/Analysis/Values/BoundBuiltinMethodInfo.cs
@@ -53,7 +53,7 @@ namespace Microsoft.PythonTools.Analysis.Values {
         public override IAnalysisSet Call(Node node, AnalysisUnit unit, IAnalysisSet[] args, NameExpression[] keywordArgNames) {
             // Check if method returns self
             var returnType = _method.ReturnTypes.GetInstanceType();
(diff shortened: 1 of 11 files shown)


Python/Product/Analysis/Values/BoundBuiltinMethodInfo.cs
```diff
@@ -50,7 +51,12 @@ public BoundBuiltinMethodInfo(IPythonBoundFunction function, PythonAnalyzer proj
         }
 
         public override IAnalysisSet Call(Node node, AnalysisUnit unit, IAnalysisSet[] args, NameExpression[] keywordArgNames) {
-            return _method.ReturnTypes.GetInstanceType();
+            // Check if method returns self
+            var returnType = _method.ReturnTypes.GetInstanceType();
+            if (args.Length > 0 && returnType.FirstOrDefault() is BuiltinInstanceInfo biif && biif.PythonType == _method.Function?.DeclaringType) {
```
zooba: There's a `.Split` extension method for `returnType` here that makes it easier to handle multiple return values properly (split with a predicate, then union and return). Otherwise, I like the approach

PR #4624, opened 2018-08-09
Populate document outline with all symbols, not just top level

MikhailArkhipov:
https://github.com/Microsoft/vscode-python/issues/2334
https://github.com/Microsoft/vscode-python/issues/2050


Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -40,30 +42,61 @@ await GetModuleVariablesAsync(entry as ProjectEntry, opts, @params.query)
...
+                    yield return m;
                 }
             }
         }
 
+        private static IEnumerable<MemberResult> GetScopeVariables(ModuleAnalysis analysis, InterpreterScope scope, GetMemberOptions opts) {
+            var res = analysis.GetAllAvailableMembersFromScope(scope, opts).Concat(GetChildScopesVariables(analysis, scope, opts));
+            foreach (var m in res) {
+                yield return m;
+            }
```
zooba: `return res`?

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -40,30 +42,61 @@ await GetModuleVariablesAsync(entry as ProjectEntry, opts, @params.query)
...
+            .Concat(GetChildScopesVariables(analysis, analysis.Scope, opts));
+        }
+
+        private static IEnumerable<MemberResult> GetChildScopesVariables(ModuleAnalysis analysis, InterpreterScope scope, GetMemberOptions opts) {
+            foreach (var childScope in scope.Children) {
+                var res = GetScopeVariables(analysis, childScope, opts);
+                foreach (var m in res) {
+                    yield return m;
                 }
             }
```
zooba: `return scope.Children.SelectMany(c => GetScopeVariables(analysis, c, opts))`?

Python/Product/Analysis/ModuleAnalysis.cs
```diff
@@ -710,6 +692,40 @@ private class ErrorWalker : PythonWalker {
...
+                if (scopeResult.TryGetValue(kvp.Key, out var values)) {
+                    values.AddRange(vars);
+                } else {
+                    scopeResult[kvp.Key] = values = new List<AnalysisValue>(vars);
+                }
+                values.Add(new SyntheticDefinitionInfo(
+                    kvp.Key,
+                    null,
+                    kvp.Value.Definitions.Select(e => e.GetLocationInfo())
+                ));
```
zooba: Do we need this if there have been variables added? Or do we need it if the value has no definitions?

Python/Product/Analysis/Values/BuiltinNamespace.cs
```diff
@@ -133,6 +133,6 @@ class BuiltinNamespace<MemberContainerType> : AnalysisValue where MemberContaine
             return false;
         }
 
-        public override int GetHashCode() => new { hc1 = GetType().GetHashCode(), hc2 = _type.GetHashCode() }.GetHashCode();
+        public override int GetHashCode() => new { hc1 = GetType()?.GetHashCode() ?? 0, hc2 = _type?.GetHashCode() ?? 0 }.GetHashCode();
```
zooba: Think I just saw a slightly different fix in another PR, but now I look at it, we shouldn't have to call `GetHashCode()` on the members here. The anonymous type `GetHashCode` will do it (and handle nulls properly)

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: Yep :-)

Python/Product/Analysis/Values/BuiltinNamespace.cs
MikhailArkhipov: Yes, will merge

Python/Product/Analysis/ModuleAnalysis.cs
MikhailArkhipov: It was existing code that applied to a single scope and now called for each scope recursively.

PR #4632, opened 2018-08-13
[15.9] Fixes #4625 General pane in output window is unlocalised

### LINKED ISSUE #4625: General pane in output window is unlocalised
When we happen to be the first extension to create the General pane in the Output Window, we create it with an unlocalized title.

We should make sure that we use a properly localized resource string.

zooba:
Fixes #4625 General pane in output window is unlocalised
Fixes string using existing resources with same text.

### DIFF
--- a/Python/Product/Cookiecutter/Shared/Infrastructure/OutputWindowRedirector.cs
+++ b/Python/Product/Cookiecutter/Shared/Infrastructure/OutputWindowRedirector.cs
@@ -58,7 +58,7 @@ namespace Microsoft.CookiecutterTools.Infrastructure {
         /// not be found or created.</exception>
         public static OutputWindowRedirector GetGeneral(IServiceProvider provider) {
             if (_generalPane == null) {
-                _generalPane = Get(provider, VSConstants.OutputWindowPaneGuid.GeneralPane_guid, "General");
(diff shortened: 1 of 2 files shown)


Python/Product/Cookiecutter/Shared/Infrastructure/OutputWindowRedirector.cs
```diff
@@ -58,7 +58,7 @@ sealed class OutputWindowRedirector : Redirector {
         /// not be found or created.</exception>
         public static OutputWindowRedirector GetGeneral(IServiceProvider provider) {
             if (_generalPane == null) {
-                _generalPane = Get(provider, VSConstants.OutputWindowPaneGuid.GeneralPane_guid, "General");
+                _generalPane = Get(provider, VSConstants.OutputWindowPaneGuid.GeneralPane_guid, Strings.Settings_General_Category);
```
MikhailArkhipov: I assume these are already defined

zooba: Already defined and localized. The fix in master adds specific new strings, but we won't have time to get those fully updated for 15.9

PR #4652, opened 2018-08-15
Improve reliability of document outline population in VS Code

MikhailArkhipov:
Addresses https://github.com/Microsoft/vscode-python/issues/2224

Symbols nature is currently not always known until analysis completes. However, blindly waiting for analysis to complete does not make sense. So added cancellable wait - VSC does not cancel request for document symbols, but user command (as in Palette -> Go To Symbol) can be canceled.

VSC does not provide API to update document outline. It simple asks for symbols when document changes.

### DIFF
--- a/Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
+++ b/Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
@@ -208,8 +208,9 @@ namespace Microsoft.Python.LanguageServer.Implementation {
         }
 
         [JsonRpcMethod("workspace/symbol")]
-        public async Task<SymbolInformation[]> WorkspaceSymbols(JToken token) {
+        public async Task<SymbolInformation[]> WorkspaceSymbols(JToken token, CancellationToken cancellationToken) {
             await _prioritizer.DefaultPriorityAsync();
+            await WaitForCompleteAnalysisAsync(cancellationToken);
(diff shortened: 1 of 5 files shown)


Python/Product/VSCode/AnalysisVsc/LanguageServer.cs
```diff
@@ -354,6 +355,8 @@ public Task<CompletionItem> CompletionItemResolve(JToken token)
         [JsonRpcMethod("textDocument/documentSymbol")]
         public async Task<SymbolInformation[]> DocumentSymbol(JToken token, CancellationToken cancellationToken) {
             await _prioritizer.DefaultPriorityAsync(cancellationToken);
+            // This call is also used by VSC document outline and it needs correct information
+            await WaitForCompleteAnalysisAsync(cancellationToken);
```
AlexanderSher: Both `_server.WorkspaceSymbols` and `_server.DocumentSymbol` call `entry.GetAnalysisAsync` through `GetModuleVariablesAsync` with 50ms timeout. Shouldn't it work by increasing the timeout?

The problem with `_server.WaitForCompleteAnalysisAsync` is that it isn't reliable. It waits sequentially for three queues to become empty but doesn't stop them from adding new elements, so while it wait for `AnalysisQueue` to become empty, `ParseQueue` may get new item that will clean analysis data, which will result in a bug similar to #4560.

PR #4656, opened 2018-08-16
Reduce noise in LS output

MikhailArkhipov:
Address user feedback per https://github.com/Microsoft/vscode-python/issues/2405
- Reduce noise in the output window. 
- Related setting was added to VSC extension.
- Default level is now 'error'.
- Added 'general' level for messages that we always want to be displayed.

### DIFF
--- a/Python/Product/Analysis/Infrastructure/StringExtensions.cs
+++ b/Python/Product/Analysis/Infrastructure/StringExtensions.cs
@@ -151,6 +151,6 @@ namespace Microsoft.PythonTools.Analysis.Infrastructure {
         }
 
         public static bool EqualsIgnoreCase(this string s, string other)
-            => string.Compare(s, other, StringComparison.OrdinalIgnoreCase) == 0;
+            => string.Equals(s, other, StringComparison.OrdinalIgnoreCase);
     }
 }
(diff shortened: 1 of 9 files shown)


Python/Product/Analysis/LanguageServer/Enums.cs
```diff
@@ -71,9 +71,30 @@ public enum TextDocumentSyncKind : int {
     }
 
     public enum MessageType : int {
+        /// <summary>
+        /// General language server output relevant to the user
+        /// such as information on Python interpreter type.
+        /// </summary>
+        General = 0,
```
zooba: This value isn't defined in LSP, so should start with an underscore (for consistency with other unspecified values)

Python/Product/Analysis/Infrastructure/StringExtensions.cs
```diff
@@ -149,5 +149,8 @@ static class StringExtensions {
         public static int IndexOfOrdinal(this string s, string value, int startIndex = 0, bool ignoreCase = false) {
             return s?.IndexOf(value, startIndex, ignoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal) ?? -1;
         }
+
+        public static bool EqualsIgnoreCase(this string s, string other)
+            => string.Compare(s, other, StringComparison.OrdinalIgnoreCase) == 0;
```
AlexanderSher: `string.Equals(s, other, StringComparison.OrdinalIgnoreCase)`

PR #4657, opened 2018-08-16
Fixes for debug-attach tests

zooba:
Potentially fixes tests expecting 16.0 compiler.
Fixes debug messages not marshalling to correct thread.
Improves logging for debug tests.

### DIFF
--- a/Python/Product/Debugger/Debugger/DebugConnection.cs
+++ b/Python/Product/Debugger/Debugger/DebugConnection.cs
@@ -21,6 +21,7 @@ using System.IO;
 using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.PythonTools.Common.Infrastructure;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.Ipc.Json;
 using LDP = Microsoft.PythonTools.Debugger.LegacyDebuggerProtocol;
@@ -34,7 +35,7 @@ namespace Microsoft.PythonTools.Debugger {
         private TextWriter _debugLog;
         private Connection _connection;
         private Thread _eventThread;
-        private Thread _debuggerThread;
+        private AsyncThread _debuggerThread;
         private readonly object _isListeningLock = new object();
         private readonly object _eventHandlingLock = new object();
         private readonly ConcurrentQueue<EventReceivedEventArgs> _eventsPending = new ConcurrentQueue<EventReceivedEventArgs>();
@@ -80,7 +81,7 @@ namespace Microsoft.PythonTools.Debugger {
             // Process events in a separate thread from the one that is processing messages
             // so that event handling code that needs access to the UI thread don't end up racing with other
             // code on the UI thread which may be waiting for a response to a request.
-            _debugLog.WriteLine("PythonProcess enqueuing event: " + e.Name);
+            _debugLog.WriteLine("[{0:s}] PythonProcess enqueuing event: {1}".FormatUI(DateTime.Now, e.Name));
             _eventsPending.Enqueue(e);
             _eventsPendingWakeUp.Set();
         }
@@ -93,7 +94,7 @@ namespace Microsoft.PythonTools.Debugger {
             _eventThread.Name = "Python Debugger Event Handling " + _processGuid;
             _eventThread.Start();
 
-            _debuggerThread = new Thread(MessageProcessingThread);
+            _debuggerThread = new AsyncThread(MessageProcessingThreadAsync);
             _debuggerThread.Name = "Python Debugger Message Processing " + _processGuid;
             _debuggerThread.Start();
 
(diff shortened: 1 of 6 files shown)


Python/Product/Debugger/Debugger/DebugConnection.cs
```diff
@@ -93,7 +94,7 @@ class DebugConnection : IDisposable {
             _eventThread.Name = "Python Debugger Event Handling " + _processGuid;
             _eventThread.Start();
 
-            _debuggerThread = new Thread(MessageProcessingThread);
+            _debuggerThread = new AsyncThread(MessageProcessingThreadAsync);
```
AlexanderSher: I'm not sure I understand why do you need the thread at all. Before the change, code execution would've jumped into the thread pool and never returned back. `_debuggerThread.Join` was pretty much similar to `Task.Wait`. Did we have any bugs because of that?

Python/Tests/DebuggerTests/AttachTests.cs
```diff
@@ -453,18 +453,26 @@ import sys
             try {
                 using (var dumpWriter = new MiniDumpWriter(p)) {
                     Thread.Sleep(1500);
+                    Assert.IsFalse(p.HasExited);
 
-                    var attached = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
-                    var bpHit = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+                    var attached = new TaskCompletionSource<bool>();
```
AlexanderSher: We have a thing called `EventTaskSource`. It is in tests now, but can be moved here. Its purpose is to safely await for the event to fire once after subscription with continuation on the thread the event has fired.

Python/Product/Debugger/Debugger/DebugConnection.cs
zooba: That's what it was doing, and it seemed to be deadlocking in the test (which is also jumping into the thread pool).

Having a dedicated thread to come back to after each real await is the point of the thread. This makes sure it actually works.

Python/Tests/DebuggerTests/AttachTests.cs
zooba: Looks like it needs to be extended to also allow cancellation by a different event. Can't guarantee I'll get to that extra work anytime soon.

AlexanderSher: Cancellation? Meaning that if one event happens before another test should fail?

zooba: That's a good point - we can just fail the test on the other event :)

But I tried switching this over to the EventTaskSource (ignoring the other event) and the deadlock came back...

Python/Product/Debugger/Debugger/DebugConnection.cs
AlexanderSher: Jumping into thread pool can't cause a deadlock by itself, while binding task continuations to a single thread does it with guarantee for any task that is synchronously waited on that same thread. Conversion from free-threaded to a single-threaded is a major breaking change.

What test is got blocked and what are the deadlock callstacks?

zooba: So it's the debugger message processing thread that gets blocked, and I'm pretty sure since we're using slim concurrency objects instead of kernel level ones that the thread pool can try and schedule a task on an existing thread without realizing that it's blocked. 

Ensuring that debugger messages are all processed on the same thread doesn't generally change concurrency - it actually fixes the implementation to match the design, ensuring that messages _always_ have a thread to come back to that is only ever blocked by another message or by forcing a task to the UI thread. It's a single long-running task, not a transient free-threaded threadpool type task, and keeping it on its own thread is meant to isolate it from whatever other threadpool threads may be doing, thereby simplifying concurrency rather than complicating it.

zooba: The tests are the AttachNewThread tests that wait for a breakpoint to be hit. While they're waiting for the breakpoint, no messages are being processed on the debugger thread (without this fix). Then when it stops waiting, all the messages are processed at once. Adding the timestamps and setting breakpoints in the message handling loop (and seeing which thread it kept moving to - not the one it's supposed to stay on!) is what told me the problem.

AlexanderSher: Let me dig into that a bit. Is it only reproducible on VS16?

zooba: I see it on VS15 as well. Make sure you add the timestamps to the log messages (as in this PR).

MinKPark:
@zooba is this PR still relevant? I wonder since it has been stalled for a long while.

huguesv:
Closing this, since these issues are with the legacy debugger which we are no longer spending time on at this point.

PR #4661, opened 2018-08-17
Support hierarchical document symbols in LSP

MikhailArkhipov:
Per LSP 4.4+, VS Code 1.26+
https://github.com/Microsoft/vscode-python/issues/2384

Questions: 
- Do we need old DocumentSymbol method? VS does not appear to use it.
- Just change IServer or keep method and make IServer2?
- Better way to establish child-parent relations?

### DIFF
--- a/Python/Product/Analysis/LanguageServer/IServer.cs
+++ b/Python/Product/Analysis/LanguageServer/IServer.cs
@@ -79,4 +79,8 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
 
         Task ReloadModulesAsync(CancellationToken token);
     }
+
+    public interface IServer2 : IServer {
+        Task<DocumentSymbol[]> HierarchicalDocumentSymbol(DocumentSymbolParams @params, CancellationToken cancellationToken);
+    }
 }
--- a/Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
@@ -21,6 +21,7 @@ using System.Threading.Tasks;
 using Microsoft.PythonTools.Analysis.Analyzer;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
+using Microsoft.PythonTools.Parsing.Ast;
 
 namespace Microsoft.PythonTools.Analysis.LanguageServer {
     public sealed partial class Server {
@@ -81,7 +82,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
 
         private static IEnumerable<MemberResult> GetChildScopesVariables(ModuleAnalysis analysis, InterpreterScope scope, GetMemberOptions opts, int currentDepth)
             => currentDepth < _symbolHierarchyDepthLimit
-                ? scope.Children.SelectMany(c => GetScopeVariables(analysis, c, opts, currentDepth)) 
+                ? scope.Children.SelectMany(c => GetScopeVariables(analysis, c, opts, currentDepth))
                 : Enumerable.Empty<MemberResult>();
 
         private static IEnumerable<MemberResult> GetScopeVariables(ModuleAnalysis analysis, InterpreterScope scope, GetMemberOptions opts, int currentDepth)
@@ -138,7 +139,8 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
                 name = m.Name,
                 detail = m.Name,
                 kind = ToSymbolKind(m.MemberType),
-                deprecated = false
+                deprecated = false,
+                _functionKind = GetFunctionKind(m)
             };
 
             if (childMap.TryGetValue(m, out var children)) {
@@ -156,6 +158,23 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
                 res.selectionRange = res.range;
             }
             return res;
+        }
+
+        private static string GetFunctionKind(MemberResult m) {
+            if (m.MemberType == PythonMemberType.Function) {
+                var funcDef = m.Values.FirstOrDefault(x => x is IPythonFunction)?.PythonType as FunctionDefinition;
+                if (funcDef?.Decorators != null && funcDef.Decorators.DecoratorsInternal.Length == 1) {
+                    foreach (var decorator in funcDef.Decorators.DecoratorsInternal) {
+                        if (decorator is NameExpression nameExpr) {
+                            if (nameExpr.Name == "property" || nameExpr.Name == "staticmethod" || nameExpr.Name == "classmethod") {
+                                return nameExpr.Name;
+                            }
+                        }
+                    }
+                }
+                return "function";
+            }
+            return m.MemberType == PythonMemberType.Class ? "class" : string.Empty;
         }
 
         private static SymbolKind ToSymbolKind(PythonMemberType memberType) {
--- a/Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
+++ b/Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
@@ -51,7 +51,7 @@ namespace Microsoft.PythonTools.Analysis.LanguageServer {
                 .ToArray();
         }
 
-        public async Task<DocumentSymbol[]> NewDocumentSymbol(DocumentSymbolParams @params, CancellationToken cancellationToken) {
+        public async Task<DocumentSymbol[]> HierarchyDocumentSymbol(DocumentSymbolParams @params, CancellationToken cancellationToken) {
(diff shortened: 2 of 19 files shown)


zooba:
`OutOfProcProjectAnalyzer.GetNavigations()` should be reimplemented in terms of `DocumentSymbol`, but currently it's not using it.

It might be nice to add a recursion limit parameter here, firstly because it is probably possible to get an infinitely recursive structure (from analysis), but also just to save time/space when returning the symbols. VS is unlikely to request more than two levels deep, for example.

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -96,11 +105,61 @@ private static IEnumerable<MemberResult> GetScopeVariables(ModuleAnalysis analys
...
+            }
+            return res;
+        }
+
+        private static SymbolKind ToSymbolKind(MemberResult m) {
+            switch (m.MemberType) {
                 case PythonMemberType.Unknown: return SymbolKind.None;
-                case PythonMemberType.Class: return SymbolKind.Class;
-                case PythonMemberType.Instance: return SymbolKind.Object;
+                case PythonMemberType.Class: return m.Scope is FunctionScope ? SymbolKind.Variable : SymbolKind.Class;
```
zooba: This should be a constant mapping. If nested classes have an incorrect member type, that is probably its own bug.

Alternatively, handle this extra check earlier, so that Scope doesn't have to be passed around here.

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -50,6 +51,14 @@ await GetModuleVariablesAsync(entry as ProjectEntry, opts, @params.query, 50)
                 .ToArray();
         }
 
+        public async Task<DocumentSymbol[]> NewDocumentSymbol(DocumentSymbolParams @params, CancellationToken cancellationToken) {
```
zooba: Public?

zooba: Ah, I see, this is the newer one. Personally I'd prefer a suffix `2` or `3_10` (for the LSP version), or a descriptive prefix (`HierarchyDocumentSymbol`?)

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: It a minor nit. In class method first argument comes as class although technically it is an argument in the signature. It looks a bit odd in the outline as a class inside a function. Or maybe its OK?

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: Sure

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
zooba: Well, you'd certainly want it to look like a class in this case, so perhaps the answer is to explicitly handle arguments differently?

```python
def make_type():
    class MyClass:
        pass
    return MyClass
```

MikhailArkhipov: OK

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -96,11 +109,79 @@ private static IEnumerable<MemberResult> GetScopeVariables(ModuleAnalysis analys
...
+                        new SourceLocation(loc.EndLine ?? loc.StartLine, loc.EndColumn ?? loc.StartColumn)
+                    );
+                res.selectionRange = res.range;
+            }
+            return res;
+        }
+
+        private static string GetFunctionKind(MemberResult m) {
+            if (m.MemberType == PythonMemberType.Function) {
+                var funcDef = m.Values.FirstOrDefault(x => x is IPythonFunction)?.PythonType as FunctionDefinition;
```
zooba: This always results in `null` - you will want `m.Values.OfType<FunctionInfo>()` and then check their `.IsStatic`/`.IsClassMethod`/`.IsProperty` member.

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -96,11 +109,79 @@ private static IEnumerable<MemberResult> GetScopeVariables(ModuleAnalysis analys
             return res;
         }
 
+        private DocumentSymbol[] ToDocumentSymbols(List<MemberResult> members) {
+            var childMap = new Dictionary<MemberResult, List<MemberResult>>();
+            var topLevel = new List<MemberResult>();
+
+            foreach (var m in members) {
+                var parent = members.FirstOrDefault(x => x.Scope?.Node == m.Scope?.OuterScope?.Node && x.Name == m.Scope?.Name);
```
zooba: We should be able to avoid this check by more carefully collecting values from scopes, yes? And save adding scope references to the MemberResult struct?

zooba: Or arguably we can extract _nearly_ enough information from an AST walk (the names at least), and then use a direct analysis lookup to resolve those that we don't know for sure (like we do in the analysis classifier).

Python/Product/Analysis/LanguageServer/IServer.cs
```diff
@@ -79,4 +79,8 @@ public interface IServer {
 
         Task ReloadModulesAsync(CancellationToken token);
     }
+
+    public interface IServer2 : IServer {
+        Task<DocumentSymbol[]> HierarchicalDocumentSymbol(DocumentSymbolParams @params, CancellationToken cancellationToken);
+    }
```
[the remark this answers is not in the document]
MikhailArkhipov: Extension (IC) will be using it most probably since they need symbols with types.

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: OK

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: Yes, I am actually going to write document symbols fetch using AST, but that is going to be September. https://github.com/Microsoft/vscode-python/issues/2430

zooba: MemberResult is a public API, so adding/removing a field will be a breaking change.

Python/Product/Analysis/LanguageServer/IServer.cs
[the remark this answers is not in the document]
zooba: Do they need to use it via the interface? Or are they downcasting to the Server class anyway?

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: We'll need to revisit public surface going forward, add more interfaces since it gets harder to make changes as we go.

Python/Product/Analysis/LanguageServer/IServer.cs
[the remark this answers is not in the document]
MikhailArkhipov: They use `IServer` except there is some odd code that casts it to `Server` to see it is actual Python.

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
MikhailArkhipov: Can make `MemberResult2` deriving from the base technically

Python/Product/Analysis/LanguageServer/Server.WorkspaceSymbols.cs
```diff
@@ -50,6 +54,14 @@ await GetModuleVariablesAsync(entry as ProjectEntry, opts, @params.query, 50)
                 .ToArray();
         }
 
+        public override async Task<DocumentSymbol[]> HierarchicalDocumentSymbol(DocumentSymbolParams @params, CancellationToken cancellationToken) {
+            var opts = GetMemberOptions.ExcludeBuiltins | GetMemberOptions.DeclaredOnly;
+            var entry = ProjectFiles.GetEntry(@params.textDocument);
+
+            var members = await GetModuleVariablesAsync(entry as ProjectEntry, opts, string.Empty, 50);
```
AlexanderSher: `cancellationToken`

PR #4662, opened 2018-08-17
Fixes #4144 No completion for custom tags and filters in Django projects

### LINKED ISSUE #4144: No completion for custom tags and filters in Django projects
1. Create new Django Web Project
1. In the `app` folder, create new `templatetags` folder
1. Create new module `custom.py` with the following code:
```python
from django import template

register = template.Library()

@register.simple_tag
def get_title():
    return 'Getting started title'

@register.filter(name='get_num_color')
def get_num_color(value):
    if value < 1:
        return "#FF0000"
    elif value <= 3 :
        return "#FF7400"
    else:
        return "#00CC00"
```
4. Add `{% load custom %}` to the beginning of `app\templates\app\index.html`
5. Try get custom tag or filter in html:
image

zooba:
Fixes #4144 No completion for custom tags and filters in Django projects
Fixes specializations for Django 1.9 and later
Improves tests

### DIFF
--- a/Python/Product/Django.Analysis/DjangoAnalyzer.cs
+++ b/Python/Product/Django.Analysis/DjangoAnalyzer.cs
@@ -198,16 +198,21 @@ namespace Microsoft.PythonTools.Django.Analysis {
             analyzer.SpecializeFunction("django.contrib.gis.shortcuts", "render_to_kmz", RenderToStringProcessor, true);
             analyzer.SpecializeFunction("django.contrib.gis.shortcuts", "render_to_text", RenderToStringProcessor, true);
 
-            analyzer.SpecializeFunction("django.template.base.Library", "filter", FilterProcessor, true);
-            analyzer.SpecializeFunction("django.template.base.Library", "filter_function", FilterProcessor, true);
+            analyzer.SpecializeFunction("django.template.Library", "filter", FilterProcessor, true);
+            analyzer.SpecializeFunction("django.template.Library", "filter_function", FilterProcessor, true);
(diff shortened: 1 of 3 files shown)


Python/Product/Django.Analysis/DjangoAnalyzer.cs
```diff
@@ -198,17 +198,22 @@ internal static class Commands {
...
+            analyzer.SpecializeFunction("django.template.Library", "filter_function", FilterProcessor, true);
 
-            analyzer.SpecializeFunction("django.template.base.Library", "tag", TagProcessor, true);
-            analyzer.SpecializeFunction("django.template.base.Library", "tag_function", TagProcessor, true);
-            analyzer.SpecializeFunction("django.template.base.Library", "assignment_tag", TagProcessor, true);
-            analyzer.SpecializeFunction("django.template.base.Library", "simple_tag", TagProcessor, true);
+            analyzer.SpecializeFunction("django.template.Library", "tag", TagProcessor, true);
+            analyzer.SpecializeFunction("django.template.Library", "tag_function", TagProcessor, true);
+            analyzer.SpecializeFunction("django.template.Library", "assignment_tag", TagProcessor, true);
+            analyzer.SpecializeFunction("django.template.Library", "simple_tag", TagProcessor, true);
```
zooba: This works for both because when `django.template` is imported, it'll import `Library` from the correct module and then overwrite the member. So even if you import it directly from `django.template.library.Library`, you still get the specialization.

PR #4677, opened 2018-08-22
Merge with trunk (external profiler driver disabled by default)

rgesteve:


Python/Product/Profiling/Profiling.csproj
```diff
@@ -35,12 +35,22 @@
     <AssemblyName>Microsoft.PythonTools.Profiling</AssemblyName>
     <UseVSSDK>true</UseVSSDK>
     <CreateVsixContainer>true</CreateVsixContainer>
+    <DefineConstants Condition="$(IncludeExternalProfilerDriver) == 'true'">EXTERNAL_PROFILER_DRIVER;$(DefineConstants)</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Platform)' == 'x86' ">
     <PlatformTarget>x86</PlatformTarget>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="DiagnosticsHub.Packaging.Interop, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
```
huguesv: Is HintPath necessary? If so, then this won't work. If there's no NuGet package, we may have to check in the dlls into source control.

Python/Product/Profiling/Profiling/LaunchProfiling.xaml
```diff
@@ -152,6 +152,14 @@
                 </Grid>
             </GroupBox>
         </Grid>
+
+        <Grid Grid.Row="1">
+            <CheckBox Grid.Row="1" Margin="12 12 12 4"
+                      Content="{x:Static ui:Strings.LaunchProfiling_MixedModeProfiling}"
+                      IsEnabled="{Binding IsVTuneAvailable}"
```
huguesv: `IsVTuneAvailable` may not be compiled in. I think you should keep it always compiled in.

We should set the `Visibility` attribute so that if support is not compiled in, the control is Hidden+Collapsed.

There's examples elsewhere in PTVS sources that show how to set Visibility binding to a bool property.

Python/Product/Profiling/Profiling/LaunchProfiling.xaml.cs
```diff
@@ -82,6 +85,7 @@ public partial class LaunchProfiling : DialogWindowVersioningWorkaround {
         private void CancelClick(object sender, RoutedEventArgs e) {
             this.DialogResult = false;
             Close();
+            // ProfiledProcess.PackageTrace();
```
huguesv: remove?

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -303,9 +316,75 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+#endif
+                RunProfiler(session, config, openReport);
+#if EXTERNAL_PROFILER_DRIVER
+            }
+#endif
         }
 
+#if EXTERNAL_PROFILER_DRIVER
+        private static void RunVTune(SessionNode session, LaunchConfiguration config, bool openReport) {
+#if false
```
huguesv: remove?

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -303,9 +316,75 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+                    } else {
+                        MessageBox.Show($"The Python interpreter in question is: [{interpreter}]");
+                    }
+                }
+            }
+#endif
+
+            var interpreter = config.GetInterpreterPath();
+            if (!File.Exists(interpreter)) {
+                MessageBox.Show("Can't find specified python interpreter.");
```
huguesv: Add to Strings.resx, and include the path in the message. Second param to Show should be Strings.ProductTitle.

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -303,9 +316,75 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+                return;
+            }
+
+            string outPathDir = Path.GetTempPath();
```
huguesv: Every running instance of VS will stomp on each other, because GetTempPath() always returns the same folder. You'll need to create an instance specific subfolder under outPathDir to avoid collision.

ex: everyone is writing to `c:\users\huvalo\apdata\local\temp\sample.dwjson`

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -303,9 +316,75 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+                }),
+                WorkingDirectory = config.WorkingDirectory,
+            };
+
+            var proc = new Process { StartInfo = procInfo };
+            var dte = (EnvDTE.DTE)session._serviceProvider.GetService(typeof(EnvDTE.DTE));
+            proc.EnableRaisingEvents = true;
+            proc.Exited += (_, args) => {
+                if (!File.Exists(Path.Combine(outPathDir, "Sample.dwjson"))) {
+                    MessageBox.Show($"Something happened, cannot find output file");
```
huguesv: Message should be moved to Strings.resx, and probably include the path to the file that couldn't be found. Don't forget Strings.ProductTitle too.

Python/Product/Profiling/PythonProfilingPackage.cs
rgesteve: Would you suggest of a way to get the VS instance id?  Or a reasonable proxy?  Thanks!

Python/Product/Profiling/Profiling/LaunchProfiling.xaml
rgesteve: Fixed in a57dc046b , perhaps a little artlessly.  This is because there's a sequence of decisions more than a single boolean (whether external profiler driver is compiled, if so, whether vtune is installed, and if so, whether the user wants to make use of it)

Python/Product/Profiling/Profiling.csproj
rgesteve: Looks like `HintPath` is not needed, at least not in 15.8/Community, where I tested.  The removed property is in d618c1e0f.

Python/Product/ExternalProfilerDriver/VTuneUtils.cs
```diff
@@ -0,0 +1,79 @@
...
+// permissions and limitations under the License.
+
+using System;
+using System.IO;
+using System.Collections.Generic;
+using System.Linq;
+using System.Diagnostics;
+
+namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver {
+    public static class Utils
```
huguesv: wrong brace formatting in this file

Python/Product/Profiling/Profiling/LaunchProfiling.xaml
```diff
@@ -152,6 +166,24 @@
...
+<!--
+        <Grid Grid.Row="1">
+            <CheckBox Grid.Row="1" Margin="12 12 12 4"
+                      Content="{x:Static ui:Strings.LaunchProfiling_MixedModeProfiling}"
+                      Visibility="{Binding IsExternalProfilerEnabled}"
+                      IsEnabled="{Binding IsVTuneAvailable}"
+                      IsChecked="{Binding UseVTune}" />
+        </Grid>
+        -->
+        <Grid Grid.Row="1">
```
huguesv: That looks good to me. Please remove the commented out version.

Python/Product/Profiling/Profiling/LaunchProfiling.xaml
```diff
@@ -16,7 +16,21 @@
     TextOptions.TextFormattingMode="Display"
     ResizeMode="CanResizeWithGrip">
     <Window.Resources>
+    <!--
```
huguesv: Please remove this commented out code.

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -70,6 +72,9 @@ internal sealed class PythonProfilingPackage : AsyncPackage {
         internal static readonly string PerformanceFileFilter = Strings.PerformanceReportFilesFilter;
         private AutomationProfiling _profilingAutomation;
         private static OleMenuCommand _stopCommand, _startCommand;
+#if EXTERNAL_PROFILER_DRIVER
+        private const string ExternalProfilerDriverExe = "ExternalProfilerDriver.exe";
```
huguesv: Were there any discussions on the naming of this executable? @zooba
It has a generic name but its implementation is specific to VTune.
It's also not using the "Microsoft.PythonTools" like we do for our analyzer executable. I understand not using the prefix for executables that the user interacts with directly (less typing).

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -303,9 +316,59 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
+#if EXTERNAL_PROFILER_DRIVER
+        private static void RunVTune(SessionNode session, LaunchConfiguration config, bool openReport) {
+
+            var interpreter = config.GetInterpreterPath();
+            if (!File.Exists(interpreter)) {
+                MessageBox.Show(Strings.CannotFindPythonInterpreter, Strings.ProductTitle);
+                return;
+            }
+
+            string outPathDir = Path.GetTempPath();
```
huguesv: This `outPathDir` needs it own unique subfolder. You use it below `Path.Combine(outPathDir, "Sample.dwjson")`, so that has to be unique per run.

I see you've made changes elsewhere to generate datetime based subfolder, I don't know if those were necessary or not. Also note that using Path.GetRandomFileName may be simpler than all the date formatting, unless you find that useful to track down the files in the temp folder.
Example from TestData.CalculateTempRoot
```
                path = Path.GetTempPath();
                var subpath = Path.Combine(path, Path.GetRandomFileName());
                while (Directory.Exists(subpath) || File.Exists(subpath)) {
                    subpath = Path.Combine(path, Path.GetRandomFileName());
                }
                path = subpath;
```

Python/Product/Profiling/Profiling/LaunchProfiling.xaml
rgesteve: So sorry about this, please find it corrected in c8561d933

Python/Product/ExternalProfilerDriver/VTuneUtils.cs
rgesteve: Corrected in 1d87a5d0a

Python/Product/Profiling/PythonProfilingPackage.cs
huguesv: I'm thinking Microsoft.PythonTools.Profiling.VTuneDriver.exe

rgesteve: Works for me, I'll make the change

Python/Product/ExternalProfilerDriver/DWJsonModel.cs
```diff
@@ -0,0 +1,179 @@
...
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.PythonTools.Profiling.ExternalProfilerDriver {
+
+    // Most numerical integer data is represented in dwjson as a 64-bit unsigned integer
+    // Due to JSON limitations, this is represented as a 2D named tuple, where the fields are
+    // "l" (low) and "h" (high), each containing a 32-bit *signed* integer.
+    // In this way, the max value (0xffffffffffffffff) is represented as {"h" : -1, "l" : -1}
+    public class LongInt {
```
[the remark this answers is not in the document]
rgesteve: @huguesv , I've started work on string localization in 105bb6e79, please take a look?  In particular, I'm not sure if the use of string interpolation is appropriate, please let me know.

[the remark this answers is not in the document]
rgesteve: Localized all strings I could identify, work in ffccbe713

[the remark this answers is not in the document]
huguesv: Strings should not be concatenated in code, because the concatenation order may differ based on language.
We usually have the localizable string contain the format parameter.
```
<data name="ProcessRunnerInferiorOutputPrefix" xml:space="preserve">
  <value>From process: {0}</value>
</data>
```
Then:
```
Console.WriteLine(string.Format(CultureInfo.CurrentCulture, Strings.ProcessRunnerInferiorOutputPrefix, e.Message));
```

`CultureInfo.CurrentCulture` is needed for the call to `Format`, otherwise we'll have static analysis errors.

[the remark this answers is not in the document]
huguesv: BTW I know this is a pain, I had to extract all the strings from code for the whole product over a year ago...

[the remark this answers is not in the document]
rgesteve: Oh, I don't question the value (and begrudge the time) it's just that I really *really* would like to see this merged this week. :)  I thank you for the effort of going through the strings in the rest of the codebase, by the way, *that* must've been quite a chore.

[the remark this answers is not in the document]
rgesteve: Replaced all string interpolation with string.format as recommended in 13a67afc0

Python/Product/Profiling/Profiling.csproj
rgesteve: Can't find the relevant comment, but aa835d2d6 now removes the `public` qualifier from the types in `ExternalProfilerDriver` when not necessary.

Python/Product/Profiling/PythonProfilingPackage.cs
rgesteve: Actually, if we could postpone this change 'til after the merge, it'd be very useful.  Changing the name of the executable involves changes in the signing and skip validation infrastructures that I'm not sure I can make on my own.  I've added a comment in 50e290aa5 so that this change does not fall through the cracks.

rgesteve: Also, the idea is that ExternalProfilerDriver becomes a component that knows how to deal with more than one profiler, parses its input and creates diagsessions.  VTune is just the first one.  That was the motivation for the name.  I'm still confident that we can get teach it of at least one more profiler.

huguesv: Agreed to postpone.

PR #4690, opened 2018-08-24
Import standard modules without __init__.py

MikhailArkhipov:
Fixes https://github.com/Microsoft/vscode-python/issues/2319

Basically pretend empty `__init__.py` exist if higher than 3.3

image


Python/Product/Analysis/Interpreter/Ast/AstPythonModule.cs
```diff
@@ -41,7 +41,9 @@ PythonLanguageVersion langVersion
             PythonLanguageVersion langVersion,
             string moduleFullName
         ) {
-            using (var stream = new FileStream(sourceFile, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)) {
+            using(var stream = Directory.Exists(sourceFile) 
+                ? new MemoryStream()// Module without __init__.py, create empty stream
```
AlexanderSher: So this is basically null-stream, right?

MikhailArkhipov: Yes, like empty file signaling that this folder is a module.

PR #4691, opened 2018-08-24
Fix build

MikhailArkhipov:
Apparently I didn't push the latest


AlexanderSher:
Is it just formatting difference?

MikhailArkhipov:
No, error was that MemoryStream and FileStream cannot be used in the same `? :` expression as they are different types.

PR #4702, opened 2018-09-05
Remove most of the internal dependencies on the Analysis and Analyzer

AlexanderSher:


Python/Tests/IronPython/IronPythonAnalysisTest.cs
```diff
@@ -490,24 +490,24 @@ import wpf
...
-                        p.Complete();
-                    }
-                }
-
-                pyEntry.Analyze(CancellationToken.None);
-            }
+            //using (var analyzer = CreateAnalyzer()) {
+            //    string xamlPath = TestData.GetPath(@"TestData\Xaml\EmptyXName.xaml");
+            //    string pyPath = TestData.GetPath(@"TestData\Xaml\EmptyXName.py");
+            //    var xamlEntry = analyzer.Analyzer.AddXamlFile(xamlPath);
```
AlexanderSher: This test was commented out because of the `AddXamlFile` method. It should be actually removed from `Analysis` with the rest of the `IronPython`-specific code, but it requires some changes in public interfaces. This doesn't block separation by itself though, cause it is called only from Analyzer.

Python/Product/TestAdapter/QualifiedFunctionNameWalker.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// class A:
+    ///   def b(self):
+    ///     def c():
+    ///       class D:
+    ///         def e(self):
+    ///           pass
+    /// </code>
+    /// And with the current statement being <c>pass</c>, the qualified name is "D.e in c in A.b".
+    /// </example>
+    class QualifiedFunctionNameWalker : PythonWalker {
```
AlexanderSher: This is actually a rename, GitHub just doesn't show it properly.

Python/Product/Analysis/Interpreter/Ast/AstPythonInterpreterFactory.cs
```diff
@@ -29,7 +29,7 @@
 using Microsoft.PythonTools.Parsing;
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
-    class AstPythonInterpreterFactory : IPythonInterpreterFactory, IPythonInterpreterFactoryWithLog, ICustomInterpreterSerialization, IDisposable {
+    public class AstPythonInterpreterFactory : IPythonInterpreterFactory, IPythonInterpreterFactoryWithLog, ICustomInterpreterSerialization, IDisposable {
```
AlexanderSher: This has to be public for `IronPythonAstInterpreterFactory` to be derived from it.

Python/Product/Analysis/Infrastructure/UriEqualityComparer.cs
```diff
@@ -18,7 +18,7 @@
 using System.Collections.Generic;
 
 namespace Microsoft.PythonTools.Analysis.Infrastructure {
-    sealed class UriEqualityComparer : IEqualityComparer<Uri> {
+    public sealed class UriEqualityComparer : IEqualityComparer<Uri> {
```
AlexanderSher: Small types like `UriEqualityComparer` can be copied to PTVS-assemly instead of being public.

Python/Product/TestAdapter/QualifiedFunctionNameWalker.cs
huguesv: There are 2 copies of it now, in TestAdapter and Debugger projects. We could have the Debugger project Add as Link instead to avoid the copy. Note: Legacy debugger, which needs this, will be removed in the future

AlexanderSher: Ok

Python/Product/Analysis/Infrastructure/UriEqualityComparer.cs
huguesv: This one is only used in PythonTools project, we could move it there.

Python/Product/VSInterpreters/Interpreter/InterpreterRegistryService.cs
```diff
@@ -379,9 +379,9 @@ private sealed class InterpretersEnumerable : IEnumerable<IPythonInterpreterFact
                 return;
             }
 
-            var fallback = typeof(AnalysisOnlyInterpreterFactory);
+            var fallback = typeof(IPythonInterpreterFactory);
             assembly = fallback.Assembly.Location;
-            typeName = fallback.FullName;
+            typeName = "Microsoft.PythonTools.Interpreter.AnalysisOnlyInterpreterFactory";
```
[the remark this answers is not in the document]
AlexanderSher: There is `PythonAnalyzer.BuiltinModule` internal property

Python/Product/Analysis/Infrastructure/UriEqualityComparer.cs
AlexanderSher: Done

Python/Product/PythonTools/PythonTools/Intellisense/AnalysisVariable.cs
```diff
@@ -0,0 +1,15 @@
...
+            Location = location;
+            Type = type;
+            Version = version;
+        }
+
+        public LocationInfo Location { get; }
+        public VariableType Type { get; }
+        public int? Version { get; }
+    }
+}
```
huguesv: The original implementation also has Equals and GetHashCode. Can we be sure they are not needed?

AlexanderSher: Yes. They are never compared and never used as keys in a dictionary.

Python/Tests/IronPython/IronPythonAnalysisTest.cs
huguesv: Looks like you could change signature to return IProjectEntry instead of IXamlProjectEntry. Then it's easier to make it public and keep the test alive. I'm just worried we'll forget about this commented out test.

AlexanderSher: There are more changes required here. I need to remove IronPython bits from Analysis, cause they require separate compilation for .net framework. But I want it to be a separate PR.

PR #4716, opened 2018-09-20
Propagate return types of derived methods to their bases

### LINKED ISSUE #4715: Empty methods with @abc.abstractmethod and @abc.abstractproperty should infer type based on derived classes
Repro:

```python
import abc

class A:
    @abc.abstractmethod
    def virt():
        pass

class B(A):
    def virt():
        return 42

a = A()
b = a.virt()
```

type of b is resolved as `void`, should be `int`.

The example is artificial and sort of illegal, but it explains the concept. A better one would be when some function has known return type `A`, and on its return value `.virt()` is called.

lostmsu:
This change fixes #4715 

Rationale:
If `class A(B)`, and both have method `foo`, then any variable `v` of type `B` can potentially hold value of type `A`. Which means, that `v.foo()` can return a value of type that of return value of `A.foo` or `B.foo`.

To achieve that, when function return type is determined by `DDG`, it is also added as a potential return type for all of the function's bases (e.g. functions, that this function overrides).

In addition to that, this change also improves `DDG.LookupBaseMethods`, which previously only took into account built-in classes.

### DIFF
--- a/Python/Product/Analysis/Analyzer/DDG.cs
+++ b/Python/Product/Analysis/Analyzer/DDG.cs
@@ -463,13 +463,13 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
             return false;
         }
 
-        internal List<AnalysisValue> LookupBaseMethods(string name, IEnumerable<IAnalysisSet> bases, Node node, AnalysisUnit unit) {
+        internal List<AnalysisValue> LookupBaseMethods(string name, IEnumerable<IAnalysisSet> mro, Node node, AnalysisUnit unit) {
             var result = new List<AnalysisValue>();
-            foreach (var b in bases) {
-                foreach (var curType in b) {
-                    var klass = curType as BuiltinClassInfo;
-                    if (klass != null) {
-                        var value = klass.GetMember(node, unit, name);
+            foreach (var @class in mro.Skip(1)) {
+                foreach (var curType in @class) {
+                    bool isClass = curType is ClassInfo || curType is BuiltinClassInfo;
+                    if (isClass) {
+                        var value = curType.GetMember(node, unit, name);
                         if (value != null) {
                             result.AddRange(value);
                         }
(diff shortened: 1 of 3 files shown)


lostmsu:
A note: with this change, `AnalysisTest.SubclassFindAllRefs` starts failing:

> Assert.Fail failed. References did not match.
> 
> Actual:
> new VariableLocation(4, 14, VariableType.Reference), 
> new VariableLocation(6, 5, VariableType.Value), 
> new VariableLocation(6, 9, VariableType.Definition), 
> new VariableLocation(11, 5, VariableType.Reference), 
> new VariableLocation(11, 9, VariableType.Definition)
> 
> Expected but missing:
> new VariableLocation(11, 5, VariableType.Value)
> 
> Unexpected:
> new VariableLocation(11, 5, VariableType.Reference)

The tested code for the reference:

```python
class Base(object):
    def __init__(self):
        self.fob()
    
    def fob(self): 
        pass
    
    
class Derived(Base):
    def fob(self): 
        'x'
```

I've analyzed the failure.

This change causes a new VariableLocation to be introduced by the analysis, of type `VariableType.Reference` (see test output). The expected `VariableType.Value` is also present in the analysis, however test function `PythonAnalysis.UniquifyReferences` prefers `VariableType.Reference` to `VariableType.Value`, which results in masking the later.

I do not understand the expected behavior here, so, please, advise if the test code should be fixed, or my change is producing unexpected result.

huguesv:
Thank you for the PR!

@AlexanderSher has ported your changes in https://github.com/Microsoft/python-language-server/pull/115

We're still in the process of moving over to using it from PTVS, which is why it is still in the PTVS code base at the moment. We'll be consuming it as a nuget package.

We'll continue discussion of merging your changes in the above PR.

huguesv:
These changes have now been merged at https://github.com/Microsoft/python-language-server/pull/115

Thanks!

PR #4718, opened 2018-09-21
Some accessibility fixes for environments window, language server options

huguesv:
Fix narrator for language server option browse button (I checked the tab order, it is correct)
Fix contrast for environments window listbox selected items (it was a slightly different blue than seen in the rest of VS - toolbox, error list, etc)
Remove the help button from environments view

### DIFF
--- a/Python/Product/EnvironmentsList/ToolWindow.xaml
+++ b/Python/Product/EnvironmentsList/ToolWindow.xaml
@@ -190,12 +190,6 @@
                             Focusable="False"
                             ContentTemplate="{StaticResource EnvironmentHeader}" />
 
-            <CheckBox x:Name="HelpVisibility_Horizontal"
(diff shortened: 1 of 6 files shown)


Python/Product/EnvironmentsList/ToolWindow.xaml
```diff
@@ -206,13 +200,6 @@
                     <RowDefinition Height="*" />
                 </Grid.RowDefinitions>
 
-                <ContentControl Grid.Row="0"
```
AlexanderSher: There is only one item left in the grid. Do you need that grid at all?

huguesv: Don't need it, I can remove it.

huguesv: @AlexanderSher : I removed the grid.

PR #4719, opened 2018-09-21
Remove Server Explorer Azure debug attach command

huguesv:


DonJayamanne:
Are we dropping support for debugging python on azure?

huguesv:
No, this is only removing the convenience command for attaching by clicking on app services in server explorer. They are removing their server explorer integration.

Debugging can still be done via the standard debug attach dialog, you just need to know the URL to use.

PR #4729, opened 2018-09-26
Remove the use of smart tag interfaces from tests.

huguesv:
Also replaced usage of the words SmartTag by LightBulb.

VS is deleting the smart tag interfaces.


Common/Tests/Utilities.UI/UI/EditorWindow.cs
```diff
@@ -172,23 +172,17 @@ public EditorWindow(VisualStudioApp app, string filename, AutomationElement elem
...
         }
-#if DEV14_OR_LATER
-#pragma warning disable 0618
-#endif
-        public void StartSmartTagSessionNoSession() {
-            ShowSmartTag();
+
+        public void StartLightBulbSessionNoSession() {
+            ShowLightBulb();
             Thread.Sleep(100);
```
AlexanderSher: What is this `Thread.Sleep(100)` for?

huguesv: Don't know. This is really old.

Issue #3118, opened 2017-10-03
Bug in Pykinect function NuiImageGetColorPixelCoordinatesFromDepthPixel


TjWhale:
I was trying to get the function

NuiImageGetColorPixelCoordinatesFromDepthPixel

which is here, 

https://github.com/Microsoft/PTVS/blob/master/Python/Product/PyKinect/PyKinect/pykinect/nui/_interop.py

line 192 to work in PyKinect and it kept throwing the error "byref(x)" is not defined. I found this page about the issue

http://pytools.codeplex.com/workitem/1078

and it said to change byref(x) and byref(y) to ctypes.byref(x) and ctypes.byref(y) in _interop.py and when I did this the function worked fine so I think it might be a bug. 

If you look at line 186 in _interop.py you can see ctypes.byref(res) being used there.

Hope this is helpful :)

zooba:
That certainly looks like a bug.

We don't get to bugs in our sample libraries very often, but if someone brings a pull request then we'll happily merge it.

rv404674:
Hi @tjwhale, is this bug still open. I am a newcomer and would like to contribute.

TjWhale:
@rv404674 it looks like it is. I got it to work locally by making the change I suggested so I think if you make a pull request it should get accepted. Worth testing first of course. Good luck.

rv404674:
hey @TjWhale , i have opened a PR for the same. Can you please take a look when you get some time. Thanks. :)

huguesv:
Thanks everyone. I merged https://github.com/Microsoft/PTVS/pull/4747.

PR #4750, opened 2018-10-01
Fix NuilImageGetColorPixelCoordinatesFromDepthPixel #3118

rv404674:


huguesv:
Thanks. There was an identical PR opened a few days earlier with the same fix, which I just merged: https://github.com/Microsoft/PTVS/pull/4747

PR #4768, opened 2018-10-08
Environments update

### LINKED ISSUE #4730: Add environment dialog
New dialog that consolidates adding existing environment and creating new environments tasks into one command / dialog.

The design will allow it to be used in the context of "Open Folder", but that work will be tracked separately.

### LINKED ISSUE #4731: Create conda and virtual env information bar
When a project includes an environment.yml or requirements.txt, look at the existing environments associated with the project and show an info bar to propose creating either a conda or virtual env if we determine it is likely needed. Can be turned off.

This replaces the existing task dialog that appears on creation of a new project.

### LINKED ISSUE #4732: Install packages from requirements.txt information bar
When the project home folder has an requirements.txt, and the associated project is not the global default, look at the active environment packages, and if a package listed in requirements.txt is missing from the environment, show an info bar that proposes to install them.

huguesv:
Fix #4730 
Fix #4731 
Fix #4732

- New dialog for adding environments, which replaces the functionality found in previously separate UI (add existing env, add virtual env, add conda env, add custom env)
- Info bars for creating environments and installing packages (the latter replaces the new project wizard task dialog)
- Options for customizing conda location. There's only 2 MEF exports for conda locator, but a 3rd one will be added if/when we bundle miniconda (will be a separate PR).
- Change the Add Conda Env and Add Custom Env items in the environments window into a single Add Environment toolbar button. Also remove the Help item from that list. This fixes accessibility bugs with the environments window.

Edit: If you want to save time reviewing, you can go over ModernStyles.xaml quickly. Most of it is existing code copied from either the windows template or our controls.xaml (didn't want to reuse that one so I could make any necessary modern style specific changes without affecting the old UI).


Common/Tests/Utilities.UI/UI/RadioButton.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class RadioButton : AutomationWrapper {
+        public string Name { get; set; }
+
+        public RadioButton(AutomationElement element)
+            : base(element) {
+            Name = (string)Element.GetCurrentPropertyValue(AutomationElement.NameProperty);
+        }
+
+        public void SetSelected() {
+            Assert.IsTrue((bool)Element.GetCurrentPropertyValue(AutomationElement.IsSelectionItemPatternAvailableProperty), "Element is not a radio button");
```
AlexanderSher: `SelectionItemPattern` isn't specific to radio button

Python/Product/Common/Infrastructure/ProcessOutput.cs
```diff
@@ -716,6 +716,26 @@ sealed class ProcessOutput : IDisposable {
             return true;
         }
 
+        private async Task<int> WaitAsync() {
+            return await this;
+        }
+
+        public async Task<int> WaitAsync(CancellationToken ct, bool killProcessOnCancel) {
```
AlexanderSher: If you want to add cancellation handling for a task that you can't cancel explicitly, it is enough to use `ContinueWith`:
```cs
Foo(CancellationToken ct) => Foo().ContinueWith(t => t.GetAwaiter().GetResult(), ct);
```
And if some action is required on cancellation:
```cs
try {
    await Foo().ContinueWith(t => t.GetAwaiter().GetResult(), ct);
} catch (OperationCanceledException) when (ct.IsCancellationRequested) {
    SomeAction();
    throw;
}
```

But that isn't the case here. We have an `Exited` event, so we can either use `EventTaskSource` (it is now defined only in tests), or do the thing manually:
```cs
var tcs = new TaskCompletionSource<int>();
Exited += (o, e) => tcs.TrySetResult(0);
try {
    if (!_process.HasExited) {
        tcs.RegisterForCancellation(ct).UnregisterOnCompletion(tcs.Task);
        await tcs.Task;        
    }
    return _process.ExitCode;
} catch (OperationCanceledException) when (ct.IsCancellationRequested) {
    if (killProcessOnCancel) {
        Kill();
    }
    throw;
}
```
and `GetAwaiter` should use `WaitAsync`, not visa versa.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
```diff
@@ -0,0 +1,405 @@
...
+    <UserControl.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="pack://application:,,,/Microsoft.PythonTools;component/PythonTools/Wpf/ModernStyles.xaml" />
+                <ResourceDictionary>
+                    <wpf:Lambda x:Key="BoolToVisible">(bool b) => b ? Visibility.Visible : Visibility.Hidden</wpf:Lambda>
+                    <wpf:Lambda x:Key="BoolToHidden">(bool b) => b ? Visibility.Hidden : Visibility.Visible</wpf:Lambda>
+
+                    <DataTemplate x:Key="ProjectTemplate">
+                        <StackPanel>
```
AlexanderSher: Stack panel with one element?

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
```diff
@@ -0,0 +1,405 @@
...
+            CanExecute="BrowsePackages_CanExecute"
+            Executed="BrowsePackages_Executed" />
+    </UserControl.CommandBindings>
+    <UserControl.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="pack://application:,,,/Microsoft.PythonTools;component/PythonTools/Wpf/ModernStyles.xaml" />
+                <ResourceDictionary>
+                    <wpf:Lambda x:Key="BoolToVisible">(bool b) => b ? Visibility.Visible : Visibility.Hidden</wpf:Lambda>
+                    <wpf:Lambda x:Key="BoolToHidden">(bool b) => b ? Visibility.Hidden : Visibility.Visible</wpf:Lambda>
```
AlexanderSher: In RTVS, we had a set of predefined lambda converters: https://github.com/Microsoft/RTVS/blob/master/src/R/Wpf/Impl/Converters.cs
It helps to avoid defining these common things in every control or view

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
```diff
@@ -0,0 +1,405 @@
...
+                                                </Grid>
+                                            </DataTemplate>
+                                        </ItemsControl.ItemTemplate>
+                                    </ItemsControl>
+                                </StackPanel>
+                            </ScrollViewer>
+                        </Grid>
+                        <DataTemplate.Triggers>
+                            <DataTrigger Binding="{Binding Path=CondaPreview.IsInProgress}" Value="True">
+                                <Setter TargetName="ProgressBar" Property="UIElement.Visibility" Value="Visible"/>
```
AlexanderSher: `Property="Visibility"` should work

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
```diff
@@ -0,0 +1,405 @@
...
+        <Grid Visibility="{Binding Path=IsCondaMissing, Converter={StaticResource BoolToVisible}}">
+            <TextBlock
+                HorizontalAlignment="Center"
+                VerticalAlignment="Center"
+                Text="{x:Static common:Strings.CondaNotFoundError}"
+                TextWrapping="Wrap"/>
+        </Grid>
+        <Grid Visibility="{Binding Path=IsCondaMissing, Converter={StaticResource BoolToHidden}}">
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition Width="514*"/>
```
AlexanderSher: `514*/188*` - is it to preserve a proportion?

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentOperation.cs
```diff
@@ -0,0 +1,281 @@
...
+                    CanBeCanceled = false,
+                    ProgressText = Strings.CondaStatusCenterCreateProgressPreparing,
+                    PercentComplete = null,
+                }
+            );
+
+            var ui = new CondaUI(taskHandler, _outputWindow, _showAndActiveOutputWindow);
+            _statusBar.SetText(Strings.CondaStatusBarCreateStarted.FormatUI(_envName));
+
+            var task = Task.Run(() => CreateCondaEnvironmentAsync(ui, taskHandler, ct));
```
AlexanderSher: Shoudn't we use `taskHandler.UserCancellation` here?

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentOperation.cs
```diff
@@ -0,0 +1,281 @@
...
+                ui.OnErrorTextReceived(_condaMgr, ex.Message);
+                throw;
+            }
+        }
+
+        private async Task<IPythonInterpreterFactory> CreateFactoryAsync(CondaUI ui, ITaskHandler taskHandler, CancellationToken ct) {
+            var tcs = new TaskCompletionSource<IPythonInterpreterFactory>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            EventHandler interpretersChanged = (object sender, EventArgs e) => {
+                var first = _registry.Interpreters
```
AlexanderSher: Don't we know the `IPythonInterpreterFactoryProvider` for the factory?

Python/Product/PythonTools/PythonTools/Environments/EnvironmentViewBase.cs
```diff
@@ -0,0 +1,207 @@
...
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Data;
+using Microsoft.PythonTools.Interpreter;
+
+namespace Microsoft.PythonTools.Environments {
+    abstract class EnvironmentViewBase : DependencyObject, INotifyDataErrorInfo, IDisposable {
```
AlexanderSher: Is it a view or a data context?

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentOperation.cs
huguesv: I would like to get rid of this code, because of the risk of involved in missing the event.

We do know it's `CondaEnvironmentFactoryProvider`, but that factory refreshes after we detect a file system update and fire off a timer to refresh. Meaning, if I query the conda factory provider after creation, it might be too early, it may not have refreshed yet.

I could change it to query/retry/sleep for a certain amount of time instead. In practice the refresh should kick off within a couple of seconds max after the conda mgr function call has returned.

Python/Product/PythonTools/PythonTools/Environments/EnvironmentViewBase.cs
huguesv: The old classes were all named with the "View" suffix so that's why I kept it. If you have a suffix preference let me know.

huguesv: It's data context.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
huguesv: correct

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
huguesv: I'll remove in the other 2 files as well.

Common/Tests/Utilities.UI/UI/RadioButton.cs
huguesv: True. The assert message not wrong though.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
huguesv: Thanks. I'm tired of defining them everywhere. I'm integrating those now.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
```diff
@@ -0,0 +1,405 @@
...
+                            </ScrollViewer>
+                        </Grid>
+                        <DataTemplate.Triggers>
+                            <DataTrigger Binding="{Binding Path=CondaPreview.IsInProgress}" Value="True">
+                                <Setter TargetName="ProgressBar" Property="UIElement.Visibility" Value="Visible"/>
+                            </DataTrigger>
+                            <DataTrigger Binding="{Binding Path=CondaPreview.IsNoPackages}" Value="True">
+                                <Setter TargetName="NoPackagesMessage" Property="UIElement.Visibility" Value="Visible"/>
+                            </DataTrigger>
+                            <DataTrigger Binding="{Binding Path=CondaPreview.HasPreviewError}" Value="True">
```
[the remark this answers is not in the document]
huguesv: They can, especially now that I've copied the multi lambdas. That's the one that was less convoluted to do with triggers. I'll change them, that should allow me to get rid of the UserControl+DataTemplate that I added purely for the ability to define data triggers.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
huguesv: It did. The whole thing is gone now, switched to using converters. It's much nicer.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentOperation.cs
AlexanderSher: Ok, let me take a look on the `CondaEnvironmentFactoryProvider`

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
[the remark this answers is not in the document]
huguesv: I'm cleaning up the virtual env and existing env pages the same way.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentOperation.cs
AlexanderSher: Ok, it looks like we can add `CondaEnvironmentFactoryProvider.GetFirstFactoryAsync(CancellationToken ct)` that will return first available interpreter that is found in `DiscoverInterpreterFactories`.

Python/Product/PythonTools/PythonTools/Environments/AddEnvironmentDialog.xaml.cs
```diff
@@ -0,0 +1,273 @@
...
+            }
+
+            var sln = (IVsSolution)site.GetService(typeof(SVsSolution));
+            var projects = sln?.EnumerateLoadedPythonProjects().ToArray();
+
+            var projectViews = projects
+                .Select((projectNode) => new ProjectView(projectNode))
+                .ToArray();
+
+            var selectedProjectView = projectViews.SingleOrDefault(pv => pv.Node == project);
```
zooba: We should handle the `InvalidOperationException` from here. (It'll happen when we least expect it...)

Python/Product/PythonTools/PythonTools/Environments/AddInstalledEnvironmentView.cs
```diff
@@ -0,0 +1,54 @@
...
+            PageName = Strings.AddInstalledEnvironmentTabHeader;
+            AcceptCaption = Strings.AddInstalledEnvironmentLaunch;
+            IsAcceptShieldVisible = true;
+        }
+
+        public override Task ApplyAsync() {
+            _logger?.LogEvent(PythonLogEvent.InstallEnv, null);
+
+            var setupService = Site.GetService(typeof(SVsSetupCompositionService)) as IVsSetupCompositionService;
+            string installerPath = setupService?.InstallerPath;
```
zooba: I assume the text for this option says something like "Launch Visual Studio installer"? (I lost track of where it's defined.)

Python/Product/PythonTools/PythonTools/Environments/AddVirtualEnvironmentOperation.cs
```diff
@@ -0,0 +1,186 @@
...
+                        }
+
+                        if (_viewInEnvWindow) {
+                            await InterpreterListToolWindow.OpenAtAsync(_site, factory);
+                        }
+                    });
+                }
+            } catch (Exception ex) when (!ex.IsCriticalException()) {
+                failed = true;
+                throw;
```
zooba: You could make this `failed = false` at the end of the `try` block, and then remove the `catch` entirely.

Python/Product/PythonTools/PythonTools/Environments/ProjectView.cs
```diff
@@ -0,0 +1,43 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Linq;
+using System.Windows;
+using Microsoft.PythonTools.Project;
+
+namespace Microsoft.PythonTools.Environments {
+    class ProjectView : DependencyObject {
```
zooba: This shouldn't need to be a DependencyObject

Python/Product/PythonTools/PythonTools/Environments/UserSpecifiedCondaLocator.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+using System.ComponentModel.Composition;
+using Microsoft.PythonTools.Interpreter;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.PythonTools.Environments {
+    [Export(typeof(ICondaLocator))]
+    [ExportMetadata("Priority", 100)]
+    public class UserSpecifiedCondaLocator : ICondaLocator {
```
zooba: `sealed`

Python/Product/PythonTools/PythonToolsService.cs
```diff
@@ -100,7 +100,7 @@ public sealed class PythonToolsService : IDisposable {
             _idleManager = new IdleManager(container);
             _advancedOptions = new Lazy<AdvancedEditorOptions>(CreateAdvancedEditorOptions);
             _debuggerOptions = new Lazy<DebuggerOptions>(CreateDebuggerOptions);
-            _experimentalOptions = new Lazy<Options.ExperimentalOptions>(CreateExperimentalOptions);
```
zooba: Is this a deliberate choice to no longer have experimental options?

Python/Product/VSInterpreters/Interpreter/ICondaLocatorProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+namespace Microsoft.PythonTools.Interpreter {
+
+    public interface ICondaLocatorProvider {
```
zooba: Difference between this and `ICondaLocator`?

huguesv:
I'm done with the xaml related stuff, I pushed the changes. I will look at the rest now. Thanks @AlexanderSher  and @zooba !

AlexanderSher:
I'll take a look at the changes tomorrow.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentOperation.cs
```diff
@@ -0,0 +1,281 @@
...
+
+                try {
+                    if (useEnvFile) {
+                        if (!await _condaMgr.CreateFromEnvironmentFileAsync(
+                            _envName,
+                            _envFilePath,
+                            ui,
+                            ct
+                        )) {
+                            throw new ApplicationException(Strings.CondaStatusCenterCreateFailure);
```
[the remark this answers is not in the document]
huguesv: I had modeled it after `IPackageManager` which does a similar thing for `ExecuteAsync`, `InstallAsync`, etc.

If I change those 2 (`CreateFromEnvironmentFileAsync` and `CreateAsync`, I would want to change the other methods on `ICondaEnvironmentManager` for consistency, and I'm worried I'll break things if I rush this in. I prefer not to do this.

Python/Product/PythonTools/PythonTools/Environments/AddInstalledEnvironmentView.cs
huguesv: Yeah, we went with "Open Visual Studio Installer"

Python/Product/VSInterpreters/Interpreter/ICondaLocatorProvider.cs
huguesv: There's one instance of the provider, but multiple instances of locator.

Provider figures out which locator to use. This is based on priority, from highest to lowest:
Tools/Option/Python/Conda path > Bundled miniconda path > latest installed Anaconda/Miniconda path

You are making me realize that I should have the locator provider return a locator instead of returning the conda path of the appropriate locator. That would make it clearer. I'll make the change and add comments too.

Python/Product/PythonTools/PythonToolsService.cs
huguesv: Deliberate. We'll put future experimental options where they belong based on feature area, and put experimental suffix in the checkbox label.

huguesv: Also to work better side-by-side, we have to start using the same VS registry that the other options already use. To read/write via command line, we can use vsregedit.exe

Python/Product/Common/Infrastructure/ProcessOutput.cs
```diff
@@ -716,6 +716,26 @@ sealed class ProcessOutput : IDisposable {
             return true;
         }
 
+        private async Task<int> WaitAsync() {
+            return await this;
```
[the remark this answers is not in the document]
huguesv: I'm removing it

Python/Product/Common/Infrastructure/ProcessOutput.cs
huguesv: Thanks! I've undone my changes to `ProcessOutput` and put this code where I needed it in `CondaEnvironmentManager`.

huguesv:
I pushed another set of changes to address the remaining feedback.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentOperation.cs
```diff
@@ -0,0 +1,281 @@
...
+
+                // Wait for the factory to refresh
+                taskHandler.Progress.Report(new TaskProgressData() {
+                    CanBeCanceled = false,
+                    ProgressText = Strings.CondaStatusCenterCreateProgressDetecting,
+                    PercentComplete = 90,
+                });
+
+                var timeoutTask = Task.Delay(TimeSpan.FromSeconds(30));
+                if (await Task.WhenAny(tcs.Task, timeoutTask) == timeoutTask) {
```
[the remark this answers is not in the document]
AlexanderSher: You just need to remember that it isn't 100% universal. If continued task can actually be canceled, continuation task will be faulted instead of canceled because of this: https://github.com/dotnet/coreclr/blob/913428d5915a9729c9405f57a75e7f912f9d29a5/src/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L2505 (.net framework implementation has similar code).

This is not a problem for a code like `await task.ContinueWith(t => t.GetAwaiter().GetResult(), ct)`, because the calling async code will convert it back into cancellation, but if you want to have an extension method, you can either use `Unwrap`:
```cs
task.ContinueWith(t => t, ct, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default).Unwrap();
```
which preserves the state, but requires one more instance of `Task`, or use `TaskContinuationOptions.NotOnCanceled`:
```
task.ContinueWith(t => t.GetAwaiter().GetResult(), ct, TaskContinuationOptions.ExecuteSynchronously | TaskContinuationOptions.NotOnCanceled, TaskScheduler.Default);
```
which has a problem of swallowing `CancellationToken` from original `task`, but there are other cases when it is swallowed.

`TaskContinuationOptions.ExecuteSynchronously` is kinda desirable, cause it saves unnecessary scheduling, especially when `task` is already completed.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentView.cs
```diff
@@ -0,0 +1,445 @@
...
+                    }
+                    p.HasPreviewError = !string.IsNullOrEmpty(msg);
+                    p.ErrorMessage = msg;
+                    p.IsInProgress = false;
+                    p.IsPackages = p.Packages.Count > 0;
+                    p.IsNoPackages = !p.IsPackages && !p.HasPreviewError;
+
+                    CondaPreview = p;
+                });
+            } catch (OperationCanceledException) {
```
AlexanderSher: If you catch OCE here, calling async method will continue execution. Is that expected?

huguesv: The calling method is a timer callback, and runs this method on thread pool and never waits on it. So that is fine.

Python/Product/PythonTools/PythonTools/Environments/AddVirtualEnvironmentOperation.cs
```diff
@@ -0,0 +1,183 @@
...
+                    _customEnvName,
+                    _useVEnv
+                );
+
+                if (factory != null) {
+                    if (_installReqs && File.Exists(_reqsPath)) {
+                        await InstallPackagesAsync(taskHandler, factory);
+                    }
+
+                    await _site.GetUIThread().InvokeTask(async () => {
```
AlexanderSher: In some cases we use `GetUIThread().InvokeTask()`, and in some other it is `Dispatcher.Invoke`.

huguesv: The `DependencyObject` derived classes use dispatcher since it's provided by `DependencyObject`, but the other classes (the operation classes) use `GetUIThread`.

I figured that using the dispatcher was a 'lesser' dependency than getting the ui thread service. Is it better to use `GetUIThread` there too?

Python/Product/PythonTools/PythonTools/Environments/AddVirtualEnvironmentView.cs
```diff
@@ -0,0 +1,521 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Windows;
+using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Interpreter;
+using Microsoft.PythonTools.Project;
+
+namespace Microsoft.PythonTools.Environments {
+    sealed class AddVirtualEnvironmentView : EnvironmentViewBase, INotifyPropertyChanged {
```
AlexanderSher: It isn't a good idea to have mix of `DependencyObject` and `INotifyPropertyChanged`. I'd actually prefer to have the latter: it is easier to debug, more lightweight, better handled inside collections and not that slow in recent versions of WPF

huguesv: Turns out it's not used anymore, it was part of the old virtual env view class. I'll remove it as a base.

As far as switching everything to `INotifyPropertyChanged`, this is too risky / last minute for this upcoming release. I agree with you though, there isn't much benefit to DependencyObject, and it's a lot more line sof code.

huguesv: The SafeGetValue/SafeSetValue also appear to not be needed. All property access is done from main thread

Python/Product/PythonTools/PythonTools/Environments/AddVirtualEnvironmentOperation.cs
```diff
@@ -0,0 +1,183 @@
...
+        public async Task RunAsync() {
+            var outputWindow = OutputWindowRedirector.GetGeneral(_site);
+            var taskHandler = _statusCenter.PreRegister(
+                new TaskHandlerOptions() {
+                    ActionsAfterCompletion = CompletionActions.RetainAndNotifyOnFaulted | CompletionActions.RetainAndNotifyOnRanToCompletion,
+                    Title = Strings.VirtualEnvStatusCenterCreateTitle.FormatUI(PathUtils.GetFileOrDirectoryName(_virtualEnvPath)),
+                    DisplayTaskDetails = (t) => { outputWindow.ShowAndActivate(); }
+                },
+                new TaskProgressData() {
+                    CanBeCanceled = false,
```
AlexanderSher: So why cancellation isn't allowed here?

huguesv: We never allowed cancel of this operation in the past.
Killing the process while it's creating/installing will end up with an unusable env on disk.

We could cancel between creation and installation though, since that's 2 consecutive processes that are launched. But sometimes there's no installation step.

So giving them a cancel button that does pretty much nothing may just lead to more frustration?

I'll put it on the list to consider.

PR #4775, opened 2018-10-10
Replaced VTune report parser with a more capable one that doesn't get confused with C++ overloads/demangled names

rgesteve:
Also includes payload to list line numbers where functions were defined from pdbs.


Python/Product/ExternalProfilerDriver/VTuneToDWJSON.cs
```diff
@@ -131,7 +138,25 @@ class VTuneToDWJSON {
...
+                if (d.ContainsKey(mods[i].name)) {
+                    var unique_files = d[mods[i].name].Values.Select(fi => fi.SourceFile).Distinct();
+                    mods[i].fileIdMapping = unique_files.Zip(Enumerable.Range(0, int.MaxValue), (v,j) => new FileIDMapSpec { id = j, file = v}).ToList();
+                    Dictionary<string, int> fileTOC = mods[i].fileIdMapping.ToDictionary(x => x.file, x => x.id);
+                    var ranges = mods[i].ranges;
+                    for(int j = 0; j < ranges.Count(); j++) {
+                        if (d[mods[i].name].ContainsKey(ranges[j].name)) {
+                            var ff = d[mods[i].name][ranges[j].name];
+                            var ls = new LineSpec { fileId = fileTOC[ff.SourceFile], lineBegin = (int)ff.LineNumber };
+                            var lss = new List<LineSpec>(); lss.Add(ls);
```
huguesv: Would prefer `lss.Add(ls)` on its own line. Otherwise it's easy to miss.

Python/Product/ExternalProfilerDriver/VTuneToDWJSON.cs
```diff
@@ -209,6 +234,70 @@ class VTuneToDWJSON {
...
+        ///
+        /// TODO: maybe this function should take a list of possible directories to search?
+        /// </summary>
+        public static Dictionary<string, Dictionary<string, FunctionSourceLocation>> AddLineNumbers(ref Dictionary<string, Dictionary<string, BaseSizeTuple> > orig, string symbolPath)
+        {
+            if (!Directory.Exists(symbolPath)) {
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.VTuneSympathNotFound, symbolPath));
+            }
+
+            Dictionary<string, Dictionary<string, FunctionSourceLocation> > sourcelocs = new Dictionary<string, Dictionary<string, FunctionSourceLocation> >();
```
huguesv: `var` is great when type names are long like this

Python/Product/ExternalProfilerDriver/PdbListFunctions.cs
```diff
@@ -0,0 +1,122 @@
...
+                    if (++i >= s_msdiaGuids.Length) throw;
+                }
+            }
+        }
+
+        public IEnumerable<FunctionSourceLocation> FunctionLocations()
+        {
+            IDiaEnumSymbols results;
+
+            // findChildren(IDiaSymbol parent, SymTagEnum symTag, string name, uint compareFlags, out IDiaEnumSymbols ppResult);
```
huguesv: do you still need the comment for later?

Python/Product/ExternalProfilerDriver/VTuneStackParser.cs
```diff
@@ -21,6 +21,9 @@
 using System.Linq;
 using System.Globalization;
 using System.Text.RegularExpressions;
+using System.Text;
```
huguesv: sorting (should be one above)

Python/Product/ExternalProfilerDriver/VTuneStackParser.cs
```diff
@@ -34,7 +37,12 @@ class PerformanceSample {
 
         public PerformanceSample(string function, string cpuTime, string module, string functionFull, string sourceFile, string startAddress) {
             Function = function;
-            CPUTime = Single.Parse(cpuTime);
+            try {
+                CPUTime = Single.Parse(cpuTime);
+            } catch (Exception) { // TODO -- should use a more specific exception
```
huguesv: check https://docs.microsoft.com/en-us/dotnet/api/system.single.parse?view=netframework-4.7.2#System_Single_Parse_System_String_System_IFormatProvider_ for the exceptions to catch

You should also specify an IFormatProvider. If this comes from a file, then I hope this is culture invariant? If that's the case, then pass `CultureInfo.InvariantCulture`

Python/Product/ExternalProfilerDriver/VTuneStackParser.cs
```diff
@@ -143,4 +151,61 @@ static class VTuneStackParser {
         }
 
     }
+
+    class VTuneStackParserForCPP
+    {
```
huguesv: Some brace formatting in this file too

Python/Product/ExternalProfilerDriver/VTuneStackParser.cs
```diff
@@ -143,4 +151,61 @@ static class VTuneStackParser {
...
+        public static IEnumerable<SampleWithTrace> ParseFromFile(string filename)
+        {
+            if (!File.Exists(filename)) {
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMsgPathDoesNotExist, filename));
+            }
+
+            SampleWithTrace current = null;
+            List<PerformanceSample> currentStack = new List<PerformanceSample>();
```
huguesv: `var`

Python/Product/ExternalProfilerDriver/VTuneStackParser.cs
```diff
@@ -143,4 +151,61 @@ static class VTuneStackParser {
...
+
+            uint index = 0;
+            using (var reader = new StreamReader(filename)) {
+                using (CsvReader csvReader = new CsvReader(reader)) {
+
+                    while (csvReader.Read()) {
+                        var row = csvReader.GetRecord<dynamic>();
+                        var dict = row as IDictionary<string, object>;
+
+                            if (dict["Function"] == null || dict["Function"].ToString() == string.Empty) {
```
huguesv: this is potentially going to lookup the same entry twice.

you could use (hopefully I have it right, I didn't compile this):
`if (!dict.TryGetValue("Function", out object funcObj) || string.IsNullOrEmpty(funcObj?.ToString()))`

Similar for Function Stack

Python/Product/ExternalProfilerDriver/VTuneStackParser.cs
```diff
@@ -143,4 +151,61 @@ static class VTuneStackParser {
...
+                                if (dict["Function Stack"] == null || dict["Function Stack"].ToString() == string.Empty) {
+                                    current.AddStack(currentStack);
+                                    currentStack = new List<PerformanceSample>();
+                                } else {
+                                    PerformanceSample s = new PerformanceSample(dict["Function Stack"].ToString(),
```
huguesv: you can save a bit using `var` (same below)

Python/Product/ExternalProfilerDriver/VTuneToDWJSON.cs
```diff
@@ -209,6 +234,70 @@ class VTuneToDWJSON {
...
+            return mfdd;
+        }
+
+        /// <summary>
+        /// Given a two-level module/function dictionary, check in <para>symbolPath</param> if
+        /// there is information on the source file/line number the function is defined in.
+        ///
+        /// TODO: maybe this function should take a list of possible directories to search?
+        /// </summary>
+        public static Dictionary<string, Dictionary<string, FunctionSourceLocation>> AddLineNumbers(ref Dictionary<string, Dictionary<string, BaseSizeTuple> > orig, string symbolPath)
```
huguesv: `ref` is only for when you want to change the value of the parameter, which isn't done here, ie there is no `orig = newvalue`

Python/Product/ExternalProfilerDriver/VTuneToDWJSON.cs
```diff
@@ -94,11 +94,11 @@ class VTuneToDWJSON {
         /// <paramref name="filename"/>
         /// </summary>
         /// <param name="filename">The filename with the callstack report</param>
-        public static double CSReportToDWJson(string filename, string outfname) {
+        public static double CSReportToDWJson(string filename, string outfname, string sympath = "") {
```
huguesv: the names of the params could be more specific, especially the first one.

the 3rd param is called `symbolPath` in `AddLineNumbers`, that would be better than `sympath`

Python/Product/ExternalProfilerDriver/VTuneToDWJSON.cs
```diff
@@ -209,6 +234,70 @@ class VTuneToDWJSON {
...
+                    string fnd = Utils.FindFileInDir(modfname, rootSymbolPath);
+                    // 2. Getting the symbols in the file
+                    SymbolReader symReader = SymbolReader.Load(fnd);
+                    var syms = symReader.FunctionLocations().ToList();
+                    // 3. Get the information only for the functions that appear in the trace
+                    var funlocs = orig[modk].Join(syms, f => f.Key, fi => fi.Function, (f, fi) => fi).ToList();
+
+                    if (funlocs.Count() > 0) {
+                        sourcelocs[modk] = new Dictionary<string, FunctionSourceLocation>();
+                        //foreach ( var x in (orig[modk].Join(syms, f => f.Key, fi => fi.Function, (f, fi) => fi)) ) {
```
huguesv: delete old commented out code?

Python/Product/ExternalProfilerDriver/VTuneStackParser.cs
```diff
@@ -143,4 +151,61 @@ static class VTuneStackParser {
...
+                                                                                dict["Function (Full)"].ToString(),
+                                                                                dict["Source File"].ToString(),
+                                                                                dict["Start Address"].ToString());
+                                    currentStack.Add(s);
+                                }
+                            } else {
+                                if (current != null) {
+                                    yield return current;
+                                }
+                                PerformanceSample s = new PerformanceSample(dict["Function"].ToString(),
```
huguesv: since you are repeating these column names 2 or 3 times, you should create constants.

it could be an inner class:

```
static class Columns {
    const string Function = "Function";
    const string FunctionStack = "Function Stack";
// etc
}
```

rgesteve: Please see #4906 instead, it includes many of the revisions suggested here

RaymonGulati1:
@rgesteve, What's the update on this pull request? Is this a feature you still want to integrate into PTVS?

rgesteve:
Yes, this is to integrate into PTVS for VTune support.  @huguesv had asked for unit tests and this PR update includes some.  More are forthcoming.

rgesteve:
Sorry, read the wrong end of this, tests are in PR #4906, the current one can be closed, only left open to reference earlier comments in the code.

MinKPark:
@rgesteve Are you still working on this PR? 
I wonder whether I can discard this PR.

huguesv:
The other PR has all the changes from this one. I'll close it and hopefully we can still see the review comments, which I hadn't repeated in the PR.

PR #4825, opened 2018-10-30
Environment toolbar switcher

huguesv:
Add a toolbar that shows the active environment for current file/project/workspace.

You can use it to switch to a different environment.

Includes additional buttons to open the add environment dialog, package management view, and the environments window.

image

Loose Python file, or Python file from a C#, C++ or other non-Python project:
- Shows/changes global default environment
- List includes all globally installed environments

Workspace:
- Shows/changes the interpreter specified in PythonSettings.json. If none set, then global default environment is shown.
- List includes all globally installed environments
- List includes all virtual environments located under workspace folder

Python Project:
- shows/changes the interpreter in the .pyproj project file. If none set, then global default environment is shown.
- List includes all environments that are referenced in the .pyproj project file. This means the list will be empty when project doesn't have any references and relies on global default.


huguesv:
This is causing NullReferenceException on latest dev16. Need to look into it.

Python/Product/PythonTools/PythonTools/Commands/AddEnvironmentCommand.cs
```diff
@@ -0,0 +1,49 @@
...
+            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+            _envSwitchMgr = serviceProvider.GetPythonToolsService().EnvironmentSwitcherManager;
+        }
+
+        public CommandStatus Status => CommandStatus.SupportedAndEnabled;
+
+        public async Task InvokeAsync() {
+            var project = _envSwitchMgr.Project;
+            if (project == null) {
+                var sln = (IVsSolution)_serviceProvider.GetService(typeof(SVsSolution));
```
MikhailArkhipov: Is `IVsSolution` thread safe? It didn't used to be.

Python/Product/PythonTools/PythonTools/Commands/SwitchToEnvironmentCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+        private readonly EnvironmentSwitcherManager _envSwitchMgr;
+        private IPythonInterpreterFactory[] _allFactories;
+        private IPythonInterpreterFactory _currentFactory;
+
+        public SwitchToEnvironmentCommand(IServiceProvider serviceProvider) {
+            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+            _envSwitchMgr = serviceProvider.GetPythonToolsService().EnvironmentSwitcherManager;
+        }
+
+        public int MaxCount => 64; // 0x4090 - 0x4050
```
MikhailArkhipov: const

Python/Product/PythonTools/PythonTools/Commands/SwitchToEnvironmentCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+                ? CommandStatus.SupportedAndEnabled | CommandStatus.Latched
+                : CommandStatus.SupportedAndEnabled;
+        }
+
+        public string GetText(int index) {
+            return _allFactories[index].Configuration.Description;
+        }
+
+        public async Task InvokeAsync(int index) {
+            await _envSwitchMgr.SwitchToFactoryAsync(_allFactories[index]);
```
MikhailArkhipov: Maybe just
```cs
public Task InvokeAsync(int index) =>_envSwi`tchMgr.SwitchToFactoryAsync(_allFactories[index]);
```

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
```diff
@@ -0,0 +1,238 @@
...
+        private IVsHierarchy _previousHier;
+        private PythonProjectNode _project;
+
+        public event EventHandler EnvironmentsChanged;
+
+        public EnvironmentSwitcherManager(IServiceProvider serviceProvider) {
+            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+            _optionsService = serviceProvider.GetComponentModel().GetService<IInterpreterOptionsService>();
+            _registryService = serviceProvider.GetComponentModel().GetService<IInterpreterRegistryService>();
+            _monitorSelection = serviceProvider.GetService(typeof(IVsMonitorSelection)) as IVsMonitorSelection;
```
MikhailArkhipov: Probably `SVsMonitorSelection`

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// <summary>
+        /// List of factories to present to the user, based on the current
+        /// project, workspace or loose document.
+        /// </summary>
+        /// <remarks>
+        /// For a project, this is the list of environments referenced by the
+        /// project. For a loose document or workspace, this is the list of all
+        /// environments found on the machine. 
+        /// </remarks>
+        public IEnumerable<IPythonInterpreterFactory> AllFactories { get; private set; }
```
MikhailArkhipov:
```cs
public IEnumerable<IPythonInterpreterFactory> AllFactories { get; }= Enumerable.Empty<IPythonInterpreterFactory>();
```
Up to you

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
```diff
@@ -0,0 +1,238 @@
...
+                    _project.InterpreterFactoriesChanged += OnProjectInterpreterFactoriesChanged;
+                }
+            }
+        }
+
+        public string GetSwitcherCommandKeyBinding() {
+            try {
+                return KeyBindingHelper.GetGlobalKeyBinding(GuidList.guidPythonToolsCmdSet, (int)PkgCmdIDList.cmdidViewEnvironmentStatus);
+            } catch (Exception ex) when (!ex.IsCriticalException()) {
+                // I've seen NRE thrown during VS shutdown
```
MikhailArkhipov: Technically you know if VS is going away since package receives `QueryClose`

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
```diff
@@ -0,0 +1,238 @@
...
+                return null;
+            }
+        }
+
+        public Task SwitchToFactoryAsync(IPythonInterpreterFactory factory) {
```
MikhailArkhipov: It doesn't look async, does it have to be?

Python/Product/VSCommon/Infrastructure/Commands/AsyncCommandRangeToOleMenuCommandShim.cs
```diff
@@ -0,0 +1,83 @@
...
+                return;
+            }
+
+            _commandRange.InvokeAsync(index).DoNotWait();
+        }
+
+        public override bool DynamicItemMatch(int cmdId) {
+            var index = cmdId - CommandID.ID;
+
+            if (index >= 0 && index < _maxCount) {
```
MikhailArkhipov:
```cs
MatchedCommandId = index >= 0 && index < _maxCount ? cmdId : 0;
return MatchedCommandId != 0;
```

Python/Product/VSCommon/Infrastructure/Commands/PackageCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (command != null) {
+                object inArg, outArg;
+
+                var oleArgs = args as OleMenuCmdEventArgs;
+                inArg = oleArgs?.InValue;
+
+                command.Handle(inArg, out outArg);
+
+                if (oleArgs != null && oleArgs.OutValue != IntPtr.Zero) {
+                    Marshal.GetNativeVariantForObject(outArg, oleArgs.OutValue);
```
MikhailArkhipov: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getnativevariantforobject?view=netframework-4.7.2 says API is obsolete

Python/Tests/Core.UI/EnvironmentUITests.cs
```diff
@@ -507,5 +505,82 @@ TreeNode env
...
+            var globalDefault37 = PythonPaths.Python37_x64 ?? PythonPaths.Python37;
+            globalDefault37.AssertInstalled();
+
+            using (var dis = app.SelectDefaultInterpreter(globalDefault37)) {
+                // Loose Python file shows global default
+                app.OpenDocument(TestData.GetPath("TestData", "Environments", "Program.py"));
+                CheckSwitcherText(app, globalDefault37.Configuration.Description);
+
+                // No switcher visible when solution closed and no file opened
+                app.Dte.ActiveWindow.Close();
```
MikhailArkhipov: I remember these things like `ActiveWindow` can easily be nulls in DTE.

Python/Product/PythonTools/PythonTools/Commands/SwitchToEnvironmentCommand.cs
huguesv: it's a property getter implementation, IAsyncCommandRange.MaxCount

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
huguesv: I have to keep the private setter because we update the value in `RefreshFactories`

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
huguesv: I did it in anticipation of OpenFolder, where the methods we'll need to use to write the settings are all async

Python/Product/VSCommon/Infrastructure/Commands/PackageCommand.cs
huguesv: I don't see an alternative. I see we use that function elsewhere in PTVS, for combo boxes commands. Also VS repo still uses it in several places.

Python/Product/PythonTools/PythonTools/Commands/AddEnvironmentCommand.cs
huguesv: This is running on the UI thread.

Python/Product/VSCommon/Infrastructure/Commands/AsyncCommandRangeToOleMenuCommandShim.cs
huguesv: If CommandID.ID is 0 and cmdId is 0, meaning the first dynamic entry, we would return false (incorrect) with this suggested code, and true (correct) with the existing code. It's unlikely that CommandID.ID would be 0 but technically possible. I'd rather stick with this RTVS code as-is.

Python/Tests/Core.UI/EnvironmentUITests.cs
huguesv: The test just opened a document right before, so if active window is null I'm ok with test failing.

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
huguesv: I thought that existed too, but I guess it doesn't. It's not in `Microsoft.VisualStudio.Shell.Interop`.

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
huguesv: The OpenFolder scenario is in now.

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
```diff
@@ -0,0 +1,277 @@
...
+        /// environments found on the machine. 
+        /// </remarks>
+        public IEnumerable<IPythonInterpreterFactory> AllFactories { get; private set; }
+
+        public IEnvironmentSwitcherContext Context {
+            get => _context;
+            private set {
+                if (_context != null) {
+                    _context.EnvironmentsChanged -= OnInterpretersChanged;
+                    _context.Dispose();
```
AlexanderSher: That is a dangerous approach to dispose object in property setter. It is unexpected for property callers and is responsibility of the code that manages life cycle of the contexts

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
```diff
@@ -0,0 +1,277 @@
...
+            EnvironmentsChanged?.Invoke(this, EventArgs.Empty);
+        }
+
+        private void Reset() {
+            Context = null;
+            RefreshFactories();
+        }
+
+        private void SetInitialContext() {
+            IntPtr hierPtr = IntPtr.Zero;
```
AlexanderSher: It is better to have a separate managed wrapper that works with `IntPtr` and use this wrapper in `EnvironmentSwitcherManager`.

Python/Product/PythonTools/PythonToolsPackage.cs
```diff
@@ -248,6 +264,16 @@ protected override Task<object> InitializeToolWindowAsync(Type toolWindowType, i
             return base.CreateToolWindow(ref toolWindowType, id);
         }
 
+        protected override int QueryClose(out bool canClose) {
+            var res = base.QueryClose(out canClose);
+
+            if (canClose) {
+                var pyService = this.GetPythonToolsService();
```
AlexanderSher: Can it return null?

Python/Product/VSCommon/Infrastructure/Commands/AsyncCommandRangeToOleMenuCommandShim.cs
```diff
@@ -0,0 +1,83 @@
...
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+
+namespace Microsoft.PythonTools.Infrastructure.Commands {
+    internal class AsyncCommandRangeToOleMenuCommandShim : PackageCommand {
```
AlexanderSher: Is it copied from RTVS?

Python/Product/VSInterpreters/Interpreter/WorkspaceInterpreterFactoryProvider.cs
```diff
@@ -0,0 +1,389 @@
...
+                    _initialized = true;
+                    InitializeWorkspace();
+                    DiscoverInterpreterFactories();
+                }
+            }
+        }
+
+        private Task OnActiveWorkspaceChanged(object sender, EventArgs e) {
+            InitializeWorkspace();
+            DiscoverInterpreterFactories();
```
AlexanderSher: `InitializeWorkspace` and `DiscoverInterpreterFactories` are called without lock here but under lock in `EnsureInitialized`. Is that expected?

Python/Product/VSCommon/Infrastructure/Commands/AsyncCommandRangeToOleMenuCommandShim.cs
huguesv: Yes, all the command + disposable helpers come from RTVS.

Python/Product/PythonTools/PythonToolsPackage.cs
huguesv: The service is created in package initialization, which I presume will have been called if QueryClose is called.

Python/Product/PythonTools/PythonTools/Environments/EnvironmentSwitcherManager.cs
huguesv: The wrapper class sounds overkill for this, but you made me look at the RTVS code at https://github.com/Microsoft/RTVS/blob/46a8f4579e4fd56219af3bb2cf3fd15b84464874/src/Package/Impl/ProjectSystem/ProjectSystemServices.cs#L47 and I realized that I can't let C# handle the container ptr out parameter that I don't care about, because it won't get freed.

Python/Product/VSInterpreters/Interpreter/WorkspaceInterpreterFactoryProvider.cs
huguesv: You're right, there are some missing locks

Python/Product/Common/Infrastructure/Disposable.cs
```diff
@@ -0,0 +1,103 @@
...
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
```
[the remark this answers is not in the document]
huguesv: We'll do a separate PR to fix them all.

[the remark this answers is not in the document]
huguesv: https://github.com/Microsoft/PTVS/issues/4880

huguesv:
@MikhailArkhipov @AlexanderSher I've updated this to use the toolbar instead of the status bar (as a result of the UX board review). I plan on merging Monday after lunch time. If you want to further comment, please do before then.

PR #4882, opened 2018-11-28
Fixes #4734 Live Share support

### LINKED ISSUE #4734: LiveShare support
Two parts to this:
* enable VS to be a LiveShare client via the Python Language Server
* prevent regular Python support interfering when VS is a LiveShare host

zooba:
Fixes #4734 Live Share support
Adds an optional MEF component for providing Live Share callback handling. This allows us to forward the LSP requests to the correct analyzer.
Also suppresses our integration when Live Share's content type is in the buffer. This means that the regular LSP handling will get it and we do not want to interfere.


zooba:
Still to do:
 - [ ] Replace explicit content type with constant
 - [ ] Diagnose why text changes are not synchronized before requesting completions
 - [ ] Decide how to deal with VS's LSP IntelliSense not sorting completions

@srivatsn - open to contributions or suggestions if you have any.

PR #4902, opened 2018-12-06
Mark the miniconda distro folder as vendored.

### LINKED ISSUE #4900: Set .gitattributes for vendored code
https://github.com/github/linguist#overrides

Use this for the miniconda vendored code.

huguesv:
Fix #4900 

https://github.com/github/linguist#overrides

Didn't bother doing that with typeshed folder, since Analysis project will be removed soon.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -13,4 +13,4 @@
 *.bat eol=crlf
 
 # Language detection
-PTVS/Python/Product/Miniconda/Miniconda3-x64/ linguist-vendored
+Python/Product/Miniconda/Miniconda3-x64/* linguist-vendored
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -11,3 +11,6 @@
 
 # Some Windows-specific files should always be CRLF
 *.bat eol=crlf
+
+# Language detection
+PTVS/Python/Product/Miniconda/Miniconda3-x64/ linguist-vendored
```
jakebailey: This needs to rooted at the repo, which means you shouldn't include `PTVS`. I also think you still need the wildcard at the end, i.e.:

```
Python/Product/Miniconda/Miniconda3-x64/* linguist-vendored
```

huguesv: You're right, fixing

huguesv: Done

Issue #4974, opened 2019-01-09
Modify database connection dialog to use updated version


RaymonGulati1:
Update the API which creates a data connection dialog

Issue #4964, opened 2019-01-07
Update django dependency in templates to 2.0.3 minimum


huguesv:
There's a vulnerability alert (at https://github.com/microsoft/ptvs/network/alerts), we should update both of these to >=2.0.3 instead of >=2.0
```
Python/Templates/Django/ProjectTemplates/Python/Web/StarterDjangoProject/requirements.txt
Python/Templates/Django/ProjectTemplates/Python/Web/WebRoleDjango/requirements.txt
```

huguesv:
I think there's more templates. Django blank & django polls are probably affected too.

PR #4988, opened 2019-01-16
Issue 4867

RaymonGulati1:
Removed the ODBC dialog window from cookie cutter project


huguesv:
Microsoft.Data.ConnectionUI.Dialog.dll can be deleted from source control.
ConnectionStringConverter.cs can be deleted.

RaymonGulati1:
From what I can tell, "Microsoft.Data.ConnectionUI" is not being used either. Should I remove this as well?

RaymonGulati1:
Fix #4974 
Fix #4964

Issue #4991, opened 2019-01-18
Enable C++ compiler warning level 4


huguesv:
We're using level 3, and there's a number of warnings in level 4 that are mandatory for compliance.

PR #4992, opened 2019-01-18
Compile C++ projects with warning level 4.

huguesv:
Fix https://github.com/Microsoft/PTVS/issues/4991

### DIFF
--- a/Python/Product/DebuggerHelper/trace.cpp
+++ b/Python/Product/DebuggerHelper/trace.cpp
@@ -453,19 +453,14 @@ static void TraceLine(void* frame) {
         }
     }
 
-#pragma warning(push)
-#pragma warning(disable:4244) // '=': conversion from 'unsigned int' to 'volatile uint8_t', possible loss of data
-
     // See the large comment at the declaration of breakpointData for details of how the below synchronization scheme works.
-    unsigned iData;
+    uint8_t iData;
     do {
         iData = currentBreakpointData;
         // BreakpointManager.WriteBreakpoints may run at this point and change currentBreakpointData ...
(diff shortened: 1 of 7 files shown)


Python/Product/DebuggerHelper/trace.cpp
```diff
@@ -445,6 +453,9 @@ static void TraceLine(void* frame) {
         }
     }
 
+#pragma warning(push)
+#pragma warning(disable:4244) // '=': conversion from 'unsigned int' to 'volatile uint8_t', possible loss of data
+
```
huguesv: This one should be fixed, not suppressed (it's a mandatory warning for security compliance)

I don't understand why `iData` was made to be a larger type (unint32 vs uint8), and if that was intentional or not.

Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -1178,14 +1199,15 @@ bool DoAttach(HMODULE module, ConnectionInfo& connInfo, bool isDebug) {
                         // all of the work here needs to be minimal - in particular we shouldn't
                         // ever evaluate user defined code as we could end up switching to this
                         // thread on the main thread and corrupting state.
-                        auto prevThreadState = getThreadTls(threadStateIndex);
```
huguesv: There was a warning that the result is not used, and I'm assuming that this function has no side-effect and it's pointless to call it if we don't use the result.

Python/Product/DebuggerHelper/trace.cpp
int19h: No particular reason that I can think of - should be safe to just change the type of `iData` to `uint8_t`.

PR #5017, opened 2019-01-26
Fix build in language-server branch

huguesv:
Update dirs.proj files for projects that were removed as part of language server integration. Also update the setup layout for renamed analysis files.


huguesv:
Oh there's shared files in analysis test folder, I have to undo a couple of deletes.

Issue #4945, opened 2018-12-21
Missing package check doesn't understand django requirements


huguesv:
Create new django project
Create virtual env, accept the default of installing from requirements.txt
Close project
Reopen project (make sure to open a python file)

Result:
Info bar prompt that there are missing packages. This is not true, the info bar should not appear.

This works fine for bottle projects where the requirements.txt contents is simply "bottle<1"

It doesn't work for this django requirements:
```
django >= 2.1, < 3.0
```

See `PipRequirementsUtilsTests.AnyPackageMissing` for the tests we currently have, which include a case similar to above that passes.

Regular expression is at `PipRequirementsUtils.FindRequirementsRegex`

![image](https://user-images.githubusercontent.com/1696845/50366241-68b6fa00-052d-11e9-9d37-490ed8396503.png)

Edit: I just changed the test temporarily. This first assert passes and the second fails.
```
            Assert.IsFalse(PipRequirementsUtils.AnyPackageMissing(
                new[] { "django >=1.9, <2" },
                new[] { new PackageSpec("Django", "1.8") }
            ));

            Assert.IsFalse(PipRequirementsUtils.AnyPackageMissing(
                new[] { "django >= 2.1, < 3.0" },
                new[] { new PackageSpec("Django", "2.1.4") }
            ));
```

Issue #5035, opened 2019-01-30
Change preview status string in virtual env and conda pages


huguesv:
Instead of "Please wait..."
Say "Building preview..."

PR #5037, opened 2019-01-31
Fix regex that parses requirments.txt and update unit tests

RaymonGulati1:
Fix: #4945 

Ignore head branch name.

### DIFF
--- a/Python/Tests/VSInterpretersTests/PipRequirementsUtilsTests.cs
+++ b/Python/Tests/VSInterpretersTests/PipRequirementsUtilsTests.cs
@@ -25,10 +25,9 @@ namespace VSInterpretersTests {
     public class PipRequirementsUtilsTests {
         [TestMethod, Priority(0)]
         public void MergeRequirements() {
-            IEnumerable<string> a;
             // Comments should be preserved, only package specs should change.
             AssertUtil.AreEqual(
-                a = PipRequirementsUtils.MergeRequirements(new[] {
+                PipRequirementsUtils.MergeRequirements(new[] {
                     "a # with a comment",
                     "B==0.2",
                     "# just a comment B==01234",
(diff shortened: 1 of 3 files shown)


Python/Tests/VSInterpretersTests/PipRequirementsUtilsTests.cs
```diff
@@ -215,6 +181,30 @@ public class PipRequirementsUtilsTests {
...
+            ));
+
+            Assert.IsFalse(PipRequirementsUtils.AnyPackageMissing(
+                new[] { "django >= 2.1.5, <= 3.0 " },
+                new[] { new PackageSpec("Django", "2.2") }
+            ));
+            Assert.IsFalse(PipRequirementsUtils.AnyPackageMissing(
+                new[] { "django>=2.1.5,<=3.0#comment" },
+                new[] { new PackageSpec("Django", "2.2") }
+            ));
```
huguesv: please add IsFalse tests for these:
```
-r morereqs.txt
git+https://myvcs.com/some_dependency
git+https://myvcs.com/some_dependency@sometag#egg=SomeDependency
```

RaymonGulati1: Since the "AnyPackagesMissing" cannot parse those 3 cases, shouldn't this be "Assert.IsTrue()"?

huguesv: No. We don't want to prompt the user that packages are missing if they have those things in their requirements.txt.

RaymonGulati1: Will be updated shortly.

RaymonGulati1:
Fix #5035

PR #5074, opened 2019-02-08
Add REPL window tests for workspaces support.

### LINKED ISSUE #5069: Open folder: Send to REPL UI tests

huguesv:
Fix #5069 

Only test changes, no product change.

### DIFF
--- a/Python/Tests/ReplWindowUITests/ReplWindowSendUITests.cs
+++ b/Python/Tests/ReplWindowUITests/ReplWindowSendUITests.cs
@@ -14,10 +14,17 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using System;
+using System.IO;
+using System.Linq;
 using System.Text;
+using System.Threading;
+using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Interpreter;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Editor;
+using TestUtilities;
 using TestUtilities.UI;
 using TestUtilities.UI.Python;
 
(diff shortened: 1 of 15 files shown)


Python/Tests/ReplWindowUITests/ReplWindowSendUITests.cs
```diff
@@ -87,7 +94,7 @@ ...     print(y ** x)
         /// while it's executing.
         /// </summary>
         public void SendToInteractiveDelayed(PythonVisualStudioApp app) {
-            RunOne(app, "Delayed.py",
+            RunFromProject(app, "Delayed.py",
```
RaymonGulati1: How about adding "RunFromWorkspace" scenarios to these tests as well?

Python/Tests/ReplWindowUITests/ReplWindowSendUITests.cs
```diff
@@ -159,24 +166,101 @@ ...     time.sleep(5)
...
-        private void RunOne(PythonVisualStudioApp app, string filename, params SendToStep[] inputs) {
-            // SendToInteractive.pyproj uses Python 3.5 32-bit
-            var settings = ReplWindowSettings.FindSettingsForInterpreter("Python35");
-            var sln = app.CopyProjectForTest(@"TestData\SendToInteractive.sln");
-            var project = app.OpenProject(sln);
-            var program = project.ProjectItems.Item(filename);
-            var window = program.Open();
+        private void RunFromProject(PythonVisualStudioApp app, string filename, params SendToStep[] inputs) {
+            PythonPaths.Python27_x64.AssertInstalled();
+            PythonPaths.Python37_x64.AssertInstalled();
```
RaymonGulati1: This will fail if the user does not have python 27 and 37 installed. Is there some way to look for the interpreters the user has installed and use that?

Python/Tests/ReplWindowUITests/ReplWindowSendUITests.cs
huguesv: I only added tests that triggers the differences in the product code when using a workspace. These other tests could be adapted for workspaces but wouldn't add to the code coverage (and would take up time to run).

Python/Tests/ReplWindowUITests/ReplWindowSendUITests.cs
huguesv: In this case, user = a ptvs developer = us, and python 2.7 and 3.7 are the bare minimum needed/expected to be installed in order to run tests. The AssertInstalled is so we have a nice error if we try to run the test and forgot to install the correct CPython.

PR #5091, opened 2019-02-13
Allow identifier renaming and extract method for Python identifiers in 3.0+

### LINKED ISSUE #3520: Allow renaming to all valid characters
Currently refactoring names (extract method, rename) are restricted to Python 2-compatible identifiers.

We should consider enabling all valid characters when the target language is Python 3.

(Reported at https://developercommunity.visualstudio.com/content/problem/174906/代码重命名不能重命名中文.html )

RaymonGulati1:
Fix #3520 
Internal issues: 749790 and 749787

{Do not merge until QB approval is granted}


huguesv:
I've merged all the fixes into 16.0 as one PR, so we don't need this one anymore.

PR #5112, opened 2019-02-19
Update calls to Process.Start()

### LINKED ISSUE #4995: Update calls to Process.Start
Unless we're launching the interpreter that the user typed in, we should be making sure to use full paths when launching executable.

Example of one that needs to be updated:
https://github.com/Microsoft/PTVS/blob/1d04f01b7b902a9e1051b4080770b4a27e6e97e7/Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml.cs#L81

We can use `Path.Combine(Environment.SystemDirectory, "explorer.exe")` to specify the full path.

Alternative: `string filepath= "%WINDIR%\\system32\\explorer.exe"; Environment.ExpandEnvironmentVariables(filepath);`

RaymonGulati1:
Fix #4995 

Update calls to Process.Start()

### DIFF
--- a/Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
+++ b/Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
@@ -77,7 +77,7 @@ namespace Microsoft.PythonTools.InterpreterList {
             _outputWindow = OutputWindowRedirector.GetGeneral(_site);
             Debug.Assert(_outputWindow != null);
             _statusBar = _site.GetService(typeof(SVsStatusbar)) as IVsStatusbar;
-            
+
             var list = new ToolWindow();
             list.ViewCreated += List_ViewCreated;
             list.ViewSelected += List_ViewSelected;
@@ -225,7 +225,7 @@ namespace Microsoft.PythonTools.InterpreterList {
 
             var psi = new ProcessStartInfo();
             psi.UseShellExecute = false;
-            psi.FileName = "explorer.exe";
+            psi.FileName = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Windows), "explorer.exe");
             psi.Arguments = "\"" + path + "\"";
 
             Process.Start(psi).Dispose();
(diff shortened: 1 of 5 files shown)


Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
```diff
@@ -523,13 +523,14 @@ class CondaEnvironmentManagerUI : ICondaEnvironmentManagerUI {
...
+            var psi = new ProcessStartInfo(Path.Combine(Environment.SystemDirectory, "cmd.exe")) {
+                Arguments = string.Join(" ", new[] {
                 "/S",
                 "/K",
                 pathCmd + string.Format("title {0} environment", view.Description)
-            }.Select(ProcessOutput.QuoteSingleArgument));
-            psi.WorkingDirectory = view.PrefixPath;
+            }.Select(ProcessOutput.QuoteSingleArgument)),
+                WorkingDirectory = view.PrefixPath
+            };
```
huguesv: This is hard to read with this indentation. Arguments and WorkingDirectory should be the only 2 lines at that indent level.

Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
```diff
@@ -545,14 +546,15 @@ class CondaEnvironmentManagerUI : ICondaEnvironmentManagerUI {
...
-            psi.WorkingDirectory = view.PrefixPath;
+            var psi = new ProcessStartInfo(Path.Combine(Environment.SystemDirectory, "WindowsPowerShell", "v1.0", "powershell.exe")) {
+                Arguments = string.Join(" ", new[] {
+            "-NoLogo",
+            "-NoExit",
+            "-Command",
+            pathCmd + string.Format("(Get-Host).UI.RawUI.WindowTitle = '{0} environment'", view.Description)
+            }.Select(ProcessOutput.QuoteSingleArgument)),
+                WorkingDirectory = view.PrefixPath
+            };
```
huguesv: Same comment as above. The older code was easier to read.

Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
RaymonGulati1: Updated. Not sure if this is exactly what you wanted, but I feel that it's much easier to read.

Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
RaymonGulati1: Same as above.

PR #5119, opened 2019-02-22
Activate conda environments before executing them.

### LINKED ISSUE #4820: Activate conda environment on launch, python install, etc
Conda environment with python 3.7 and numpy fail to load the dll when running the conda environment's python.exe without first activating the conda environment.

This didn't happen before (with Python 3.6).

We will have to activate conda environment before we launch, and anywhere we run the python interpreter (like install packages, profiling).

### LINKED ISSUE #5111: incompactable with conda 4.6
## environments:
conda 4.6

This throws an exception as follows.
image

## Probable causation:
conda 4.6 has a different logic of loading `activate SOME_VIRTUAL_ENV`

### LINKED ISSUE #5115: Fail to create conda environment with conda 4.5.12 or later
`conda create` now only succeeds if the conda base is activated.

huguesv:
Fix #5115 
Fix #5111 
Fix #4820 

(they're all variations of we need to activate conda environment for X to work)

### DIFF
--- a/Python/Product/PythonTools/PythonToolsService.cs
+++ b/Python/Product/PythonTools/PythonToolsService.cs
@@ -676,8 +676,9 @@ namespace Microsoft.PythonTools {
             // right now to minimize risk/complexity so close to release.
             if (CondaUtils.IsCondaEnvironment(config.Interpreter.GetPrefixPath())) {
                 var condaExe = CondaUtils.GetRootCondaExecutablePath(Site);
-                if (File.Exists(condaExe)) {
-                    var condaEnv = CondaUtils.GetActivatedEnvironmentVariables(condaExe, config.InterpreterPath);
+                var prefixPath = config.Interpreter.GetPrefixPath();
+                if (File.Exists(condaExe) && Directory.Exists(prefixPath)) {
+                    var condaEnv = CondaUtils.GetActivatedEnvironmentVariables(condaExe, prefixPath);
                     baseEnv = PathUtils.MergeEnvironments(baseEnv.AsEnumerable<string, string>(), condaEnv, "Path", pathVar);
                 }
             }
--- a/Python/Product/PythonTools/PythonToolsService.cs
+++ b/Python/Product/PythonTools/PythonToolsService.cs
@@ -678,7 +678,7 @@ namespace Microsoft.PythonTools {
                 var condaExe = CondaUtils.GetRootCondaExecutablePath(Site);
                 var prefixPath = config.Interpreter.GetPrefixPath();
                 if (File.Exists(condaExe) && Directory.Exists(prefixPath)) {
-                    var condaEnv = CondaUtils.GetActivatedEnvironmentVariables(condaExe, prefixPath);
+                    var condaEnv = CondaUtils.CaptureActivationEnvironmentVariablesForPrefix(condaExe, prefixPath);
                     baseEnv = PathUtils.MergeEnvironments(baseEnv.AsEnumerable<string, string>(), condaEnv, "Path", pathVar);
                 }
             }
--- a/Python/Product/PythonTools/PythonToolsService.cs
+++ b/Python/Product/PythonTools/PythonToolsService.cs
@@ -670,6 +670,18 @@ namespace Microsoft.PythonTools {
                 pathVar = "PYTHONPATH";
             }
             baseEnv[pathVar] = string.Empty;
+
+            // TODO: We could introduce a cache so that we don't activate the
+            // environment + capture the env variables every time. Not doing this
+            // right now to minimize risk/complexity so close to release.
+            if (CondaUtils.IsCondaEnvironment(config.Interpreter.GetPrefixPath())) {
+                var condaExe = CondaUtils.GetRootCondaExecutablePath(Site);
+                if (File.Exists(condaExe)) {
+                    var condaEnv = CondaUtils.GetActivatedEnvironmentVariables(condaExe, config.InterpreterPath);
+                    baseEnv = PathUtils.MergeEnvironments(baseEnv.AsEnumerable<string, string>(), condaEnv, "Path", pathVar);
+                }
+            }
+
             var env = PathUtils.MergeEnvironments(
                 baseEnv.AsEnumerable<string, string>(),
                 config.GetEnvironmentVariables(),
--- a/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
+++ b/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
@@ -311,7 +311,7 @@ namespace Microsoft.PythonTools.Interpreter {
 
         private async Task EnsureActivatedAsync() {
             if (_activatedEnvironmentVariables == null) {
-                var env = await CondaUtils.GetActivatedEnvironmentVariablesAsync(CondaPath);
+                var env = await CondaUtils.CaptureActivationEnvironmentVariablesForRootAsync(CondaPath);
                 _activatedEnvironmentVariables = env.Union(UnbufferedEnv).ToArray();
             }
         }
--- a/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
+++ b/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
@@ -36,7 +36,9 @@ namespace Microsoft.PythonTools.Interpreter {
             new KeyValuePair<string, string>("PYTHONUNBUFFERED", "1")
         };
 
-        internal CondaEnvironmentManager(string condaPath) {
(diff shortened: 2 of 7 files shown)


Python/Product/PythonTools/PythonToolsService.cs
```diff
@@ -670,6 +670,18 @@ public sealed class PythonToolsService : IDisposable {
                 pathVar = "PYTHONPATH";
             }
             baseEnv[pathVar] = string.Empty;
+
+            // TODO: We could introduce a cache so that we don't activate the
+            // environment + capture the env variables every time. Not doing this
+            // right now to minimize risk/complexity so close to release.
```
RaymonGulati1: Did you create an issue for it?

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
```diff
@@ -50,15 +52,21 @@ sealed class CondaEnvironmentManager : ICondaEnvironmentManager, IDisposable {
...
             var condaPath = CondaUtils.GetRootCondaExecutablePath(serviceProvider);
             if (!string.IsNullOrEmpty(condaPath)) {
-                return new CondaEnvironmentManager(condaPath);
+                return Create(condaPath);
             }
 
             return null;
         }
 
+        public static CondaEnvironmentManager Create(string condaPath) {
```
RaymonGulati1: More descriptive name please. "CreateCondaEnvironmentManager"?

Python/Product/VSInterpreters/PackageManager/CondaUtils.cs
```diff
@@ -45,5 +53,73 @@ static class CondaUtils {
...
             return provider?.FindLocator()?.CondaExecutablePath;
         }
+
+        /// <summary>
+        /// Determine if an environment is created/managed by conda and thus should be activated.
+        /// </summary>
+        /// <param name="prefixPath">Path to the environment.</param>
+        /// <returns><c>true</c> if it is a conda environment.</returns>
+        internal static bool IsCondaEnvironment(string prefixPath) {
+            return Directory.Exists(Path.Combine(prefixPath, "conda-meta"));
```
RaymonGulati1: Shouldn't we check if Conda environment contains python.exe?

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
```diff
@@ -291,8 +309,15 @@ sealed class CondaEnvironmentManager : ICondaEnvironmentManager, IDisposable {
             }
         }
 
+        private async Task EnsureActivatedAsync() {
+            if (_activatedEnvironmentVariables == null) {
+                var env = await CondaUtils.GetActivatedEnvironmentVariablesAsync(CondaPath);
```
RaymonGulati1: If "activate.bat" does not exist in the Conda environment, it will return an empty KeyValuePair [Line 82]. Should we be error checking here? Same for other cases relating to "GetActivatedEnvironmentVariables", "ParseEnvironmentVariables", etc.

Python/Product/VSInterpreters/PackageManager/CondaUtils.cs
```diff
@@ -45,5 +53,73 @@ static class CondaUtils {
...
+        }
+
+        /// <summary>
+        /// Activate the base conda environment, capture and return its environment variables
+        /// </summary>
+        /// <param name="condaPath">Path to the base conda environment's conda.exe</param>
+        /// <returns>List of environment variables.</returns>
+        internal async static Task<IEnumerable<KeyValuePair<string, string>>> GetActivatedEnvironmentVariablesAsync(string condaPath) {
```
RaymonGulati1: Slightly confusing with the function name and the function below. Change name to be more descriptive to specify that this is the base interpreter?

Python/Product/VSInterpreters/PackageManager/CondaUtils.cs
```diff
@@ -45,5 +53,73 @@ static class CondaUtils {
...
+            return Enumerable.Empty<KeyValuePair<string, string>>();
+        }
+
+        /// <summary>
+        /// Activate the specified conda environment, capture and return its environment variables.
+        /// </summary>
+        /// <param name="condaPath">Path to the base conda environment's conda.exe</param>
+        /// <param name="prefixPath">Path to the conda environment to activate.</param>
+        /// <returns>List of environment variables.</returns>
+        internal static IEnumerable<KeyValuePair<string, string>> GetActivatedEnvironmentVariables(string condaPath, string prefixPath) {
```
RaymonGulati1: "ActivateAndGetEnvironmentVariables"? Current name implies it's already activated.

Python/Tests/Core/CondaEnvironmentManagerTests.cs
```diff
@@ -112,30 +112,6 @@ public class CondaEnvironmentManagerTests {
             AssertCondaMetaFiles(envPath, "python-2.7.*.json");
         }
 
-        [TestMethod, Priority(0)]
```
RaymonGulati1: Create additional tests to replace this one?

Python/Product/PythonTools/PythonToolsService.cs
huguesv: Done, thanks
https://github.com/Microsoft/PTVS/issues/5124

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
huguesv: `Create` is standard naming convention in .NET for this scenario.

`CondaEnvironmentManager.CreateCondaEnvironmentManager` would be long and redundant.

Python/Tests/Core/CondaEnvironmentManagerTests.cs
huguesv: I've removed the test because conda changed behavior between 4.5.11 and later versions. Rather than fail when attempting to create an environment that already exists, they now delete+create. So I figure it's not worth it anymore.

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
huguesv: I made it work this way because:
1. Returning empty environment variables in the case of failing activation is "safer" and falls back to old behavior (it will use VS' process environment variables). I prefer safer solution right now before release (no unanticipated crash due to missing error handling)
2. Some scenarios will work without activation, so forcing a failure could make things worse ie. block the user when things would work just fine.
3. Failing to activate will lead to a known class of errors "failed to load module" that are easy for us to identify.
4. We're past the localization deadline so I'd prefer not introducing error messages at this time if possible/reasonable.

Python/Product/VSInterpreters/PackageManager/CondaUtils.cs
huguesv: I'll add it.

Python/Product/VSInterpreters/PackageManager/CondaUtils.cs
huguesv: Method name with "And" in them is usually a bad sign. I'll try to come up with something.

huguesv: I think I'll go with `CaptureActivationEnvironmentVariables`. That doesn't imply that it's already activated.

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
RaymonGulati1: 3. If the activation fails, will this known class of errors display an error stating that activate environment failed? My concern is that we would confuse the user if we don't display some kind of error message whenever we fail to create the environment.

Python/Tests/Core/CondaEnvironmentManagerTests.cs
RaymonGulati1: So, if a user accidentally creates a new environment with the same name as existing environment, their previous environment would be deleted. Do we give them some kind of warning or confirmation option to prevent them from accidentally doing this?

huguesv: Add environment dialog checks for invalid or conflicting names, and doesn't allow you to proceed with creation if you specify an already existing name.

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentManager.cs
huguesv: The user gets an error message right now because we don't activate. With this change we will continue to display the same error message if activation fails. So this won't be more confusing than it is now.

PR #5126, opened 2019-02-26
Update readme for VS 2019.

### LINKED ISSUE #5116: Updated README for VS 2019
Update README for PTVS repo to include VS 2019

huguesv:
Fix #5116

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,6 @@ We welcome code contributions - see [Contributing](https://github.com/Microsoft/
 
 [![Join the chat at https://gitter.im/Microsoft/PTVS](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/Microsoft/PTVS?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-![Build status for master branch](https://mseng.visualstudio.com/_apis/public/build/definitions/b21de0ba-dffd-474c-b888-7112d5e6397b/2921/badge)
-
 ## Visual Studio 2017 and Visual Studio 2019 installation
 
 The Python development workload is available as part of [Visual Studio](https://aka.ms/PTVS) Community, Professional and Enterprise. To install, run the [normal VS installer](https://visualstudio.com/vs/downloads?wt.mc_id=github_microsoft_com) and select either the **Python development workload** or the **Data science and analytical applications workload**.
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
 # Python tools for Visual Studio
 
-Python support for [Visual Studio](https://aka.ms/PTVS) is developed and managed here.
+Python support for Visual Studio is developed and managed here.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,9 +10,9 @@ Feel free to file issues or ask questions on our [issue tracker](http://github.c
 
 ![Build status for master branch](https://mseng.visualstudio.com/_apis/public/build/definitions/b21de0ba-dffd-474c-b888-7112d5e6397b/2921/badge)
```
RaymonGulati1: This badge doesn't provide any valuable information to the end user, what do you think about removing it? Users expect the master branch to be successfully built. If it's not, it's a bad impression and we're publicly advertising it.

README.md
```diff
@@ -10,9 +10,9 @@ Feel free to file issues or ask questions on our [issue tracker](http://github.c
```
RaymonGulati1: Why are we supporting a chat that we barely use? It would be better to steer the users to one spot--maybe create a dedicated post in developer community where users can ask questions? It will allow us to respond and organize information so that it can be easily referenced by future users.

README.md
```diff
@@ -1,6 +1,6 @@
 # Python tools for Visual Studio
 
-Python support for [Visual Studio 2017](https://aka.ms/PTVS) is developed and managed here.
+Python support for [Visual Studio](https://aka.ms/PTVS) is developed and managed here.
 
 Visit our page on [VisualStudio.com](https://aka.ms/PTVS) for an overview and download links, and check out our documentation on [docs.microsoft.com](https://aka.ms/PTVSDocs) and our (old) [Getting Started](http://aka.ms/PTVSTutorial) series on YouTube.
```
RaymonGulati1: The "aka.ms/PTVS" link on this line and above are the same line which duplicates information for no benefit. Please modify either one of them. 

Also, add the links in line rather than describing and linking separately. For example: "You can visit our overview page (link), download links (link), documentation (link), and (old) YouTube tutorials (link).

README.md
huguesv: This isn't what I check to see if the build is passing/failing, so I don't personally care either way. If our build is not passing and I don't notice, and a user lets me know, I'll be happy about it.

It's a bit of advertising for Azure DevOps... it would be more useful if the build was public and you could navigate to it via the badge.

README.md
huguesv: Some users asked for it. I suppose some like having a lower impact way of asking a question without officially "filing" something somewhere.

We can look into shutting it down.

README.md
RaymonGulati1: I understand the logic behind azure advertisement, but it's kind of useless right now because it just leads to a page with a picture. How about making it point [here](https://azure.microsoft.com/en-us/services/devops/)? 

The public build is a better--if there's no easy way to do it right now, we can consider it down the line.

README.md
RaymonGulati1: Let's take the conversation offline and make the change later if need be.

README.md
RaymonGulati1: I understand the logic behind azure advertisement, but it's kind of useless right now because it just leads to a page with a picture. How about making it point [here](https://azure.microsoft.com/en-us/services/devops/)?

The public build is a better--if there's no easy way to do it right now, we can consider it down the line.

huguesv: I've removed it. We can add it back if/when it ever provides value.

PR #5138, opened 2019-03-01
Fix outlining

### LINKED ISSUE #4703: Collapsed outlining has multiple classes/methods on the same line
Input code:
```
class Class1:
    def __init__(self, hello):
        self._hello = hello

    def method1a(self):
        for handler in self._hello:
            if hasattr(handler, 'file'):
                handler.file.close()


class Class2:
    def __init__(self):
        self._list = []

    def method2a(self, num_bytes):
        if num_bytes > 40:
            raise NotImplementedError(
                "Sorry!"
            )


class Class3:
    """
    HELLO
    """
    def __init__(self, stream):
        self.stream = stream

    def __next__(self):
        try:
            data = self.stream.read(1)
        except InputStreamExhausted:
            raise StopIteration()
        if data:
            return data
        else:
            raise StopIteration()

    def __iter__(self):
        return self

```
Collapse to definitions:
image

Click the '+' and you get this:
image

Reported at https://developercommunity.visualstudio.com/content/problem/328855/python-global-class-collapse-merge-method-defs.html

huguesv:
Fix #4703 by integrating updated language server

See language server PR at https://github.com/Microsoft/python-language-server/pull/681

Also fixes tests:
- Fix MEF composition error in mock tests.
- Fix go to navigation tests to reflect the updated language server behavior for get definition.

### DIFF
--- a/Build/15.0/packages.config
+++ b/Build/15.0/packages.config
@@ -68,5 +68,5 @@
   <package id="Microsoft.VisualStudio.Editor.Implementation" version="15.8.525" />
   <package id="Microsoft.VisualStudio.Platform.VSEditor" version="15.8.525" />
   <package id="Microsoft.VisualStudio.Text.Internal" version="15.8.525" />
-  <package id="Microsoft.VisualStudio.Python.LanguageServer" version="0.1.28" />
+  <package id="Microsoft.VisualStudio.Python.LanguageServer" version="0.1.29" />
 </packages>
--- a/Common/Tests/MockVsTests/MockVsFolderWorkspaceService.cs
+++ b/Common/Tests/MockVsTests/MockVsFolderWorkspaceService.cs
@@ -1,0 +1,29 @@
+﻿// Visual Studio Shared Project
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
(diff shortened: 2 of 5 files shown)


Build/15.0/packages.config
```diff
@@ -68,5 +68,5 @@
   <package id="Microsoft.VisualStudio.Editor.Implementation" version="15.8.525" />
   <package id="Microsoft.VisualStudio.Platform.VSEditor" version="15.8.525" />
   <package id="Microsoft.VisualStudio.Text.Internal" version="15.8.525" />
-  <package id="Microsoft.VisualStudio.Python.LanguageServer" version="0.1.28" />
+  <package id="Microsoft.VisualStudio.Python.LanguageServer" version="0.1.29" />
```
RaymonGulati1: Since the updated language server is going into 16.0, why update 15.0?

Common/Tests/MockVsTests/MockVsFolderWorkspaceService.cs
```diff
@@ -0,0 +1,29 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.Workspace;
+using Microsoft.VisualStudio.Workspace.VSIntegration.Contracts;
+
+namespace Microsoft.VisualStudioTools.MockVsTests {
```
RaymonGulati1: What's the point of file? Where is it used?

Build/15.0/packages.config
huguesv: While this file exists, I'd rather keep it in sync. Not ready to delete the file yet, would rather wait until 16.0 is out.

Common/Tests/MockVsTests/MockVsFolderWorkspaceService.cs
huguesv: It satisfies a MEF import. Since the real implementation is not exported by VS assembly when running mock tests, we need at least one class in a loaded assembly to export this interface.

PR #5148, opened 2019-03-06
C++ static analysis 4999

### LINKED ISSUE #4999: C++ static analysis warnings
Fix warnings reported by prefast tool.

Internal bug 773848

RaymonGulati1:
Resolving the warnings generated by VS recommended rule set.
Fix #4999

### DIFF
--- a/Python/Product/DebuggerHelper/trace.cpp
+++ b/Python/Product/DebuggerHelper/trace.cpp
@@ -241,32 +241,32 @@ volatile uint32_t evalLoopSEHCode; // if a structured exception occured during e
 
 __declspec(dllexport)
 bool StringEquals27(const DebuggerString* debuggerString, const void* pyString) {
-    int32_t my_length = debuggerString->length;
+	int32_t my_length = debuggerString->length;
     const wchar_t* my_data = debuggerString->data;
 
     // In 2.7, we have to be able to compare against either ASCII or Unicode strings, so check type and branch.
     auto ob_type = ReadField<const void*>(pyString, fieldOffsets.PyObject.ob_type);
     if (reinterpret_cast<uint64_t>(ob_type) == types.PyBytes_Type) {
-        auto ob_size = ReadField<SSIZE_T>(pyString, fieldOffsets.PyVarObject.ob_size);
+		auto ob_size = ReadField<SSIZE_T>(pyString, fieldOffsets.PyVarObject.ob_size);
         if (ob_size != my_length) {
             return false;
         }
 
         const char* data = reinterpret_cast<const char*>(pyString) + fieldOffsets.PyBytesObject.ob_sval;
-        for (int32_t i = 0; i < my_length; ++i) {
+		for (int32_t i = 0; i < my_length; ++i) {
             if (data[i] != my_data[i]) {
                 return false;
             }
         }
         return true;
     } else if (reinterpret_cast<uint64_t>(ob_type) == types.PyUnicode_Type) {
-        auto length = ReadField<SSIZE_T>(pyString, fieldOffsets.PyUnicodeObject27.length);
+		auto length = ReadField<SSIZE_T>(pyString, fieldOffsets.PyUnicodeObject27.length);
         if (length != my_length) {
             return false;
         }
 
         const wchar_t* data = ReadField<const wchar_t*>(pyString, fieldOffsets.PyUnicodeObject27.str);
-        return memcmp(data, my_data, my_length * 2) == 0;
+        return memcmp(data, my_data, my_length * ((size_t) 2)) == 0;
     } else {
         return false;
     }
@@ -281,7 +281,7 @@ bool StringEquals33(const DebuggerString* debuggerString, const void* pyString) 
         return false;
     }
 
-    int32_t my_length = debuggerString->length;
+	int32_t my_length = debuggerString->length;
     const wchar_t* my_data = debuggerString->data;
 
 #pragma warning(push)
@@ -311,7 +311,7 @@ bool StringEquals33(const DebuggerString* debuggerString, const void* pyString) 
             return false;
         }
 
-        return memcmp(wstr, my_data, my_length * 2) == 0;
+        return memcmp(wstr, my_data, my_length * ((size_t) 2)) == 0;
     }
 
     auto length = ReadField<SSIZE_T>(pyString, fieldOffsets.PyUnicodeObject33.length);
@@ -329,7 +329,7 @@ bool StringEquals33(const DebuggerString* debuggerString, const void* pyString) 
     }
 
     if (kind == 2) {
-        return memcmp(data, my_data, my_length * 2) == 0;
+        return memcmp(data, my_data, my_length * ((size_t) 2)) == 0;
     } else if (kind == 1 || ascii) {
         auto asciiData = reinterpret_cast<const char*>(data);
         for (int32_t i = 0; i < my_length; ++i) {
@@ -524,22 +524,20 @@ static void TraceReturn(void* frame) {
 #pragma warning(disable:4211) // nonstandard extension used: redefined extern to static
 
 static void ReleasePendingObjects() {
-    if (objectsToRelease) {
-        auto otr = reinterpret_cast<ObjectToRelease*>(InterlockedExchange64(&objectsToRelease, 0));
-        while (otr != nullptr) {
-            // Releasing an object may trigger execution of its __del__ function, which will cause re-entry to this code,
-            // so null out the reference before releasing, and check for nulls in the list.
-            auto obj = otr->pyObject;
-            if (obj != 0) {
-                otr->pyObject = 0;
-                Py_DecRef(reinterpret_cast<PyObject*>(obj));
-            }
+	auto otr = reinterpret_cast<ObjectToRelease*>(InterlockedExchange64(&objectsToRelease, 0));
+	while (otr != nullptr) {
+		// Releasing an object may trigger execution of its __del__ function, which will cause re-entry to this code,
+		// so null out the reference before releasing, and check for nulls in the list.
+		auto obj = otr->pyObject;
+		if (obj != 0) {
+			otr->pyObject = 0;
+			Py_DecRef(reinterpret_cast<PyObject*>(obj));
+		}
 
-            auto next = reinterpret_cast<ObjectToRelease*>(otr->next);
-            VirtualFree(otr, 0, MEM_RELEASE);
-            otr = next;
-        }
-    }
+		auto next = reinterpret_cast<ObjectToRelease*>(otr->next);
+		VirtualFree(otr, 0, MEM_RELEASE);
+		otr = next;
+	}
 }
 
 #pragma warning(pop)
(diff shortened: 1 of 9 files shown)


Python/Product/DebuggerHelper/trace.cpp
```diff
@@ -266,7 +266,7 @@ bool StringEquals27(const DebuggerString* debuggerString, const void* pyString)
         }
 
         const wchar_t* data = ReadField<const wchar_t*>(pyString, fieldOffsets.PyUnicodeObject27.str);
-        return memcmp(data, my_data, my_length * 2) == 0;
+        return memcmp(data, my_data, my_length * ((size_t) 2)) == 0;
```
RaymonGulati1: image

Python/Product/DebuggerHelper/trace.cpp
```diff
@@ -524,21 +524,19 @@ static void TraceReturn(void* frame) {
...
-            auto obj = otr->pyObject;
-            if (obj != 0) {
-                otr->pyObject = 0;
-                Py_DecRef(reinterpret_cast<PyObject*>(obj));
-            }
-
-            auto next = reinterpret_cast<ObjectToRelease*>(otr->next);
-            VirtualFree(otr, 0, MEM_RELEASE);
-            otr = next;
+    auto otr = reinterpret_cast<ObjectToRelease*>(InterlockedExchange64(&objectsToRelease, 0));
```
RaymonGulati1: image

Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -315,7 +315,7 @@ __success(return) BOOL EnumProcessModulesHelper(
 
             auto cur = start;
             *lpcbNeeded = 0;
-
+            auto lphCurrentModule = lphModule;
```
RaymonGulati1: image

RaymonGulati1: image

Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -324,16 +324,17 @@ __success(return) BOOL EnumProcessModulesHelper(
...
+                    *lphCurrentModule = (HMODULE)curTable->DllBase;
+                    lphCurrentModule++;
                 }
 
                 (*lpcbNeeded) += sizeof(SIZE_T);
                 cur = (LDR_DATA_TABLE_ENTRY*)((LIST_ENTRY*)cur)->Flink;
             } while (cur != start && cur != 0);
 
-            return *lpcbNeeded <= cb;
+            return (lphCurrentModule == lphModule) ? FALSE : *lpcbNeeded <= cb;
```
RaymonGulati1: image

RaymonGulati1: image

Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -456,13 +457,16 @@ class PrivateHeapAllocator {
     pointer allocate(size_type size, allocator<void>::const_pointer hint = 0) {
         UNREFERENCED_PARAMETER(hint);
 
-        if (g_heap == nullptr) {
-            g_heap = HeapCreate(0, 0, 0);
-        }
+    g_heap = (g_heap == nullptr) ? HeapCreate(0, 0, 0) : g_heap;
```
RaymonGulati1: image

Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -406,7 +407,7 @@ char* ReadCodeFromFile(wchar_t* filePath) {
     auto length = filestr.tellg();
     filestr.seekg(0, ios::beg);
 
-    int len = (int)length;
+    SSIZE_T len = (int)length;
```
RaymonGulati1: image

Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -597,7 +601,8 @@ class ConnectionInfo {
     MemoryBuffer *Buffer;
     bool Succeeded;
 
-    ConnectionInfo() : Succeeded(false) {
+    ConnectionInfo() :
```
RaymonGulati1: image

Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -1033,7 +1038,7 @@ bool DoAttach(HMODULE module, ConnectionInfo& connInfo, bool isDebug) {
                 }
                 ResumeThreads(suspendedThreads);
             } while (!threadsInited() &&
-                (TICKS_DIFF(startTickCount, GetTickCount())) < (ticksPerSecond * 20) &&
+                (TICKS_DIFF(startTickCount, GetTickCount64())) < (ticksPerSecond * 20) &&
```
RaymonGulati1: image

Python/Product/VsPyProf/PythonApi.cpp
```diff
@@ -345,11 +345,12 @@ void VsPyProf::GetModuleName(wstring name, wstring& finalName) {
 
             // Re-assemble to C:\Fob\Oar\Baz
             wchar_t newName[MAX_PATH];
+            newName[0] = '\0';
```
RaymonGulati1: image

Python/Product/VsPyProf/PythonApi.cpp
```diff
@@ -424,7 +425,7 @@ bool VsPyProf::GetBuiltinToken(PyObject* codeObj, DWORD_PTR& func, DWORD_PTR& mo
 
                 // In Python3k module methods apparently have the module as their self, modules don't 
                 // actually have any interesting methods so we can always filter.
-                if (type != PyModule_Type) {
+                if (type != nullptr && type != PyModule_Type) {
```
RaymonGulati1: image

Python/Product/VsPyProf/PythonApi.cpp
```diff
@@ -542,7 +543,8 @@ VsPyProf::VsPyProf(HMODULE pythonModule, int majorVersion, int minorVersion, Ent
     PyModule_Type(pyModuleType),
     PyInstance_Type(pyInstType),
     _asUnicode(asUnicode),
-    _unicodeGetLength(unicodeGetLength) {
+    _unicodeGetLength(unicodeGetLength), 
+    _refCount(0) {
```
RaymonGulati1: image

Python/Product/DebuggerHelper/trace.cpp
int19h: This is a false positive, because `objectsToRelease` are only ever going to be set from the debugger while the debuggee (and this code inside it) is frozen - and it's marked `volatile` to indicate that such external modification can take place. The use of `InterlockedExchange` here is to ensure that the swap is atomic even in the face of debugger pausing the process - if you look at the code, there's no other interlocked operation on this variable.

The change will work, but this function is called on a hot path (from `TraceFunc`, which invokes it on every call - i.e. this is called for every line of Python code traced), so avoiding an atomic operation if we can do the check without is beneficial to perf.

Python/Product/DebuggerHelper/trace.cpp
int19h: Can you do a `static_cast` here and in other places? There's still a bunch of C casts around, but I've been trying to stick to C++-style for all new code here.

Python/Product/PyDebugAttach/PyDebugAttach.cpp
int19h: I'm confused about how this one fixes that. Isn't the real problem here that `lphModule` isn't properly annotated as an array with `cb` as size, and so static analysis expects there to be exactly one element written via it, no more and no less?

But also, this is pre-Win7 workaround. @huguesv, do we even support that still? If not, let's just delete this whole function, and the caller can then call `EnumProcessModules` directly.

Python/Product/PyDebugAttach/PyDebugAttach.cpp
int19h: What's the operator * it's complaining about? I don't see any multiplications here.

As changed, this adds a potential integer overflow in the for-loop below, because the index variable is compared `<= len`, but is itself `int`, so it'll loop forever. The original code is broken for files >sizeof(int), but only in a sense that it would not load the complete file - it would at least return.

(This is all strictly hypothetical, since we're loading specific files here that we know are small enough that `int` is fine in practice. Maybe this just needs a check-and-error-return after the line that assigns `length`? Then the rest of it doesn't need to worry about dealing with size mismatches, overflows etc. The types here can be tricky to reconcile properly, because there's no guarantee that `streamoff` is not larger than `ssize_t`, so doing this strictly correctly requires a great deal more checks.)

int19h: Although given what this code really does, I think it could be made much simpler if it doesn't try to micro-optimize for allocations (which it shouldn't, because this isn't a hot path). If the function is changed to return `std::string`, then the entire thing after opening the file can be replaced with:
```cpp
std::string s(std::istreambuf_iterator<char>(filestr), {});
s.erase(std::remove(s.begin(), s.end(), '\r'), str.end());
return s;
```
And the caller can call `data()` explicitly to get the raw pointer.

int19h: Or alternatively:
```cpp
std::string s;
std::copy_if(
    std::istreambuf_iterator<char>(filestr), {},
    std::back_inserter(s),
    [](auto ch) { return ch != '\r'; }
);
```

Python/Product/PyDebugAttach/PyDebugAttach.cpp
huguesv: It appears that Win7 SP1 and Server 2012 R2 are the oldest OSes that we support:
https://docs.microsoft.com/en-us/visualstudio/releases/2019/system-requirements

If that's pre-Win7, then it could go away?

Python/Product/DebuggerHelper/trace.cpp
huguesv: Makes sense. Let's add that explanation as a comment, and suppress that warning locally.

Python/Product/DebuggerHelper/trace.cpp
RaymonGulati1: Will be updated in next commit.

Python/Product/DebuggerHelper/trace.cpp
RaymonGulati1: Will be updated in next commit

Python/Product/PyDebugAttach/PyDebugAttach.cpp
RaymonGulati1: Removed and directly calling EnumProcessModules.

Python/Product/PyDebugAttach/PyDebugAttach.cpp
RaymonGulati1: Updated in latest commit

RaymonGulati1:
@int19h, changed applied. Ready for round 2.

PR #5152, opened 2019-03-08
Allow users to paste information into the interactive experimental debugger window

### LINKED ISSUE #4380: Copy and pasting to python debug interactive when during debugging
While debugging I am not able to copy or paste from VS.  Cannot even paste from other programs like notepad to the debug interactive console.  I only use python, so I am not sure if it is specific to python or not.

There is no error message produced,  just canot copy or past to the debug interactive window.

Seem to happen independent of the version of python.  I use Python 2.7 and 3.6  both 32 & 64 bit of each.  Same issue regardless of python version.

Filed at https://developercommunity.visualstudio.com/content/problem/273649/copy-and-pasting-to-python-debug-interactive-when.html

RaymonGulati1:
Fix #4380

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
+++ b/Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
@@ -40,8 +40,7 @@ namespace Microsoft.PythonTools.Repl {
         IInteractiveEvaluator,
         IPythonInteractiveEvaluator,
         IMultipleScopeEvaluator,
-        IPythonInteractiveIntellisense
-    {
+        IPythonInteractiveIntellisense {
         private PythonDebugProcessReplEvaluator _activeEvaluator;
         private readonly Dictionary<int, PythonDebugProcessReplEvaluator> _evaluators = new Dictionary<int, PythonDebugProcessReplEvaluator>(); // process id to evaluator
         private readonly Dictionary<int, Task> _attachingTasks = new Dictionary<int, Task>();
@@ -121,7 +120,7 @@ namespace Microsoft.PythonTools.Repl {
             }
             if (_activeEvaluator != null) {
                 return _activeEvaluator.CanExecuteCode(text);
-            }else if (CustomDebugAdapterProtocolExtension.CanUseExperimental()) {
+            } else if (CustomDebugAdapterProtocolExtension.CanUseExperimental()) {
                 return CanExecuteCodeExperimental(text);
             }
             return true;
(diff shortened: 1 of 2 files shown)


Python/Product/PythonTools/PythonTools/Repl/PythonDebugReplEvaluator.cs
```diff
@@ -203,7 +202,8 @@ internal class PythonDebugReplEvaluator :
             if (_activeEvaluator != null) {
                 return _activeEvaluator.FormatClipboard();
             }
-            return String.Empty;
+            //Formatting for experimental debugger
```
huguesv: I don't think that comment is necessarily true. If it's `null` it could also be that there is no debugging happening currently, regardless of which debugger is selected in tools/options.

`CustomDebugAdapterProtocolExtension.CanUseExperimental()` lets you know if experimental debugger is currently debugging.

There's nothing debugger specific about paste, so we could unconditionally call `PythonCommonInteractiveEvaluator.FormatClipBoard`and take out the previous `if` statement ie not call `_activateEvaluator.FormatClipboard`.

RaymonGulati1: I agree with removing the comment. 

I considered changing "FormatClipboard()" to static function in PythonCommonInteractiveEvaluator.cs and calling that function directly through SelectableReplEvaluator.cs (line 356) and PythonDebugReplEvaluator.cs (line 203) through their respective CurrentWindows. However, there may be differences in the "CurrentWindow" property in PythonDebugReplEvaluator.cs and PythonCommonInteractiveEvaluator.cs. If we start interchanging them (using PythonCommonInteractiveEvaluator.cs property everywhere except FormatClipboard()) within the instance object, we could potentially end up with unexpected problems. 

I'm forced to use CurrentWindow property from PythonDebugReplEvaluator.cs for experimental debugger, but I want to avoid doing it for sections of code which don't strictly require it. And I don't want to fix something that is not broken :)

huguesv: It's going to be the same Window from both places. `PythonDebugReplEvaluator` passes its `CurrentWindow` to `PythonDebugProcessReplEvaluator` (which derives from `PythonCommonInteractiveEvaluator`).

`PythonDebugReplEvaluator` will get a cleanup when we remove legacy debugger. We can remove that `if` now or later when all the other `if`s go away, doesn't matter much.

RaymonGulati1: Modified. 

I've kept the FormatClipboard() in PythonCommonInteractiveEvaluator.cs because it implements IInteractiveEvaluator which requires FormatClipboard. This is being called by SelectableReplEvaluator.cs.

Issue #4784, opened 2018-10-11
Failure creating virtual env with base interpreter IronPython


huguesv:
Since that doesn't work out of the box, we could remove it from the list. Let the user go to command line.

```
Virtual environment is being created at 'c:\users\vslab1\source\repos\PythonApplication2\PythonApplication2\envi2764' 
Cannot find file C:\Program Files\IronPython 2.7\include (bad symlink) 
New python executable in envi2764\bin\ipy.exe 
Also creating executable in envi2764\bin\python.exe 
Unhandled exception: 
Traceback (most recent call last): 
  File "C:\Program Files\IronPython 2.7\Lib\runpy.py", line 187, in run_module 
  File "C:\Program Files\IronPython 2.7\Lib\runpy.py", line 81, in _run_module_code 
  File "C:\Program Files\IronPython 2.7\Lib\runpy.py", line 72, in _run_code 
  File "C:\Program Files\IronPython 2.7\Lib\site-packages\virtualenv.py", line 2352, in <module> 
  File "C:\Program Files\IronPython 2.7\Lib\site-packages\virtualenv.py", line 816, in main 
  File "C:\Program Files\IronPython 2.7\Lib\site-packages\virtualenv.py", line 983, in create_environment 
  File "C:\Program Files\IronPython 2.7\Lib\site-packages\virtualenv.py", line 1416, in install_python 
AttributeError: 'module' object has no attribute 'symlink' 
Virtual environment was not created at 'c:\users\vslab1\source\repos\PythonApplication2\PythonApplication2\envi2764'. Exit code: 1
```

Issue #4783, opened 2018-10-11
Failure creating virtual env with base interpreter itself a virtualenv


huguesv:
Create a virtual env and register as global.
Create another virtual env based on that one.

We could remove those as an option for base interpreter. There's no way to differentiate based on interpreter configuraiton other than look on disk. Will that degrade performance too much for it to be worth it? Maybe add a flag on interpreter factory that is set when it is created. So at least that happens only once?

```
Virtual environment is being created at 'c:\users\vslab1\source\repos\PythonApplication3\PythonApplication3\env3764g-derived' 
Using base prefix 'C:\\Program Files (x86)\\Microsoft Visual Studio\\Shared\\Python37_64' 
c:\users\vslab1\source\repos\PythonApplication1\PythonApplication1\env3764g\lib\site-packages\virtualenv.py:1087: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses 
  import imp 
New python executable in env3764g-derived\Scripts\python.exe 
Installing setuptools, pip... 
Traceback (most recent call last): 
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\runpy.py", line 193, in _run_module_as_main 
  Complete output from command c:\users\vslab1\sour...d\Scripts\python.exe -c "import sys, pip; sys...d\"] + sys.argv[1:]))" setuptools pip: 
  Traceback (most recent call last): 
  File "<string>", line 1, in <module> 
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load 
    "__main__", mod_spec) 
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked 
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\runpy.py", line 85, in _run_code 
  File "<frozen importlib._bootstrap>", line 668, in _load_unlocked 
  File "<frozen importlib._bootstrap>", line 638, in _load_backward_compatible 
  File "c:\users\vslab1\source\repos\PythonApplication1\PythonApplication1\env3764g\lib\site-packages\virtualenv_support\pip-6.1.1-py2.py3-none-any.whl\pip\__init__.py", line 13, in <module> 
    exec(code, run_globals) 
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load 
[...]

PR #5159, opened 2019-03-13
[WIP] Filters interpreters from "Add Environment" window

### LINKED ISSUE #4781: Performance is bad in virtual env page when "Base interpreter" is IronPython or Anaconda
To 2.7 64-bit: about 49s 
To 2.7 32-bit: about 39s 
To Anaconda 5.3.0 (3.7, 64-bit): about 30s

RaymonGulati1:
Remove the ability to create virtual environment based on certain base interpreters. 

Removed base interpreters are Minconda/Anaconda interpreter and virtual environments, python virtual environment, and iron python interpreter. 

Fix #4781 #4783 #4784

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
+++ b/Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
@@ -44,7 +44,7 @@ namespace Microsoft.PythonTools.Environments {
             var res = knownProviders.Configurations
                 .Where(PythonInterpreterFactoryExtensions.IsUIVisible)
                 .Where(PythonInterpreterFactoryExtensions.IsRunnable)
-                .Where(configuration => FilterInterpreters(configuration, includeVirtualEnv, includeCondaEnv, includeIronPythonEnv))
+                .Where(configuration => FilterInterpreter(configuration, includeVirtualEnv, includeCondaEnv, includeIronPythonEnv))
                 .OrderBy(c => c.Description)
                 .ThenBy(c => c.Version)
                 .Select(c => new InterpreterView(c.Id, c.Description, c.InterpreterPath, c.Version.ToString(), c.ArchitectureString, project));
@@ -62,14 +62,14 @@ namespace Microsoft.PythonTools.Environments {
             return res;
         }
 
-        private static bool FilterInterpreters(
+        internal static bool FilterInterpreter(
             InterpreterConfiguration interpreterConfiguration,
             bool includeVirtualEnv,
             bool includeCondaEnv,
             bool includeIronPythonEnv
         ) {
 
-            if (!includeVirtualEnv && VirtualEnv.IsVirtualEnvironment(interpreterConfiguration.GetPrefixPath())) {
+            if (!includeVirtualEnv && VirtualEnv.IsPythonVirtualEnv(interpreterConfiguration.GetPrefixPath())) {
                 return false;
             }
 
--- a/Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
+++ b/Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
@@ -29,8 +29,11 @@ namespace Microsoft.PythonTools.Environments {
 
         public static IEnumerable<InterpreterView> GetInterpreters(
             IServiceProvider serviceProvider,
-            PythonProjectNode project, 
-            bool onlyGlobalEnvironments = false
+            PythonProjectNode project,
+            bool onlyGlobalEnvironments = false,
+            bool includeVirtualEnv = true,
+            bool includeCondaEnv = true,
+            bool includeIronPythonEnv = true
         ) {
             if (serviceProvider == null) {
                 throw new ArgumentNullException(nameof(serviceProvider));
@@ -41,6 +44,7 @@ namespace Microsoft.PythonTools.Environments {
             var res = knownProviders.Configurations
                 .Where(PythonInterpreterFactoryExtensions.IsUIVisible)
                 .Where(PythonInterpreterFactoryExtensions.IsRunnable)
+                .Where(configuration => FilterInterpreters(configuration, includeVirtualEnv, includeCondaEnv, includeIronPythonEnv))
                 .OrderBy(c => c.Description)
                 .ThenBy(c => c.Version)
                 .Select(c => new InterpreterView(c.Id, c.Description, c.InterpreterPath, c.Version.ToString(), c.ArchitectureString, project));
@@ -56,6 +60,28 @@ namespace Microsoft.PythonTools.Environments {
             }
 
             return res;
+        }
+
+        private static bool FilterInterpreters(
+            InterpreterConfiguration interpreterConfiguration,
+            bool includeVirtualEnv,
+            bool includeCondaEnv,
+            bool includeIronPythonEnv
+        ) {
+
+            if (!includeVirtualEnv && VirtualEnv.IsVirtualEnvironment(interpreterConfiguration.GetPrefixPath())) {
+                return false;
+            }
+
+            if (!includeCondaEnv && CondaUtils.IsCondaEnvironment(interpreterConfiguration.GetPrefixPath())) {
+                return false;
+            }
+
+            if (!includeIronPythonEnv && interpreterConfiguration.IsIronPython()) {
+                return false;
+            }
+
+            return true;
         }
 
         private static string FormatInvalidId(string id) {
--- a/Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
+++ b/Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
@@ -38,11 +38,10 @@ namespace Microsoft.PythonTools.Environments {
             }
 
             var knownProviders = serviceProvider.GetComponentModel().GetService<IInterpreterRegistryService>();
-
             var res = knownProviders.Configurations
                 .Where(PythonInterpreterFactoryExtensions.IsUIVisible)
                 .Where(PythonInterpreterFactoryExtensions.IsRunnable)
-                .Where(configuration => ExcludeInterpreter(configuration, excludeInterpreters) == false)
+                .Where(configuration => !ExcludeInterpreter(configuration, excludeInterpreters))
                 .OrderBy(c => c.Description)
                 .ThenBy(c => c.Version)
                 .Select(c => new InterpreterView(c.Id, c.Description, c.InterpreterPath, c.Version.ToString(), c.ArchitectureString, project));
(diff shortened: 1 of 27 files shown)


Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
```diff
@@ -58,6 +62,28 @@ sealed class InterpreterView : DependencyObject {
             return res;
         }
 
+        private static bool FilterInterpreters(
+            InterpreterConfiguration interpreterConfiguration,
+            bool includeVirtualEnv,
+            bool includeCondaEnv,
+            bool includeIronPythonEnv
+        ) {
```
huguesv: Method name is plural, but it only processes one item. I'd go with something like `ShouldIncludeConfiguration`, or at least something not plural.

Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
```diff
@@ -425,5 +425,13 @@ IInterpreterRegistryService service
             }
             return libPath;
         }
+        
+        internal static bool IsVirtualEnvironment(string prefixPath) {
+            if(string.IsNullOrEmpty(prefixPath)) {
+                return false;
+            }
+
+            return (File.Exists(Path.Combine(prefixPath, "pyvenv.cfg")) || File.Exists(Path.Combine(FindLibPath(prefixPath) ?? "", "orig-prefix.txt")));
```
huguesv: If `FindLibPath` is `null`, you'll be checking for "orig-prefix.txt" in the current directory (which is unpredictable). Best to do this in multiple statements, don't do the `File.Exists` if folder is `null`.

Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
```diff
@@ -425,5 +425,13 @@ IInterpreterRegistryService service
             }
             return libPath;
         }
+        
+        internal static bool IsVirtualEnvironment(string prefixPath) {
+            if(string.IsNullOrEmpty(prefixPath)) {
```
huguesv: Style: Use space between `if` and `(`

Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
RaymonGulati1: Will be updated in next commit

RaymonGulati1:
@huguesv Ready for round 2

Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
```diff
@@ -29,8 +29,9 @@ sealed class InterpreterView : DependencyObject {
 
         public static IEnumerable<InterpreterView> GetInterpreters(
             IServiceProvider serviceProvider,
-            PythonProjectNode project, 
-            bool onlyGlobalEnvironments = false
+            PythonProjectNode project,
+            bool onlyGlobalEnvironments = false,
```
huguesv: `onlyGlobalEnvironments` is a form of exclusion so it feels like it should be replaced by the new parameter.

Python/Tests/Core/EnvironmentListTests.cs
```diff
@@ -771,6 +772,121 @@ public class EnvironmentListTests {
...
+
+            Assert.IsNotNull(pythonInterpreter, "Python 3.6 or 3.7 not found");
+            FilterPythonInterpreterEnv(pythonInterpreter);
+        }
+
+        [TestMethod, Priority(0)]
+        public void FilterConda2Interpreter() {
+            PythonVersion condaInterpreter = PythonPaths.Anaconda27_x64 ?? PythonPaths.Anaconda27;
+
+            Assert.IsNotNull(condaInterpreter, "Anaconda 2.6 or 2.7 not found");
```
huguesv: In all these tests, if the machine doesn't have the required Python installed, we don't want to fail.

Use Assert.Inconclusive(), because we can't tell whether it would fail or not if it would be present.

This is what we do in all other tests.

Python/Tests/Core/WorkspaceInterpreterFactoryTests.cs
```diff
@@ -211,9 +211,9 @@ public class WorkspaceInterpreterFactoryTests {
             return new WorkspaceTestHelper.MockWorkspaceContext(new WorkspaceTestHelper.MockWorkspace(workspacePath));
         }
 
-        private static void CreatePythonVirtualEnv(string pythonInterpreterPath, string workspacePath, string envName) {
+        internal static void CreatePythonVirtualEnv(string pythonInterpreterPath, string workspacePath, string envName, string environmentCreationParameter = "venv") {
```
huguesv: `virtualEnvModule` would be a more descriptive name for what it is you are passing in, rather than `environmentCreationParameter`

Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
```diff
@@ -425,5 +425,19 @@ IInterpreterRegistryService service
...
             return libPath;
         }
+        
+        internal static bool IsPythonVirtualEnv(string prefixPath) {
+            if (string.IsNullOrEmpty(prefixPath)) {
+                return false;
+            }
+            
+            string libPath = FindLibPath(prefixPath);
+            if(libPath == null) {
```
huguesv: space after if

Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
```diff
@@ -425,5 +425,19 @@ IInterpreterRegistryService service
...
+                return false;
+            }
+
+            //pyenv.cfg detects python 3 and orig-prefix.txt detects python 2
+            return (File.Exists(Path.Combine(prefixPath, "pyvenv.cfg")) || File.Exists(Path.Combine(libPath, "orig-prefix.txt")));
```
huguesv: Python 3 that has a pyvenv.cfg but no lib folder will return false because it never gets to this line. Maybe that's fine, if lib folder isn't there it's probably not a venv after all.

Python/Tests/Core/EnvironmentListTests.cs
```diff
@@ -771,6 +772,121 @@ public class EnvironmentListTests {
             TestFileNameEllipsis("", "", "", "");
         }
 
+        [TestMethod, Priority(0)]
+        public void FilterPython2Interpreter() {
```
huguesv: We try to keep the version at the very end of the test class name / method name.

Since the method you are actually testing is "ExcludeInterpreter", I'd probably name them "ExcludeInterpreterPython2", "ExcludeInterpreterPython3", "ExcludeInterpreterConda2", etc.

Python/Tests/Core/EnvironmentListTests.cs
```diff
@@ -771,6 +772,121 @@ public class EnvironmentListTests {
...
+                                                PythonPaths.Anaconda36;
+
+            Assert.IsNotNull(condaInterpreter, "Anaconda 3.6 or 3.7 not found");
+            FilterCondaInterpreter(condaInterpreter);
+        }
+
+        [TestMethod, Priority(0)]
+        public void FilterIronPython2Interpreter() {
+            PythonVersion ironPythonInterpreter = PythonPaths.IronPython27_x64 ?? PythonPaths.IronPython27;
+            Assert.IsNotNull(ironPythonInterpreter, "Iron python 2.7 not found");
```
huguesv: IronPython always in one word

Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
```diff
@@ -41,6 +42,7 @@ sealed class InterpreterView : DependencyObject {
             var res = knownProviders.Configurations
                 .Where(PythonInterpreterFactoryExtensions.IsUIVisible)
                 .Where(PythonInterpreterFactoryExtensions.IsRunnable)
+                .Where(configuration => ExcludeInterpreter(configuration, excludeInterpreters) == false)
```
huguesv: !ExcludeInterpreter (no need for `== false`)

Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
```diff
@@ -58,6 +60,35 @@ sealed class InterpreterView : DependencyObject {
...
+                return true;
+            }
+
+            return false;
+        }
+
+        [Flags]
+        internal enum InterpreterFilter {
+            None = 0,
+            ExcludePythonEnv = 1,
```
huguesv: ExcludePythonEnv => ExcludeVirtualEnv

Python/Tests/Core/EnvironmentListTests.cs
huguesv: Actually, we have an extension method for that. Just call `condaInterpreter.AssertInstalled()`

Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
RaymonGulati1: onlyGlobalEnvironments requires IInterpreterRegistryService which is not accepted by ExcludeInterpreter. I could modify ExcludeInterpreter, but it would be somewhat weird (IInterpreterRegistry only passed in case of ExcludeGlobalInterpreter) and I'd have to write additional tests for ExcludeGlobalInterpreter cases. I don't think the cost vs benefit is justified when everything works fine right now.

Python/Tests/Core/WorkspaceInterpreterFactoryTests.cs
```diff
@@ -206,19 +206,20 @@ public class WorkspaceInterpreterFactoryTests {
...
             File.WriteAllText(Path.Combine(workspacePath, "app.py"), string.Empty);
 
-            CreatePythonVirtualEnv(python.InterpreterPath, workspacePath, envName);
+            virtualEnvModule(python.InterpreterPath, workspacePath, envName);
 
             return new WorkspaceTestHelper.MockWorkspaceContext(new WorkspaceTestHelper.MockWorkspace(workspacePath));
         }
 
-        private static void CreatePythonVirtualEnv(string pythonInterpreterPath, string workspacePath, string envName) {
+        internal static void virtualEnvModule(string pythonInterpreterPath, string workspacePath, string envName, string environmentCreationParameter = "venv") {
```
huguesv: `CreatePythonVirtualEnv` was fine for the function name. It's the parameter name that should be `virtualEnvModule` instead of `environmentCreationParameter`

RaymonGulati1: Corrected.

PR #5161, opened 2019-03-13
Add package install and environment creation info bars support for workspaces

### LINKED ISSUE #4851: Open Folder: Support install packages and create environment info bars
When a Python file is opened when in Folder View, do a similar check for environments.txt/environment.yml that is currently done for projects.

huguesv:
Fix #4851


Python/Product/PythonTools/PythonTools/Project/CondaEnvCreateInfoBar.cs
```diff
@@ -27,51 +29,62 @@
...
-                // Propose to recreate one of the conda references, since they are all missing
-                existingName = condaNotFoundNames.First();
-            } else if (!foundConda.Any() && !string.IsNullOrEmpty(yamlPath)) {
-                // Propose to create a new one, since there's a yaml file and no conda references
-                existingName = null;
-            } else {
-                // Nothing to do
-                return;
+            var yamlPath = Project?.GetEnvironmentYmlPath() ?? Workspace?.GetEnvironmentYmlPath();
+            string existingName = null;
```
RaymonGulati1: if(Project != null) is false, then this string will remain null and may cause problems in "AddEnvironmentDialog.ShowAddCondaEnvironmentDialogAsync"

Python/Product/PythonTools/PythonTools/Project/CondaEnvCreateInfoBar.cs
```diff
@@ -81,7 +94,18 @@ public CondaEnvCreateInfoBar(PythonProjectNode projectNode)
                         Action = CondaEnvCreateInfoBarActions.Create,
```
RaymonGulati1: The CheckAsync function is unusually big. Please split it into smaller chunks for easier readability. For example, create the "Action create" and "Action projectIgnore" in a different function and pass them back.

Python/Product/PythonTools/PythonTools.csproj
```diff
@@ -391,6 +391,7 @@
       <DependentUpon>AddEnvironmentDialog.xaml</DependentUpon>
```
RaymonGulati1: I've put some review comments in CondaEnvCreateInfoBar.cs, but they apply to the other 2 classes as well. Such as "The "Project?.Caption" contains a property "buildProject" which can be null..." and "Declare the "projectOrWorkspacename..."

Python/Product/PythonTools/PythonTools/Project/CondaEnvCreateInfoBar.cs
```diff
@@ -27,51 +29,62 @@
 
 namespace Microsoft.PythonTools.Project {
     internal sealed class CondaEnvCreateInfoBar : PythonProjectInfoBar {
-        public CondaEnvCreateInfoBar(PythonProjectNode projectNode)
-            : base(projectNode) {
+        public CondaEnvCreateInfoBar(IServiceProvider site, PythonProjectNode projectNode, IPythonWorkspaceContext workspace)
+            : base(site, projectNode, workspace) {
         }
 
         public override async Task CheckAsync() {
```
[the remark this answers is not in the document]
RaymonGulati1: This function contains the logic for two cases (Project is null or workspace is null) which makes the code error prone and difficult to follow. How about creating two functions to separate the logic?

Python/Product/PythonTools/PythonTools/Project/PackageInstallInfoBar.cs
```diff
@@ -91,7 +93,9 @@ public PackageInstallInfoBar(PythonProjectNode projectNode)
                         Action = PackageInstallInfoBarActions.Ignore,
```
RaymonGulati1: I want to keep the log events for workspace and projects separately. So, one log event for projects and one logging event for workspaces. This will give us another datapoint about how many people are using project vs workspace, as well the similarities or dissimilarities. 

Doesn't have to be in this PR--file an issue?

Python/Product/PythonTools/PythonTools.csproj
RaymonGulati1: In the future, it would also be nice if you could separate the component changes into different components. One commit for package install, another for virtual env creation, etc. It make it easier to code review :smile:

PR #5183, opened 2019-03-21
Add localization tool workaround

### LINKED ISSUE #5182: CSS Authoring breaks the localization validation process
Hi,
As part of our validation process we noticed that the validation chokes on the Microsoft.PythonTools.WebRole.dll. This is because this DLL has some CSS fragment that are not supported by our tool set. You are using a start which is not supported by our tool set currently.
Thank you for changing the authoring so that we can see the selector (Replacing * with the appropriate tag.

Thank you very much

    <style type="text/css">
        * {
            font-family: sans-serif;
        }
        .secret {

huguesv:
Fix #5182 

Update embedded CSS in html resource to not use * selector, which was breaking some internal tools used by our localization team.

(I've sent this fix to them for validation and it solves their issue)

### DIFF
--- a/Python/Product/WebRole/WebSocketProxy.html
+++ b/Python/Product/WebRole/WebSocketProxy.html
@@ -3,16 +3,6 @@
 <head>
     <meta charset="utf-8" />
     <title>PTVS remote debugging proxy for Microsoft Azure Web Sites</title>
-    <style type="text/css">
-        html
-        {
-            font-family: sans-serif;
-        }
-        .secret {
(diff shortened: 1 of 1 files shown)


huguesv:
I got build errors from CI. Not ready to merge.

huguesv:
Decided to remove it altogether. The build now passes.

It's a rarely seen debug page in deprecated functionality (Python Azure App Service running on Windows). Customizing the font in which the page renders is not important.

PR #5195, opened 2019-03-26
Enables set next statement feature.

karthiknadig:


huguesv:
Our UI tests caught a regression with breakpoints on Python 2.7 as well as problems with set next statement (adapter bug).

Let's wait for the breakpoint bug fix, and run the tests again before merging this.

karthiknadig:
This fix to ptvsd should improve the behavior, but does not entirely fix it.
https://github.com/Microsoft/ptvsd/pull/1267

huguesv:
We'll close and do another PR when the issues are fixed.

In the new PR, you should enable `DebuggerUITests.DebugProjectUITests.SetNextLine` for VSC debugger, ie. remove this code from the top of the test:
```
            if (useVsCodeDebugger) {
                // https://github.com/Microsoft/ptvsd/issues/18
                // This is not implemented in the experimental debugger
                Assert.Inconclusive();
                return;
            }

```

PR #5244, opened 2019-04-17
Find interpreters that are installed from the store

### LINKED ISSUE #5189: Detect Python when installed as a Store app
When Python is installed as a Store app, it does not expose the normal registry keys (because Store apps aren't allowed to write to the regular registry). It _does_ expose the key when queried directly (e.g. as `HKCU\Software\Python\PythonCore\3.7`), but enumerating `HKCU\Software\Python\PythonCore` key does not find it (because of deliberate limitations in Windows).

One option is to add an explicit query for 3.7 and later keys, which raises the question of "how much later" (probably stop at 3.9 for now I'd say). Alternatively, it could use the [`FindPackagesForUser`](https://docs.microsoft.com/en-us/uwp/api/windows.management.deployment.packagemanager.findpackagesforuser) API and checking for publisher id `qbz5n2kfra8p0` (alternatively, Package Family Name `PythonSoftwareFoundation.Python.3.7_qbz5n2kfra8p0`, but note that this includes the version number, so it would have to be a partial match). The install location should be available as part of this query, and `python.exe` is in there.

huguesv:
Always look for fixed registry keys in addition to enumerating sub keys, since store apps registry keys are not returned when enumerating.

Fix #5189

### DIFF
--- a/Python/Product/VSInterpreters/Interpreter/PythonRegistrySearch.cs
+++ b/Python/Product/VSInterpreters/Interpreter/PythonRegistrySearch.cs
@@ -33,6 +33,15 @@ namespace Microsoft.PythonTools.Interpreter {
         public const string CompanyPropertyKey = "Company";
         public const string SupportUrlPropertyKey = "SupportUrl";
 
+        // When Python is installed as a Store app, it does not expose
+        // the normal registry keys (because Store apps aren't allowed
+        // to write to the regular registry). It does expose the key
+        // when queried directly (e.g. as HKCU\Software\Python\PythonCore\3.7),
+        // but enumerating HKCU\Software\Python\PythonCore key does not
+        // find it (because of deliberate limitations in Windows).
+        private static readonly string[] StoreAppTags = new[] { "3.7", "3.8", "3.9" };
+        private static readonly string[] StoreAppCompanies = new[] { PythonCoreCompany };
+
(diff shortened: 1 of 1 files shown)


Python/Product/VSInterpreters/Interpreter/PythonRegistrySearch.cs
```diff
@@ -33,6 +33,15 @@ class PythonRegistrySearch {
         public const string CompanyPropertyKey = "Company";
         public const string SupportUrlPropertyKey = "SupportUrl";
 
+        // When Python is installed as a Store app, it does not expose
+        // the normal registry keys (because Store apps aren't allowed
+        // to write to the regular registry). It does expose the key
+        // when queried directly (e.g. as HKCU\Software\Python\PythonCore\3.7),
+        // but enumerating HKCU\Software\Python\PythonCore key does not
+        // find it (because of deliberate limitations in Windows).
+        private static readonly string[] StoreAppTags = new[] { "3.7", "3.8", "3.9" };
```
RaymonGulati1: What about Python 3.6? 2.6? 2.7?

huguesv: We'll never have Windows store versions of those.

PR #5247, opened 2019-04-19
Install Python in Conda Environment creation with no packages

### LINKED ISSUE #4799: Add more input validation to add conda environment
The "no packages" in particular will probably cause confusion. This will lead to an environment without python, the user will wonder why it doesn't appear in the environment window.

It will be difficult without waiting for the conda preview to determine if the environment will have python or not, but one thing is for sure is that if packages text box is empty, it will not have python.

RaymonGulati1:
Fix #4799

### DIFF
--- a/Python/Product/Common/Strings.resx
+++ b/Python/Product/Common/Strings.resx
@@ -2264,7 +2264,7 @@ Expand selection to the full expression?</value>
   <data name="AddEnvironmentCancelButton" xml:space="preserve">
     <value>_Cancel</value>
   </data>
-    <data name="AddEnvironmentValidationErrors" xml:space="preserve">
+  <data name="AddEnvironmentValidationErrors" xml:space="preserve">
     <value>Some fields have invalid values.
 
 {0}</value>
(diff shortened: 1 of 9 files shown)


Python/Product/Common/Strings.resx
```diff
@@ -2441,7 +2441,7 @@ Expand selection to the full expression?</value>
     <value>On_e or more Anaconda package names</value>
   </data>
   <data name="AddCondaPackagesWatermark" xml:space="preserve">
-    <value>ex. numpy pandas</value>
```
huguesv: lowercase was correct, please undo

Python/Product/Common/Strings.resx
```diff
@@ -2450,7 +2450,7 @@ Expand selection to the full expression?</value>
     <value>Add packages</value>
   </data>
   <data name="AddCondaPackagesHelpText" xml:space="preserve">
-    <value>List of Conda packages to install when creating the Conda environment. Entries should be separated by spaces. Example: numpy pandas</value>
+    <value>List of Conda packages to install when creating the Conda environment. Entries should be separated by spaces. Example: Numpy Pandas</value>
```
huguesv: same - lowercase was correct

RaymonGulati1: Fixed

Python/Product/Common/Strings.resx
RaymonGulati1: Fixed

Issue #5233, opened 2019-04-12
Interactive mode hangs: Windows 10, VS 2019, 64 bit


sriksrid:
Hi

I have tried two different installations of Python 3 (Aanconda3 2019.03, and a barebones 3.7 with ipython, pandas, matplotlib installed) on Windows 10 (clean install). In both cases, if I set the checkbox "Use IPython interactive mode", the interactive window hangs as soon as I type any command (or even no command, just an enter to increment the counter). Resetting the process outputs three lines:
```
Resetting Python state.
[IPKernelApp] WARNING | Parent appears to have exited, shutting down.
The interactive Python process has exited.
>> {I press enter}
In [2]: {I press enter}
hangs.
```

I also noticed that the module dropdown (in non IPython mode) shows "__main__", while the IPython mode doesn't show this drop down at all. How should I use the IPython interactive mode with VS 2019?

Thanks.

huguesv:
Thanks for letting us know about this.

It works with Anaconda 2018.12 but now with 2019.03 I see the behavior you described.

huguesv:
The problem appears when using pyzmq==18.0.1 (currently the latest)

You can downgrade that package to 17.1.2 and things appear to work again.

sriksrid:
Thank you. That works great.

PR #5252, opened 2019-04-23
fix for repl hang when using  with pyzmq => 18.0.0

bschnurr:
- Both execute(..) and get_shell_msg(..) act on the same shell socket and having a blocking call to get_shell_msg(..) no longer works with the newer pyzmq.
- temporary fix was to downgrade pyzmq to 17.1.3
- Changed get_shell_msg(..) to use a timeout instead of blocking.

related to bug  #5233

### DIFF
--- a/Python/Product/PythonTools/ptvsd/repl/jupyter_client.py
+++ b/Python/Product/PythonTools/ptvsd/repl/jupyter_client.py
@@ -343,26 +343,30 @@ class JupyterClientBackend(ReplBackend):
                 if self.exit_requested:
                     break
 
-                m = Message(client.get_shell_msg(block=True))
-                msg_id = m.msg_id
-                msg_type = m.msg_type
+                try:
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/ptvsd/repl/jupyter_client.py
```diff
@@ -343,26 +343,30 @@ def __shell_threadproc(self, client):
...
+                        self.send_prompt('\n' + ps1, ps2, allow_multiple_statements=True)
+
+                    parent_id = m.parent_header['msg_id', None]
+                    if parent_id:
+                        on_reply = on_replies.pop((parent_id, msg_type), ())
+                        for callable in on_reply:
+                            callable(m)
+                except Empty:
+                    pass
+                except:
```
huguesv: since you're not doing anything but re-raise, you can just delete this last except block

bschnurr:
https://github.com/jupyter/jupyter_client/issues/436

Issue #5197, opened 2019-03-28
Fix the newly detected Roslyn Errors


RaymonGulati1:
Roslyn has detected some new errors related to ExtractLambdasFromXml.cs and we need to resolve them. 

For example, change "} catch (Exception ex)" to "catch (Exception ex) when (!ex.IsCriticalException())"

[13.txt](https://github.com/Microsoft/PTVS/files/3019343/13.txt)

PR #5266, opened 2019-04-26
Fix the newly detected Roslyn Errors #5197

bschnurr:
-had to add a workaround msbuild property due to build machine using VS 2017 with msbuild 15.0

first main issue
 ##[error]E:\A\_work\139\s\Python\Product\Cookiecutter\ExtractLambdasFromXaml.cs(114,15): Error CA1031: Modify 'Execute' to catch a more specific exception type, or rethrow the exception. [E:\A\_work\139\s\Common\Tools\BuildTasks\BuildTasks.csproj]
 ExtractLambdasFromXaml.cs(114,15): error CA1031: Modify 'Execute' to catch a more specific exception type, or rethrow the exception. [E:\A\_work\139\s\Common\Tools\BuildTasks\BuildTasks.csproj] [E:\A\_work\139\s\Python\Product\Cookiecutter\Cookiecutter_ljs5eifo_wpftmp.csproj]

second issue with Roslyn itself
"CSC(0,0): Error AD0001: Analyzer 'Microsoft.NetCore.Analyzers.Runtime.DisposableFieldsShouldBeDisposed' threw an exception of type 'System.InvalidOperationException' with message 'Feature 'Flow-Analysis' is disabled.'. [E:\A\_work\529\s\Common\Tools\BuildTasks\BuildTasks.csproj]"

Roslyn Policy run
https://dev.azure.com/mseng/DSTools/_build/results?buildId=9340237&view=logs&s=96ac2280-8cb4-5df5-99de-dd2da759617d

### DIFF
--- a/Build/Common.Build.CSharp.settings
+++ b/Build/Common.Build.CSharp.settings
@@ -84,6 +84,7 @@
     <SignManifests>false</SignManifests>
     <AssemblyOriginatorKeyFile>$(TargetsPath)\FinalPublicKey.snk</AssemblyOriginatorKeyFile>
     <DelaySign>true</DelaySign>
+    <Features>$(Features);flow-analysis</Features> 
   </PropertyGroup>
 
 </Project>
--- a/Common/Tools/BuildTasks/ExtractLambdasFromXaml.cs
+++ b/Common/Tools/BuildTasks/ExtractLambdasFromXaml.cs
@@ -113,7 +113,7 @@ namespace Microsoft.VisualStudioTools.BuildTasks {
                 return GenerateOutput();
             } catch (Exception ex) {
                 LogError(null, 0, 0, ex.Message);
-                return false;
+                throw;
             }
         }
 


Common/Tools/BuildTasks/ExtractLambdasFromXaml.cs
```diff
@@ -113,7 +113,7 @@ private struct LambdaInfo {
                 return GenerateOutput();
             } catch (Exception ex) {
                 LogError(null, 0, 0, ex.Message);
-                return false;
+                throw;
```
bschnurr: Roslyn says to be more specific or to rethrow. Since this is used in the build I choose to log and rethrow and this avoids bringing higher level projects into the buildtask.proj

Build/Common.Build.CSharp.settings
```diff
@@ -84,6 +84,7 @@
     <SignManifests>false</SignManifests>
     <AssemblyOriginatorKeyFile>$(TargetsPath)\FinalPublicKey.snk</AssemblyOriginatorKeyFile>
     <DelaySign>true</DelaySign>
+    <Features>$(Features);flow-analysis</Features>
```
bschnurr: this is my workaround after reading this post https://github.com/dotnet/roslyn-analyzers/issues/2244. I think another solution would be to have all projects include Microsoft.CodeAnalysis.FxCopAnalyzers directly as a nuget package because that package will add this build property automatically.

In our case the build pipeline task is the one who is grabbing Microsoft.CodeAnalysis.FxCopAnalyzers so we don't get the benefit of the property being added to our projects.

Common/Tools/BuildTasks/ExtractLambdasFromXaml.cs
RaymonGulati1: I would prefer it if the code logged specific exceptions (IsCriticalException(...)) and returned false. If there is an exception not caught in the specific exception list, then the exception can propagate up the call stack. This is consistent with existing behavior of other code in the file.

huguesv: It's okay to throw when implementing Execute.
https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.itask.execute?view=netframework-4.8#Microsoft_Build_Framework_ITask_Execute

bschnurr: Yup.. just did a test. changed the task to throw.

PR #5290, opened 2019-05-08
Fix #5219: Attach to a running python.exe does not work

### LINKED ISSUE #5219: Attach to a running python.exe does not work
VS Version: 28805.205.d16.1
Python Tool Version: 16.1.19081.2

Repro step:
1.Create a python application project.
2.Typing code:
input('>')
print('hello world')
3.Set breakpoint on the line "print....."
4.Set Python 3.7(64bit) as current environment.
5.Ctrl+F5 to debug>Attach to Process, select the appropriate python.exe>Attach

Except result:
Attach process successfully , and we can type enter in the console to stop at the breakpoint.

actual result:
Straight out, no console input, no breakpoint hit
image

note: Not every attach fails, which happens occasionally. If it can be reproduced, you can try to switch the python environment and try again.

int19h:
Account for terminating null when allocating string buffer.

### DIFF
--- a/Python/Product/PyDebugAttach/PyDebugAttach.cpp
+++ b/Python/Product/PyDebugAttach/PyDebugAttach.cpp
@@ -783,7 +783,7 @@ bool LoadAndEvaluateCode(
 
     dictSetItem(globalsDict, "__builtins__", getBuiltins());
     auto size = WideCharToMultiByte(CP_UTF8, 0, filePath, (DWORD)wcslen(filePath), NULL, 0, NULL, NULL);
-    char* filenameBuffer = new char[size];
+    char* filenameBuffer = new char[size + 1];
     if (WideCharToMultiByte(CP_UTF8, 0, filePath, (DWORD)wcslen(filePath), filenameBuffer, size, NULL, NULL) != 0) {
         filenameBuffer[size] = 0;
         dictSetItem(globalsDict, "__file__", strFromString(filenameBuffer));


Python/Product/PyDebugAttach/PyDebugAttach.cpp
```diff
@@ -783,7 +783,7 @@ bool LoadAndEvaluateCode(
 
     dictSetItem(globalsDict, "__builtins__", getBuiltins());
     auto size = WideCharToMultiByte(CP_UTF8, 0, filePath, (DWORD)wcslen(filePath), NULL, 0, NULL, NULL);
-    char* filenameBuffer = new char[size];
+    char* filenameBuffer = new char[size + 1];
     if (WideCharToMultiByte(CP_UTF8, 0, filePath, (DWORD)wcslen(filePath), filenameBuffer, size, NULL, NULL) != 0) {
         filenameBuffer[size] = 0;
```
huguesv: According to docs, the different strFromString implementations take a copy of the buffer, which means we'd also be missing a `delete[] filenameBufer` after `dictSetItem`?

int19h: Looks like it. I would consider this change to be minfix that takes care of the crash - ideally we should re-revert the static analysis fixes, and then do the same thing here (i.e. use `std::string` etc) to avoid all this mess. This would have also caught this bug immediately in debug builds.

huguesv: Yeah, not too concerned about those few bytes leaking for now. Let's do the minfix for 16.1.

PR #5301, opened 2019-05-14
Update NuGet packages and reference assemblies.

### LINKED ISSUE #5297: Use 16.x NuGet packages and remove checked in assemblies
Uncomment and update version for packages in `Build\16.0\packages.config`, remove the assemblies from `Python\References\Dev16.0`

huguesv:
Fix #5297


Python/Tests/DjangoUITests/DjangoUITests.csproj
```diff
@@ -85,11 +85,6 @@
     <Reference Include="Microsoft.VisualStudio.Threading">
       <SpecificVersion>False</SpecificVersion>
     </Reference>
-    <Reference Include="Microsoft.Web.Core" Condition=" '$(VSTarget)' == '15.0' " />
```
RaymonGulati1: The references for 15.0 have been removed, but the build pipelines still use 15.0 VS. Can you confirm if everything works when building with 15.0?

Same for the other changes.

Build/Common.Build.CSharp.targets
```diff
@@ -19,6 +19,8 @@
       <_NugetAssemblies Include="$(PackagesPath)\Microsoft.VisualStudio.*\lib\net45\Microsoft.VisualStudio.*.dll" />
       <_NugetAssemblies Include="$(PackagesPath)\Microsoft.VisualStudio.*\lib\net46\Microsoft.VisualStudio.*.dll" />
       <_NugetAssemblies Include="$(PackagesPath)\Microsoft.VisualStudio.*\lib\net461\Microsoft.VisualStudio.*.dll" />
+      <_NugetAssemblies Include="$(PackagesPath)\Microsoft.VisualStudio.*\lib\net472\Microsoft.VisualStudio.*.dll" />
```
RaymonGulati1: What are these new assembles for?

Build/PreBuild.ps1
```diff
@@ -8,6 +8,7 @@ $need_symlink = @(
     "python2",
     "MicroBuild.Core",
     "Microsoft.VSSDK.BuildTools",
+    "Microsoft.VSSDK.Debugger.VSDConfigTool",
```
RaymonGulati1: Why?

Python/Tests/DjangoUITests/DjangoUITests.csproj
huguesv: It still builds successfully on my pipeline which uses pool VSEng-MicroBuildVS2017, see the latest build log from PTVS-GitHub-Huguesv-Dev16 pipeline.

We do not need to target VS 2017 anymore from master branch, so I'm not maintaining that anymore (on master or 16.x branches).

Build/Common.Build.CSharp.targets
huguesv: Some of the updated NuGet packages are distributing .NET 4.72 and .NET Standard builds, and since VS 2019 extensions target 4.72, that's what we end up pulling from NuGet. We need those entries in order to find those assemblies.

Build/PreBuild.ps1
huguesv: vsdconfig executable was pulled out of the build tools package and moved into its own package. See the email I sent yesterday.

PR #5317, opened 2019-05-21
Add diagnostics to report a problem

### LINKED ISSUE #5199: Automatically add PTVS diagnostics log files to Report a Problem
Feedback tool team is adding an extensibility point so we can add extension specific logs.

huguesv:
Fix #5199 

Write a diagnostics file and include it in the list of attachments when user does Send Feedback -> Report a Problem.

Add all of the Python options to the diagnostics.


Python/Product/PythonTools/PythonTools/DiagnosticsProvider.cs
```diff
@@ -176,6 +177,65 @@ internal sealed class DiagnosticsProvider {
...
+            writer.WriteLine("    UseSmartHistory: {0}", pyService.InteractiveOptions.UseSmartHistory);
+            writer.WriteLine();
+
+            writer.WriteLine("Debug Interactive Options:");
+            writer.WriteLine("    CompletionMode: {0}", pyService.DebugInteractiveOptions.CompletionMode);
+            writer.WriteLine("    LiveCompletionsOnly: {0}", pyService.DebugInteractiveOptions.LiveCompletionsOnly);
+            writer.WriteLine("    Scripts: {0}", pyService.DebugInteractiveOptions.Scripts);
+            writer.WriteLine("    UseSmartHistory: {0}", pyService.DebugInteractiveOptions.UseSmartHistory);
+            writer.WriteLine();
+
```
bschnurr: I wonder if it worth making these options available as a list?

huguesv: I don't foresee other scenarios needing to do this now, but if that comes up again we could introduce a list then.

PR #5327, opened 2019-05-23
Update django template editor to match web editor API

### LINKED ISSUE #5326: Update django IntelliSense for web editor breaking changes
Some of the internal APIs that we are using are changing, we need to update our workload at the same time as the web editor team.

huguesv:
Fix #5326


Python/Product/Django/Intellisense/TemplateCommandFactory.cs
```diff
@@ -14,17 +14,14 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+#if !DEV16_OR_LATER
```
ToddGrun: If you're not worrying about Dev15 builds, can you just get rid of this file?

huguesv: Agreed, I'll remove the django dev15 only code.

Python/Product/Django/DjangoEditorFactory.cs
```diff
@@ -16,11 +16,7 @@
 
 using System;
 using System.Runtime.InteropServices;
-#if DEV16_OR_LATER
```
MinKPark: does it mean we don't support less version than Dev16 or the package is just work for both? I wonder.

huguesv: The builds are specific to a version of VS.

We have the 15.9 branch which targets VS 2017.

The master branch only targets VS 2019.

Until VS 2019 was GA, it was useful to allow build/run master branch on VS 2017, but now that VS 2019 is officially out, we don't need to keep maintaining support for VS 2017 in master branch.

PR #5346, opened 2019-05-31
Fix minor test issues and upgrade FastCgiTests to newer Python versions

### LINKED ISSUE #5333: Fix minor unit test issues
VSTextContext tries to discover devenv.exe so it can run VS, but in a certain case, it looks for the wrong path. It looks in "IDE/Common7" whereas it should look in "Common7/IDE". 
image

DjangoUITests.csproj builds DebuggerUITests.csproj, but it should only reference it, not build it.
[
image
](url)

Change unit test name from "FastCgiTest" to "FastCgiTests" to be consistent with other names.

RaymonGulati1:
Fix #5333 

Ignore branch name


Python/Tests/FastCgi/FastCgiTests2x.cs
```diff
@@ -354,7 +354,7 @@ public class FastCgiTests {
 
         public virtual PythonVersion PythonVersion {
             get {
-                return PythonPaths.Python27 ?? PythonPaths.Python27_x64;
+                return PythonPaths.Python27_x64 ?? PythonPaths.Python27;
```
MinKPark: 64bit is our default?

RaymonGulati1: Correct

PR #5363, opened 2019-06-07
Remove tests using Python versions 2.6, 3.1-3.4. 

### LINKED ISSUE #5313: Update Python versions for Unit Tests
Update the tests for the newer version of Python (3.7) and remove python versions that are no longer supported (2.6, 3.1-3.4).

RaymonGulati1:
Also updating tests to use the latest version (where applicable).

Fix #5313

### DIFF
--- a/Python/Tests/Core.UI/EnvironmentUITests.cs
+++ b/Python/Tests/Core.UI/EnvironmentUITests.cs
@@ -57,7 +57,7 @@ namespace PythonToolsUITests {
             return app.SelectDefaultInterpreter(
                 PythonPaths.Python37 ?? PythonPaths.Python37_x64 ??
                 PythonPaths.Python36 ?? PythonPaths.Python36_x64 ??
-                PythonPaths.Python35 ?? PythonPaths.Python35_x64 ??
+                PythonPaths.Python35 ?? PythonPaths.Python35_x64
             );
         }
 
--- a/Python/Tests/Core.UI/EnvironmentUITests.cs
+++ b/Python/Tests/Core.UI/EnvironmentUITests.cs
@@ -55,13 +55,10 @@ namespace PythonToolsUITests {
 
         static DefaultInterpreterSetter InitPython3(PythonVisualStudioApp app) {
             return app.SelectDefaultInterpreter(
-                PythonPaths.Python37_x64 ??
-                PythonPaths.Python37 ??
-                PythonPaths.Python36_x64 ??
-                PythonPaths.Python36 ??
-                PythonPaths.Python35_x64 ??
-                PythonPaths.Python35
-            );;
+                PythonPaths.Python37 ?? PythonPaths.Python37_x64 ??
+                PythonPaths.Python36 ?? PythonPaths.Python36_x64 ??
+                PythonPaths.Python35 ?? PythonPaths.Python35_x64 ??
+            );
         }
 
         static EnvDTE.Project CreateTemporaryProject(VisualStudioApp app, [CallerMemberName] string projectName = null) {
--- a/Python/Tests/Core.UI/EnvironmentUITests.cs
+++ b/Python/Tests/Core.UI/EnvironmentUITests.cs
@@ -55,12 +55,13 @@ namespace PythonToolsUITests {
 
         static DefaultInterpreterSetter InitPython3(PythonVisualStudioApp app) {
             return app.SelectDefaultInterpreter(
-                PythonPaths.Python37 ?? PythonPaths.Python37_x64 ??
-                PythonPaths.Python36 ?? PythonPaths.Python36_x64 ??
-                PythonPaths.Python35 ?? PythonPaths.Python35_x64 ??
-                PythonPaths.Python34 ?? PythonPaths.Python34_x64 ??
-                PythonPaths.Python33 ?? PythonPaths.Python33_x64
-            );
+                PythonPaths.Python37_x64 ??
+                PythonPaths.Python37 ??
+                PythonPaths.Python36_x64 ??
+                PythonPaths.Python36 ??
+                PythonPaths.Python35_x64 ??
+                PythonPaths.Python35
+            );;
         }
 
         static EnvDTE.Project CreateTemporaryProject(VisualStudioApp app, [CallerMemberName] string projectName = null) {
@@ -369,7 +370,12 @@ namespace PythonToolsUITests {
         }
 
         public void ProjectAddExistingVEnvLocal(PythonVisualStudioApp app) {
-            var python = PythonPaths.Python37 ?? PythonPaths.Python36 ?? PythonPaths.Python35 ?? PythonPaths.Python34 ?? PythonPaths.Python33;
+            var python =    PythonPaths.Python37_x64 ??
+                            PythonPaths.Python37 ??
+                            PythonPaths.Python36_x64 ??
+                            PythonPaths.Python36 ??
+                            PythonPaths.Python35_x64 ??
+                            PythonPaths.Python35;
             python.AssertInstalled();
 
             var project = CreateTemporaryProject(app);
@@ -391,7 +397,12 @@ version = 3.{1}.0", python.PrefixPath, python.Version.ToVersion().Minor));
         }
 
         public void ProjectAddCustomEnvLocal(PythonVisualStudioApp app) {
-            var python = PythonPaths.Python37 ?? PythonPaths.Python36 ?? PythonPaths.Python35 ?? PythonPaths.Python34 ?? PythonPaths.Python33;
+            var python =    PythonPaths.Python37_x64 ??
+                            PythonPaths.Python37 ??
+                            PythonPaths.Python36_x64 ??
+                            PythonPaths.Python36 ??
+                            PythonPaths.Python35_x64 ??
+                            PythonPaths.Python35;
             python.AssertInstalled();
 
             var project = CreateTemporaryProject(app);
@@ -408,7 +419,12 @@ version = 3.{1}.0", python.PrefixPath, python.Version.ToVersion().Minor));
         }
 
         public void ProjectAddExistingEnv(PythonVisualStudioApp app) {
-            var python = PythonPaths.Python37 ?? PythonPaths.Python36 ?? PythonPaths.Python35 ?? PythonPaths.Python34 ?? PythonPaths.Python33;
+            var python =    PythonPaths.Python37_x64 ??
+                            PythonPaths.Python37 ??
+                            PythonPaths.Python36_x64 ??
+                            PythonPaths.Python36 ??
+                            PythonPaths.Python35_x64 ??
+                            PythonPaths.Python35;
             python.AssertInstalled();
 
             var project = CreateTemporaryProject(app);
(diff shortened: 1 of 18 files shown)


MinKPark:
@zooba since Hugues is not available. do you mind reviewing this PR? it would be really appreciated.

Python/Tests/Core.UI/WebProjectTests.cs
```diff
@@ -220,6 +220,7 @@ public class WebProjectTests {
             app.WaitForOutputWindowText("Build", "1 failed");
         }
 
+        /*
```
MinKPark: the entire method is commented. does it mean that this code is not needed? if so, how about removing the lines than commenting out?

Python/Tests/Core/DebugReplEvaluatorTests.cs
```diff
@@ -51,7 +51,7 @@ public class DebugReplEvaluatorTests {
 
         internal virtual PythonVersion Version {
             get {
-                return PythonPaths.Python26 ?? PythonPaths.Python26_x64;
+                return PythonPaths.Python27 ?? PythonPaths.Python27_x64;
```
zooba: As this is a virtual property, presumably there's an override somewhere for running the 2.7 tests that you'll need to remove?

Python/Tests/Core/DebugReplEvaluatorTests.cs
```diff
@@ -33,7 +33,7 @@
 
 namespace PythonToolsTests {
     [TestClass]
```
MinKPark: I wonder whether we still need this "TestClass" attribute.

Python/Tests/Core.UI/EnvironmentUITests.cs
```diff
@@ -57,9 +57,7 @@ public class EnvironmentUITests {
             return app.SelectDefaultInterpreter(
                 PythonPaths.Python37 ?? PythonPaths.Python37_x64 ??
```
MinKPark: in other place, I could see that you start searching 64bit first then 32bit. however here we are using 32bit first. why?

Python/Tests/Core.UI/EnvironmentUITests.cs
```diff
@@ -57,9 +57,7 @@ public class EnvironmentUITests {
...
                 PythonPaths.Python36 ?? PythonPaths.Python36_x64 ??
-                PythonPaths.Python35 ?? PythonPaths.Python35_x64 ??
-                PythonPaths.Python34 ?? PythonPaths.Python34_x64 ??
-                PythonPaths.Python33 ?? PythonPaths.Python33_x64
+                PythonPaths.Python35 ?? PythonPaths.Python35_x64
```
zooba: This change (and the similar ones below) are probably better written as `PythonPaths.Versions.First()` (or whatever the right logic is - I don't remember right now, but it might be `.Last()` or `.First(p => p.IsCPython)`)

Python/Tests/DebuggerTests/BaseDebuggerTests.cs
```diff
@@ -29,6 +29,7 @@
 using TestUtilities;
 
 namespace DebuggerTests {
+    [TestClass, Ignore]
```
zooba: Is this necessary? This isn't actually a test class.

Python/Tests/ProfilingTests/ProfilingTests.cs
```diff
@@ -57,9 +57,8 @@ public class ProfilingTests {
                 Assert.IsTrue(File.Exists(testFile), "Did not find " + testFile);
             }
 
-            // Test in 3.4 for tokenize.open and 3.1 for tokenize.detect_encoding
             // Python 2.x uses execfile() and we do not handle encoding at all
-            foreach (var python in new[] { PythonPaths.Python31, PythonPaths.Python34 }) {
+            foreach (var python in new[] { PythonPaths.Python37, PythonPaths.Python36 }) {
```
zooba: These tests were for functionality related to those particular versions. There should be other tests covering 3.7 and 3.6 already, so this whole test can probably just be removed.

Python/Tests/Core.UI/EnvironmentUITests.cs
zooba: Okay, saw the logic below and remembered - you'll want `PythonPaths.Versions.Last(p => p.Version > PythonLanguageVersion.Vxx)`

Python/Tests/DebuggerUITestsRunner/DebugProjectUITests.cs
```diff
@@ -266,7 +266,7 @@ public class DebugProjectUITestsExperimental27 : DebugProjectUITestsExperimental
     }
 
     [TestClass]
-    public class DebugProjectUITestsExperimental36 : DebugProjectUITestsExperimental {
-        protected override string Interpreter => "Python36|Python36_x64";
+    public class DebugProjectUITestsExperimental37 : DebugProjectUITestsExperimental {
+        protected override string Interpreter => "Python37|Python37_x64";
```
MinKPark: is this for the last version of Python 3.x? I wonder.

zooba: Oh, I noticed that and forgot to mention anything. I'm pretty sure this one has to match one of the test projects under TestData. If so, you'll have to update those to use Python 3.7 as well.

Python/Tests/Core.UI/EnvironmentUITests.cs
zooba: It's very likely "because it doesn't matter". In a lot of tests we just wanted to ensure that a version of Python 3 (or 2) is active, but without having to force contributors into a certain configuration.

Python/Tests/Core.UI/WebProjectTests.cs
RaymonGulati1: None of the tests in this class are run because the "TestMethod" attribute has been commented out for all of them, including this function. I commented out this test because removing the older Python versions causes a syntax error on line 227. I can't figure out exactly what this test is doing and unsure about how to properly change it to the newer version. Whoever enables these tests can uncomment the function and fix the syntax error.

Python/Tests/Core.UI/EnvironmentUITests.cs
RaymonGulati1: There is no official rule on 32bit vs 64bit, though I prefer 64bit over 32bit.

Python/Tests/Core/DebugReplEvaluatorTests.cs
RaymonGulati1: Not needed. Will be removed in next commit.

Python/Tests/Core/DebugReplEvaluatorTests.cs
RaymonGulati1: This class has been turned into abstract, so it will not run the 2.7 tests which continue to run in "DebugReplEvaluatorTests27". This is consistent with the other code (such as "DebuggerTests") where the base class is abstract so no tests will ever be run on it. The "correct" way of doing this would be to convert this to an abstract property, but I don't want to break away from the norm for this one case. Perhaps I'll change it when i go through all the broken tests to fix them.

RaymonGulati1: This class has been turned into abstract, so it will not run the 2.7 tests which continue to run in "DebugReplEvaluatorTests27". This is consistent with the other code (such as "DebuggerTests") where the base class is abstract so no tests will ever be run on it. The "correct" way of doing this would be to convert this to an abstract property, but I don't want to break away from the norm for this one case. Perhaps I'll change it when i go through all the broken tests to fix them.

Python/Tests/DebuggerTests/BaseDebuggerTests.cs
RaymonGulati1: I did it to be consistent with "DebuggerTests" class. Upon further investigation, I see that it is not required and I've removed it. Will be updated in next commit.

Python/Tests/ProfilingTests/ProfilingTests.cs
RaymonGulati1: This is the only test case for this scenario. I have commented it out and will revisit it when I fix the broken tests.

Python/Tests/Core.UI/EnvironmentUITests.cs
RaymonGulati1: I've discussed this with Hugues and we decided to keep this as is for the time being. I'll add it to my list of things to revisit when I fix the broken tests.

Python/Tests/DebuggerUITestsRunner/DebugProjectUITests.cs
RaymonGulati1: @zooba, do you happen to know which one? I can't find it and most of the tests are passing under 37.

RaymonGulati1: @zooba, do you happen to know which one? I can't find it and most of the tests are passing under 37.

Python/Tests/Core.UI/WebProjectTests.cs
MinKPark: then, can you add your comment above into this method header? it will let others know why this method is commented out. 
btw, you don't have to explicitly comment out "TestMethod" attribute since you already comment out the entire method with "/* */". 

---
In reply to: [292125876](https://github.com/microsoft/PTVS/pull/5363#discussion_r292125876) [](ancestors = 292125876)

RaymonGulati1: I added it to my commit yesterday. Do you see it?

Python/Tests/Core.UI/EnvironmentUITests.cs
MinKPark: how about having a single approach? it would let us debug possible issues easier since we will use same CPU architecture through all test cases.

RaymonGulati1: It would take a decent amount of work to go through all the tests and change it, so I'm not sure if the cost/benefit analysis is worth it at the moment. Moving forward, we can make it such that all tests should default to 64bit and then 32bit (or vice versa). Whenever someone changes an area of code, they can update the tests to 64bit/32bit. How does that sound?

RaymonGulati1: By single approach, do you mean always running the tests on either 32bit or 64bit, but not both?

Python/Tests/DebuggerUITestsRunner/DebugProjectUITests.cs
zooba: Looking through the referenced project files, it looks like `TestData\SysPath\SysPath.pyproj` and `DebugAttach\DebugAttach.pyproj` expect to use Python 2.7, but all the rest are going to use whatever is the active default. So this is probably fine.

Python/Tests/Core.UI/EnvironmentUITests.cs
MinKPark: I meant "64bit first and 32bit" like your logic. it would let us run the same tests on different configuration based on what is available. also let us test both cases based on machine config.

RaymonGulati1: Agreed. Let's follow this guideline form now on.

PR #5378, opened 2019-06-18
Removing failing tests

### LINKED ISSUE #5355: Identify and filter failing P0, P2, and P3 UI and unit tests
Change their priority so they can be filtered out. For example, failing P0 tests became P10, P2 becomes P20, P3 becomes P3.

RaymonGulati1:
Fix #5355 

Changing the priorities on failing P0, P2, and P3 tests so they can be easily filtered out. 

Due to the unreliability of P2 and P3 tests, there may be some tests which are still failing but they not been filtered out. The primary purpose of this issue is to find and filter failing P0 tests. 

There may be some P0 UI tests that pass on a dev machine but fail on pipeline. They are being investigated.

### DIFF
--- a/Python/Tests/TestRunnerInterop/VsTestContext.cs
+++ b/Python/Tests/TestRunnerInterop/VsTestContext.cs
@@ -22,6 +22,10 @@ using System.Threading;
 
 namespace TestRunnerInterop {
     public sealed class VsTestContext : IDisposable {
+        public const int P0_FAILING_UI_TEST = 12;
+        public const int P2_FAILING_UI_TEST = 22;
+        public const int P3_FAILING_UI_TEST = 32;
+
         private readonly string _testDataRoot;
         private VsInstance _vs;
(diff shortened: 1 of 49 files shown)


Python/Tests/TestRunnerInterop/VsTestContext.cs
```diff
@@ -22,6 +22,10 @@
 
 namespace TestRunnerInterop {
     public sealed class VsTestContext : IDisposable {
+        public const int P0_FAILING_UI_TEST = 12;
```
bschnurr: wish we could have just one copy of these.. but I understand.

bschnurr: wish we could have just one copy of these.. but I understand.

bschnurr: nm.. different values

PR #5384, opened 2019-06-25
Fix script path to include working directory and script name

### LINKED ISSUE #5373: Cannot debug the python file in opened folder
VS Version: 16.2.0 Preview 4.0 [29013.205.d16.2stg]
Python Tool Version: 16.2.19150.1

Repro step:
1. Create a folder
2. Open this folder in VS
3. Add a python file
5. Press F5 

Actual Result:
Error message popup after pressing F5
image

Expect Result:
It should no error popup

RaymonGulati1:
Fix #5373

### DIFF
--- a/Python/Product/Workspace/PythonLaunchDebugTargetProvider.cs
+++ b/Python/Product/Workspace/PythonLaunchDebugTargetProvider.cs
@@ -134,7 +134,7 @@ namespace Microsoft.PythonTools.Workspace {
             var launchConfig = new LaunchConfiguration(config) {
                 InterpreterPath = config == null ? path : null,
                 InterpreterArguments = settings.GetValue(InterpreterArgumentsKey, string.Empty),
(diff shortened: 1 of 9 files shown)


Python/Product/Workspace/PythonLaunchDebugTargetProvider.cs
```diff
@@ -134,7 +134,7 @@ class PythonLaunchDebugTargetProvider : ILaunchDebugTargetProvider {
             var launchConfig = new LaunchConfiguration(config) {
                 InterpreterPath = config == null ? path : null,
                 InterpreterArguments = settings.GetValue(InterpreterArgumentsKey, string.Empty),
-                ScriptName = scriptName,
+                ScriptPath = Path.Combine(workingDir, scriptName),
```
huguesv: We should only do this if it's not already a full path.
You can check with `Path.IsPathRooted`

RaymonGulati1: Fixed

PR #5395, opened 2019-07-01
Issue 3907 updated debugger for test explorer

### LINKED ISSUE #3907: Test Explorer experimental debugger support
Once remote debugging is implemented, it might 'just work'. Right now it's ignoring the experimental debugger option and always using the old debugger.

RaymonGulati1:
Fix #3907


Python/Product/TestAdapter.Executor/TestExecutor.cs
```diff
@@ -877,10 +892,12 @@ enum PythonDebugMode {
             PythonAndNative
         }
 
-        private static string PtvsdSearchPath {
-            get {
+        private static string GetDebuggerPath(bool useLegacyDebugger) {
```
huguesv: Please keep the "Search" in the name "GetDebuggerSearchPath", because this isn't returning the path to the debugger folder, it's returning the search path for it (ie its parent folder).

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
```diff
@@ -228,8 +228,9 @@ def main():
     global _channel
 
     parser = OptionParser(prog = 'visualstudio_py_testlauncher', usage = 'Usage: %prog [<option>] <test names>... ')
-    parser.add_option('-s', '--secret', metavar='<secret>', help='restrict server to only allow clients that specify <secret> when connecting')
+    parser.add_option('-s', '--secret', metavar='<secret>', help='restrict server to only allow clients that specify <secret> when connecting (legacy debugger only)')
     parser.add_option('-p', '--port', type='int', metavar='<port>', help='listen for debugger connections on <port>')
+    parser.add_option('-d', '--debugger-path', type='str', metavar='<debugger_path>', help='Path to the debugger directory')
```
huguesv: Same as above, this should be called debugger search path.

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
```diff
@@ -239,7 +240,9 @@ def main():
     (opts, _) = parser.parse_args()
     
     sys.path[0] = os.getcwd()
-    
+    if(opts.debugger_path is not None):
```
huguesv: Remove parentheses

huguesv: Also, if it's empty, it probably shouldn't be added to `sys.path`, so you can remove the `is not None` part.

PR #5397, opened 2019-07-01
For new debugger, break on system exit exception of 0

### LINKED ISSUE #4713: sys.exit(0) throws exception in VS 2017 Community with Python 64 bit 3.6
I have been looking for a solution to this annoying problem for a day or so now,
the code I am running is a project that I wrote a few months ago and used to work in python on visual studio, which I have since updated and am guessing has changed a lot. I am using VS 2017 Community with Python 64 bit 3.6 all fully updated in the VS installer.

Now for some reason, I get exactly the same error as here https://github.com/Microsoft/PTVS/issues/888
I get the problem this person has with his own code where the option "Break on SystemExit exception with exit code of zero" is cleared and it still breaks with the exception.

I can generate the exception easily with the following code, so I am pretty sure it is nothing to do with my code unless I am missing something new.
```
import sys 
sys.exit(0)
```
My projects code used to work with no arguments but now for some reason needs a 0 or raises a further exception for not allowing a none type.

Thankyou

RaymonGulati1:
Fix #4713

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapterLauncher.cs
+++ b/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapterLauncher.cs
@@ -27,11 +27,11 @@ namespace Microsoft.PythonTools.Debugger {
     [Guid(DebugAdapterLauncherCLSIDNoBraces)]
     public sealed class DebugAdapterLauncher : IAdapterLauncher {
         public const string DebugAdapterLauncherCLSIDNoBraces = "C2990BF1-A87B-4459-9478-322482C535D6";
-        public const string DebugAdapterLauncherCLSID = "{"+ DebugAdapterLauncherCLSIDNoBraces + "}";
+        public const string DebugAdapterLauncherCLSID = "{" + DebugAdapterLauncherCLSIDNoBraces + "}";
         public const string VSCodeDebugEngineId = "{86432F39-ADFD-4C56-AA8F-AF8FCDC66039}";
         public static Guid VSCodeDebugEngine = new Guid(VSCodeDebugEngineId);
 
-        public DebugAdapterLauncher(){}
+        public DebugAdapterLauncher() { }
 
         public void Initialize(IDebugAdapterHostContext context) {
         }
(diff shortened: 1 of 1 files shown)


int19h:
@karthiknadig I wonder if VS should switch to `debugOptions`, given that we will be deprecating the other stuff eventually?

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapterLauncher.cs
```diff
@@ -56,7 +56,8 @@ public sealed class DebugAdapterLauncher : IAdapterLauncher {
                 JObject obj = new JObject()
                 {
                     ["remote"] = uri,
-                    ["options"] = "SHOW_RETURN_VALUE={0}".FormatInvariant(debugService.ShowFunctionReturnValue)
+                    ["options"] = "SHOW_RETURN_VALUE={0};BREAK_SYSTEMEXIT_ZERO={1}"
+                        .FormatInvariant(debugService.ShowFunctionReturnValue, debugService.BreakOnSystemExitZero)
```
int19h: If you watch to switch to the (newer and cleaner) `debugOptions` API:
https://github.com/microsoft/ptvsd/blob/9666eac569bb170d191896c5feb083d0a6ff1bf4/README.md#launch-request-arguments

int19h: In terms of compatibility with older versions of ptvsd, this has been around for over a year:
https://github.com/microsoft/ptvsd/issues/245

karthiknadig:
Yep. it should be fairly easy to switch to `debugOptions` from the old style options.

huguesv:
The readme doesn't show it, but the debugOptions equivalent is `BreakOnSystemExitZero`
https://github.com/microsoft/ptvsd/blob/59ef76317bd8a8c4b0a3b50b5fe21a605eb3c8b0/src/ptvsd/wrapper.py#L372

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapterLauncher.cs
RaymonGulati1: Updated to use new API

PR #5447, opened 2019-07-17
PTVS debuggers will not debug PTVS files while in release mode

### LINKED ISSUE #5406: Debuggers should not debug visualstudio_py_testlauncher.py
Whenever the legacy debugger (run with debugging and then attach TCP) or new debugger (run with debugging or attach TCP) debugs the unit tests, it will also debug visualstudio_py_testlauncher.py. Change it so that this script is not debugged.

RaymonGulati1:
Fix #5406

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapterLauncher.cs
+++ b/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapterLauncher.cs
@@ -15,6 +15,8 @@
 // permissions and limitations under the License.
 
 using System;
+using System.Diagnostics;
+using System.IO;
 using System.Runtime.InteropServices;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.VisualStudio.Debugger.DebugAdapterHost.Interfaces;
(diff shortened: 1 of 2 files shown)


Python/Product/PythonTools/PythonTools/Debugger/DebugAdapterLauncher.cs
```diff
@@ -50,24 +52,45 @@ public sealed class DebugAdapterLauncher : IAdapterLauncher {
...
+            if (debugService.BreakOnSystemExitZero) {
+                debugOptions.Add("BreakOnSystemExitZero");
+            }
+
+            if (debugOptions.Count > 0) {
+                launchJson["debugOptions"] = debugOptions;
+            }
+        }
+
+        [Conditional("DEBUG")]
```
huguesv: This attribute isn't right. We want this to work in release builds of PTVS as well.

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
```diff
@@ -177,6 +177,12 @@ def addUnexpectedSuccess(self, test):
         self._setResult(test, 'passed')
 
     def _setResult(self, test, outcome, trace = None):
+        if ("ptvsd" in sys.modules
```
huguesv: This is a very weird place to be adding this. The code in `main()` already sets the entry point and dont debug lists. Did that not work?

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapterLauncher.cs
```diff
@@ -50,24 +52,45 @@ public sealed class DebugAdapterLauncher : IAdapterLauncher {
...
+        private void AddRules(JObject launchJson) {
+            string PTVSDirectory = PathUtils.GetParent(typeof(DebugAdapterLauncher).Assembly.Location);
```
huguesv: camel case for local variables

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
RaymonGulati1: In the case where a python test script is launched and immediately attached, the code in main is executed. 

If the python test script is run without debugging and the user later attaches to it (legacy TCP or PID), this code will execute which prevents the user from debugging visualstudio_py_testlauncher.py

int19h: You should be able to do this bit on top level, outside of main. It doesn't block anything, it will work even if ptvsd is not loaded, and the only side effect that it has, you want to always occur if ptvsd *is* loaded, regardless of the reason why it happens.

int19h: Never mind - if it's on global level (or in main), then the check for ptvsd can indicate that it's not loaded, but then it gets loaded later in various attach scenarios... but since this module is already loaded, the code won't re-run.

So it has to be on some code path that is guaranteed to be re-executed periodically. Putting it here will make it perform the check after every test, but before any other code runs in this file.

I'm not sure this is 100% foolproof, since by the time we get here, this code is already running... so `DONT_DEBUG` will not apply at least to this very frame. I think the user might end up here if they do Step Out, or Step Over on the last line of their test, if they attached during that test.

A 100% foolproof way would be to register an import hook to detect exactly when ptvsd gets loaded. However, this would require a great deal more code, and very convoluted code at that. It's probably not worth the bother for such a niche scenario.

huguesv: We haven't been hiding this code from debugging for this scenario (run + manual attach) since we first allowed debugging when running tests, and no one ever filed an issue for that. I don't see the need to be fixing this now in legacy debugger when we're so close to removing support.

If this was clean code, sure, but this is an ugly hack.

PR #5458, opened 2019-07-22
Updating TestContainerDiscoverer to defer notifying VS of changes

bschnurr:
- fixes issue when editing mulitiple unittest properties at the save time causes double refresh of test discovery
- removed _isRefresh guard on notifiation sending, because I moved notifying out of LoadProject and RemoveProject and they return bool didChange instead that way when doing solution full refresh we dont try to sent any notifications.
- also removed calling RemoveTestContainers() from OnTestPropertiesChanged(..) because we are going to trigger a refresh, which will clear everything. previously I also deleted the files in this handler but now notifications are no longer guarding on _isRefresh

### DIFF
--- a/Python/Product/TestAdapter/TestContainerDiscoverer.cs
+++ b/Python/Product/TestAdapter/TestContainerDiscoverer.cs
@@ -30,6 +30,7 @@ using Microsoft.VisualStudioTools.TestAdapter;
 using Microsoft.PythonTools.Interpreter;
 using System.Collections.Concurrent;
 using System.Diagnostics;
+using System.Threading;
 
 namespace Microsoft.PythonTools.TestAdapter {
     [Export(typeof(ITestContainerDiscoverer))]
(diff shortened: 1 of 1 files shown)


Python/Product/TestAdapter/TestContainerDiscoverer.cs
```diff
@@ -188,8 +190,15 @@ class TestContainerDiscoverer : ITestContainerDiscoverer, IDisposable {
...
         private void NotifyContainerChanged() {
+            try {
+                _deferredTestChangeNotification.Change(500, Timeout.Infinite);
+            } catch (ObjectDisposedException) {
+            }
+        }
+
+        private void OnDeferredTestChanged(object state) {
             // guard against triggering multiple updates during initial load until setup is complete
-            if (!_firstLoad && !_isRefresh) {
```
bschnurr: this !_isRefresh guard was pretty shady. Before I had to call Notify but then set _isRefresh = true afterwards so that it wouldn't get ignored.   Now during the refresh SetupSolution() wont trigger notifications so we dont need to guard this

Python/Product/TestAdapter/TestContainerDiscoverer.cs
```diff
@@ -114,18 +116,18 @@ class TestContainerDiscoverer : ITestContainerDiscoverer, IDisposable {
...
+                    (_firstLoad || _isRefresh)) {
                     // The first time through, we don't know about any loaded
                     // projects.
                     ThreadHelper.JoinableTaskFactory.Run(async () => {
                         await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
                         if (_firstLoad || _isRefresh) {
-                            SetupSolution();
                             _firstLoad = false;
                             _isRefresh = false;
+
```
bschnurr: safer to reset these before calling the functions below.  this all runs in SwitchToMainThreadAsync so that after allowing another refresh it will still have to wait to SetupSolution() to complete

Issue #5431, opened 2019-07-12
PTVS UI Test Suite for unittest


huguesv:
Old test implementation is no longer working. Update/rewrite for the new unittest support.

Issue #5288, opened 2019-05-06
Update unit test support for Pytest


bschnurr:
- VSCode has a number of tests in python for testing pytest results (not sure how to integrate these)
- Add new C# unit tests around pytest usage to TestAdapterTests
- Add end-to-end UI tests to PythonToolsUITests <- already done

PR #5460, opened 2019-07-22
Test executor and discoverer tests

huguesv:
Convert remaining tests.
Some refactoring in tests and cleanup of old test data.
Write UI tests for 'unittest'.

Fix https://github.com/microsoft/PTVS/issues/5431
Fix https://github.com/microsoft/PTVS/issues/5288


Python/Tests/TestData/TestExecutor/test_cancel.py
```diff
@@ -0,0 +1,18 @@
...
+
+class CancelTests(unittest.TestCase):
+    def test_sleep_1(self):
+        time.sleep(0.1)
+
+    def test_sleep_2(self):
+        time.sleep(5)
+
+    def test_sleep_3(self):
+        time.sleep(5)
```
bschnurr: why is sleep_3 the same as sleep_2?  is that seconds?

huguesv: yes in seconds. as discussed privately, there are 2 long running tests to give enough time for us to cancel.

Issue #5454, opened 2019-07-19
Unittest framework test discovery should not require installing pytest


huguesv:
Right now the discovery python script unconditionally loads pytest and fails if not found, even though it doesn't need it.

PR #5465, opened 2019-07-23
Bschnurr/fix import pytest

bschnurr:
Unittest framework test discovery should not require installing pytest  …
- Still needed to add parse options for pytest event if it wasn't installed so that the user wouldn't see an error that would drive them to convert to unittest.
run_adapter.py discover: error: argument tool: invalid choice: 'pytest' (choose from 'unittest')

now the error they will see is:

"C:\USERS\BSCHNURR\APPDATA\LOCAL\MICROSOFT\VISUALSTUDIO\16.0_FE4FEA96EXP\EXTENSIONS\MICROSOFT CORPORATION\PYTHON\16.0.0\PythonFiles\testing_tools\adapter\pytest\_discovery.py", line 9, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'

related bug #5454

### DIFF
--- a/Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/__main__.py
+++ b/Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/__main__.py
@@ -6,29 +6,53 @@ from __future__ import absolute_import
 import argparse
 import sys
 
-from . import pytest, report, unittest
+from . import report, unittest
 from .errors import UnsupportedToolError, UnsupportedCommandError
+
+
+def default_subparser(cmd, name, parent):
+    parser = parent.add_parser(name)
+    if cmd == 'discover':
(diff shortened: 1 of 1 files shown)


Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/__main__.py
```diff
@@ -6,19 +6,45 @@
...
+    # Handle the case where the user hasn't installed pytest but we still want to see it appear as an option
+    try:
+        from . import pytest
+        return pytest.add_cli_subparser(cmd, name, parent)
+    except ImportError:
+        return default_subparser(cmd, name, parent)
+
+
+def pytest_discover(pytestargs=None, **_ignored):
+    # Delay importing pytest until actaully used
```
huguesv: typo

PR #5470, opened 2019-07-24
making sure we clear the pytest cache when doing discovery

bschnurr:


bschnurr:
still doesn't completely solve the bug related to moving the solution folder.

PR #5473, opened 2019-07-24
Update VisualStudioInstance.cs

zhouwangyang:


huguesv:
As discussed by email, my suspicion is that generated projects don't load as expected correctly the first time, but correctly the second time, is that the first time, an upgrade occurs.
We should rather try to fix the cause of the upgrade to stop that from happening (if that is indeed the case).

huguesv:
No longer needed, the project upgrade problem should be fixed with https://github.com/microsoft/PTVS/pull/5527

Issue #5462, opened 2019-07-23
Unittest discovery not working on Python 2.7


huguesv:
1. Create new project
2. Set active environment to Python 2.7
3. Make sure pytest is installed (workaround for bug #5454)
4. Add new item, unit test, "test_hello1.py"
5. Delete "PythonApplication1.py" from the project
6. Change test framework in project settings to "unittest"

Result: no test discovered

Tests are discovered if you do the same on Python 3.7

Output window:
```
[7/22/2019 5:00:09.980 PM Informational] Store opened in 0.018 sec.
[7/22/2019 5:00:10.843 PM Informational] ---------- Discovery started ----------
[7/22/2019 5:00:11.718 PM Informational] cd C:\Users\huvalo\source\repos\PythonApplication62\
[7/22/2019 5:00:11.719 PM Informational] set PYTHONPATH=C:\Users\huvalo\source\repos\PythonApplication62\;c:\users\huvalo\source\repos\pythonapplication62
[7/22/2019 5:00:11.719 PM Informational] C:\Python27_x64\python.exe "C:\USERS\HUVALO\APPDATA\LOCAL\MICROSOFT\VISUALSTUDIO\16.0_D1270DCBEXP\EXTENSIONS\MICROSOFT CORPORATION\PYTHON\16.0.0\PythonFiles\testing_tools\run_adapter.py" discover unittest -- . test*.py
[7/22/2019 5:00:12.616 PM Warning] No test is available in c:\users\huvalo\source\repos\pythonapplication62\test_hello1.py c:\users\huvalo\source\repos\pythonapplication62\test_hello2.py. Make sure that test discoverer & executors are registered and platform & framework version settings are appropriate and try again.
[7/22/2019 5:00:12.647 PM Informational] ========== Discovery finished: 0 tests found (0:00:01.8015701) ==========

```

![image](https://user-images.githubusercontent.com/1696845/61672987-92dac100-aca2-11e9-8df4-d28ff4ac0c49.png)

PR #5474, opened 2019-07-24
fix for Unittest discovery not working on Python 2.7

bschnurr:
fix bug #5462

### DIFF
--- a/Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/unittest/_discovery.py
+++ b/Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/unittest/_discovery.py
@@ -49,7 +49,7 @@ def discover(pytestargs=None, hidestdio=False,
                             pass
 
                     if (isinstance(obj, types.FunctionType) or
-                        isinstance(obj, types.UnboundMethodType)):
+                        ((sys.version_info[0] < 3) and isinstance(obj, types.UnboundMethodType))):
                         _, lineno = inspect.getsourcelines(obj)
                         setattr(test, 'lineno', lineno)
             except:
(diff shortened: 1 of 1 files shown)


Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/unittest/_discovery.py
```diff
@@ -40,15 +40,16 @@ def discover(pytestargs=None, hidestdio=False,
...
                     obj = module
                     for part in parts:
                         try:
                             parent, obj = obj, getattr(obj, part)
                         except AttributeError as e:
                             pass
 
-                    if isinstance(obj, types.FunctionType):
+                    if (isinstance(obj, types.FunctionType) or
+                        isinstance(obj, types.UnboundMethodType)):
```
bschnurr: apparently looking at loadTestsFromName from 2.7 python the type on test function is UnboundedMethod

https://github.com/python/cpython/blob/2.7/Lib/unittest/loader.py#L59

huguesv: `UnboundMethodType` will fail on Python 3. You should check it's (Python 2 and `UnboundMethodType`).

```
C:\Users\huvalo>py -3
Python 3.8.0a4 (v3.8.0a4:c1004b8546, May  7 2019, 13:53:12) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import types
>>> types.UnboundMethodType
  File "<stdin>", line 1, in <module>
AttributeError: module 'types' has no attribute 'UnboundMethodType'
>>> quit()

C:\Users\huvalo>py -2.7
Python 2.7.16 (v2.7.16:413a49145e, Mar  4 2019, 01:37:19) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import types
>>> types.UnboundMethodType
<type 'instancemethod'>
>>>
```

huguesv:
Please make sure to run the TestDiscovererTests unit tests for both 2.7 and 3.7.

`TestDiscovererTests27.DiscoverUnittest` used to fail because of this bug, so it should now pass. Well, all tests should be green or yellow.

Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/unittest/_discovery.py
bschnurr: ah good catch

PR #5481, opened 2019-07-25
Remove TestAdapter.Analysis (including TestCaseInfo)

huguesv:


huguesv:
Got a build error from build machine, looks like I have to update more setup/localization for the deleted assembly.

PR #5489, opened 2019-07-26
Update to ptvsd 4.3.0

karthiknadig:


karthiknadig:
Going with this instead https://github.com/microsoft/PTVS/pull/5490

PR #5496, opened 2019-07-29
Implement info bar for test framework

### LINKED ISSUE #5486: Info bar for when framework is none but *test*.py file(s) present
One of the suggested ideas from https://github.com/microsoft/PTVS/issues/5485

RaymonGulati1:
Fix #5486


Python/Product/PythonTools/PythonTools/Project/ConfigureTestFrameworkInfoBar.cs
```diff
@@ -32,85 +32,133 @@
...
 
-            Create(new InfoBarModel(messages, acceptAction, KnownMonikers.StatusInformation));
+            var packagesList = await Task.WhenAll(getPackageManagersTask);
+            return packagesList.Any(packages => packages.Any(package =>
+                String.Equals(package.Name, nameof(TestFrameworkType.Pytest), StringComparison.OrdinalIgnoreCase))
+            );
+        }
+
+        private bool PythonTestFileFound() {
+            return PathUtils.EnumerateFiles(_infoBarData.RootDirectory, "*test*.py", true).Any();
```
huguesv: For projects, this will result in false positives if the test files are on disk but not in the project.
For workspaces, every file is considered "in project" except for the Python environment folder. Ex: User creates a virtual environment 'env' under their workspace root folder and then install packages that have 'test' files. That shouldn't trigger a prompt (on next project load).

Python/Product/Common/Strings.resx
```diff
@@ -2909,4 +2909,11 @@ Please specify at least one package.</value>
   <data name="WorkspaceText" xml:space="preserve">
     <value>workspace</value>
   </data>
+  <data name="UnitTestEnableInfoBarAction" xml:space="preserve">
+    <value>Enable unittest</value>
+  </data>
+  <data name="PythonTestFileDetected" xml:space="preserve">
+    <value>A file containing Python tests was detected in the {0} {1}, but the testing framework is disabled</value>
```
huguesv: Needs a '.' at the end of sentence.

The UI in project settings, and the property in msbuild and json file both refer to this as the "test framework", not "testing framework". I think we should keep it consistent and change the sentence to test framework.

"containing Python tests" is a little too confident given we didn't even look inside the file, but I guess that gets the point across.

Python/Product/PythonTools/PythonTools/Project/ConfigureTestFrameworkInfoBar.cs
RaymonGulati1: https://github.com/microsoft/PTVS/issues/5508

RaymonGulati1:
Bug is filed under #5508

PR #5505, opened 2019-07-29
Caching the TestEnvironment all TestAdapter tests now run in 2mins.

bschnurr:


Python/Tests/TestAdapterTests/TestExecutorTests.cs
```diff
@@ -39,6 +39,11 @@ public abstract class TestExecutorTests {
 
         protected abstract PythonVersion Version { get; }
 
+        [ClassCleanup]
+        public static void Cleanup() {
+            TestEnvironment.Clear();
```
bschnurr: Class is all tests related to python version x.  This clears all environments created by tests in that class

PR #5509, opened 2019-07-31
Trigger discovery on interpreter/package changes

### LINKED ISSUE #5424: Pytest: should trigger test discovery when interpreter factory changed or installed packages changed
If you open a project that has pytest enabled but no pytest installed, then install pytest in your environment, the tests are not discovered once pytest has installed.

If you change the environment between env that have pytest installed and others that don't, discovery should be triggered.

So we should trigger discovery when:
- current interpreter has changed (listen for that event)
- installed packages in current interpreter changed (IPackageManager.InstalledPackagesChanged)

Note that the interpreter can change without necessarily changing any project property. Example, a project uses the default global interpreter, and the user changes the global default to be a different interpreter.

huguesv:
Fix #5424

See commit messages (it may be helpful to look at them individually).

### DIFF
--- a/Python/Product/TestAdapter/TestContainerDiscovererProject.cs
+++ b/Python/Product/TestAdapter/TestContainerDiscovererProject.cs
@@ -15,11 +15,15 @@
 // permissions and limitations under the License.
 
 using System;
-using System.IO;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
+using System.Diagnostics;
+using System.IO;
 using System.Linq;
+using System.Threading;
 using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Interpreter;
 using Microsoft.PythonTools.Projects;
 using Microsoft.PythonTools.TestAdapter.Model;
 using Microsoft.VisualStudio.Shell;
@@ -27,10 +31,6 @@ using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestWindow.Extensibility;
 using Microsoft.VisualStudioTools;
 using Microsoft.VisualStudioTools.TestAdapter;
-using Microsoft.PythonTools.Interpreter;
-using System.Collections.Concurrent;
-using System.Diagnostics;
-using System.Threading;
 
 namespace Microsoft.PythonTools.TestAdapter {
     [Export(typeof(ITestContainerDiscoverer))]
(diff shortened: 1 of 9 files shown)


Python/Product/TestAdapter/TestContainerDiscovererProject.cs
```diff
@@ -274,6 +288,7 @@ class TestContainerDiscovererProject : ITestContainerDiscoverer, IDisposable {
             var pyProj = PythonProject.FromObject(e.Project);
             if (pyProj != null) {
                 pyProj.ProjectPropertyChanged -= OnTestPropertiesChanged;
+                pyProj.ActiveInterpreterChanged -= OnActiveInterpreterChanged;
```
bschnurr: do we need to _packageManagerEventSink.Unwatch(project)  here?

Python/Product/VSInterpreters/Interpreter/PythonWorkspaceContext.cs
```diff
@@ -265,15 +265,16 @@ class PythonWorkspaceContext : IPythonWorkspaceContext {
...
                 var oldSearchPaths = _searchPaths;
-                searchPathsChanged = !oldSearchPaths.SequenceEqual(_searchPaths);
                 _searchPaths = ReadSearchPathsSetting();
 
                 var oldTestFramework = _testFramework;
                 _testFramework = GetStringProperty(TestFrameworkProperty);
+
+                interpreterChanged = oldInterpreter != _interpreter;
+                searchPathsChanged = !oldSearchPaths.SequenceEqual(_searchPaths);
                 testSettingsChanged = !String.Equals(oldTestFramework, _testFramework);
```
bschnurr: good catch.

Python/Product/VSInterpreters/PackageManager/CPythonPipPackageManagerProvider.cs
```diff
@@ -35,27 +35,42 @@ class PipCommandsV27AndLater : PipPackageManagerCommands {
...
+                    }
+                }
+            }
+
+            if (pm != null) {
+                yield return pm;
+            }
+        }
+
+        private IPackageManager TryCreatePackageManager(IPythonInterpreterFactory factory) {
```
bschnurr: best to check for factory being null

Python/Product/TestAdapter/TestContainerDiscovererProject.cs
huguesv: I don't think so. When any change requires refresh, we clear all. Then every project will add its interpreter factory package managers.

bschnurr: ahh..  I need to look again if calling refresh is really necessary when a project unloads. for now this works

PR #5512, opened 2019-07-31
Add error handling in environment window.

### LINKED ISSUE #5511: Handle exceptions when launching python executable from environment window
Launching python executable may fail due to access denied error, and that will crash/freeze VS since the `StartInterpreter_Executed` command handler doesn't catch any exception.

Reported by @zooba

huguesv:
Fix #5511 

Added some try/catch, as well as some `?.Dispose` which we've started doing elsewhere as a precaution, since `Process.Start` can return `null`.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
+++ b/Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
@@ -228,7 +228,7 @@ namespace Microsoft.PythonTools.InterpreterList {
             psi.FileName = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Windows), "explorer.exe");
             psi.Arguments = "\"" + path + "\"";
 
-            Process.Start(psi).Dispose();
+            Process.Start(psi)?.Dispose();
             e.Handled = true;
         }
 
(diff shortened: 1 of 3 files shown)


Python/Product/PythonTools/PythonTools/InterpreterList/InterpreterListToolWindow.cs
```diff
@@ -519,20 +524,25 @@ class CondaEnvironmentManagerUI : ICondaEnvironmentManagerUI {
...
                     "/K",
                     pathCmd + string.Format("title {0} environment", view.Description)
                 }.Select(ProcessOutput.QuoteSingleArgument)),
-                WorkingDirectory = view.PrefixPath
-            };
+                    WorkingDirectory = view.PrefixPath
+                };
 
-            Process.Start(psi).Dispose();
+                Process.Start(psi)?.Dispose();
```
bschnurr: best to warp Process in a using?

huguesv: It should be equivalent, given there's no code between `Start` and `Dispose`. 

IMO it looks slightly nicer than
```
using (Process.Start(psi)) { }
```

(plus we have it as `Process.Start()?.Dispose` elsewhere in our sources)

PR #5519, opened 2019-08-01
Make the package managers multi instances again

### LINKED ISSUE #5517: Single instance of package manager makes everything unstable
I made changes to package managers so they return the same instance per-factory, so multiple parts of the product can register for events on them (and actually get the events).

But this is causing slowdowns somewhere in PipPackageCache, and just generally breaking stuff.

huguesv:
Fix #5517 

Will create a bug for the functionality that this is now breaking (install package doesn't trigger test discovery)


huguesv:
Filed https://github.com/microsoft/PTVS/issues/5520

PR #5521, opened 2019-08-01
fix for Tooltips in package manager search results slows down VS, bec…

### LINKED ISSUE #5518: Tooltips in package manager search results slows down VS, becomes unusable
Open package manager UI
Type pytest in search bar
Hover the mouse over the results

Result: at some point, everything becomes so slow, it is no longer usable.

bschnurr:
…omes unusable fix #5518

circular triggering when p.Description was null.
package.description was first  null

then retriggered
then package.description was set to empty.string
then hit the first if again.

### DIFF
--- a/Python/Product/EnvironmentsList/PipPackageView.cs
+++ b/Python/Product/EnvironmentsList/PipPackageView.cs
@@ -45,17 +45,18 @@ namespace Microsoft.PythonTools.EnvironmentsList {
             }
 
             try {
-                var oldDescription = _package.Description;
                 var p = await _provider.GetInstallablePackageAsync(_package, CancellationToken.None);
                 if (p.IsValid) {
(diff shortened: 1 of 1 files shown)


Python/Product/EnvironmentsList/PipPackageView.cs
```diff
@@ -45,17 +45,18 @@ internal sealed class PipPackageView : INotifyPropertyChanged {
...
                         OnPropertyChanged("UpgradeVersion");
                     }
-                    if (p.Description != _package.Description) {
+                    
+                    if (!String.IsNullOrEmpty(p.Description) && p.Description != _package.Description) {
                         _package.Description = p.Description;
                         OnPropertyChanged("Description");
-                    } else if (string.IsNullOrEmpty(p.Description)) {
+                    }
+                    else if (_package.Description == null) {
```
bschnurr: check if valid and different first

else if set to empty string

PR #5530, opened 2019-08-02
fix for pytest execution hitting 32k command line limt.

bschnurr:
- now passing pytestIds to run in a file just like unittest execution
- testlaunch.py now parses out testids from any testlist file

### DIFF
--- a/Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
+++ b/Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
@@ -76,7 +76,7 @@ namespace Microsoft.PythonTools.TestAdapter.Services {
         }
 
         public void Dispose() {
-          
+
         }
 
         public string[] GetArguments(IEnumerable<TestCase> tests, string outputfile) {
@@ -94,12 +94,11 @@ namespace Microsoft.PythonTools.TestAdapter.Services {
             // a test list on disk so that we do not overflow the 
             // 32K argument limit.
             var testIds = tests.Select(t => t.GetPropertyValue<string>(Pytest.Constants.PytestIdProperty, default));
-            if (testIds.Count() > 5 ) {
-                var testListFilePath = TestUtils.CreateTestList(testIds);
+            if (testIds.Count() > 5) {
+                var testListFilePath = TestUtils.CreateTestListFile(testIds);
                 arguments.Add(testListFilePath);
-            }
-            else {
-                arguments.Add("blankTestList");
+            } else {
+                arguments.Add("dummyfilename"); //expected not to exist, but script excepts something
                 foreach (var testId in testIds) {
                     arguments.Add(testId);
                 }
@@ -116,7 +115,7 @@ namespace Microsoft.PythonTools.TestAdapter.Services {
             debugPort = 0;
 
             if (_debugMode == PythonDebugMode.PythonOnly) {
-                if(_projectSettings.UseLegacyDebugger) {
+                if (_projectSettings.UseLegacyDebugger) {
                     var secretBuffer = new byte[24];
                     RandomNumberGenerator.Create().GetNonZeroBytes(secretBuffer);
                     debugSecret = Convert.ToBase64String(secretBuffer)
@@ -132,7 +131,7 @@ namespace Microsoft.PythonTools.TestAdapter.Services {
         private Dictionary<string, string> InitializeEnvironment(IEnumerable<TestCase> tests) {
             var pythonPathVar = _projectSettings.PathEnv;
             var pythonPath = GetSearchPaths(tests, _projectSettings);
-            var env  = new Dictionary<string, string>();
+            var env = new Dictionary<string, string>();
 
             if (!string.IsNullOrWhiteSpace(pythonPathVar)) {
                 env[pythonPathVar] = pythonPath;
@@ -187,7 +186,7 @@ namespace Microsoft.PythonTools.TestAdapter.Services {
                                     if (proc.Wait(TimeSpan.FromMilliseconds(500))) {
                                         break;
                                     }
-                                 }
+                                }
                             }
 
                             proc.Wait();
@@ -205,7 +204,7 @@ namespace Microsoft.PythonTools.TestAdapter.Services {
             } catch (Exception e) {
                 Error(e.ToString());
             }
-            
+
             return ouputFile;
         }
 
--- a/Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
+++ b/Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
@@ -22,10 +22,9 @@ using System.Linq;
 using System.Net;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
-using System.Threading;
-using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Infrastructure;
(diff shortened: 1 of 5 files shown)


Python/Product/TestAdapter.Executor/Utils/TestUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Text;
+
+namespace Microsoft.PythonTools.TestAdapter.Utils {
+    static class TestUtils {
+
+        // For a small set of tests, we'll pass them on the command
+        // line. Once we exceed a certain (arbitrary) number, create
+        // a test list on disk so that we do not overflow the 
+        // 32K argument limit.
+        internal static string CreateTestList(IEnumerable<string> tests) {
```
huguesv: `testList` sounds like a list, not a file path. I see you used `testListFilePath` elsewhere, which is what I'd do here too.

And you can implement it this way:
```
var testListFilePath = Path.GetTempFileName();
File.WriteAllLines(testListFilePath, tests, new UTF8Encoding(false));
return testListFilePath;
```

Python/Product/TestAdapter/testlauncher.py
```diff
@@ -73,12 +74,21 @@ def load_debugger(secret, port, debugger_search_path):
             input()
         sys.exit(1)
 
-def run(testRunner, args):
+def run(testRunner,test_list, args):
```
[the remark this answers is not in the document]
huguesv: and name that makes it sound like a path, not a list

Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
```diff
@@ -88,17 +87,27 @@ sealed class TestRedirector : Redirector {
...
+            // line. Once we exceed a certain (arbitrary) number, create
+            // a test list on disk so that we do not overflow the 
+            // 32K argument limit.
+            var testIds = tests.Select(t => t.GetPropertyValue<string>(Pytest.Constants.PytestIdProperty, default));
+            if (testIds.Count() > 5 ) {
+                var testListFilePath = TestUtils.CreateTestList(testIds);
+                arguments.Add(testListFilePath);
+            }
+            else {
+                arguments.Add("blankTestList");
```
huguesv: is this a magic string recognized by pytest? or you're counting on it trying to open the file and ignoring it since it doesn't exist? worth a comment in the code.

Python/Product/TestAdapter.Executor/Utils/TestUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+// permissions and limitations under the License.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.PythonTools.TestAdapter.Utils {
+    static class TestUtils {
+
+        // For a small set of tests, we'll pass them on the command
```
huguesv: comment implies the method will make a decision as to whether to create a file or not, but that is not the case. Remove the comment?

Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
```diff
@@ -88,17 +87,27 @@ sealed class TestRedirector : Redirector {
...
                 GetDebuggerSearchPath(_projectSettings.UseLegacyDebugger),
-                String.Format("--junitxml={0}", outputfile)
             };
 
-            foreach (TestCase test in tests) {
-                var executionTestPath = test.GetPropertyValue<string>(Pytest.Constants.PytestIdProperty, default);
-                if (String.IsNullOrEmpty(executionTestPath)) {
-                    Debug.WriteLine("Pytest execution path missing for testcase {0}", test.FullyQualifiedName);
-                    continue;
+            // For a small set of tests, we'll pass them on the command
```
bschnurr: Do we care about deleting the temp testListFilePath?

its nicer for debugging not to delete it

Python/Product/TestAdapter/testlauncher.py
```diff
@@ -33,10 +33,11 @@ def parse_argv():
     3. debugSecret
     4. debugPort
     5. Debugger search path
-    6. Rest of the arguments are passed into the test runner.
+    6. TestFile List
```
huguesv: TestList file

Python/Product/TestAdapter.Executor/Utils/TestUtils.cs
huguesv: Also, I see there's another `CreateTestList` somewhere else. It would be good to make it use this one.

Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
huguesv: I don't mind it

Python/Product/TestAdapter/testlauncher.py
```diff
@@ -73,12 +74,21 @@ def load_debugger(secret, port, debugger_search_path):
...
+def run(testRunner,test_list, args):
     """Runs the test
     testRunner -- test runner to be used `pytest` or `nose`
     args -- arguments passed into the test runner
     """
 
+    all_tests = []
+    if test_list and os.path.exists(test_list):
+        with open(test_list, 'r', encoding='utf-8') as tests:
+            all_tests.extend(t.strip() for t in tests)
```
bschnurr: I think i'll change this to append to args directly.. ie. remove all_test variable

Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
bschnurr: ya, just needed a filename to ignore.. i'll add a comment

Python/Product/TestAdapter/testlauncher.py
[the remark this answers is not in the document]
bschnurr: ya i was keeping the name the same as in visualstudio_py_testluancher.py

should i change both scripts?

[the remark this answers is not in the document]
bschnurr: never mind.. i'm not using the argsparser.. i can change to name to anything

Python/Product/TestAdapter.Executor/Utils/TestUtils.cs
bschnurr: yup i updated the old location

bschnurr: TestExecutorUnitTest.cs

Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
```diff
@@ -88,17 +87,26 @@ sealed class TestRedirector : Redirector {
...
+            // line. Once we exceed a certain (arbitrary) number, create
+            // a test list on disk so that we do not overflow the 
+            // 32K argument limit.
+            var testIds = tests.Select(t => t.GetPropertyValue<string>(Pytest.Constants.PytestIdProperty, default));
+            if (testIds.Count() > 5) {
+                var testListFilePath = TestUtils.CreateTestListFile(testIds);
+                arguments.Add(testListFilePath);
+            } else {
+                arguments.Add("dummyfilename"); //expected not to exist, but script excepts something
```
huguesv: excepts -> expects

PR #5559, opened 2019-08-08
Convert test property page to WPF.

### LINKED ISSUE #5480: Convert project properties test page to WPF

huguesv:
Fix #5480 

Uses `System.Windows.Forms.Integration.ElementHost` to handle the translation between our property page framework code designed to work with windows forms/win32 and the wpf control.


Python/Product/PythonTools/PythonTools/Project/PythonTestPropertyPageView.xaml
```diff
@@ -0,0 +1,119 @@
...
+                        Grid.Column="1"
+                        Text="{Binding Path=UnitTestPattern, UpdateSourceTrigger=PropertyChanged}"
+                        AutomationProperties.LabeledBy="{Binding ElementName=PatternLabel}"
+                        AutomationProperties.HelpText="{x:Static common:Strings.PythonTestPropertyPageUnittestPatternHelpText}"/>
+                </Grid>
+            </DataTemplate>
+        </ResourceDictionary>
+    </UserControl.Resources>
+
+    <StackPanel Grid.IsSharedSizeScope="True">
```
bschnurr: do you really need this outter stackpanel?

huguesv: I need it because I want to have the framework specific content below the framework label/combo box, and you can only have one element at the root, so a container is needed.

There's always several ways to do anything in WPF ;)

I could have placed the content control as a 2-column spanning second row in the grid. But I feel it's simpler with the stack panel.

PR #5565, opened 2019-08-09
Changed discovery test warning 

bschnurr:
Changing runtime settings to write project/workspace info once for all sub testcontainers

-Now we also use a dictionary lookup, instead of linear search,  for verifying filepaths from discovery match a TestContainer.

### DIFF
--- a/Python/Product/TestAdapter.Executor/Config/PythonProjectSettings.cs
+++ b/Python/Product/TestAdapter.Executor/Config/PythonProjectSettings.cs
@@ -48,6 +48,9 @@ namespace Microsoft.PythonTools.TestAdapter.Config {
             EnableNativeCodeDebugging = nativeDebugging;
             SearchPath = new List<string>();
             Environment = new Dictionary<string, string>();
+            // Mapping of full file path to full file path which was assigned to the TestContainer.  
+            // The pytest adapter discovery is returning lowercase paths and Test Explorer needs TestCase sources and TestContainer sources
+            // to match or else tests wont be removed when you unload a project.
             TestContainerSources = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             IsWorkspace = isWorkspace;
             UseLegacyDebugger = useLegacyDebugger;
--- a/Python/Product/TestAdapter.Executor/Config/PythonProjectSettings.cs
+++ b/Python/Product/TestAdapter.Executor/Config/PythonProjectSettings.cs
@@ -23,7 +23,7 @@ namespace Microsoft.PythonTools.TestAdapter.Config {
         public readonly bool EnableNativeCodeDebugging;
         public readonly List<string> SearchPath;
         public readonly Dictionary<string, string> Environment;
-        public readonly List<string> Sources;
+        public readonly Dictionary<string, string> TestContainerSources;
         public readonly bool IsWorkspace;
         public readonly bool UseLegacyDebugger;
         public readonly TestFrameworkType TestFramwork;
@@ -48,7 +48,7 @@ namespace Microsoft.PythonTools.TestAdapter.Config {
             EnableNativeCodeDebugging = nativeDebugging;
             SearchPath = new List<string>();
             Environment = new Dictionary<string, string>();
-            Sources = new List<string>();
+            TestContainerSources = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             IsWorkspace = isWorkspace;
             UseLegacyDebugger = useLegacyDebugger;
             TestFramwork = TestFrameworkType.None;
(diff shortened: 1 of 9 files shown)


Python/Product/TestAdapter.Executor/Config/PythonProjectSettings.cs
```diff
@@ -83,31 +84,12 @@ public class PythonProjectSettings : IEquatable<PythonProjectSettings> {
                 UseLegacyDebugger == other.UseLegacyDebugger &&
                 TestFramwork == other.TestFramwork &&
                 UnitTestPattern == other.UnitTestPattern &&
-                UnitTestRootDir == other.UnitTestRootDir) {
-                if (SearchPath.Count == other.SearchPath.Count &&
```
bschnurr: these collection values shouldn't matter.  projectName and projectHome should be all we need

Python/Product/TestAdapter.Executor/Pytest/TestDiscovererPytest.cs
```diff
@@ -41,8 +41,7 @@ internal class TestDiscovererPytest : IPythonTestDiscoverer {
         public void DiscoverTests(
             IEnumerable<string> sources,
             IMessageLogger logger,
-            ITestCaseDiscoverySink discoverySink,
-            Dictionary<string, PythonProjectSettings> sourceToProjectSettings
```
bschnurr: Previous change i started to pass this Dictionary<string, PythonProjectSettings> sourceToProjectSettings around everywhere.  now i realize i dont need to do this.

I can add the TestContainer Source to the PythonProjectSettings class

Python/Product/TestAdapter/PythonRunSettings.cs
```diff
@@ -282,5 +218,80 @@ ref outObj
 
             return inputRunSettingDocument;
         }
+
+
+        bool WriteProjectInfoForContainer(System.Xml.XmlWriter writer, TestContainer container, ILogger log) {
```
bschnurr: this is just the same  code moved into a function

Python/Product/TestAdapter.Executor/Config/RunSettingsUtil.cs
```diff
@@ -44,6 +44,7 @@ public class RunSettingsUtil {
 
                 foreach (XPathNavigator test in project.Select("Test")) {
                     string testFile = test.GetAttribute("file", "");
+                    projSettings.TestContainerSources.Add(testFile, testFile);
```
bschnurr: this is where i build a dictionary of all Testcontainer source files.

Python/Product/TestAdapter.Executor/Config/PythonProjectSettings.cs
```diff
@@ -48,7 +48,7 @@ public class PythonProjectSettings : IEquatable<PythonProjectSettings> {
             EnableNativeCodeDebugging = nativeDebugging;
             SearchPath = new List<string>();
             Environment = new Dictionary<string, string>();
-            Sources = new List<string>();
+            TestContainerSources = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
bschnurr: new dictionary to make sure we get the non modified source path

Python/Product/TestAdapter/PythonRunSettings.cs
```diff
@@ -134,86 +134,22 @@ ref outObj
...
 
-                            writer.WriteStartElement("SearchPaths");
-                            foreach (var path in config.SearchPaths) {
-                                writer.WriteStartElement("Search");
-                                writer.WriteAttributeString("value", path);
-                                writer.WriteEndElement();
+                            if (didFindProjectInfo) {
+                                writer.WriteStartElement("Test");
+                                writer.WriteAttributeString("file", container.Source);
+                                writer.WriteEndElement(); // Test
```
bschnurr: I could change "Test" to "TestContainer" this is leftover code from when we used to pass the actually tests

Python/Product/TestAdapter.Executor/UnitTest/TestDiscovererUnitTest.cs
```diff
@@ -112,21 +111,19 @@ internal class TestDiscovererUnitTest : IPythonTestDiscoverer {
...
-            ITestCaseDiscoverySink discoverySink,
-            Dictionary<string, PythonProjectSettings> sourceToProjectSettings
+            ITestCaseDiscoverySink discoverySink
         ) {
             bool showConfigurationHint = false;
             foreach (var test in unitTestResults?.SelectMany(result => result.Tests.Select(test => test)).MaybeEnumerate()) {
                 try {
-                    if(!sourceToProjectSettings.ContainsKey(test.Source)) {
+                    // Note: Test Explorer will show a key not found exception if we use a source path that doesn't match a test container's source.
+                    if (_settings.TestContainerSources.TryGetValue(test.Source, out string testContainerSourcePath)) {
```
huguesv: I think you can use `out _` for when you don't actually need the value

Python/Product/TestAdapter.Executor/Config/PythonProjectSettings.cs
huguesv: I think it deserves an explanation here as to what the key and value are, and a bit why

bschnurr: good point

Issue #5488, opened 2019-07-26
Pytest discovery should ignore environment folder


huguesv:
I have a virtual environment under my workspace, and I enabled pytest.

I'm getting an error from test discovery (and no test discovered):
```
[7/26/2019 12:55:11.508 PM Error] Not a valid Python module: c:\projects\github\temp\pytest\env\lib\site-packages\setuptools\site-patch.py

```

because it's searching for tests under the virtual environment folder.

We should skip any file that's under the active environment's folder.

RaymonGulati1:
Fixed in https://github.com/microsoft/PTVS/pull/5571

PR #5571, opened 2019-08-12
Exclude test files in virtual env or workspace cache

RaymonGulati1:
Issue #5488

### DIFF
--- a/Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
+++ b/Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
@@ -176,7 +176,7 @@ namespace Microsoft.PythonTools.TestAdapter {
                 _testFilesUpdateWatcher.AddDirectoryWatch(workspace.Location);
                 oldWatcher?.Dispose();
 
-                foreach (var file in FilterInvalidFiles(projInfo)) {
+                foreach (var file in GetTestFiles(projInfo)) {
                     projInfo.AddTestContainer(this, file);
                 }
 
@@ -291,7 +291,7 @@ namespace Microsoft.PythonTools.TestAdapter {
             NotifyContainerChanged();
         }
 
-        private IEnumerable<string> FilterInvalidFiles(ProjectInfo projectInfo) {
+        private IEnumerable<string> GetTestFiles(ProjectInfo projectInfo) {
             List<string> validFiles = Directory.EnumerateFiles(projectInfo.ProjectHome, "*.py", SearchOption.TopDirectoryOnly).ToList();
             List<InterpreterConfiguration> workspaceInterpreterFactories = _interpreterRegistryService.Configurations
                 .Where(x => PathUtils.IsSubpathOf(projectInfo.ProjectHome, x.InterpreterPath))
--- a/Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
+++ b/Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
@@ -37,6 +37,7 @@ namespace Microsoft.PythonTools.TestAdapter {
         private readonly IPythonWorkspaceContextProvider _workspaceContextProvider;
         private readonly ConcurrentDictionary<string, ProjectInfo> _projectMap;
         private readonly PackageManagerEventSink _packageManagerEventSink;
+        private readonly IInterpreterRegistryService _interpreterRegistryService;
         private readonly Timer _deferredTestChangeNotification;
         private bool _firstLoad, _isDisposed, _isRefresh;
         private TestFilesUpdateWatcher _testFilesUpdateWatcher;
@@ -46,12 +47,14 @@ namespace Microsoft.PythonTools.TestAdapter {
             [Import(typeof(SVsServiceProvider))]IServiceProvider serviceProvider,
             [Import(typeof(IOperationState))]IOperationState operationState,
             [Import] IPythonWorkspaceContextProvider workspaceContextProvider,
-            [Import] IInterpreterOptionsService interpreterOptionsService
+            [Import] IInterpreterOptionsService interpreterOptionsService,
+            [Import] IInterpreterRegistryService interpreterRegistryService
         ) {
             _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
             _projectMap = new ConcurrentDictionary<string, ProjectInfo>();
             _packageManagerEventSink = new PackageManagerEventSink(interpreterOptionsService);
             _packageManagerEventSink.InstalledPackagesChanged += OnInstalledPackagesChanged;
+            _interpreterRegistryService = interpreterRegistryService;
             _deferredTestChangeNotification = new Timer(OnDeferredTestChanged);
             _firstLoad = true;
             _isRefresh = false;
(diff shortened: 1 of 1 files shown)


Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
```diff
@@ -173,7 +176,9 @@ class TestContainerDiscovererWorkspace : ITestContainerDiscoverer, IDisposable {
                 _testFilesUpdateWatcher.AddDirectoryWatch(workspace.Location);
                 oldWatcher?.Dispose();
 
-                var files = Directory.EnumerateFiles(workspace.Location, "*.*", SearchOption.AllDirectories);
+                var files = Directory.EnumerateFiles(workspace.Location, "*.py", SearchOption.AllDirectories)
```
huguesv: I think you should save the results of `interpreterRegistryService.Configurations.Where(x => PathUtils.IsSubpathOf(projectInfo.ProjectHome, x.InterpreterPath))` here in a variable and reuse it, so you don't have to recalculate that for potentially 100s or 1000s of files.

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
```diff
@@ -173,7 +176,9 @@ class TestContainerDiscovererWorkspace : ITestContainerDiscoverer, IDisposable {
...
+                    .Where(x => !IsFileExcluded(projInfo, x)
+                );
```
bschnurr: Did you consider using EnumerateDirectories and checking for InterpreterPath? 
Maybe a more common case that enivronments folder is one of the folders directory under workspace.
then calling EnumerateFiles on the remaining folders.. 
I would assume its a faster then checking every file.

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
RaymonGulati1: Updated

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
RaymonGulati1: Updated

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
```diff
@@ -173,8 +176,7 @@ class TestContainerDiscovererWorkspace : ITestContainerDiscoverer, IDisposable {
                 _testFilesUpdateWatcher.AddDirectoryWatch(workspace.Location);
                 oldWatcher?.Dispose();
 
-                var files = Directory.EnumerateFiles(workspace.Location, "*.*", SearchOption.AllDirectories);
-                foreach (var file in files) {
+                foreach (var file in FilterInvalidFiles(projInfo)) {
```
huguesv: You should call it `GetContainerFiles`

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
```diff
@@ -288,5 +290,31 @@ class TestContainerDiscovererWorkspace : ITestContainerDiscoverer, IDisposable {
             }
             NotifyContainerChanged();
         }
+
+        private IEnumerable<string> FilterInvalidFiles(ProjectInfo projectInfo) {
+            List<string> validFiles = Directory.EnumerateFiles(projectInfo.ProjectHome, "*.py", SearchOption.TopDirectoryOnly).ToList();
```
huguesv: use var for both

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
```diff
@@ -288,5 +290,31 @@ class TestContainerDiscovererWorkspace : ITestContainerDiscoverer, IDisposable {
...
+            List<InterpreterConfiguration> workspaceInterpreterFactories = _interpreterRegistryService.Configurations
+                .Where(x => PathUtils.IsSubpathOf(projectInfo.ProjectHome, x.InterpreterPath))
+                .ToList();
+
+            foreach (var directory in Directory.EnumerateDirectories(projectInfo.ProjectHome, "*", SearchOption.TopDirectoryOnly)) {
```
huguesv: Those are the defaults, you can just do `Directory.EnumerateDirectories(projectInfo.ProjectHome)`

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
```diff
@@ -288,5 +290,31 @@ class TestContainerDiscovererWorkspace : ITestContainerDiscoverer, IDisposable {
...
+        private IEnumerable<string> GetTestFiles(ProjectInfo projectInfo) {
+            List<string> validFiles = Directory.EnumerateFiles(projectInfo.ProjectHome, "*.py", SearchOption.TopDirectoryOnly).ToList();
+            List<InterpreterConfiguration> workspaceInterpreterFactories = _interpreterRegistryService.Configurations
+                .Where(x => PathUtils.IsSubpathOf(projectInfo.ProjectHome, x.InterpreterPath))
+                .ToList();
+
+            foreach (var directory in Directory.EnumerateDirectories(projectInfo.ProjectHome, "*", SearchOption.TopDirectoryOnly)) {
+
+                if (!workspaceInterpreterFactories.Any(x => PathUtils.IsSameDirectory(x.GetPrefixPath(), directory)) &&
+                    !PathUtils.IsSameDirectory(directory, Path.Combine(projectInfo.ProjectHome, ".vs"))
```
bschnurr: could move this outside the loop. Path.Combine(projectInfo.ProjectHome, ".vs")

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
```diff
@@ -288,5 +290,31 @@ class TestContainerDiscovererWorkspace : ITestContainerDiscoverer, IDisposable {
             }
             NotifyContainerChanged();
         }
+
+        private IEnumerable<string> GetTestFiles(ProjectInfo projectInfo) {
+            List<string> validFiles = Directory.EnumerateFiles(projectInfo.ProjectHome, "*.py", SearchOption.TopDirectoryOnly).ToList();
```
bschnurr: Does validFiles.AddRange still work if we leave it as an IEnumerable?  ie. remove the ToList().

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
```diff
@@ -288,5 +290,31 @@ class TestContainerDiscovererWorkspace : ITestContainerDiscoverer, IDisposable {
             }
             NotifyContainerChanged();
         }
+
+        private IEnumerable<string> FilterInvalidFiles(ProjectInfo projectInfo) {
+            List<string> validFiles = Directory.EnumerateFiles(projectInfo.ProjectHome, "*.py", SearchOption.TopDirectoryOnly).ToList();
+            List<InterpreterConfiguration> workspaceInterpreterFactories = _interpreterRegistryService.Configurations
```
huguesv: They're configurations, not factories, so I'd call it `workspaceConfigs`.

Python/Product/TestAdapter/TestContainerDiscovererWorkspace.cs
RaymonGulati1: There is no AddRange for IEnumerable, so I've replaced it with [IEnumerable.Concat()](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.concat?view=netframework-4.8).

PR #5576, opened 2019-08-14
Fix bugs in tests, remove python 36 tests, change info bar close sequ…

### LINKED ISSUE #4386: Failed to execute tests because the process has already exited
- Create tests with more than 3 tests, also with failed, passed and skip(@unittest.skip("reason")) distributions
- Run all tests, all the tests were in specific group as expected
- Click "Run passed tests"/"Run failed tests"/"Repeat last run" several times(with random steps)
- Observe the test results

image

Expected: All the tests results should always be the same as the first time running.
Actual: The failed/passed tests skipped unexpectedly, and output show error "Failed to execute tests because the process has already exited".

image

Note: After this error occurred, click "repeat last run", the failed/passed tests would failed/passed as expected, but the skipped tests show a wrong test name and clock icon

image

### LINKED ISSUE #5538: Test Adapter tests are randomly failing
These 4 tests are related to the new python testing framework and they are randomly passing or failing. They only fail if all 4 python versions of the tests are run together (for example, runUnittest on 27, 35, 36, and 37). The tests are not running in parallel.

[broken unit tests.txt](https://github.com/microsoft/PTVS/files/3469713/broken.unit.tests.txt)

image

RaymonGulati1:
Fix #5538 
Fix #4386

### DIFF
--- a/Python/Product/TestAdapter.Executor/UnitTest/TestExecutorUnitTest.cs
+++ b/Python/Product/TestAdapter.Executor/UnitTest/TestExecutorUnitTest.cs
@@ -318,7 +318,7 @@ namespace Microsoft.PythonTools.TestAdapter {
             private readonly string _debugSecret;
             private readonly int _debugPort;
             private readonly ManualResetEvent _cancelRequested;
-            private readonly AutoResetEvent _connected = new AutoResetEvent(false);
+            private readonly ManualResetEvent _connected = new ManualResetEvent(false);
             private readonly AutoResetEvent _done = new AutoResetEvent(false);
             private Connection _connection;
             private readonly Socket _socket;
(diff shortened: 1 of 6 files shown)


Python/Product/TestAdapter.Executor/UnitTest/TestExecutorUnitTest.cs
```diff
@@ -530,8 +531,9 @@ sealed class TestRunner : IDisposable {
                         // If there's an error in the launcher script,
                         // it will terminate without connecting back.
                         WaitHandle.WaitAny(new WaitHandle[] { _connected, proc.WaitHandle });
+                        bool processConnected = _connected.WaitOne(1);
```
bschnurr: I wonder if proc.WaitHandle was null sometimes causing the WaitAny to exit too quickly.

huguesv: It was going inside the if statement which checks for `ExitCode`, so it wasn't returning from `WaitAny` too quickly, it was just really already done executing the script.

huguesv: I just tested what happens when you pass `null` for one of the array elements to `WaitAny`, and it throws an `ArgumentNullException`

PR #5578, opened 2019-08-14
Modify ItemVisibility to make it work.

zhouwangyang:


Common/Tests/Utilities.UI/UI/VisualStudioInstance.cs
```diff
@@ -53,6 +53,22 @@ public class VisualStudioInstance : IDisposable, IVisualStudioInstance {
             _solutionExplorer = _app.OpenSolutionExplorer();
             SelectSolutionNode();
         }
+        
+       //SelectSolutionNode for item visibility 
+        public VisualStudioInstance(SolutionFile solution, VisualStudioApp app, string item)
```
bschnurr: item not being used?

zhouwangyang: yes, this is an overload of the ToVs method, and this parameter is used to call the overloaded method.

Python/Tests/ProjectUITests/BasicProjectTests.cs
```diff
@@ -1239,8 +1239,11 @@ public class BasicProjectTests {
                 );
 
                 var solutionFile = SolutionFile.Generate("HelloWorld", baseProj, imported);
-                using (var solution = solutionFile.ToVs(app)) {
-                    Assert.IsNotNull(solution.WaitForItem("HelloWorld", "VisibleItem.txt"), "VisibleItem.txt not found");
+                //using (var solution = solutionFile.ToVs(app)) {
+                //Sets parameters and calls the overload method of ToVS
+                using (var solution = solutionFile.ToVs(app,"true")){
+                    //Assert.IsNotNull(solution.WaitForItem("HelloWorld", "VisibleItem.txt"), "VisibleItem.txt not found");
+                    Assert.IsNull(solution.WaitForItem("HelloWorld", "VisibleItem.txt"), "VisibleItem.txt not found");
```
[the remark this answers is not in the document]
zhouwangyang: Ah, you misunderstood me, I don't want to circumvent this problem, I made the change because I think this file (" VisibleItem.txt ") should be can't find, I guess if you write wrong, so I changed this code to confirm with you , this case is a test of visibility on the file, the file cannot be loaded because of the definition of the project, its "Visible" property is set "false", so it is not Visible, which is why will display 1 of 2  projects, It's not because of an upgrade or something that the file can't be loaded, so I guess the code here is probably is null instead of is not null.
 @image
Of course I know, if your code is fine, this file should be found, and that may be something else

huguesv:
No longer need this. Proper fix is in https://github.com/microsoft/PTVS/pull/5671

PR #5582, opened 2019-08-15
Test discoverer and executor tests

### LINKED ISSUE #5494: Add discovery/execution test for unittest runTest()
https://docs.python.org/3/library/unittest.html#unittest.TestLoader.loadTestsFromTestCase

Instead of defining `test_` methods on the class derived from `TestCase`, implement a `runTest` method.

We should add that scenario to our suite of discovery and execution tests.

### LINKED ISSUE #5537: Add unittest discovery test for root dir and pattern
We currently don't have test coverage for the unittest configuration.

Both are passed to unittest like this:
```
suites = unittest.defaultTestLoader.discover(pytestargs[0], pytestargs[1])
```

Loader does pattern matching via `fnmatch.fnmatch`:
https://docs.python.org/3/library/fnmatch.html

Good news is that it does appear to work:
image

huguesv:
Fix #5537 Add unittest discovery test for root dir and pattern
Fix #5494 Add discovery/execution test for unittest runTest()

Also fixes a test that was being ignored due to failing, it now passes

### DIFF
--- a/Python/Tests/TestAdapterTests/TestExecutorTests.cs
+++ b/Python/Tests/TestAdapterTests/TestExecutorTests.cs
@@ -71,22 +71,32 @@ namespace TestAdapterTests {
         public void RunUnittest() {
             var testEnv = TestEnvironment.GetOrCreate(Version, FrameworkUnittest);
 
-            var testFilePath = Path.Combine(testEnv.SourceFolderPath, "test_ut.py");
-            File.Copy(TestData.GetPath("TestData", "TestDiscoverer", "BasicUnittest", "test_ut.py"), testFilePath);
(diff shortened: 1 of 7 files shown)


Python/Tests/TestAdapterTests/TestExecutorTests.cs
```diff
@@ -396,7 +405,7 @@ class ManyTest(unittest.TestCase):
             var expectedTests = new[] {
                 new TestInfo(
                     "test_relative_import",
-                    "test_relative_import.py::RelativeImportTests::test_relative_import",
+                    "relativeimportpackage\\test_relative_import.py::RelativeImportTests::test_relative_import",
```
bschnurr: nice

PR #5607, opened 2019-08-17
Add support for code coverage for pytest.

### LINKED ISSUE #5444: Pytest: Support for code coverage
If code coverage is requested, import the coverage package and generate the file. See the existing support for coverage that we have for unittest.

huguesv:
Fix #5444 

In order to avoid duplicating a bunch of code, I moved several methods from `TestExecutorUnitTest` to `ExecutorService` (as static methods) so it can be reused between both unittest and pytest execution code.

image

### DIFF
--- a/Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
+++ b/Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
@@ -22,6 +22,9 @@ using System.Linq;
 using System.Net;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
+using System.Xml;
+using System.Xml.XPath;
+using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.TestAdapter.Config;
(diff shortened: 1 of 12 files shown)


Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
```diff
@@ -116,6 +114,153 @@ sealed class TestRedirector : Redirector {
...
+            // the filename, class, and method.  We include each one if we're
+            // running from a single filename/class/method.  When we have multiple
+            // we drop the identifying names.  If we have multiple files we
+            // go to the top level directory...  If all else fails we do "pycov".
+            foreach (var test in tests) {
+                string testFile, testClass, testMethod;
+                TestReader.ParseFullyQualifiedTestName(
+                    test.FullyQualifiedName,
+                    out testFile,
+                    out testClass,
```
bschnurr: For pytest global test functions, i set the testClass portion of FQN to be the filename.. It would be good to add a testcase for this to make sure it works ok.

huguesv: Added test.

PR #5615, opened 2019-08-20
Enable native code debugging for pytest.

### LINKED ISSUE #5613: pytest ignores "Enable native code debugging"
Create new python project
Add new item, unit test
Place a breakpoint in the test

In project properties, check Enable native code debugging
image

Set test framework to pytest
Select the test and debug test

When you hit the breakpoint, look at processes window, you'll see the Python debugger is attached
image

Expected: the "Native, Python" debugger should be attached:
image

Note that if you set the test framework to unittest instead of pytest, then it works as expected.

huguesv:
Refactor the debug settings and debug attach so it is reused for both pytest and unittest.
Copied the code from the unittest launcher script to the pytest launcher script.

Fix #5613

### DIFF
--- a/Python/Product/TestAdapter.Executor/UnitTest/TestExecutorUnitTest.cs
+++ b/Python/Product/TestAdapter.Executor/UnitTest/TestExecutorUnitTest.cs
@@ -43,10 +43,6 @@ namespace Microsoft.PythonTools.TestAdapter {
         Justification = "object owned by VS")]
     [ExtensionUri(PythonConstants.UnitTestExecutorUriString)]
     class TestExecutorUnitTest : ITestExecutor {
-        private static readonly Guid PythonRemoteDebugPortSupplierUnsecuredId = new Guid("{FEB76325-D127-4E02-B59D-B16D93D46CF5}");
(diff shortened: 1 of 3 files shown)


Python/Product/TestAdapter.Executor/UnitTest/TestExecutorUnitTest.cs
```diff
@@ -200,27 +196,10 @@ sealed class TestRunner : IDisposable {
...
-                        _debugSecret = Convert.ToBase64String(secretBuffer)
-                            .Replace('+', '-')
-                            .Replace('/', '_')
-                            .TrimEnd('=');
-                    } else {
-                        _debugSecret = "";
-                    }
+                ExecutorService.GetDebugSettings(_app, _context, _settings, out _debugMode, out _debugSecret, out _debugPort);
 
-                    SocketUtils.GetRandomPortListener(IPAddress.Loopback, out _debugPort).Stop();
```
bschnurr: I think GetRandomPortListener().stop() is still needed right?

huguesv: it's in ExecutorService.GetDebugSettings, it doesn't show in the diff.

PR #5628, opened 2019-08-21
pytest not found in azure-cli Fix #5626

### LINKED ISSUE #5626: pytest not found in azure-cli
`git clone https://github.com/azure/azure-cli`

Open workspace
Create 3.7 virtual env, do not install requirements.txt (clear that textbox)
`pip install azdev`
from command line, activate virtual env and: `azdev setup -c`

Enable pytest:

2176 tests discovered

Execution is broken. Testcase not found.
```
[8/20/2019 1:19:22.333 PM Error] Testcase not found for test result: <testcase classname="azure.cli.command_modules.billing.tests.latest.test_billing_commands.AzureBillingServiceScenarioTest" file="azure\cli\command_modules\billing\tests\latest\test_billing_commands.py" line="40" name="test_get_latest_invoice" time="0.893"></testcase>
```

What I've tried for pythonsettings.json
```
{
  "Interpreter": "env\\Scripts\\python.exe",
  "TestFramework": "pytest",
  "SearchPaths": [
    "src\\azure-cli",
    "src\\azure-cli-command_modules-nspkg",
    "src\\azure-cli-core",
    "src\\azure-cli-nspkg",
    "src\\azure-cli-telemetry",
    "src\\azure-cli-testsdk"
  ]
}
```

bschnurr:
Issues: Pytest execute will search up the parent folders for a config .ini file which will change the rootdir from the project to the subfolder.  This will cause a mismatch between discovered test filepaths and results.

now force both disocvery  and execution to set the rootdir so that .ini location wont override it.

also note it appears the user can't set the rootdir from inside the .ini file
ie
 addopt = --rootdir=XXX
 or
 rootdir = XXX

both appear to have no effect so it should be safe to force it to be the project/workspace location.

Users can still set "testpaths = testing" to optimize the search but the paths will be relative to the project/workspace dir from now on.

https://docs.pytest.org/en/2.8.7/customize.html

### DIFF
--- a/Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
+++ b/Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
@@ -84,7 +84,7 @@ namespace Microsoft.PythonTools.TestAdapter.Services {
 
         }
 
-        public string[] GetArguments(IEnumerable<TestCase> tests, string outputfile, string coveragePath) {
+        public string[] GetArguments(IEnumerable<TestCase> tests, string outputfile, string coveragePath, PythonProjectSettings settings) {
             var arguments = new List<string> {
                 TestLauncherPath,
                 _projectSettings.WorkingDirectory,
@@ -113,6 +113,7 @@ namespace Microsoft.PythonTools.TestAdapter.Services {
 
             // output results to xml file
             arguments.Add(String.Format("--junitxml={0}", outputfile));
+            arguments.Add(String.Format("--rootdir={0}", settings.ProjectHome));
 
             return arguments.ToArray();
         }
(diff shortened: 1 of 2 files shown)


Python/Product/TestAdapter.Executor/Services/ExecutorService.cs
```diff
@@ -84,7 +84,7 @@ sealed class TestRedirector : Redirector {
 
         }
 
-        public string[] GetArguments(IEnumerable<TestCase> tests, string outputfile, string coveragePath) {
+        public string[] GetArguments(IEnumerable<TestCase> tests, string outputfile, string coveragePath, PythonProjectSettings settings) {
```
huguesv: You already have access to `_projectSettings` so no need to add it as a parameter to this method?

bschnurr: since we are removing pytest from this class I thought it best to pass it in

PR #5699, opened 2019-09-04
add env activate before test discovery and execution

### LINKED ISSUE #5289: Run unit test does not activate conda environment - ImportError
Set your project environment to latest Anaconda

Add this code to a unit test:
```
import scipy
```

From Test Explorer, Run unit test

Result:
ImportError: DLL load failed: The specified module could not be found

Note that if you run the test file using F5, it activate the environment and succeeds loading scipy.

bschnurr:
Run unit test does not activate conda environment - ImportError  Fix  #5289

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Extensions.cs
+++ b/Python/Product/PythonTools/PythonTools/Extensions.cs
@@ -236,100 +236,6 @@ namespace Microsoft.PythonTools {
             }
         }
 
-        /// <summary>
-        /// Get the items present in the project
-        /// </summary>
-        public static IEnumerable<string> GetProjectItems(this IVsProject project) {
-            Debug.Assert(project != null, "Project is not null");
-
-            // Each item in VS OM is IVSHierarchy. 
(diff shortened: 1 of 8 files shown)


Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -236,100 +236,6 @@ static class Extensions {
             }
         }
 
-        /// <summary>
```
bschnurr: removing duplicate old code. this was moved into TestAdapter.Executor

Python/Product/PythonTools/PythonToolsService.cs
```diff
@@ -661,52 +661,7 @@ public sealed class PythonToolsService : IDisposable {
         #endregion
 
         internal Dictionary<string, string> GetFullEnvironment(LaunchConfiguration config) {
```
bschnurr: moved this code below into a new function which gets environmental variables by calling activate

PR #5708, opened 2019-09-06
Add support for profiling on Python 3.8.

### LINKED ISSUE #5341: Null argument error when selecting other interpeter in profiling dialog
Start VS
Launch Python profiling dialog
Select "Other" interpreter
Put path to python.exe from Python 2.7
Put full path to script pystone.py
Press OK

image

Result: null path error

This is a regression in the code that checks to activate conda environment.

### LINKED ISSUE #5683: Support profiling on Python 3.8
```
  File "c:\users\huvalo\appdata\local\microsoft\visualstudio\16.0_b2fdd7deexp\extensions\microsoft corporation\python - profiling\16.0.0\proflaun.py", line 35, in <module>
    vspyprof.profile(__file__, globals(), locals(), profdll)
  File "c:\users\huvalo\appdata\local\microsoft\visualstudio\16.0_b2fdd7deexp\extensions\microsoft corporation\python - profiling\16.0.0\vspyprof.py", line 74, in profile
    raise NotImplementedError("Profiling is currently not supported for " + sys.version)
NotImplementedError: Profiling is currently not supported for 3.8.0b3 (tags/v3.8.0b3:4336222, Jul 29 2019, 16:33:20) [MSC v.1916 32 bit (Intel)]
Press any key to continue . . .
```

huguesv:
Fix #5683 
Fix #5341

I added version specific project profiling tests, to ensure that Python 2.7...3.8 are covered (and the 3.8 test fails without my changes)

image

### DIFF
--- a/Python/Product/VsPyProf/python.h
+++ b/Python/Product/VsPyProf/python.h
@@ -30,7 +30,8 @@ enum PythonVersion {
     PythonVersion_34 = 0x0304,
     PythonVersion_35 = 0x0305,
     PythonVersion_36 = 0x0306,
-    PythonVersion_37 = 0x0307
+    PythonVersion_37 = 0x0307,
+    PythonVersion_38 = 0x0308
 };
 
 typedef const char* (GetVersionFunc)();
@@ -57,6 +58,7 @@ static PythonVersion GetPythonVersion(HMODULE hMod) {
                 case '5': return PythonVersion_35;
                 case '6': return PythonVersion_36;
                 case '7': return PythonVersion_37;
+                case '8': return PythonVersion_38;
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


Python/Product/VsPyProf/python.h
```diff
@@ -231,15 +233,47 @@ class PyCodeObject37 : public PyObject {
...
+// 3.8
+class PyCodeObject38 : public PyObject {
+public:
+    int co_argcount;            /* #arguments, except *args */
+    int co_posonlyargcount;     /* #positional only arguments */
+    int co_kwonlyargcount;      /* #keyword only arguments */
+    int co_nlocals;             /* #local variables */
+    int co_stacksize;           /* #entries needed for evaluation stack */
+    int co_flags;               /* CO_..., see below */
+    int co_firstlineno;         /* first source line number */
```
bschnurr: shouldn't we initialize all these pointers below to null?

Python/Product/VsPyProf/python.h
```diff
@@ -231,15 +233,47 @@ class PyCodeObject37 : public PyObject {
...
-        return version >= PythonVersion_37;
+        return version == PythonVersion_37;
+    }
+};
+
+// 3.8
+class PyCodeObject38 : public PyObject {
+public:
+    int co_argcount;            /* #arguments, except *args */
+    int co_posonlyargcount;     /* #positional only arguments */
```
huguesv: @RaymonGulati1 fyi this `co_posonlyargcount` is new in 3.8

Python/Product/VsPyProf/python.h
bschnurr: in PythonApi.cpp I see null checking but these could be pointing to garbage

  argNames = (PyTupleObject*)((PyCodeObject25_27*)codeObj)->co_varnames;

if (argCount != 0 && argNames && argNames->ob_type == PyTuple_Type) {

huguesv: no, we never instantiate this, we only need the structure of the class so we can offset correctly into existing objects in memory

bschnurr: or is there some kind of special allocator for these?

PR #5721, opened 2019-09-11
Unable to cancel a running pytest Fix #5542

### LINKED ISSUE #5542: Unable to cancel a running pytest
**Describe the bug**
Currently hitting cancel still waits for the current pytest run to finish.

We support cancelling in between running project test groups but we need to also be able kill the running pytest process.

bschnurr:
- This is just a force process kill for now
- Added the same cancel test that unittest has


huguesv:
Either connect this to an issue or remember to close it manually when you merge.

PR #5731, opened 2019-09-16
Adding UI tests for debbugging Test Explorer tests.

### LINKED ISSUE #5633: Add UI test for debug test
We have integration tests in `TestExplorerTests` class.

Debugging is not covered by the unit tests (and it would be difficult to since it needs to talk to an instance of VS), then we should add integration tests for it.

Cover: 
- unittest and pytest
- ptvsd 4.x and native code debugging (don't bother with legacy)
- set a breakpoint in a test, debug test, wait for break mode, validate location and loaded debugger

bschnurr:
Fix #5633

Tests cover:
pytest/unittest
project/workspace

### DIFF
--- a/Common/Tests/Utilities.UI/UI/VisualStudioApp.cs
+++ b/Common/Tests/Utilities.UI/UI/VisualStudioApp.cs
@@ -65,8 +65,7 @@ namespace TestUtilities.UI {
                 using (var dlg = new AutomationDialog(this, AutomationElement.FromHandle(hwnd))) {
                     //startup window popup, press button "Continue without code"
                     var cwc = dlg.FindByName("Continue without code");
-                    if (cwc != null)
-                    {
+                    if (cwc != null) {
                         Invoke(cwc);
                         WaitForInputIdle();
                         return;
@@ -426,7 +425,7 @@ namespace TestUtilities.UI {
             return doc.Selection.Text;
         }
 
-        public void WaitForOutputWindowText(string name, string containsText, int timeout=5000) {
+        public void WaitForOutputWindowText(string name, string containsText, int timeout = 5000) {
             for (int i = 0; i < timeout; i += 500) {
                 var text = GetOutputWindowText(name);
                 if (text.Contains(containsText)) {
@@ -446,7 +445,7 @@ namespace TestUtilities.UI {
                 if (uiShell == null) {
                     return;
                 }
-                
+
                 IntPtr hwnd;
                 uiShell.GetDialogOwnerHwnd(out hwnd);
 
@@ -982,7 +981,7 @@ namespace TestUtilities.UI {
             }
 
             var projects = EnumerateLoadedProjects(solution).ToList();
-            
+
             if (expectedProjects != null && expectedProjects.Value != projects.Count) {
                 // if we have other files open we can end up with a bonus project...
                 Assert.AreEqual(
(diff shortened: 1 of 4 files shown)


Python/Tests/Core.UI/TestExplorerTests.cs
```diff
@@ -122,6 +169,31 @@ public class TestExplorerTests {
...
+            app.Dte.Debugger.Breakpoints.Add(File: test.SourceFile, Line: breakLineno);
+
+            Console.WriteLine($"Debugging test {test.Name}");
+
+            testExplorer.DebugAll();
+
+            app.WaitForMode(EnvDTE.dbgDebugMode.dbgBreakMode);
+
+            Assert.IsNotNull(app.Dte.Debugger.BreakpointLastHit);
+            Assert.AreEqual(breakLineno, app.Dte.Debugger.BreakpointLastHit.FileLine);
```
bschnurr: currently only checking breakpoint was hit and line.  Couldn't find a way to check which debugger is loaded.

Python/Tests/PythonToolsUITestsRunner/TestExplorerTests.cs
```diff
@@ -62,5 +62,29 @@ public class TestExplorerTests {
...
+        [TestCategory("Installed")]
+        public void DebugUnittestProject() {
+            _vs.RunTest(nameof(PythonToolsUITests.TestExplorerTests.DebugUnittestProject));
+        }
+
+        [TestMethod, Priority(0)]
+        [TestCategory("Installed")]
+        public void DebugUnittestWorkspace() {
+            _vs.RunTest(nameof(PythonToolsUITests.TestExplorerTests.DebugUnittestWorkspace));
+        }
```
bschnurr: Note: no native debugging tests yet

Common/Tests/Utilities.UI/UI/VisualStudioApp.cs
```diff
@@ -65,8 +65,7 @@ public VisualStudioApp(IServiceProvider site)
                 using (var dlg = new AutomationDialog(this, AutomationElement.FromHandle(hwnd))) {
                     //startup window popup, press button "Continue without code"
                     var cwc = dlg.FindByName("Continue without code");
-                    if (cwc != null)
-                    {
+                    if (cwc != null) {
```
bschnurr: ran reformat on the document

Python/Tests/Core.UI/TestExplorerTests.cs
```diff
@@ -122,6 +169,31 @@ public class TestExplorerTests {
             return workspaceFolderPath;
         }
 
+        private static void DebugTest(PythonVisualStudioApp app, TestInfo test, TestInfo[] tests) {
```
bschnurr: i'm only debugging a single test, but I still need all tests so that I can wait for discovery to finish and check it discovered all the tests.

Python/Tests/Core.UI/TestExplorerTests.cs
```diff
@@ -122,6 +169,31 @@ public class TestExplorerTests {
...
+
+            Console.WriteLine($"Debugging test {test.Name}");
+
+            testExplorer.DebugAll();
+
+            app.WaitForMode(EnvDTE.dbgDebugMode.dbgBreakMode);
+
+            Assert.IsNotNull(app.Dte.Debugger.BreakpointLastHit);
+            Assert.AreEqual(breakLineno, app.Dte.Debugger.BreakpointLastHit.FileLine);
+        }
```
huguesv: The test ends while being stopped at a breakpoint. How does that behave? Say you run all 4 debug ui tests?

bschnurr: I think you're right... when I ran these all it didn't have a problem but it might for other tests. i'll see what the regular debug tests do for cleanup

Python/Tests/Utilities.Python/PythonTestExplorer.cs
```diff
@@ -151,6 +152,7 @@ public PythonTestExplorer(VisualStudioApp app, AutomationElement element, Automa
 
         public void ClearSearchBar() {
             _searchBar.SetValue("");
+            Thread.Sleep(500);
```
bschnurr: new sleep.. started seeing errors related to the serachbar not clearing fast enough

Python/Tests/Core.UI/TestExplorerTests.cs
bschnurr: updated

Python/Tests/Core.UI/TestExplorerTests.cs
```diff
@@ -122,6 +169,33 @@ public class TestExplorerTests {
...
+            Console.WriteLine($"Debugging test {test.Name}");
+
+            testExplorer.DebugAll();
+
+            app.WaitForMode(EnvDTE.dbgDebugMode.dbgBreakMode);
+
+            Assert.IsNotNull(app.Dte.Debugger.BreakpointLastHit);
+            Assert.AreEqual(breakLineno, app.Dte.Debugger.BreakpointLastHit.FileLine);
+
+            app.Dte.Debugger.Stop(WaitForDesignMode: true);
```
bschnurr: Now we stop the debugger

PR #5749, opened 2019-09-19
Removing "Test container source not found" warning

### LINKED ISSUE #5743: "Test container source not found" warning appears too often
**Describe the bug**
Now that Test Explorer has partially fixed the issue with discovery thinking every file has changed, we are seeing this warning due to finding tests that we now think dont exist in the project because TE is now sending us only the file/s that have changed.

**Steps to Reproduce**
1. New python project with 2 test files
2. Modify one file and save
3. notice output window

**Expected behavior**
warning should not appear because both files have been added to the project.

image

VS Version: [Can be found under "Help->About Microsoft Visual Studio"]
PTVS version: [Can be found under "Help->About Microsoft Visual Studio->Installed products"]
Python version:
Debugger Type (if applicable): [New debugger or legacy debugger?]

**Diagnostics file**
The diagnostics file contains information such as the Python version, environments, loaded assemblies, configuration options, etc. This file makes it easier for us to reproduce and fix bugs, so we would really appreciate it if you could provide it to us. This file can be generated by going to Tools->Options->Python->Diagnostics.
You can attach the file to this issue or email it to us at ptvshelp@microsoft.com. If you email it, please add a link to this issue.

bschnurr:
- Now that Test Explorer is no longer sending our discovery task a list of all sources, our logic around warning the user when tests are found in files not in the project is not valid.

- I spoke with Eric about incremental pytest, limiting discovery to just the source files that changed, but I still dont think its a good idea, because users can specify filename patterns in their pytest.ini and then when we pass files in to pytest as arguments they bypass those settings.

Fix #5743

### DIFF
--- a/Python/Product/TestAdapter.Executor/Pytest/PytestExtensions.cs
+++ b/Python/Product/TestAdapter.Executor/Pytest/PytestExtensions.cs
@@ -44,17 +44,8 @@ namespace Microsoft.PythonTools.TestAdapter.Pytest {
 
         public static TestCase ToVsTestCase(
             this PytestTest test,
-            string source,
-            int line,
-            Dictionary<string, PytestParent> parentMap,
             string projectHome
         ) {
-            if (parentMap == null) {
-                throw new ArgumentException(nameof(parentMap));
-            }
-            if (String.IsNullOrWhiteSpace(source)) {
-                throw new ArgumentException(nameof(source) + " " + test.ToString());
-            }
             if (String.IsNullOrWhiteSpace(projectHome)) {
                 throw new ArgumentException(nameof(projectHome));
             }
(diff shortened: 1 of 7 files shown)


Python/Product/TestAdapter.Executor/Pytest/PytestExtensions.cs
```diff
@@ -44,31 +44,29 @@ static internal class PyTestExtensions {
 
         public static TestCase ToVsTestCase(
             this PytestTest test,
```
bschnurr: moving these values below inside ToVsTestCase

Python/Product/TestAdapter.Executor/Pytest/TestDiscovererPytest.cs
```diff
@@ -94,41 +109,27 @@ ITestCaseDiscoverySink discoverySink
...
-                        //Test Explorer will show a key not found exception if we use a source path that doesn't match a test container's source.
-                        if (_settings.TestContainerSources.TryGetValue(parsedFullSourcePath, out string testContainerSourcePath)) {
-                            TestCase tc = test.ToVsTestCase(
-                                testContainerSourcePath, 
-                                line, 
-                                parentMap, 
-                                _settings.ProjectHome);
-                            discoverySink?.SendTestCase(tc);
-                        } else {
-                            Warn(Strings.ErrorTestContainerNotFound.FormatUI(_settings.ProjectHome, test.ToString()));
```
bschnurr: removing warnings...

PR #5762, opened 2019-09-25
Unittest test discovery misses tests in folder after adding __init__.…

### LINKED ISSUE #5751: Unittest test discovery misses tests in folder after adding __init__.py
**Describe the bug**
[A clear and concise description of the bug]

**Steps to Reproduce**
1. Create a new python project 
2. set unittest as the testframework
3. add test_1.py 
4. add a new folder eg. "Uppercase" and add a new test "test_Upper.py"
5. Note test_Upper.py is missing in Test Explorer after discovery runs
6. Add "__init__.py" to "Uppercase" folder
7. Note test_Upper.py is still missing
8. edit test_Upper.py and save
9. test now appears in Test Explorer

**Expected behavior**
After adding __init__.py to a test folder, discovery should find all the new tests the first time, without requiring the user to modify each file.

[Add any other information such as screenshots, error message, or stack trace]
image

bschnurr:
…py Fix #5751

 - Now that Test Explorer is no longer thinking every source has changed, our discovery results are being filtered to just those files that have changed.

- However there are cases like adding a __init__.py file that could affect discovery for other files in the same folder (or below). To catch these changes we now force a rebuild of all testcontainers in the solution.

### DIFF
--- a/Python/Product/TestAdapter/TestContainerDiscovererProject.cs
+++ b/Python/Product/TestAdapter/TestContainerDiscovererProject.cs
@@ -113,7 +113,7 @@ namespace Microsoft.PythonTools.TestAdapter {
 
         public IEnumerable<ITestContainer> TestContainers {
             get {
-                if (IsSolutionMode() && 
+                if (IsSolutionMode() &&
                     (_firstLoad || _isRefresh)) {
                     // The first time through, we don't know about any loaded
                     // projects.
@@ -123,10 +123,10 @@ namespace Microsoft.PythonTools.TestAdapter {
                             _firstLoad = false;
                             _isRefresh = false;
 
-                             _projectMap.Clear();
+                            _projectMap.Clear();
                             _packageManagerEventSink.UnwatchAll();
 
-                             SetupSolution();
+                            SetupSolution();
                         }
                     });
                 }
@@ -241,7 +241,7 @@ namespace Microsoft.PythonTools.TestAdapter {
             var projInfo = new ProjectInfo(pyProj);
             _projectMap[projInfo.ProjectHome] = projInfo;
             var files = FilteredTestOrSettingsFiles(vsProject);
-            UpdateSolutionTestContainersAndFileWatchers(files, projInfo, isAdd: true);
+            UpdateContainersAndListeners(files, projInfo, isAdd: true);
             return files.Any();
         }
 
@@ -300,7 +300,7 @@ namespace Microsoft.PythonTools.TestAdapter {
                 && _projectMap.TryRemove(projectHome, out ProjectInfo projToRemove)) {
 
                 var testFilesToRemove = projToRemove.GetAllContainers().Select(t => t.Source);
-                UpdateSolutionTestContainersAndFileWatchers(testFilesToRemove, projToRemove, isAdd: false);
+                UpdateContainersAndListeners(testFilesToRemove, projToRemove, isAdd: false);
                 projToRemove.Dispose();
                 return testFilesToRemove.Any();
             }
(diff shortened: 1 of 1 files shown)


Python/Product/TestAdapter/TestContainerDiscovererProject.cs
```diff
@@ -241,7 +241,7 @@ class TestContainerDiscovererProject : ITestContainerDiscoverer, IDisposable {
             var projInfo = new ProjectInfo(pyProj);
             _projectMap[projInfo.ProjectHome] = projInfo;
             var files = FilteredTestOrSettingsFiles(vsProject);
-            UpdateSolutionTestContainersAndFileWatchers(files, projInfo, isAdd: true);
+            UpdateContainersAndListeners(files, projInfo, isAdd: true);
```
bschnurr: rename.
previous name was from when openfolder and solution logic were both in this file

Python/Product/TestAdapter/TestContainerDiscovererProject.cs
```diff
@@ -309,10 +309,13 @@ class TestContainerDiscovererProject : ITestContainerDiscoverer, IDisposable {
 
         private IEnumerable<string> FilteredTestOrSettingsFiles(IVsProject project) {
             return project.GetProjectItems()
-                .Where(s => IsTestFileOrSetting(s));
```
bschnurr: this function isnt really needed

Python/Product/TestAdapter/TestContainerDiscovererProject.cs
```diff
@@ -328,81 +331,101 @@ class TestContainerDiscovererProject : ITestContainerDiscoverer, IDisposable {
...
-                    case TestFileChangedReason.Removed:
-                        _testFilesUpdateWatcher.RemoveWatch(e.File);
-                        break;
-                }
+                HandleSettingsChanged(e);
+            }
+            // VS uses temp files we need to ignore
+            else if (IsTestFile(e.File)) {
+                HandleSourcesChanged(e);
+            }
```
bschnurr: tried to condense this logic

Python/Product/TestAdapter/TestContainerDiscovererProject.cs
```diff
@@ -328,81 +331,101 @@ class TestContainerDiscovererProject : ITestContainerDiscoverer, IDisposable {
...
+                    return foundProjectInfo;
                 }
-                NotifyContainerChanged();
             }
-            
+
+            //Renamed  old files are no longer in the project so linear search
+            var projectInfo = _projectMap.Values.FirstOrDefault(p => p.TryGetContainer(file, out _));
+            return projectInfo;
+        }
```
bschnurr: there was a bug in the previous logic to find the projectInfo when a file was renamed because its no longer in the 
VsProject but its still in our _projectMap of testcontainers.

Also some events send a IVsProject and so do not so moving this complicated logic into its own function

Python/Product/TestAdapter/TestContainerDiscovererProject.cs
```diff
@@ -328,81 +331,101 @@ class TestContainerDiscovererProject : ITestContainerDiscoverer, IDisposable {
...
+                    //In this case we just have to fire TestContainerChnaged event
+                    break;
             }
 
-            // VS uses temp files we need to ignore
-            if (!IsTestFile(e.File))
-                return;
+            if (ShouldRebuild(e.File) || ShouldRebuild(e.OldFile)) {
+                _isRefresh = true;
+            }
```
bschnurr: this is the fix for when adding  __init__.py

Its a bit overkill, I was going to just grab the files in the same directory, but then I should also check child folders.

Alternatively I could try unloading and reloading the project or manually looping through the project to update the timestamp on the testscontainers.

PR #5767, opened 2019-09-27
TestExplorer displays extra child nodes for pytest parameter sets con…

### LINKED ISSUE #5658: TestExplorer displays extra child nodes for pytest parameter sets containing floats
**Describe the bug**
any pytest parameter set containing floats wont be parsed correctly by Test Explorer since it splits on "." and "::"

**Steps to Reproduce**
1. clone https://github.com/microsoft/ptvsd
2. enable pytest
3. search for test "test_any"

**Expected behavior**
No additional child nodes.
Fully Qualify names need to be delimited on "." and "::"

[Add any other information such as screenshots, error message, or stack trace]
image

bschnurr:
…taining floats Fix #5658

Previously we added an override to pytest's id generation to fix strings with "." and replace it with "_" in parameter set defintions.

This fix is for floats/doubles in parameter sets, but now we fix the displaystring and FullyQualifiedName we give test explorer, because they split on "."

### DIFF
--- a/Python/Product/TestAdapter.Executor/Pytest/PytestExtensions.cs
+++ b/Python/Product/TestAdapter.Executor/Pytest/PytestExtensions.cs
@@ -64,13 +64,13 @@ namespace Microsoft.PythonTools.TestAdapter.Pytest {
             var pytestId = CreateProperCasedPytestId(sourceFullPath, projectHome, test.Id);
             var fullyQualifiedName = CreateFullyQualifiedTestNameFromId(sourceFullPath, pytestId);
             var tc = new TestCase(fullyQualifiedName, PythonConstants.PytestExecutorUri, sourceFullPath) {
-                DisplayName = test.Name,
+                DisplayName = FixupParameterSets(test.Name),
                 LineNumber = line,
                 CodeFilePath = sourceFullPath
             };
 
             tc.SetPropertyValue(Constants.PytestIdProperty, pytestId);
-          
+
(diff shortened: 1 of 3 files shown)


Python/Product/TestAdapter.Executor/Pytest/PytestExtensions.cs
```diff
@@ -64,13 +64,13 @@ string projectHome
             var pytestId = CreateProperCasedPytestId(sourceFullPath, projectHome, test.Id);
             var fullyQualifiedName = CreateFullyQualifiedTestNameFromId(sourceFullPath, pytestId);
             var tc = new TestCase(fullyQualifiedName, PythonConstants.PytestExecutorUri, sourceFullPath) {
-                DisplayName = test.Name,
+                DisplayName = FixupParameterSets(test.Name),
```
bschnurr: Not a huge fan of calling FixupParameterSets in two spots. I think ideally we'd get displayname from FQN

Python/Product/TestAdapter.Executor/Pytest/PytestExtensions.cs
```diff
@@ -64,13 +64,13 @@ string projectHome
...
                 LineNumber = line,
                 CodeFilePath = sourceFullPath
             };
 
             tc.SetPropertyValue(Constants.PytestIdProperty, pytestId);
```
bschnurr: Note we dont modify the actual pytestId.. that is the string that execution passes on command line to pytest.  this is just a fix for Test Explorer

PR #5772, opened 2019-09-30
Assigning priorities to tests

### LINKED ISSUE #5707: Mark tests that should be run on PR creation/merge and commit to master
Change the priorities of tests such that we have a core set of tests that run on every PR creation and update. These tests should be quick and cover all the basic cases. 

We also want to mark a set of tests as non-critical, but still important to run. These tests should cover edge cases and run whenever a PR is merged into master.

RaymonGulati1:
Fix #5707


Python/Tests/DjangoTests/DjangoTemplateParserTests.cs
```diff
@@ -26,13 +26,15 @@
 using Microsoft.VisualStudio.Text;
 using Microsoft.PythonTools.Django.TemplateParsing.DjangoBlocks;
 using Microsoft.PythonTools.Django.TemplateParsing;
+using TestUtilities;
+using Classification = Microsoft.PythonTools.Django.TemplateParsing.Classification;
```
huguesv: there doesn't seem to be any other change in this file related to this using, is it unnecessary?

Python/Tests/DjangoTests/DjangoTemplateParserTests.cs
```diff
@@ -106,7 +108,7 @@ public class DjangoTemplateParserTests {
 
 #region Block parser tests
 
-        [TestMethod, Priority(0)]
+        [TestMethod, Priority(TestExtensions.IMPORTANT_UNIT_TEST)]
```
huguesv: these parser tests should be super fast, we could keep all of them in PRs

Python/Tests/Core/Pep440VersionTests.cs
```diff
@@ -82,7 +82,7 @@ public class Pep440VersionTests {
 1!1.0 == 01!1.0
 ".Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
 
-        [TestMethod, Priority(0)]
+        [TestMethod, Priority(TestExtensions.IMPORTANT_UNIT_TEST)]
```
huguesv: these pep440 tests should be super fast, we should keep all of them

Python/Tests/Core/PathUtilsTests.cs
```diff
@@ -31,7 +31,7 @@ public class PathUtilsTests {
             AssertListener.Initialize();
         }
 
-        [TestMethod, Priority(0)]
+        [TestMethod, Priority(TestExtensions.IMPORTANT_UNIT_TEST)]
```
huguesv: i'd like to keep all these pathutilstests

Python/Tests/DebuggerTests/DebuggerTests.cs
```diff
@@ -1641,7 +1641,7 @@ class ExceptionHandlerInfo {
             }
         }
 
-        [TestMethod, Priority(0)]
+        [TestMethod, Priority(TestExtensions.CORE_UNIT_TEST)]
```
huguesv: based on the category 60s, looks like `TestExceptions` may be a slow test, if it is slow, you could demote it.

Common/Tests/Utilities/TestExtensions.cs
```diff
@@ -24,6 +24,12 @@
 
 namespace TestUtilities {
     public static class TestExtensions {
+        public const int CORE_UNIT_TEST = 0;          // Run on every PR request going into master
+        public const int IMPORTANT_UNIT_TEST = 1;     // Run on every commit into master
```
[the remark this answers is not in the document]
RaymonGulati1: I'm a bit hesitant to use "Priority" because these numbers don't inherently mean anything. When the UI tests are implemented, they can't have the same priority as unit tests which implies they are less or more priority than unit test which is not really the case. I've changed the 2nd variable name to "SUPPLEMENTARY_UNIT_TEST". The new name and comment should be sufficiently descriptive.

Common/Tests/Utilities/TestExtensions.cs
```diff
@@ -24,6 +24,12 @@
...
+        
+        public const int P2_UNIT_TEST = 20;
```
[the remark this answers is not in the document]
RaymonGulati1: Changed

Python/Tests/Core/AutoIndentTests.cs
```diff
@@ -47,7 +47,7 @@ public class AutoIndentTests {
             Assert.AreEqual(6, AutoIndent.GetIndentation(" \t p", 4));
         }
 
-        [TestMethod, Priority(0)]
+        [TestMethod, Priority(TestExtensions.IMPORTANT_UNIT_TEST)]
```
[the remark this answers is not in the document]
RaymonGulati1: Changed

Python/Tests/DjangoTests/DjangoTemplateParserTests.cs
RaymonGulati1: Required because adding TestUtilities causes ambiguity for "Classification" which is a imported in "…Django.TemplateParsing" and TestUtilities.

image

Python/Tests/DjangoTests/DjangoTemplateParserTests.cs
RaymonGulati1: Changed

Python/Tests/Core/PathUtilsTests.cs
RaymonGulati1: Changed

Python/Tests/DebuggerTests/DebuggerTests.cs
RaymonGulati1: Changed

Common/Tests/Utilities/TestExtensions.cs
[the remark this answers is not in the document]
huguesv: When it comes to priority, P0, P1, P2 is the standard way of referring to them. If you look at the issues labels on our repo, we have priority:P0, priority:P1, priority:P2 and priority:release-blocking.

priority:release-blocking was an attempt at giving a name to a priority, thinking that it would be more understandable. It wasn't. No one is used to that, and now it's been abandoned, partly due to the inconsistency with other repos our team owns which use P0 as the highest possible priority.

Core is a term used in the product to represent the core part of the extension, there are folders named core, so someone could think that it somehow refers to that.

If you want different values for unit test and ui tests, you can still achieve that while having priorities that are easily ranked.

UI tests can use a different constant class:

```
public static class UnitTestPriority {
    public const int P0 = 0;
    public const int P1 = 1;
    public const int P2 = 2;
    public const int P0_FAILING = 11;
    public const int P1_FAILING = 21;
    public const int P2_FAILING = 31;
}

public static class UITestPriority {
    public const int P0 = 0; // or whatever integer value you want
    public const int P1 = 1;
    public const int P2 = 2;
    public const int P0_FAILING = 11;
    public const int P1_FAILING = 21;
    public const int P2_FAILING = 31;
}
```

Or they could be different constants in the same class:

```
public static class TestPriority {
    public const int UNIT_TEST_P0 = 0;
    public const int UNIT_TEST_P1 = 1;
    public const int UNIT_TEST_P2 = 2;
    public const int UNIT_TEST_P0_FAILING = 11;
    public const int UNIT_TEST_P1_FAILING = 21;
    public const int UNIT_TEST_P2_FAILING = 31;

    public const int UI_TEST_P0 = 0; // or whatever integer value you want
    public const int UI_TEST_P1 = 1;
    public const int UI_TEST_P2 = 2;
    public const int UI_TEST_P0_FAILING = 11;
    public const int UI_TEST_P1_FAILING = 21;
    public const int UI_TEST_P2_FAILING = 31;
}

```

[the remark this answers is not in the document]
RaymonGulati1: Changed.

Common/Tests/Utilities/UnitTestPriority.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace TestUtilities {
+    public static class UnitTestPriority {
+        public const int P0 = 0;    // Run on every PR request going into master
+        public const int P1 = 1;    // Run on every commit into master
+
+        public const int P2_UNIT_TEST = 20;
```
huguesv: no need for _UNIT_TEST suffix here

Python/Tests/Core/BuildTasksTests.cs
```diff
@@ -46,7 +46,7 @@ public class BuildTasksTests {
             AssertListener.Initialize();
         }
 
-        //[TestMethod, Priority(3)]
+        //[TestMethod, UnitTestPriority(TestExtensions.P3_UNIT_TEST)]
```
huguesv: You should either undo these or make them use the correct constant so it compiles when we uncomment

Common/Tests/Utilities/UnitTestPriority.cs
RaymonGulati1: Done

Python/Tests/Core/BuildTasksTests.cs
RaymonGulati1: Done

Common/Tests/Utilities/UnitTestPriority.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace TestUtilities {
```
huguesv: missing copyright header

PR #5773, opened 2019-09-30
Adding back caching packagemanagers so that test frameworks will run …

### LINKED ISSUE #5520: Installing/uninstalling a package should trigger test discovery
This is partially implemented but not working due to the reversal of the package manager single instance change. In summary, since everyone creates their own package manager instance, it is not possible to register for events to know something was installed. See #5517

bschnurr:
…discovery on package changes.

Installing/uninstalling a package should trigger test discovery Fix  #5520

### DIFF
--- a/Python/Product/VSInterpreters/PackageManager/CPythonCondaPackageManagerProvider.cs
+++ b/Python/Product/VSInterpreters/PackageManager/CPythonCondaPackageManagerProvider.cs
@@ -24,7 +24,7 @@ namespace Microsoft.PythonTools.Interpreter {
     [Export(typeof(IPackageManagerProvider))]
     sealed class CPythonCondaPackageManagerProvider : IPackageManagerProvider {
         private readonly IServiceProvider _site;
-        //private readonly Dictionary<IPythonInterpreterFactory, IPackageManager> _packageManagerMap;
+        private readonly Dictionary<IPythonInterpreterFactory, IPackageManager> _packageManagerMap;
         private Lazy<string> _latestCondaExe;
 
         [ImportingConstructor]
@@ -32,7 +32,7 @@ namespace Microsoft.PythonTools.Interpreter {
             [Import(typeof(SVsServiceProvider), AllowDefault = true)] IServiceProvider site = null
         ) {
             _site = site;
-            //_packageManagerMap = new Dictionary<IPythonInterpreterFactory, IPackageManager>();
+            _packageManagerMap = new Dictionary<IPythonInterpreterFactory, IPackageManager>();
 
             // This can be slow, if there are 2 or more global conda installations
             // (some conda versions have long startup time), so we only fetch it once.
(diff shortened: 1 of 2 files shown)


Python/Product/VSInterpreters/Interpreter/InterpreterOptionsService.cs
```diff
@@ -344,9 +344,9 @@ sealed class InterpreterOptionsService : IInterpreterOptionsService {
             }
 
             return _packageManagerProviders.SelectMany(p => p.Value.GetPackageManagers(factory))
+                .Where(p => p != null)
                 .GroupBy(p => p.UniqueKey)
                 .Select(g => g.OrderBy(p => p.Priority).FirstOrDefault())
-                .Where(p => p != null)
```
bschnurr: GetPackageManagers can return null so added the new where statement above so it filters out all null.

but i'm not sure if I need the  old null check still

Python/Tests/Core/EnvironmentListTests.cs
```diff
@@ -945,7 +945,12 @@ public class EnvironmentListTests {
                     python.Version.ToVersion()
                 )
             );
-            service.AddPackageManagers(factory, new CPythonPipPackageManagerProvider(null).GetPackageManagers(factory).ToList());
+            var pmList = new CPythonPipPackageManagerProvider(null)
+                .GetPackageManagers(factory)
+                .Where(p => p != null)
+                .ToList();
+
+            service.AddPackageManagers(factory, pmList);
```
bschnurr: adding null filters

Python/Tests/Core/EnvironmentListTests.cs
```diff
@@ -997,7 +1002,13 @@ public class EnvironmentListTests {
...
             );
             var condaLocatorProvider = new MockCondaLocatorProvider(new MockCondaLocator(condaExePath));
-            service.AddPackageManagers(factory, new CPythonPipPackageManagerProvider(condaLocatorProvider).GetPackageManagers(factory).Prepend(new CondaPackageManager(factory, condaExePath)).ToArray());
+            var packageProvider = new CPythonPipPackageManagerProvider(condaLocatorProvider);
+            var pmList = packageProvider.GetPackageManagers(factory)
+                .Where(p => p != null)
+                .Prepend(new CondaPackageManager(factory, condaExePath))
+                .ToList();
+
+            service.AddPackageManagers(factory, pmList);
```
bschnurr: adding null filters

Python/Product/VSInterpreters/Interpreter/InterpreterOptionsService.cs
bschnurr: I think now the old code was correct

bschnurr: I think by filtering the nulls first  the FirstOrDefault() call will always return something.

Python/Product/VSInterpreters/PackageManager/CPythonCondaPackageManagerProvider.cs
```diff
@@ -72,7 +67,7 @@ sealed class CPythonCondaPackageManagerProvider : IPackageManagerProvider {
             }
 
             var condaPath = CondaUtils.GetCondaExecutablePath(prefixPath);
-            if (string.IsNullOrEmpty(condaPath)) {
+            if (string.IsNullOrEmpty(condaPath) && _latestCondaExe.IsValueCreated) {
```
huguesv: `.Value` is what will trigger the lazy evaluation and set `.IsValueCreated` to `true`, but since `.Value` is only called inside this if statement, that will never happen. Was there unexpected behavior before your change?

Python/Tests/Core/EnvironmentListTests.cs
huguesv: Both package manager providers in this PR check and make sure they don't yield return null, so there's no need to check for that AFAIK. If they don't have package managers, they'll return empty enumerable, not an enumerable with a null entry.

Python/Product/VSInterpreters/PackageManager/CPythonCondaPackageManagerProvider.cs
bschnurr: i was worried value was null. but I think it actually handles that case. this isn't needed.

Python/Tests/Core/EnvironmentListTests.cs
bschnurr: Ah I see that now.

Issue #5617, opened 2019-08-20
Invalid cross-thread call when opening workspace


huguesv:
I got while opening a workspace.

```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
Invalid cross-thread call from thread 14

   at Microsoft.VisualStudioTools.UIThreadExtensions.MustBeCalledFromUIThread(UIThreadBase self, String message)

   at Microsoft.PythonTools.Extensions.GetPythonToolsService(IServiceProvider serviceProvider)

   at Microsoft.PythonTools.Project.CondaEnvCreateInfoBar.get_IsGloballySuppressed()

   at Microsoft.PythonTools.Project.CondaEnvCreateWorkspaceInfoBar.<CheckAsync>d__4.MoveNext()

   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[TStateMachine](TStateMachine& stateMachine)

   at Microsoft.PythonTools.Project.CondaEnvCreateWorkspaceInfoBar.CheckAsync()

   at Microsoft.PythonTools.WorkspaceInfoBarManager.<TriggerInfoBarsAsync>d__14.MoveNext()

   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[TStateMachine](TStateMachine& stateMachine)

   at Microsoft.PythonTools.WorkspaceInfoBarManager.TriggerInfoBarsAsync()

   at Microsoft.PythonTools.WorkspaceInfoBarManager.TriggerInfoBar()

   at Microsoft.PythonTools.WorkspaceInfoBar......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------

```

PR #5794, opened 2019-10-03
Info bar fix for cross thread call warning

bschnurr:
- Moved the call to get the pythontoolssevice into the constructor of the base class so that is occurs on the mainthread

Invalid cross-thread call when opening workspace Fx #5617

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonInfoBar.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonInfoBar.cs
@@ -16,8 +16,6 @@
 
 using System;
 using System.Diagnostics;
-using Microsoft.PythonTools.Infrastructure;
-using Microsoft.PythonTools.Interpreter;
 using Microsoft.PythonTools.Logging;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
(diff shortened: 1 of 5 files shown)


huguesv:
So the constructor is always called from UI thread?

You should check all derived classes and see if they use python tools service, and they should be updated to use the cache value.

bschnurr:
I guess technically its still not guaranteed to run on the main thread.

bschnurr:
I'm going to redo this

Python/Product/PythonTools/PythonTools/Project/PythonInfoBar.cs
```diff
@@ -50,6 +48,10 @@ internal abstract class PythonInfoBar : IVsInfoBarUIEvents, IDisposable {
 
         public abstract Task CheckAsync();
 
+        protected virtual bool IsGloballySuppressed  =>
+            /// Needs to be called from mainthread
+            !Site.GetPythonToolsService().GeneralOptions.PromptForEnvCreate;
+
```
huguesv: this is the generic base class, it shouldn't be looking at any specific option.
it's also not called from anywhere in this class, so doesn't need to be defined here.
should be moved back to `CondaEnvCreateInfoBar`

bschnurr: ok. will do

bschnurr: done

PR #5802, opened 2019-10-07
Python not supported info bar

### LINKED ISSUE #5787: Show info bar on project/workspace load about Python 3.8 support
Let user know of the gaps in our support.
- debugging, native debugging, profiling work
- editor new language features will show syntax errors

RaymonGulati1:
Fix #5787

### DIFF
--- a/Python/Product/Common/Strings.resx
+++ b/Python/Product/Common/Strings.resx
@@ -2943,7 +2943,7 @@ Please specify at least one package.</value>
 {2} is "project" or "workspace"
 {3} is an integer wait time in seconds</comment>
   </data>
-    <data name="PythonTestDiscovererTimeoutErrorMessage" xml:space="preserve">
+  <data name="PythonTestDiscovererTimeoutErrorMessage" xml:space="preserve">
     <value>Python test discovery timeout, possibly due to blocking call. Go to https://aka.ms/VSTestExplorerPython for more details on test discovery.</value>
   </data>
   <data name="UnitTestEnableInfoBarAction" xml:space="preserve">
@@ -2968,6 +2968,14 @@ Please specify at least one package.</value>
     <value>Results for pytest not found {0}</value>
     <comment>{0} is the expected file path</comment>
   </data>
-
-    
+  <data name="DoNotShowMessageAgain" xml:space="preserve">
+    <value>Do not show this message again</value>
+  </data>
+  <data name="PythonVersionNotSupportInfoBarText" xml:space="preserve">
+    <value>Certain features may not work properly because you are using Python {0}, but VS doesn't officially support it. </value>
+    <comment>{0} is the Python version</comment>
+  </data>
+  <data name="PythonVersionNotSupportMoreInformation" xml:space="preserve">
+    <value>More information</value>
+  </data>
 </root>
--- a/Python/Product/Common/Strings.resx
+++ b/Python/Product/Common/Strings.resx
@@ -2968,14 +2968,14 @@ Please specify at least one package.</value>
     <value>Results for pytest not found {0}</value>
     <comment>{0} is the expected file path</comment>
   </data>
-  <data name="DoNotShowMessageAgain" xml:space="preserve">
-    <value>Do not show this message again</value>
+  <data name="PythonVersionNotSupportedDontShowMessageAgain" xml:space="preserve">
+    <value>Don't show this message again</value>
   </data>
   <data name="PythonVersionNotSupportInfoBarText" xml:space="preserve">
-    <value>Certain features may not work properly because you are using Python {0}, but VS doesn't officially support it. </value>
+    <value>You are using Python {0} and some features might not work because VS doesn't support it yet.</value>
     <comment>{0} is the Python version</comment>
   </data>
-  <data name="PythonVersionNotSupportMoreInformation" xml:space="preserve">
-    <value>More information</value>
+  <data name="PythonVersionNotSupportMoreInfo" xml:space="preserve">
+    <value>More info</value>
   </data>
 </root>
--- a/Python/Product/Common/Strings.resx
+++ b/Python/Product/Common/Strings.resx
@@ -2971,8 +2971,8 @@ Please specify at least one package.</value>
   <data name="PythonVersionNotSupportedDontShowMessageAgain" xml:space="preserve">
     <value>Don't show this message again</value>
   </data>
-  <data name="PythonVersionNotSupportInfoBarText" xml:space="preserve">
-    <value>You are using Python {0} and some features might not work because VS doesn't support it yet.</value>
+  <data name="PythonVersionNotSupportedInfoBarText" xml:space="preserve">
+    <value>You are using Python {0} and some features might not work because Visual Studio doesn't support it yet.</value>
     <comment>{0} is the Python version</comment>
   </data>
   <data name="PythonVersionNotSupportMoreInfo" xml:space="preserve">
--- a/Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
@@ -27,32 +27,32 @@ using Microsoft.VisualStudio.Shell.Interop;
 
 namespace Microsoft.PythonTools.Project {
     internal class PythonNotSupportedInfoBar : PythonInfoBar {
+        private const string _moreInformationLink = @"https://go.microsoft.com/fwlink/?LinkId=2108304";
+        private readonly Version _pythonVersionNotSupported = new Version("3.8");
         private Func<IPythonInterpreterFactory> _getActiveInterpreterFunc;
-        private IPythonInterpreterFactory _activeInterpreter { get { return _getActiveInterpreterFunc(); } }
-
-        private GeneralOptions _options;
+        private IPythonInterpreterFactory _interpreterTriggeredInfoBar;
         private String _context;
 
-        private string _moreInformationLink = @"https://go.microsoft.com/fwlink/?LinkId=2108304";
-        private readonly Version _pythonVersionNotSupported = new Version("3.8");
-        private bool _infoBarTriggered = false;
-
         public PythonNotSupportedInfoBar(IServiceProvider site, string context, Func<IPythonInterpreterFactory> activeInterpreter) : base(site) {
-            _options = Site.GetPythonToolsService().GeneralOptions;
             _getActiveInterpreterFunc = activeInterpreter;
             _context = context;
         }
 
+        private IPythonInterpreterFactory ActiveInterpreter { get { return _getActiveInterpreterFunc(); } }
+
         public override Task CheckAsync() {
+            ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
             if (IsCreated ||
-                !_options.PromptForPythonVersionNotSupported ||
-                _infoBarTriggered ||
-                _activeInterpreter.Configuration.Version < _pythonVersionNotSupported
+                !Site.GetPythonToolsService().GeneralOptions.PromptForPythonVersionNotSupported ||
+                _interpreterTriggeredInfoBar != null ||
+                ActiveInterpreter == null ||
+                ActiveInterpreter.Configuration.Version < _pythonVersionNotSupported
             ) {
                 return Task.CompletedTask;
             }
 
-            var infoBarTextSpanMessage = new InfoBarTextSpan(Strings.PythonVersionNotSupportInfoBarText.FormatUI(_activeInterpreter.Configuration.Version));
+            var infoBarTextSpanMessage = new InfoBarTextSpan(Strings.PythonVersionNotSupportedInfoBarText.FormatUI(ActiveInterpreter.Configuration.Version));
             var infoBarMessage = new List<IVsInfoBarTextSpan>() { infoBarTextSpanMessage };
             var actionItems = new List<InfoBarActionItem>() {
                 new InfoBarHyperlink(Strings.PythonVersionNotSupportMoreInfo, (Action)MoreInformationAction),
@@ -61,7 +61,7 @@ namespace Microsoft.PythonTools.Project {
 
             LogEvent(PythonVersionNotSupportedInfoBarAction.Prompt);
             Create(new InfoBarModel(infoBarMessage, actionItems, KnownMonikers.StatusInformation));
-            _infoBarTriggered = true;
+            _interpreterTriggeredInfoBar = ActiveInterpreter;
 
             return Task.CompletedTask;
         }
@@ -77,17 +77,19 @@ namespace Microsoft.PythonTools.Project {
             LogEvent(PythonVersionNotSupportedInfoBarAction.Ignore);
             Close();
 
-            _options.PromptForPythonVersionNotSupported = false;
-            _options.Save();
+            Site.GetPythonToolsService().GeneralOptions.PromptForPythonVersionNotSupported = false;
+            Site.GetPythonToolsService().GeneralOptions.Save();
         }
 
         private void LogEvent(string action) {
+            var logInterpreter = _interpreterTriggeredInfoBar ?? ActiveInterpreter;
+
             Logger?.LogEvent(
                 PythonLogEvent.PythonNotSupportedInfoBar,
                 new PythonVersionNotSupportedInfoBarInfo() {
                     Action = action,
                     Context = _context,
-                    PythonVersion = _activeInterpreter.Configuration.Version
+                    PythonVersion = logInterpreter.Configuration.Version.ToString()
                 }
             );
         }
--- a/Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
@@ -33,7 +33,7 @@ namespace Microsoft.PythonTools.Project {
         private GeneralOptions _options;
         private String _context;
 
-        private string _moreInformationLink = @"https://github.com/microsoft/PTVS/issues/1"; //TODO Raymon  Change URL
+        private string _moreInformationLink = @"https://go.microsoft.com/fwlink/?LinkId=2108304";
         private readonly Version _pythonVersionNotSupported = new Version("3.8");
         private bool _infoBarTriggered = false;
 
--- a/Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
@@ -1,13 +1,29 @@
-﻿using Microsoft.PythonTools.Infrastructure;
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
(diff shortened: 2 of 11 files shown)


Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,144 @@
...
+                _workspaceContext.CurrentFactory.Configuration,
+                _workspaceContext.GetBoolProperty(_suppressInfoBarPropertyName) ?? false,
+                InfoBarContexts.Workspace
+                );
+
+            await InfoBarCreationRequest(infoBarData);
+        }
+
+        protected override async Task SetPropertyAsync(string propertyName, string propertyValue) {
+            await _workspaceContext.SetPropertyAsync(propertyName, propertyValue);
```
bschnurr: these one liners, you can skip the async and await, its more efficient to just return _workspaceContext.SetPropertyAsync

Python/Product/PythonTools/PythonTools/WorkspaceInfoBarManager.cs
```diff
@@ -128,10 +134,19 @@ class WorkspaceInfoBarManager : IVsRunningDocTableEvents, IDisposable {
...
+                _pythonSupportInfoBar.CheckAsync()
             );
         }
 
+        private async void TriggerPythonNotSupportedInforBar(object sender, EventArgs e) {
+            TriggerPythonNotSupportedInforBarAsync().HandleAllExceptions(_serviceProvider, GetType()).DoNotWait();
+        }
+
+        private async Task TriggerPythonNotSupportedInforBarAsync() {
+            await _pythonSupportInfoBar.CheckAsync();
```
bschnurr: here too.

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        private void DoNotShowAgainAction() {
+            DoNotShowAgainActionAsync().HandleAllExceptions(Site, GetType()).DoNotWait();
+        }
+
+        private async Task MoreInformationActionAsync() {
+            LogEvent(PythonVersionNotSupportedInfoBarAction.MoreInfo);
+            Close();
+
+            //TODO Raymon
```
bschnurr: todo done?

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+    }
+
+    internal sealed class PythonNotSupportedProjectInfoBar : PythonNotSupportedInfoBar {
+        private PythonProjectNode _project;
+
+        public PythonNotSupportedProjectInfoBar(IServiceProvider site, PythonProjectNode projectNode) : base(site) {
+            _project = projectNode ?? throw new ArgumentNullException(nameof(projectNode));
+
```
bschnurr: minor, extra space

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace Microsoft.PythonTools.Project {
+    internal class PythonNotSupportedInfoBar : PythonInfoBar {
+        private Func<IPythonInterpreterFactory> _getActiveInterpreterFunc;
+        private IPythonInterpreterFactory _activeInterpreter { get { return _getActiveInterpreterFunc(); } }
+
+        private GeneralOptions _options;
+        private String _context;
+
+        private string _moreInformationLink = @"https://github.com/microsoft/PTVS/issues/1"; //TODO Raymon  Change URL
```
bschnurr: minor: extra spaces above

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,95 @@
...
+        private readonly Version _pythonVersionNotSupported = new Version("3.8");
+        private bool _infoBarTriggered = false;
+
+        public PythonNotSupportedInfoBar(IServiceProvider site, string context, Func<IPythonInterpreterFactory> activeInterpreter) : base(site) {
+            _options = Site.GetPythonToolsService().GeneralOptions;
```
bschnurr: GetPythonToolsService() requires mainthread I believe, maybe better not to cache the options.

bschnurr: or add an init()

Python/Product/VSCommon/Logging/PythonToolsLoggerData.cs
```diff
@@ -167,6 +167,18 @@ static class PackageInstallInfoBarActions {
         public const string Ignore = "Ignore";
     }
 
+    sealed class PythonVersionNotSupportedInfoBarInfo : PythonToolsLoggerData {
+        public string Action { get; set; }
+        public string Context { get; set; }
+        public Version PythonVersion { get; set; }
```
huguesv: this needs to be a string, only simple types allowed, see similar cases in this file

Python/Product/Common/Strings.resx
```diff
@@ -2968,6 +2968,14 @@ Please specify at least one package.</value>
     <value>Results for pytest not found {0}</value>
     <comment>{0} is the expected file path</comment>
   </data>
-
-    
+  <data name="PythonVersionNotSupportedDontShowMessageAgain" xml:space="preserve">
+    <value>Don't show this message again</value>
+  </data>
+  <data name="PythonVersionNotSupportInfoBarText" xml:space="preserve">
+    <value>You are using Python {0} and some features might not work because VS doesn't support it yet.</value>
```
huguesv: UX folks were okay with abbreviating VS?

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
huguesv: and can be a constant

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
huguesv: as Bill says, call GetPythonToolsService() from CheckAsync (don't cache it) and call SwitchToMainThread at the top of CheckAsync

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,95 @@
...
+            _infoBarTriggered = true;
+
+            return Task.CompletedTask;
+        }
+
+        private void MoreInformationAction() {
+            LogEvent(PythonVersionNotSupportedInfoBarAction.MoreInfo);
+            Close();
+
+            VsShellUtilities.OpenBrowser(_moreInformationLink);
```
huguesv: Just making sure that this opens external browser, not the old internal one. If not, then just call `Process.Start(url).?Dispose()`

Python/Product/Common/Strings.resx
huguesv: Should rename to PythonVersionNotSupport*ed*InfoBarText

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.PythonTools.Logging;
+using Microsoft.PythonTools.Options;
+using Microsoft.VisualStudio.Imaging;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.PythonTools.Project {
+    internal class PythonNotSupportedInfoBar : PythonInfoBar {
+        private Func<IPythonInterpreterFactory> _getActiveInterpreterFunc;
+        private IPythonInterpreterFactory _activeInterpreter { get { return _getActiveInterpreterFunc(); } }
```
huguesv: private getters should still be named with pascal case and no underscore: `ActiveInterpreter`

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,95 @@
...
+            _options = Site.GetPythonToolsService().GeneralOptions;
+            _getActiveInterpreterFunc = activeInterpreter;
+            _context = context;
+        }
+
+        public override Task CheckAsync() {
+            if (IsCreated ||
+                !_options.PromptForPythonVersionNotSupported ||
+                _infoBarTriggered ||
+                _activeInterpreter.Configuration.Version < _pythonVersionNotSupported
```
huguesv: You need to protect against a null active interpreter. This will happen if there's no python installed on machine.

Python/Product/PythonTools/PythonTools/WorkspaceInfoBarManager.cs
```diff
@@ -128,10 +135,19 @@ class WorkspaceInfoBarManager : IVsRunningDocTableEvents, IDisposable {
                 _condaEnvCreateInfoBar.CheckAsync(),
                 _virtualEnvCreateInfoBar.CheckAsync(),
                 _packageInstallInfoBar.CheckAsync(),
-                _testFrameworkInfoBar.CheckAsync()
+                _testFrameworkInfoBar.CheckAsync(),
+                _pythonVersionNotSupportedInfoBar.CheckAsync()
             );
         }
 
+        private async void TriggerPythonNotSupportedInforBar(object sender, EventArgs e) {
```
huguesv: remove async keyword here

huguesv: You can just call `_pythonVersionNotSupportedInfoBar.CheckAsync().HandleAllExceptions(_serviceProvider, GetType()).DoNotWait();` right here and don't need the extra helper method.

Python/Product/Common/Strings.resx
RaymonGulati1: Recommendation is: "You're using Python 3.8 and some features might not work.", but it fails to convey the reason why it won't work, so I added that part in. Changed VS to "Visual Studio (see next commit).

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,95 @@
...
+            _options.Save();
+        }
+
+        private void LogEvent(string action) {
+            Logger?.LogEvent(
+                PythonLogEvent.PythonNotSupportedInfoBar,
+                new PythonVersionNotSupportedInfoBarInfo() {
+                    Action = action,
+                    Context = _context,
+                    PythonVersion = _activeInterpreter.Configuration.Version
```
huguesv: this re-evaluates active interpreter, which could now be different than the one that triggered the info bar (user switch to python 3.8, info bar appears, switch back to 3.7, then clicks on more info)

I guess this could be avoided if you were to cache the active interpreter when `_infoBarTriggered` and use that instead.

Python/Product/PythonTools/PythonTools/WorkspaceInfoBarManager.cs
```diff
@@ -50,17 +52,22 @@ class WorkspaceInfoBarManager : IVsRunningDocTableEvents, IDisposable {
...
 
             workspace.AddActionOnClose(_packageInstallInfoBar, (obj => ((PythonInfoBar)obj).Dispose()));
             workspace.AddActionOnClose(_condaEnvCreateInfoBar, (obj => ((PythonInfoBar)obj).Dispose()));
             workspace.AddActionOnClose(_virtualEnvCreateInfoBar, (obj => ((PythonInfoBar)obj).Dispose()));
             workspace.AddActionOnClose(_testFrameworkInfoBar, (obj => ((PythonInfoBar)obj).Dispose()));
-            _infoBarCheckTriggered = false;
+            workspace.AddActionOnClose(_pythonVersionNotSupportedInfoBar, (obj => ((PythonInfoBar)obj).Dispose()));
+
+            workspace.ActiveInterpreterChanged += TriggerPythonNotSupportedInforBar;
+            workspace.AddActionOnClose(_pythonVersionNotSupportedInfoBar, (obj) => workspace.ActiveInterpreterChanged -= TriggerPythonNotSupportedInforBar);
```
huguesv: You should check the implementation, but I think the first param is a key, and it will overwrite the previous `AddActionOnClose`, which would mean dispose wouldn't get called, so should be easy to verify.

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
RaymonGulati1: Opens external browser (edge on my computer)

Python/Product/PythonTools/PythonTools/WorkspaceInfoBarManager.cs
```diff
@@ -128,10 +137,19 @@ class WorkspaceInfoBarManager : IVsRunningDocTableEvents, IDisposable {
...
                 _virtualEnvCreateInfoBar.CheckAsync(),
                 _packageInstallInfoBar.CheckAsync(),
-                _testFrameworkInfoBar.CheckAsync()
+                _testFrameworkInfoBar.CheckAsync(),
+                _pythonVersionNotSupportedInfoBar.CheckAsync()
             );
         }
 
+        private void TriggerPythonNotSupportedInforBar(object sender, EventArgs e) {
+            TriggerPythonNotSupportedInforBarAsync().HandleAllExceptions(_serviceProvider, GetType()).DoNotWait();
```
bschnurr: Hey @huguesv isn't this a fire and forget case where we want to use   ThreadHelper.JoinableTaskFactor.runasync?

huguesv: Yes and No.

Yes, for `JoinableTaskFactory.RunAsync` but no for `ThreadHelper`. This is the case where they have mentioned we need to use our package's `JoinableTaskFactory`, not the static from `ThreadHelper`.

We do not use `JoinableTaskFactory.RunAsync` anywhere currently, we've been using `HandleAllExceptions().DoNotWait()` so far. I'd like to get our fire & forget calls changed at one point but I think we'll need a helper method on top of `RunAsync` to replicate some of the work that our helper does. I'd rather hold off for now and take the time to do it right.

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.PythonTools.Project {
+    internal class PythonNotSupportedInfoBar : PythonInfoBar {
+        private const string _moreInformationLink = @"https://go.microsoft.com/fwlink/?LinkId=2108304";
+        private readonly Version _pythonVersionNotSupported = new Version("3.8");
+        private Func<IPythonInterpreterFactory> _getActiveInterpreterFunc;
+        private IPythonInterpreterFactory _interpreterTriggeredInfoBar;
+        private String _context;
```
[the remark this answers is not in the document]
huguesv: could _context be made readonly?

Python/Product/PythonTools/PythonTools/WorkspaceInfoBarManager.cs
bschnurr: maybe HandleAllExceptions could do this

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,97 @@
...
+        private void DoNotShowAgainAction() {
+            LogEvent(PythonVersionNotSupportedInfoBarAction.Ignore);
+            Close();
+
+            Site.GetPythonToolsService().GeneralOptions.PromptForPythonVersionNotSupported = false;
+            Site.GetPythonToolsService().GeneralOptions.Save();
+        }
+
+        private void LogEvent(string action) {
+            var logInterpreter = _interpreterTriggeredInfoBar ?? ActiveInterpreter;
```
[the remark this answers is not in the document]
RaymonGulati1: Inside CheckAsync(), the logic was to log the event, create info bar, and then set "_interpreterTriggeredInfoBar". I've reorded this to set the "_interpreterTriggeredInfoBar" and then log the event which fixes it.

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        private IPythonInterpreterFactory ActiveInterpreter { get { return _getActiveInterpreterFunc(); } }
+
+        public override Task CheckAsync() {
+            ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            if (IsCreated ||
+                !Site.GetPythonToolsService().GeneralOptions.PromptForPythonVersionNotSupported ||
+                _interpreterTriggeredInfoBar != null ||
+                ActiveInterpreter == null ||
```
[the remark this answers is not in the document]
RaymonGulati1: Are 2 "if" statements required? If the activeInterpreter is null, the next condition won't be evaluted. 

Removed the ActiveInterpreter property

Python/Product/PythonTools/PythonTools/WorkspaceInfoBarManager.cs
```diff
@@ -128,10 +137,19 @@ class WorkspaceInfoBarManager : IVsRunningDocTableEvents, IDisposable {
...
+        }
+
+        private async Task TriggerPythonNotSupportedInforBarAsync() {
+            await _pythonVersionNotSupportedInfoBar.CheckAsync();
```
huguesv: async and await are unnecessary overhead here, just `return _pythonVersionNotSupportedInfoBar.CheckAsync();`

Python/Product/PythonTools/PythonTools/Project/PythonNotSupportedInfoBar.cs
```diff
@@ -0,0 +1,92 @@
...
+        private readonly string _context;
+
+        public PythonNotSupportedInfoBar(IServiceProvider site, string context, Func<IPythonInterpreterFactory> getActiveInterpreterFunc) : base(site) {
+            _getActiveInterpreterFunc = getActiveInterpreterFunc ?? throw new ArgumentNullException(nameof(getActiveInterpreterFunc));
+            _context = context ?? throw new ArgumentNullException(nameof(context));
+        }
+
+        public async override Task CheckAsync() {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+            var activeInterpreter = _getActiveInterpreterFunc();
```
huguesv: I was suggesting 2 ifs so that you don't unconditionally invoke the callback if it's already been created, or the global option is off. If that doesn't bother you, then this is fine.

PR #5812, opened 2019-10-10
Fix new test failures

### LINKED ISSUE #5811: Test failures - null reference exception from CondaEnvironmentFactoryProvider
In constructor, `ThreadHelper` is `null` because it's running outside of VS.

Tests:
WatchWorkspaceVirtualEnvRenamed
InstalledFactories
AddUpdateRemoveConfigurableFactory
AddUpdateRemoveConfigurableFactoryThroughUI
ChangeDefault

huguesv:
Fix #5811 

Replace ThreadHelper usage with a MEF import to avoid failure when running non UI tests.


huguesv:
`WatchWorkspaceVirtualEnvRenamed` failed for a different reason than before my fix "Assert.IsTrue failed. Failed to trigger discovery." but it passes on my machine.

PR #5867, opened 2019-11-08
fix button alignment, 2 items were slightly to the right

bschnurr:
Explore interactive scripts
Use Ipython interace mode

Before:
image

After:
image


huguesv:
I believe those were intentional by the original author, because those 2 options apply to the interactive window. I agree that they look wrong, and should have been a bit more to the right to make that point more obvious. But if you prefer them not indented, I'm okay with that too.

PR #5909, opened 2019-11-23
LSC: ArgumentOutOfRange exception while formatting document  Fix #5879

### LINKED ISSUE #5879: LSC: ArgumentOutOfRange exception while formatting document
See test `LspDiffTextEditFactoryTests.AddNewLineAtEndOfDoc`

Create a document that doesn't end with a newline like this one and format it with `black`.
```
1
2
3
four_hundred=4* 100
5

seven=[1,2, 3]

nine = ( 1,
2,
3,
4)
end = 13```

```

Error:
```
---------------------------
---------------------------
Specified argument was out of the range of valid values.

Parameter name: lineNumber
---------------------------
OK   
---------------------------
```

Also affected test: `FormattingUITests.FormatBlack`

bschnurr:
Fix #5879

- new idea is to append changes to previous line.  This gets around the issue of outofRange exceptions when trying to add a new line to a document.

- this also solves the existing issue in VsCode's code, where additional blank lines were being added to end of file, due to adding newline characters to every diff before processing edits

### DIFF
--- a/Python/Tests/Core.UI/FormattingUITests.cs
+++ b/Python/Tests/Core.UI/FormattingUITests.cs
@@ -68,7 +68,7 @@ def g ():    return 5 ** 2
 
 def g():
     return 5 ** 2
-", new[] { Span.FromBounds(0, 83), Span.FromBounds(93, 214) });
+", new[] { Span.FromBounds(0, 81), Span.FromBounds(91, 182), Span.FromBounds(93, 219) });
         }
 
         public void FormatYapf(PythonVisualStudioApp app) {
--- a/Python/Tests/Core.UI/FormattingUITests.cs
+++ b/Python/Tests/Core.UI/FormattingUITests.cs
@@ -66,7 +66,8 @@ def g ():    return 5 ** 2
     pass
 
 
-def g(): return 5 ** 2
+def g():
+    return 5 ** 2
 ", new[] { Span.FromBounds(0, 83), Span.FromBounds(93, 214) });
         }
 
--- a/Python/Tests/Core/LspDiffTextEditFactoryTests.cs
+++ b/Python/Tests/Core/LspDiffTextEditFactoryTests.cs
@@ -349,7 +349,6 @@ bottle.abort()
         }
 
         private void VerifyTextEdits(string text, string diff, TextEdit[] expected) {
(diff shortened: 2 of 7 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/LspEditorUtilities.cs
```diff
@@ -22,10 +22,12 @@
 namespace Microsoft.PythonTools.LanguageServerClient {
     internal static class LspEditorUtilities {
         internal static SnapshotPoint GetSnapshotPositionFromProtocolPosition(this ITextSnapshot textSnapshot, Position position) {
-            var line = textSnapshot.GetLineNumberFromPosition(position.Line);
```
bschnurr: deleted unused ```line```

Python/Product/PythonTools/PythonTools/LanguageServerClient/LspEditorUtilities.cs
```diff
@@ -22,10 +22,12 @@
...
-            var snapshotPosition = textSnapshot.GetLineFromLineNumber(position.Line).Start + position.Character;
+            if (position.Line < textSnapshot.LineCount) {
```
bschnurr: new code doesn't try to get the snapshot position if the line is out side of the snapshot.

now it will return the end position in this case

Python/Tests/Core.UI/FormattingUITests.cs
```diff
@@ -66,7 +66,8 @@ public class FormattingUITests {
     pass
 
 
-def g(): return 5 ** 2
+def g():
+    return 5 ** 2
```
bschnurr: i was going to add another fix for black formatting but instead just updating the test case expected result

Python/Tests/Core/LspDiffTextEditFactoryTests.cs
```diff
@@ -314,39 +312,39 @@ import bottle
...
-            //";
+            -nine = ( 1,
+            -2,
+            -3,
+            -4)
+            +nine = (1, 2, 3, 4)
+             end = 13
+            +
+            ",
+             new TextEdit[] { });
```
bschnurr: will update results.

PR #5918, opened 2019-11-28
LSC: Interactive window completions use incorrect expression to calcu…

### LINKED ISSUE #5874: LSC: Interactive window completions use incorrect expression to calculate live completions
Right now the code isn't able to determine the expression to evaluate in the python session in order to get completions. It's 'hard coded' to take the whole contents of the text buffer, minus the '.'.

Calculating the expression at caret position used to be done in the old analyzer when returning analysis completions, we'd get the current expression back in the results (which we'd then pass on to the python session).

We need to either implement the expression calculation in PTVS or have the language server provide it for us.

The code to fix is in `ReplRemoteCompletionBroker.GetExpressionAtPoint`.

bschnurr:
…late live completions fix #5874

- Bring over ExpressionFind and FindExpressionOptions from LS project

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Repl/Completion/ReplRemoteCompletionBroker.cs
+++ b/Python/Product/PythonTools/PythonTools/Repl/Completion/ReplRemoteCompletionBroker.cs
@@ -126,7 +126,7 @@ namespace Microsoft.PythonTools.Repl.Completion {
         ) {
             await TaskScheduler.Default;
 
-            string expression = GetExpressionAtPoint(triggerPoint);
+            string expression = GetExpressionAtPoint(triggerPoint, eval.LanguageVersion);
             if (!string.IsNullOrEmpty(expression)) {
                 var members = await eval.GetMemberNamesAsync(expression, token);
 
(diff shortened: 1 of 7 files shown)


Python/Product/PythonTools/PythonTools/Repl/Completion/ReplRemoteCompletionBroker.cs
```diff
@@ -146,11 +148,15 @@ CancellationToken token
             // this code here is just good enough to test simple cases
             // in old codebase, we used to get the expression from the analyzer
             var expression = triggerPoint.Snapshot.GetText();
-            if (expression.EndsWith(".")) {
-                expression = expression.Substring(0, expression.Length - 1);
+
+            var finder = new ExpressionFinder(expression, PythonLanguageVersion.V37, FindExpressionOptions.Complete);
```
huguesv: Language version should not be hard coded. There are couple of ways you could obtain it:

1. Cast `eval` in the calling function `as` a `PythonCommonInteractiveEvaluator`, and look at `LanguageVersion` property on that. Cast will be null for debug repl though, might be best to add the property to `IPythonInteractiveIntellisense` then we can make it accurate for all and no need to cast.

2. Use the factory available on language client: `client.Factory.Configuration.Version.LanguageVersion`

Python/Product/PythonTools/PythonTools/Repl/Completion/ReplRemoteCompletionBroker.cs
```diff
@@ -146,11 +148,15 @@ CancellationToken token
...
+            var Node = finder.GetExpression(triggerPoint.Position);
```
huguesv: camel case `node` + space after `if`

Python/Product/PythonTools/PythonTools/Repl/Completion/ReplRemoteCompletionBroker.cs
```diff
@@ -146,11 +148,15 @@ CancellationToken token
             // this code here is just good enough to test simple cases
             // in old codebase, we used to get the expression from the analyzer
             var expression = triggerPoint.Snapshot.GetText();
-            if (expression.EndsWith(".")) {
-                expression = expression.Substring(0, expression.Length - 1);
+
```
huguesv: remove comment and probably rename `expression` for `code` to avoid potential confusion, ie it may not be an expression.

bschnurr: good point

PR #5929, opened 2019-12-06
Allow use of 2 different language servers

huguesv:
Moved LS specific code out of the client and into new server class.

Configure which one is used via env var to allow testing either server. Defaults to the .NET Core server that is currently included in extension.

Configure the location of the node based server via env var since it is not included in the extension.


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageServerNodejs.cs
```diff
@@ -0,0 +1,241 @@
...
+                return true;
+            }
+
+            return version >= PythonLanguageVersion.V30 || version == PythonLanguageVersion.None;
+        }
+
+        public async override Task<Connection> ActivateAsync() {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            var nodePath = GetNodeExecutablePath();
```
bschnurr: ```nodePath ``` can be null

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageServerNodejs.cs
```diff
@@ -0,0 +1,241 @@
...
+        private string GetNodeExecutablePath() {
+            string filePath;
+
+            // Prefer the node executable from the server folder, if there is one
+            var serverFolderPath = GetServerLocation();
+            if (Directory.Exists(serverFolderPath)) {
+                filePath = Path.Combine(serverFolderPath, "node.exe");
+                if (File.Exists(filePath)) {
+                    return filePath;
+                }
```
bschnurr: do we care about logging errors?

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageServerDotNetCore.cs
```diff
@@ -0,0 +1,220 @@
...
+            };
+
+            if (process.Start()) {
+                return new Connection(process.StandardOutput.BaseStream, process.StandardInput.BaseStream);
+            }
+
+            return null;
+        }
+
+        public override object CreateInitializationOptions(string interpreterPath, string interpreterVersion, string rootPath, IEnumerable<string> searchPaths) {
```
bschnurr: should we add string checks?

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageServerNodejs.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+
+            return val != "0" && string.Compare(val, "false", StringComparison.OrdinalIgnoreCase) != 0;
+        }
+
+        private string GetNodeExecutablePath() {
+            string filePath;
+
+            // Prefer the node executable from the server folder, if there is one
+            var serverFolderPath = GetServerLocation();
```
bschnurr: ```serverFolderPath``` can be null

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageServerNodejs.cs
huguesv: right now they'll get a message box if either node or the server is missing.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageServerNodejs.cs
huguesv: null is accepted by File.Exists

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageServerNodejs.cs
huguesv: Directory.Exists accepts null

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageServerDotNetCore.cs
huguesv: searchPaths was the only that could cause a crash, I've added a check for null.

PR #5933, opened 2019-12-11
LiveShare fix

### LINKED ISSUE #5887: Handle LiveShare API breaking change for 16.5
In 16.5, we need to either 

1) switch to using multiple `ILspRequestHandler` handlers
2) make our one ILanguageServiceProviderCallback be able to accept JToken and return JToken (this will require LiveShare to make changes for this)

The problem with our existing ILanguageServiceProviderCallback is that LiveShare is passing in LSP assembly types to us, and thus require us to reference the same LSP assembly version as LiveShare. They want to remove that coupling so they can move to newer assembly versions without affecting us.

Edit: after some discussion, we'll go with 2) and wait for the LiveShare team to make the changes (probably week of the Nov 25th) and quickly fix the breaking change soon after.

huguesv:
Fix #5887 

Do not directly cast to TIn or TOut in implementation of ILanguageServiceProviderCallback, since we no longer have a guarantee that the types passed in will match the types in the version of Microsoft.VisualStudio.LanguageServer.Protocol that we reference. We now convert between the LiveShare types and ours by going through the JSON format, which is stable and backwards compatible.

### DIFF
--- a/Python/Product/LiveShare/PythonLanguageServiceProviderCallback.cs
+++ b/Python/Product/LiveShare/PythonLanguageServiceProviderCallback.cs
@@ -27,6 +27,8 @@ using Microsoft.VisualStudio.Threading;
 using Microsoft.VisualStudioTools;
 using LSP = Microsoft.VisualStudio.LanguageServer.Protocol;
 using LS = Microsoft.VisualStudio.LiveShare.LanguageServices;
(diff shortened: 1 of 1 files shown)


huguesv:
I've relaunched the build that failed and it passed the second time, though it doesn't show it here.

PR #5943, opened 2019-12-20
Add LiveShare support

### LINKED ISSUE #5847: LSC: LiveShare support
Need to investigate how to reimplement this.

I was under the impression it was going to be mostly automatic because we use LSC, but maybe there is some work needed. Since we are manually activating LSC and don't MEF export it, that is likely preventing things from working automatically.

If you have PTVS installed in the LiveShare client, an instance of LSC gets activated for the temporary LiveShare files in temp folder, using an interpreter on the LiveShare client's machine. That's of course all wrong and must be prevented.

If you don't have PTVS installed on client, then IntelliSense, go to definition, etc. don't work, so it must not be able to connect to the right LSC instance.

huguesv:
Fix #5847 

Similar LiveShare callback implementation as before, except now it delegates to `PythonLanguageClient` instead of `VsProjectAnalyzer`.

Updated the LiveShare unit tests, which run outside of VS.

That meant allowing `PythonLanguageClient` to run outside of VS, which required some refactoring. to remove a bunch of direct dependencies. Now it has fewer responsibilities and the specific support for workspace, project, global etc are separated out and `PythonLanguageClient` just deals with generic configuration and update events. So the code should be more easily understandable now.


Python/Product/LiveShare/PythonLanguageServiceProviderCallback.cs
```diff
@@ -55,89 +53,117 @@ internal class PythonLanguageServiceProviderCallback : ILanguageServiceProviderC
...
         }
 
         /// <summary>
         /// Helper function for tests, enabling this class to be tested without needing
         /// a service provider or UI thread.
         /// </summary>
-        internal void SetAnalyzer(Uri documentUri, VsProjectAnalyzer analyzer) {
-            _analyzerCache[documentUri] = analyzer;
+        internal void SetClient(Uri documentUri, PythonLanguageClient client) {
+            _clientCache[documentUri] = client;
```
bschnurr: Consider adding a check for null client here.  I think SetClient can be called with results from FindLanguageClient with might be null

Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
```diff
@@ -159,10 +159,10 @@ class SendToReplCommand : Command {
 
                     var activeView = CommonPackage.GetActiveTextView(_serviceProvider);
 
-                    IPythonInterpreterFactory factory;
-                    if (activeView != null && (factory = activeView.GetInterpreterFactoryAtCaret(_serviceProvider)) != null) {
+                    InterpreterConfiguration config;
+                    if (activeView != null && (config = activeView.GetInterpreterConfigurationAtCaret(_serviceProvider)) != null) {
```
bschnurr: is this better?
 if  (config = activeView?.GetInterpreterConfigurationAtCaret(_serviceProvider)) != null)

Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -272,39 +272,15 @@ out guid
...
+            if (client?.Configuration != null) {
+                return client.Configuration;
             }
 
-            return serviceProvider.GetPythonToolsService().InterpreterOptionsService.DefaultInterpreter;
-        }
-
-        internal static InterpreterConfiguration GetInterpreterConfigurationAtCaret(this ITextView textView, IServiceProvider serviceProvider) {
-            return textView.GetInterpreterFactoryAtCaret(serviceProvider)?.Configuration;
+            return serviceProvider.GetPythonToolsService().InterpreterOptionsService.DefaultInterpreter?.Configuration;
```
bschnurr: Looks like GetInterpreterConfigurationAtCaret is called from an eventhandler..

should we add a switch to main thread here?

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientContextFixed.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.Python.Parsing;
+using Microsoft.PythonTools.Common.Infrastructure;
+using Microsoft.PythonTools.Interpreter;
+using Microsoft.PythonTools.Project;
+using Microsoft.PythonTools.Repl;
+using Microsoft.VisualStudio.InteractiveWindow;
+
+namespace Microsoft.PythonTools.LanguageServerClient {
+    internal class PythonLanguageClientContextFixed : IPythonLanguageClientContext {
+#pragma warning disable CS0067
```
bschnurr: does this trick of adding empty add/remove help with the removing the warnings instead of disabling them? https://stackoverflow.com/questions/49994869/suppressing-is-never-used-warning-for-events-in-c-sharp

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientContextWorkspace.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public event EventHandler InterpreterChanged;
+        public event EventHandler SearchPathsChanged;
+        public event EventHandler Closed;
+
+        public PythonLanguageClientContextWorkspace(
+            IPythonWorkspaceContext pythonWorkspace,
+            string contentTypeName
+        ) {
+            _pythonWorkspace = pythonWorkspace ?? throw new ArgumentNullException(nameof(pythonWorkspace));
```
bschnurr: Its ok to hold on to a reference to the workspace? what happens if the workspace changes? is it the same object just the internals change?

Python/Product/LiveShare/PythonLanguageServiceProviderCallback.cs
huguesv: It's technically valid, all the code in this class can handle a null entry in the dictionary. We could also want to add tests later that specifically test that behavior when it's null, so I think not throwing is fine.

Python/Product/PythonTools/PythonTools/Commands/SendToReplCommand.cs
huguesv: we can do that, i'll change it

Python/Product/PythonTools/PythonTools/Extensions.cs
huguesv: Both callers of this method are from command handlers, so they are on the UI thread.

The call to `FindLanguageClient` is thread safe.

`GetPythonToolsService` throws in debug builds if called from anything but the UI thread.

Threading rules say that if this was an async method, we should do the switch to main thread. In a sync method, recommendation is to throw if you need to be on main thread but aren't on it, which `GetPythonToolsService` does, but it may take a while to discover that issue if most of the calls return early in the if statement.

I guess I'll add a throw at the top.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientContextFixed.cs
huguesv: It does, thank you!

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientContextWorkspace.cs
huguesv: When the workspace changes, this object will get disposed. This is kicked off by the `_pythonWorkspace.AddActionOnClose` that is a few lines below. The language client registers to this object's `Closed` event, it will terminate the language client and dispose of it, and the dispose of language client disposes of the context (this object).

huguesv: After that's all done, the events for the new workspace come in, where a new `PythonLanguageClientContextWorkspace` gets created.

PR #5948, opened 2020-01-06
[WIP] Updating VS to use PTVSD 5.*

RaymonGulati1:
WIP because I'm working on fixing one of the tests. Other than that, it's ready for code review.


Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -0,0 +1,213 @@
...
+            if (options.Length - lastStart > 0) {
+                res.Add(options.Substring(lastStart, options.Length - lastStart));
+            }
+            return res.ToArray();
+        }
+
+        #endregion
+
+        #region Attach
+        public static DebugAttachInfo GetTcpAttachDebugInfo(IAdapterLaunchInfo adapterLaunchInfo) {
```
huguesv: can be private

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -0,0 +1,213 @@
...
+
+            SetInterpreterOptions(debugLaunchInfo, adapterLaunchInfoJson);
+            SetProgramInfo(debugLaunchInfo, adapterLaunchInfoJson);
+            SetEnvVariables(debugLaunchInfo, adapterLaunchInfoJson);
+            SetLaunchDebugOptions(debugLaunchInfo, adapterLaunchInfoJson);
+
+            return debugLaunchInfo;
+
+        }
+        private static void SetInterpreterOptions(DebugLaunchInfo debugLaunchInfo, JObject adapterLaunchInfoJson) {
```
huguesv: add new line between methods

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs
```diff
@@ -0,0 +1,105 @@
...
+        public abstract string GetJsonString();
+    }
+
+    [JsonObject(MemberSerialization.OptIn)]
+    public class DebugLaunchInfo : DebugInfo {
+        [JsonProperty("cwd")]
+        public string Cwd { get; set; }
+
+        [JsonProperty("python")]
+        public List<string> Python { get; set; } //Interpreter path and arguments
```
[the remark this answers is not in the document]
huguesv: You could name the property `InterpreterPathAndArguments`, it's a bit long but more understandable than `Python` and then you don't need a comment.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs
```diff
@@ -0,0 +1,105 @@
...
+    public class DebugAttachInfo : DebugInfo {
+        [JsonProperty("host")]
+        public string Host { get; set; }
+
+        [JsonProperty("port")]
+        public int Port { get; set; }
+
+        public Uri RemoteUri { get; set; }
+
+        public override string GetJsonString() {
```
huguesv: implementation is the same in both derived classes, you might was well move it to the base class, or at the caller, since there is only one caller anyway.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -0,0 +1,213 @@
...
+            if (String.IsNullOrEmpty(command)) {
+                yield break;
+            }
+
+            IntPtr argPointer = CommandLineToArgvW(command, out int argumentCount);
+            string[] arguments = new string[argumentCount];
+
+            if (argPointer != IntPtr.Zero) {
+                try {
+                    for (int i = 0; i < arguments.Length; i++) {
```
huguesv: change `arguments.Length` to `argumentCount` and get rid of the otherwise unused `arguments`

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.PythonTools.Debugger {
+
+    [JsonObject(MemberSerialization.OptIn)]
+    public abstract class DebugInfo {
```
huguesv: you should be able to make these classes internal

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterProcess.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace Microsoft.PythonTools.Debugger {
+    sealed class DebugAdapterProcess {
+        private readonly ITargetHostInterop _targetInterop;
+        private readonly string _debuggerAdapterDirectory;
+
+        private ITargetHostProcess _targetHostProcess;
+        private string _webBrowserUrl;
+
+        public DebugAdapterProcess(ITargetHostInterop targetInterop, string debuggerAdapterDirectory) {
+            _targetInterop = targetInterop;
```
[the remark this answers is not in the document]
huguesv: this class is not very useful anymore, it's only constructed to call `StartProcess`. You could merge `StartProcess` into its caller and `LaunchBrowserDebugger` moved there too.
I don't know what future changes, if any, are planned

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -0,0 +1,213 @@
...
+                if (parsedOption.Length == 2) {
+                    debugLaunchInfo.DebugDjango = parsedOption[1].ToLower().Trim().Equals("true");
+                }
+            }
+
+            string webPageUrlOption = options.FirstOrDefault(x => x.StartsWith("WEB_BROWSER_URL"));
+            if (webPageUrlOption != null) {
+                string[] parsedOption = webPageUrlOption.Split('=');
+                if (parsedOption.Length == 2) {
+                    debugLaunchInfo.WebPageUrl = WebUtility.UrlDecode(parsedOption[1].ToLower().Trim());
```
huguesv: why lower the url? the user will notice this in the browser. is it new or was that done previously?

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -0,0 +1,213 @@
...
+            launchJson.WaitOnNormalExit = debugService.WaitOnNormalExit;
+            launchJson.BreakOnSystemExitZero = debugService.BreakOnSystemExitZero;
+            launchJson.DebugStdLib = debugService.DebugStdLib;
+            launchJson.ShowReturnValue = debugService.ShowFunctionReturnValue;
+        }
+
+        [DllImport("shell32.dll")]
+        static extern IntPtr CommandLineToArgvW([MarshalAs(UnmanagedType.LPWStr)] string lpCmdLine, out int pNumArgs);
+
+        private static IEnumerable<string> GetParsedCommandLineArguments(string command) {
```
int19h: This method doesn't throw if the command line couldn't be parsed - so the app will be silently invoked with no arguments at all, which will be very confusing to the user. That should be reported as an error instead.

Python/Product/TestAdapter/testlauncher.py
```diff
@@ -33,7 +33,7 @@ def load_debugger(secret, port, debugger_search_path, mixed_mode):
     # Load the debugger package
     try:
         if debugger_search_path:
-            sys.path.append(debugger_search_path)
+            sys.path.insert(1, debugger_search_path)
```
int19h: Is there any particular reason why this is inserted as the 2nd item in the search path, rather than the 1st (index 0)?

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
```diff
@@ -247,7 +247,7 @@ def main():
     
     sys.path[0] = os.getcwd()
     if opts.debugger_search_path:
-        sys.path.append(opts.debugger_search_path)
+        sys.path.insert(1, opts.debugger_search_path)
```
int19h: Same here.

huguesv:
There's an issue with bottle and flask web projects. Create new project, install packages as needed and press F5, the browser starts and attempts to load the page but it never ends. By comparison, legacy debugger works fine, also start without debugging is fine.

huguesv:
Stop debugging command doesn't work. I tried django app and regular python console app project. You'll see this after a few seconds, which is not supposed to happen:
image

If you click stop now, it will stop but you'll have leftover Python processes running.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System;
```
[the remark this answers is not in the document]
RaymonGulati1: Changed. Will be updated in next commit.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterProcess.cs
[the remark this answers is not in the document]
RaymonGulati1: 1. Will be updated in next commit. 

2. I prefer to keep it in a separate class because it's more organized and allows for cleaner integration of the upcoming changes (monitoring messages/events).

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs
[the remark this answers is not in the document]
RaymonGulati1: Will be updated in next commit

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        [JsonProperty("python")]
+        public List<string> Python { get; set; } //Interpreter path and arguments
+        //public string Python { get; set; }
+
+        [JsonProperty("console")]
+        public string Console { get; set; }
+
+        [JsonProperty("program")]
+        //public string Program { get; set; }
```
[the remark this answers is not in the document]
RaymonGulati1: Will be updated in next commit

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
RaymonGulati1: Will be updated in next commit

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
RaymonGulati1: Will be updated in next commit

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
RaymonGulati1: Will be updated in next commit

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs
RaymonGulati1: Will be updated in next commit

Python/Product/TestAdapter/testlauncher.py
RaymonGulati1: It's convention that sys.path[0] is the directory from which the script was invoked. 

Stackoverflow:
image


https://stackoverflow.com/questions/10095037/why-use-sys-path-appendpath-instead-of-sys-path-insert1-path/10095099#10095099

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
RaymonGulati1: Was not done previously--has been reverted.

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
RaymonGulati1: Same as above

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs
RaymonGulati1: Changed

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterProcess.cs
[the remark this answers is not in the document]
huguesv: If you're keeping it separate, then I would at least change debuggerAdapterDirectory to not be quoted when passed in. It is not something you would expect of a variable/field that holds a directory path. Instead I would quote it at the time you pass it to ExecuteCommandAsync, that's the thing that requires the quoted value.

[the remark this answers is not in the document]
huguesv: You're using `FirstOrDefault()` for passing in the python exe, which means it could be `null`. I wonder how `ExecuteCommandAsync` will behave in that case.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
RaymonGulati1: Will be updated in next commit

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterProcess.cs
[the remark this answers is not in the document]
RaymonGulati1: 1. Changed. 
2. How about putting a null/empty string check in place? If it is null/empty, show error box and return the ITargetHostProcess as null.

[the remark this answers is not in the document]
huguesv: Failed to launch debug adapter... sounds good enough

Python/Product/TestAdapter/testlauncher.py
int19h: I didn't spot this before, but this code should be removing the entry after importing all the top-level packages that it needs - there's no reason for it to be around there afterwards (if user re-imports explicitly, it's already in `sys.modules`).

The problem with leaving `path[0]` alone is that it may still interfere, e.g. if they have a local copy of ptvsd around.

RaymonGulati1:
> Stop debugging command doesn't work. I tried django app and regular python console app project. You'll see this after a few seconds, which is not supposed to happen:

Bug in PTVSD. Issue created [here](https://github.com/microsoft/ptvsd/issues/2026).

Python/Product/TestAdapter/testlauncher.py
RaymonGulati1: Will be updated in next commit

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
RaymonGulati1: Will be updated in next commit

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -185,29 +202,26 @@ public sealed class DebugAdapterLauncher : IAdapterLauncher {
...
-            string[] arguments = new string[argumentCount];
-
-            if (argPointer != IntPtr.Zero) {
-                try {
-                    for (int i = 0; i < arguments.Length; i++) {
-                        yield return Marshal.PtrToStringUni(Marshal.ReadIntPtr(argPointer, i * IntPtr.Size));
-                    }
+            IntPtr argPointer = CommandLineToArgvW(command, out var argumentCount);
+            if (argPointer == IntPtr.Zero) {
+                throw new System.ComponentModel.Win32Exception();
```
int19h: If the intent is to capture the `GetLastError()` code from the failed call, then the P/Invoke declaration [needs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getlastwin32error?view=netframework-4.8#remarks) `[DllImport(..., SetLastError=true)]`.

RaymonGulati1: Don't think VS needs to display the exact error to the user--just state the fact that VS can't parse. Though, the error code could be useful for debugging.

int19h: I was looking for `catch (Win32Exception)`, and didn't find it - now I see that it catches `Exception`.

The parameterless `Win32Exception()` constructor will try to retrieve the error code, and then use `FormatMessage` to turn it into a string. So it will be whatever was the last P/Invoke with `SetLastError=true`. If it's not the one above, it will be something else (quite possibly, something called from within .NET stdlib). If it's not actually displayed, it's not a big deal, but it may be confusing when debugging, because you'll see the exception happen - it'll get logged in Output, for example - but its message will be unrelated to the actual error.

RaymonGulati1: Good point. Will be updated in next commit.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterProcess.cs
```diff
@@ -0,0 +1,74 @@
...
+            _targetHostProcess = _targetInterop.ExecuteCommandAsync(pythonExePath, "\"" + _debuggerAdapterDirectory + "\"");
+
+            if (!string.IsNullOrEmpty(webBrowserUrl) && Uri.TryCreate(webBrowserUrl, UriKind.RelativeOrAbsolute, out var uri)) {
+                OnPortOpenedHandler.CreateHandler(uri.Port, null, null, () => _targetHostProcess.HasExited, LaunchBrowserDebugger);
+            }
+
+            return _targetHostProcess;
+        }
+
+        private void LaunchBrowserDebugger() {
```
karthiknadig: >LaunchBrowserDebugger [](start = 21, length = 21)

This is currently done using `OnPortOpenedHandler`, once we have a way to listen to the messages this can be done entirely using that. VS Code uses pattern matching to extract the host+port from the output event.

RaymonGulati1: I'll keep it in mind

Issue #5954, opened 2020-01-08
Unable to debug flask and bottle projects with PTVSD 5.*


RaymonGulati1:
There's an issue with bottle and flask web projects. Create new project, install packages as needed and press F5, the browser starts and attempts to load the page but it never ends. By comparison, legacy debugger works fine, also start without debugging is fine.

Related to #5948

huguesv:
I've found a workaround.

Clear the Launch URL text box under project properties debug tab. This will prevent the browser from opening automatically.
Set breakpoint in your route and start debugging.
Manually open browser and navigate to the server URL (which you can see printed in the console)
Request should succeed and breakpoint in the route handler gets hit.

Hypothesis:

It appears to be that the browser is opened too early and the request that is sent to the server may be incomplete/partial and that puts the server in a state that no longer responds to requests (refreshing or sending more requests make no difference).

This used to happen periodically in the past with older versions of ptvsd, more frequently with Edge or Internet Explorer. Maybe those browsers just started faster than Chrome, causing the issue more easily.

Maybe now that we're starting up the adapter python process then starting the app process means that the server is ready to accept requests later than it used to be in the past, causing all browsers to be sending a request too early.

If we switch to watching the output for the URL and invoke the browser only at that time, that should fix the issue. We'll need to decide what to do with regards to the Launch URL in project debug properties though, as that will effectively become deprecated feature.

karthiknadig:
This should be addressed (or at least be better) when you switch to using `output` event to launch the browser rather than pinging the port that we do now.

PR #5960, opened 2020-01-09
Delay launching web browser for debugging web projects

RaymonGulati1:
#5954

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterProcess.cs
+++ b/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterProcess.cs
@@ -16,11 +16,11 @@
 
 using System;
 using System.Runtime.InteropServices;
-using System.Timers;
+using System.Threading;
 using System.Windows.Forms;
 using Microsoft.VisualStudio.Debugger.DebugAdapterHost.Interfaces;
 using Microsoft.VisualStudio.Shell.Interop;
-using Timer = System.Timers.Timer;
+using Timer = System.Threading.Timer;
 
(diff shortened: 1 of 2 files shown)


Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterProcess.cs
```diff
@@ -43,13 +44,21 @@ internal sealed class DebugAdapterProcess {
             _webBrowserUrl = webBrowserUrl;
             _targetHostProcess = _targetInterop.ExecuteCommandAsync(pythonExePath, "\"" + _debuggerAdapterDirectory + "\"");
 
-            if (!string.IsNullOrEmpty(webBrowserUrl) && Uri.TryCreate(webBrowserUrl, UriKind.RelativeOrAbsolute, out var uri)) {
-                OnPortOpenedHandler.CreateHandler(uri.Port, null, null, () => _targetHostProcess.HasExited, LaunchBrowserDebugger);
+            if (!string.IsNullOrEmpty(webBrowserUrl) && Uri.TryCreate(webBrowserUrl, UriKind.RelativeOrAbsolute, out _)) {
+                Timer timer = new Timer(5000);
```
huguesv: Timer needs to be `Dispose()`. I suppose you could do that in the elapsed handler.

Not sure why you decided to go with this Timer class rather than the one in System.Threading that we use everywhere... Regardless all Timer classes require disposing.

RaymonGulati1: Changed

Issue #5931, opened 2019-12-06
LSC: Workspace interpreter is wrong for first workspace loaded after VS started


huguesv:
This seems to happen the first time you open a workspace after starting VS.

Open a workspace that already has a PythonSettings.json that points to a virtual env under the workspace, ex:
```
{
  "Interpreter": "env1\\Scripts\\python.exe"
}
```

Sometimes it will ignore that setting and just give you another global environment instead. You can see that by looking at the selected environment on Python toolbar. None of the workspace virtual envs are listed.

If you close the workspace, and then reload (without restarting VS), then it is correct.

Note: I cannot reproduce this in master branch

PR #5969, opened 2020-01-14
Fix workspace interpreter on first load

huguesv:
Fix intialization order of Python workspace object in relation to workspace interpreter factory provider. This fixes the workspace interpreter being incorrect for the first workspace that is loaded after VS has started. #5931

### DIFF
--- a/Python/Product/VSInterpreters/Interpreter/PythonWorkspaceContext.cs
+++ b/Python/Product/VSInterpreters/Interpreter/PythonWorkspaceContext.cs
@@ -50,7 +50,6 @@ namespace Microsoft.PythonTools.Interpreter {
         private string _unitTestRootDirectory;
         private string _unitTestPattern;
 
-
         // These are set in initialize
         private IPythonInterpreterFactory _factory;
         private bool? _factoryIsDefault;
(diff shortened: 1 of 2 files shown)


Python/Product/VSInterpreters/Interpreter/PythonWorkspaceContext.cs
```diff
@@ -63,6 +62,19 @@ class PythonWorkspaceContext : IPythonWorkspaceContext {
...
             _workspaceSettingsMgr = _workspace.GetSettingsManager();
+
+            // Initialization in 2 phases (Constructor + Initialize) is needed to
+            // break a circular dependency.
+            // We create a partially initialized object that can be used by
+            // WorkspaceInterpreterFactoryProvider to discover the interpreters
+            // in this workspace (it needs the interpreter setting to do that).
+            // Once that is done, the IPythonInterpreterFactory on this object
+            // can be resolved in Initialize.
+            _interpreter = ReadInterpreterSetting();
```
RaymonGulati1: You can put the field initialization and deceleration on the same line. Rather than "private string _interpreter;", use "private string _interpreter = ReadInterpreterSetting();"

Same for the others

bschnurr: looks like ReadInterpreterSetting() can throw ArgumentNullException if you follow its calls. should we add a catch?

bschnurr: i guess we check for null workspace above this so should be ok

huguesv: I cannot do that, as I would lose control of the initialization order. It's important for the other fields to be initialized before attempting to read the settings.

huguesv: exactly

PR #5974, opened 2020-01-15
[Accessibility] interactive window titles are now unique

bschnurr:
New interactive window titles are now unique
image

### DIFF
--- a/Python/Product/Common/Strings.resx
+++ b/Python/Product/Common/Strings.resx
@@ -1035,7 +1035,7 @@ Your web browser will not be opened.</value>
     <value>Unknown command '{0}', use '$help' for a list of commands.</value>
   </data>
   <data name="ReplCaption" xml:space="preserve">
-    <value>{0} Interactive:{1}</value>
+    <value>{0} Interactive {1}</value>
     <comment>{0} is the name of the Python environment, {1} is the window id</comment>
   </data>
   <data name="ReplCaptionNoEvaluator" xml:space="preserve">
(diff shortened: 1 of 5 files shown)


Python/Tests/Utilities.Python/PythonVisualStudioApp.cs
```diff
@@ -317,7 +317,7 @@ public PythonVisualStudioApp(IServiceProvider site)
 
         public ReplWindowProxy GetInteractiveWindow(string title, ReplWindowProxySettings settings = null) {
             var iwp = GetService<IComponentModel>(typeof(SComponentModel))?.GetService<InteractiveWindowProvider>();
-            var window = iwp?.AllOpenWindows.FirstOrDefault(w => ((ToolWindowPane)w).Caption == title);
```
huguesv: there's another one of these that you missed, in the same file

Python/Product/Common/Strings.resx
```diff
@@ -1035,8 +1035,8 @@ Your web browser will not be opened.</value>
     <value>Unknown command '{0}', use '$help' for a list of commands.</value>
   </data>
   <data name="ReplCaption" xml:space="preserve">
-    <value>{0} Interactive</value>
-    <comment>{0} is the name of the Python environment</comment>
```
huguesv: you said you like space too, should we change it?

bschnurr: yes. sorry missed that

Python/Tests/Utilities.Python/PythonVisualStudioApp.cs
bschnurr: thanks

PR #6003, opened 2020-02-05
5999 updated ptvsd to debugpy

### LINKED ISSUE #5999: Upgrade ptvsd to DebugPy

RaymonGulati1:
Fix #5999


Python/Product/TestAdapter/testlauncher.py
```diff
@@ -1,17 +1,32 @@
...
 import os
 import sys
 import traceback
 
 def main():
-    cwd, testRunner, secret, port, debugger_search_path, mixed_mode, coverage_file, test_file, args = parse_argv()
-    sys.path[0] = os.getcwd()
+    cwd, testRunner, secret, port, mixed_mode, coverage_file, test_file, args = parse_argv()
+
+    load_debugger(secret, port, mixed_mode)
```
huguesv: at this point in time it looks like you are assuming that the current working directory is debugpy's parent folder, so you can import debugpy without adding it to the path.

but is that guaranteed?

the code that launches the python process with testlauncher.py passes in the desired cwd as argument (which is parsed above), but the call to start process sets the working directory like this:
```
                using (var proc = ProcessOutput.Run(
                    _projectSettings.InterpreterPath,
                    arguments,
                    _projectSettings.WorkingDirectory,
                    env,
                    visible: false,
                    testRedirector,
                    quoteArgs: true,
                    elevate: false,
                    System.Text.Encoding.UTF8,
                    System.Text.Encoding.UTF8
                )) {
```

huguesv: oh does import debugpy work because THIS script happens to be deployed in the parent folder of debugpy?

Python/Product/PythonTools/PythonTools/Debugger/DebugPyVersionHelper.cs
```diff
@@ -25,81 +25,81 @@
...
-                Strings.InstalledPtvsdOutdatedMessage.FormatUI("unknown", PtvsdVersion.Version),
+                Strings.InstalledPtvsdOutdatedMessage.FormatUI("unknown", DebugPyVersion.Version),
                 allowDisable: false,
                 isError: false
             );
         }
 
-        public static void VerifyPtvsdVersionLegacy() {
-            ShowPtvsdMessage(
+        public static void ShowLegacyPtvsdVersionError() {
```
huguesv: This will confusing if this gets triggered.

You have an old version of ptvsd installed in the current environment (version {0}), but version {1} is included with Visual Studio. Please upgrade or uninstall ptvsd to get the latest functionality.

where {0} is 3.* and {1} is 1.0

Python/Product/PythonTools/PythonTools/Debugger/DebugPyVersionHelper.cs
```diff
@@ -25,81 +25,81 @@
...
             if (PackageVersion.TryParse(response.Debugger.Version, out PackageVersion runningVersion)) {
-                var bundledPtvsdVersion = PackageVersion.Parse(PtvsdVersion.Version);
-                if (runningVersion.CompareTo(bundledPtvsdVersion) < 0) {
-                    ShowPtvsdMessage(
+                var bundledDebugPyVersion = PackageVersion.Parse(DebugPyVersion.Version);
+                if (runningVersion.CompareTo(bundledDebugPyVersion) < 0) {
+                    ShowDebuggingErrorMessage(
                         Strings.InstalledPtvsdOutdatedTitle,
-                        Strings.InstalledPtvsdOutdatedMessage.FormatUI(response.Debugger.Version, PtvsdVersion.Version),
+                        Strings.InstalledPtvsdOutdatedMessage.FormatUI(response.Debugger.Version, DebugPyVersion.Version),
```
huguesv: this one and the next will need to say debugpy in the message rather than ptvsd.

Python/Product/PythonTools/PythonTools/Debugger/DebugPyVersionHelper.cs
huguesv: my guess is that it won't be possible to see it though, as we can't accidentally launch legacy ptvsd when we import debugpy.

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
```diff
@@ -236,19 +236,15 @@ def main():
...
     parser.add_option('-r', '--result-port', type='int', help='connect to port on localhost and send test results')
     parser.add_option('--test-list', metavar='<file>', type='str', help='read tests from this file')
     parser.add_option('--dry-run', action='store_true', help='prints a list of tests without executing them')
     (opts, _) = parser.parse_args()
-    
-    sys.path[0] = os.getcwd()
-    if opts.debugger_search_path:
-        sys.path.insert(0, opts.debugger_search_path)
 
+    sys.path.insert(1, os.getcwd())
```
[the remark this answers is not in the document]
huguesv: oh is it because at [0] it's the path to this script and if you overwrite that then you can't load debugpy relative to this script?

Python/Product/TestAdapter/testlauncher.py
RaymonGulati1: Yes, it is guaranteed. 

["As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter."](https://docs.python.org/3/library/sys.html#sys.path)

Python/Product/TestAdapter/visualstudio_py_testlauncher.py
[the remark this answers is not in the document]
RaymonGulati1: The value "sys.path[0]" is the directory of the python script that was started. The old logic was concluded because it was changing the starting script directory, then adding the debugger path, if required (in the following "if opts.debugger..." statement). The "os.getcwd()" refers to the directory of the user's python script which has simply been inserted in.

Issue #6058, opened 2020-03-28
Pytest standard output no longer captured


huguesv:
If you do any `print()` in your pytest test, it doesn't get captured in the test result.

This test detected this regression:

```
Test Name:	RunPytestSetupAndTeardown
Test Outcome:	Failed
Result Message:	
Test method TestAdapterTests.TestExecutorTests37.RunPytestSetupAndTeardown threw exception: 
System.InvalidOperationException: Sequence contains no matching element
Result StandardOutput:	
Creating temp directory for test at C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\1ze53mac.uaw\
Creating temp directory for test at C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\udbgyyiu.4ti\
"C:\Program Files\Python37\python.exe" -m venv C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\udbgyyiu.4ti\

Messages:
Informational:cd C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\1ze53mac.uaw\Source
Informational:set PYTHONPATH=C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\1ze53mac.uaw\Source
Informational:C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\udbgyyiu.4ti\scripts\python.exe C:\Projects\GitHub\huguesv\PTVS\BuildOutput\Debug16.0\raw\binaries\testlauncher.py C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\1ze53mac.uaw\Source pytest "" 0 "" "" dummyfilename .\test_teardown.py::TeardownTests::test_success .\test_teardown.py::TeardownTests::test_failure --junitxml=C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\1ze53mac.uaw\Results\junitresults_3698c606-cd40-4758-9ea1-36cabc1841b7.xml --rootdir=C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\1ze53mac.uaw\Source
Informational:============================= test session starts =============================
Informational:platform win32 -- Python 3.7.4, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
Informational:rootdir: C:\Users\huvalo\AppData\Local\Temp\s10vaaj3.njt\1ze53mac.uaw\Source
Informational:collected 2 items
[...]

PR #6078, opened 2020-04-07
New updates to pytest 5.4 removed output logging to junit files by de…

bschnurr:
…fault, turning it back on.

Removed warnings, new junit 2.0 format is coming so specify we want orginal format.

https://docs.pytest.org/en/latest/reference.html#confval-junit_logging

#6058


bschnurr:
old pytest still works

image

PR #6086, opened 2020-04-09
fix for python 3.8 from microsoft store failing to add new venv

### LINKED ISSUE #6082: Create virtual environment permanently fails when Base interpreter is Python 3.8
**Describe the bug**
After I open a Solution with `requirements.txt` in its dir Visual Studio detects it and shows a badge with the "Create virtual environment" link. I click it, change `env` to `venv` and click "Create":
image
After a few seconds process finishes with conflicting messages between "Background tasks" window and "Output" window:
image

**Steps to Reproduce**
1. Create some Python Solution.
2. Place `requirements.txt` into its dir so VS can detect it.
3. Save and reopen the Solution in VS.
4. Continue as described in "Describe the bug" section above.

**Expected behavior**
A successfully created Python Virtual Environment.

VS Version: 16.5.3
PTVS version: 16.5.20041.1-16.0
Python version: 3.8

bschnurr:
new versions of python >= 3.8 from the windows store have a different executable name (ie. python3.8.exe) in the registry InstallPath, compared to what is in the venv folder (python.exe) so search for both if interpExe is not found.

Fix #6082

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
+++ b/Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
@@ -408,10 +408,8 @@ namespace Microsoft.PythonTools.Environments {
 
             // new versions of python >= 3.8 from the windows store have a different executable name (ie. python3.8.exe) in the 
             // registry InstallPath, compared to what is in the venv folder (python.exe) so search for both if interpExe is not found.
-            interpExe = PathUtils.FindFile(prefixPath, interpExe, firstCheck: scripts);
-            interpExe = interpExe == null ? PathUtils.FindFile(prefixPath, "python.exe", firstCheck: scripts) : interpExe;
-            winterpExe = PathUtils.FindFile(prefixPath, winterpExe, firstCheck: scripts);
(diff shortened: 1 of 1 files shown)


bschnurr:
also tested previous python 3.7 still works.

Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
```diff
@@ -405,8 +405,13 @@ IInterpreterRegistryService service
             interpExe = Path.GetFileName(baseInterpreter.Configuration.InterpreterPath);
             winterpExe = Path.GetFileName(baseInterpreter.Configuration.GetWindowsInterpreterPath());
             var scripts = new[] { "Scripts", "bin" };
+
+            // new versions of python >= 3.8 from the windows store have a different executable name (ie. python3.8.exe) in the 
+            // registry InstallPath, compared to what is in the venv folder (python.exe) so search for both if interpExe is not found.
             interpExe = PathUtils.FindFile(prefixPath, interpExe, firstCheck: scripts);
+            interpExe = interpExe == null ? PathUtils.FindFile(prefixPath, "python.exe", firstCheck: scripts) : interpExe;
```
huguesv, line 412: you could do it in one statement instead of 2

interpExe = PathUtils.FindFile(...) ?? PathUtils.FindFile(...);

PR #6170, opened 2020-07-07
Update the feature/lsc branch with all the changes since around October

huguesv:
Unfortunately in that period we updated miniconda, so lots of file changes are in that subfolder.


Python/Product/Miniconda/Miniconda3-x64/Lib/_nsis.py
```diff
@@ -132,7 +187,7 @@ def run_post_install():
 
 
 allusers = (not exists(join(ROOT_PREFIX, '.nonadmin')))
-out('allusers is %s\n' % allusers)
+# out('allusers is %s\n' % allusers)
```
bschnurr, line 190: i wonder why this was commented out

huguesv: that was probably just logging, not sure why they would take it out though

PR #6196, opened 2020-08-10
Fix deadlock when loading a workspace.

huguesv:
Remove UI thread dependency from the MEF constructor, which was a violation of threading guidelines.
Ensure we are on UI thread in `OnActiveWorkspaceChanged`.

Fixes internal bug 1168894


bschnurr:
I wonder if projects will have a similar issue.

huguesv:
This one became a problem due to a change / refactoring in the workspace implementation in VS. I'm hopeful that didn't affect project system.

Issue #6181, opened 2020-07-31
Disable/Delete Microsoft.PythonTools.Analyzer.exe


manstie:
**Describe the feature you are requesting**
Allow me to delete Microsoft.PythonTools.Analyzer.exe without an error appearing or allow me to disable the feature.

**Reasons for requesting this feature**
Whenever I have my django server running this file eventually takes up all my ram (11gb from this one .exe out of 16gb).

**Describe alternatives you've considered**
I deleted the file so it doesn't run when I run my server for testing but I get an error, and there is no way to disable it.

**Additional context**
[Window Title]
devenv.exe

[Main Instruction]
An unexpected error occurred

[Content]
Please press Ctrl+C to copy the contents of this dialog and report this error to our issue tracker.

[^] Hide details  [Close]

[Expanded Information]
```
Build: 16.6.20100.1

System.ComponentModel.Win32Exception (0x80004005): The system cannot find the file specified
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at System.Diagnostics.Process.Start(ProcessStartInfo startInfo)
   at Microsoft.PythonTools.Intellisense.VsProjectAnalyzer.StartSubprocessConnection(String comment, AnalysisProcessInfo& proc)
   at Microsoft.PythonTools.Intellisense.VsProjectAnalyzer.<InitializeAsync>d__42.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PythonTools.Intellisense.VsProjectAnalyzer.<CreateForProjectAsync>d__33.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PythonTools.Project.PythonProjectNode.<CreateAnalyzerAsync>d__142.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #6197, opened 2020-08-11
Allow disabling of IntelliSense / language server

huguesv:
For users that experience analyzer crash, memory or CPU issues with their code base or Python environment.

When checked, this prevents the launch of the analyzer process, which means no IntelliSense, no refactoring, semantic colorization, etc.

image

Additionally, prevent the analyzer from restarting if it has crashed more than 5 times during the session.

Fixes https://github.com/microsoft/PTVS/issues/6181


bschnurr:
hmm. not sure "Enable" is better than "Disable", but I guess this is ok.

bschnurr:
did you test the interactive window?

huguesv:
Yes, it disables IntelliSense in interactive window as well.

I've called it Disable IntelliSense to match the name used for the similar C++ tools/option.

PR #6199, opened 2020-08-12
[16.7] Fix deadlock when loading a workspace.

huguesv:
We may have to fix this in 16.7 servicing release, so preparing the backport in case.


bschnurr:
great fix!

PR #6255, opened 2020-09-22
Create Pylance extension

MikhailArkhipov:
- Unload Core for now
- Remove XAML, IronPython
- Remove Python 3.8 warning
- Update strong name verification scripts
- Place Pylance dist in subfolder
- Update code respectively to above


huguesv (approved):
Appears to be working fine, it's launching Pylance from the vsix.

The info bar might not work as expected due to the removal of the 2.7 detection code (my project that referenced 2.7 was automatically targeting my global default 3.7 so I did not see the info bar). We can worry about that later.

PR #6257, opened 2020-09-23
Support loose files in editor

MikhailArkhipov:
- Also fixes reopening of folder with previously open files. 
- Minor code cleanup

### DIFF
--- a/Python/Product/PythonTools/PythonToolsPackage.cs
+++ b/Python/Product/PythonTools/PythonToolsPackage.cs
@@ -46,8 +46,6 @@ using Microsoft.VisualStudio.InteractiveWindow.Shell;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Text;
-using Microsoft.VisualStudio.TextManager.Interop;
-using Microsoft.VisualStudio.Threading;
 using Microsoft.VisualStudioTools;
 using Microsoft.VisualStudioTools.Navigation;
 using Microsoft.VisualStudioTools.Project;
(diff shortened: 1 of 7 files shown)


Python/Product/PythonTools/PythonToolsPackage.cs
```diff
@@ -315,17 +299,15 @@ protected override Task<object> InitializeToolWindowAsync(Type toolWindowType, i
 
             // no launcher configured, use the default one.
             Debug.Assert(defaultLaunchProvider != null);
-            return (defaultLaunchProvider != null) ?
-                serviceProvider.GetUIThread().Invoke<IProjectLauncher>(() => defaultLaunchProvider.CreateLauncher(project)) :
-                null;
+            return serviceProvider.GetUIThread().Invoke(() => defaultLaunchProvider.CreateLauncher(project));
```
huguesv, line 302: I would keep the check for null to avoid crashing in release builds

huguesv:
> Also fixes reopening of folder with previously open files.

I suppose now we start a server with the global default settings instead of the workspace settings. We'll have to fix that later.

Some background:
The `IPythonWorkspaceContextProvider` that we use to check if a workspace is present hasn't initialized yet, it initializes itself asynchronously on the workspace opened event, which is racing with the opening of the previously opened files.

MikhailArkhipov:
_The IPythonWorkspaceContextProvider that we use to check if a workspace is present hasn't initialized yet, it initializes itself asynchronously on the workspace opened event, which is racing with the opening of the previously opened files._

Yes, I can see that it is reopened as standalone. Let me see if I can fix that...

MikhailArkhipov:
I moved client context creation to later, when workspace is available so file now reopens in the correct context.

huguesv (approved):

PR #6259, opened 2020-09-24
Handle interpreter change

MikhailArkhipov:
- Send updates to Pylance
- Minor reformatting to more modern style

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
@@ -95,14 +95,15 @@ namespace Microsoft.PythonTools.LanguageServerClient {
             }
 
             _clientContext = PythonWorkspaceContextProvider.Workspace != null 
-                ? (IPythonLanguageClientContext)new PythonLanguageClientContextWorkspace(PythonWorkspaceContextProvider.Workspace, PythonCoreConstants.ContentType) 
-                : new PythonLanguageClientContextGlobal(OptionsService, PythonCoreConstants.ContentType);
+                ? (IPythonLanguageClientContext)new PythonLanguageClientContextWorkspace(PythonWorkspaceContextProvider.Workspace) 
+                : new PythonLanguageClientContextGlobal(OptionsService);
 
             _clientContext.InterpreterChanged += OnInterpreterChanged;
             _clientContext.SearchPathsChanged += OnSearchPathsChanged;
             _disposables.Add(() => {
                 _clientContext.InterpreterChanged -= OnInterpreterChanged;
                 _clientContext.SearchPathsChanged -= OnSearchPathsChanged;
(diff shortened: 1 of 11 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -103,33 +94,34 @@ public class PythonLanguageClient : ILanguageClient, ILanguageClientCustomMessag
...
-            }
+            _clientContext = PythonWorkspaceContextProvider.Workspace != null 
+                ? (IPythonLanguageClientContext)new PythonLanguageClientContextWorkspace(PythonWorkspaceContextProvider.Workspace, PythonCoreConstants.ContentType) 
+                : new PythonLanguageClientContextGlobal(OptionsService, PythonCoreConstants.ContentType);
+
+            _clientContext.InterpreterChanged += OnInterpreterChanged;
+            _clientContext.SearchPathsChanged += OnSearchPathsChanged;
+            _disposables.Add(() => {
+                _clientContext.InterpreterChanged -= OnInterpreterChanged;
+                _clientContext.SearchPathsChanged -= OnSearchPathsChanged;
```
huguesv, line 105: I've noticed that we don't have anyone disposing `_clientContext`, I think it should be added here after the event unregistration.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientContextGlobal.cs
```diff
@@ -24,20 +24,15 @@ namespace Microsoft.PythonTools.LanguageServerClient {
...
 #pragma warning disable CS0067
         public event EventHandler SearchPathsChanged;
         public event EventHandler Closed;
 #pragma warning restore CS0067
 
-        public PythonLanguageClientContextGlobal(
-            IInterpreterOptionsService optionsService,
-            string contentTypeName
-        ) {
+        public PythonLanguageClientContextGlobal(IInterpreterOptionsService optionsService, string contentTypeName) {
```
huguesv, line 35: Btw, the contentTypeName here and in other places is a leftover from MPLS where each language server instance had its own derived content type in order to associate a file with the correct server.

We no longer need this, so if you feel like it, you can remove it from here, and the IPythonLanguageClientContext interface.

MikhailArkhipov: Yeah, I removed it. I was also wondering why it was there... :-)

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
MikhailArkhipov: Yep, added disposing.

huguesv (approved):

PR #6264, opened 2020-09-29
Ability to add new items to workspace + JSON schema for settings

MikhailArkhipov:
- Add 'Add New Item' to workspace via 'Provide*' attribute
- Add `Empty Module` and `Config file` to `File | New` in `Python` section
- Pylance config schema uploaded to our blob for now
- Move attribute definitions to a separate folder


huguesv:
Does the schema work for you? I'm getting this warning:
image
No completion or validation:
image

Python/Product/Pylance/Templates/NewItem/NewFileItems.vsdir
```diff
@@ -0,0 +1,2 @@
+module.py|{6DBD7C1E-1F1B-496D-AC7C-C55DAE66C783}|#3017|1|#3017|0|#404|0|module.py
+mspythonconfig.json|{6DBD7C1E-1F1B-496D-AC7C-C55DAE66C783}|#3019|1|#3019|0|#404|0|mspythonconfig.json
```
huguesv, line 2: VS used the default name mspythonconfig1.json when I went to add a new item. I did not have an existing file causing name conflict. Is there a way to make it not append 1 unless necessary?

MikhailArkhipov: Yes, these are still couple things to deal with. 

- Pylance settings schema specifies draft#7 of JSON schema while VS editor appears to only support 6 or 4. I think we may have to manually change schema standard URL to 6 to get rid of the warning.

- VS adds `1` to anything in workspace (not project). Tested the same with `web.config` - works properly in project, adds `1` when Web site is opened as folder. I'll see what I can do, maybe there are some hidden flags.

Let me check what's up with JSON completions - I did get them, maybe something with schema caching.

Python/Product/Pylance/Templates/NewItem/mspythonconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+    "$schema": "https://pvsc.blob.core.windows.net/mspythonconfig-schema/pyrightconfig-schema.json",
+    ""
```
huguesv, line 3: This is invalid JSON. If the user adds the item and doesn't edit it, it will cause an error on the next Pylance startup/update.
```
Error: Config file "C:\Projects\Temp\OpenFolderApp\mspythonconfig.json" could not be parsed. Verify that JSON is correct.
```

How about we put a real setting there as an example? Like:
"typeCheckingMode": "basic"

Python/Product/Pylance/Templates/NewItem/NewFileItems.vsdir
huguesv: JSON completions and validation now works 👍

MikhailArkhipov: Looks like we need wizard in order to nix (1) in the config name.

Python/Product/Pylance/Templates/NewItem/mspythonconfig.json
MikhailArkhipov: Actually, `""` should be removed, leaving just the schema.

huguesv (approved):

PR #6267, opened 2020-09-30
Remove legacy debugger option

MikhailArkhipov:
Remove legacy debugger option. Does not remove actual code, only Tools | Options related part. Also removes options that are configured in Pylance differently.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterRemoteProcess.cs
+++ b/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterRemoteProcess.cs
@@ -59,7 +59,6 @@ namespace Microsoft.PythonTools.Debugger {
                     _stream = new DebugRemoteAdapterProcessStream(new NetworkStream(socket, ownsSocket: true));
                     _stream.Disconnected += OnDisconnected;
                     _stream.Initialized += OnInitialized;
-                    _stream.LegacyDebugger += OnLegacyDebugger;
                 } else {
                     Debug.WriteLine("Timed out waiting for debugger to connect.", nameof(DebugAdapterRemoteProcess));
                     logger?.LogEvent(PythonLogEvent.DebugAdapterConnectionTimeout, "Attach");
@@ -73,6 +72,8 @@ namespace Microsoft.PythonTools.Debugger {
 
         private void OnDisconnected(object sender, EventArgs e) {
             if (_stream != null) {
+                _stream.Disconnected -= OnDisconnected;
+                _stream.Initialized -= OnInitialized;
                 _stream.Dispose();
             }
             Exited?.Invoke(this, null);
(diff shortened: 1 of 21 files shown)


Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterRemoteProcess.cs
```diff
@@ -59,7 +59,6 @@ internal sealed class DebugAdapterRemoteProcess : ITargetHostProcess, IDisposabl
                     _stream = new DebugRemoteAdapterProcessStream(new NetworkStream(socket, ownsSocket: true));
                     _stream.Disconnected += OnDisconnected;
                     _stream.Initialized += OnInitialized;
-                    _stream.LegacyDebugger += OnLegacyDebugger;
```
huguesv, line 62: This event comes from `DebugAdapterRemoteProcess.ConnectSocket` to inform user if they are connecting to a remote machine where the legacy ptvsd is running in the remote process, ie. they did an `import ptvsd; ...; ptvsd.wait_for_attach()` instead of `import debugpy; ...; debugpy.wait_for_client();`

It does that by checking the name of the messages it is receiving, it's easy because the legacy messages start with 'legacy'.

If you feel we don't need to worry about that anymore, then we should delete the code in DebugAdapterRemoteProcess that looks for the legacy messages and stop firing that LegacyDebugger event.

Python/Product/PythonTools/PythonTools/Options/PythonDebuggingOptions.cs
```diff
@@ -139,13 +136,5 @@ public sealed class PythonDebuggingOptions {
...
             set;
         }
-
-        /// <summary>
-        /// True to use the legacy debugger. Default is false.
-        /// </summary>
-        public bool UseLegacyDebugger {
-            get;
-            set;
-        }
```
huguesv, line 149: We should delete `DebugProjectUITestsLegacyPtvsd` class now. That means the `UseVsCodeDebugger` there can be removed, and in the actual test code (DebuggerUITests/DebugProjectUITests) the `bool useVsCodeDebugger` parameter to each method can be removed. We can do that in another PR since that's a lot of changes.

FYI debug/attach to process still uses legacy ptvsd (it doesn't look at the setting), this is a feature we still need to implement for debugpy. So AttachUITests is to remain as-is until then.

MikhailArkhipov: Yes, actual code removal is going to be large change. This one pretty much just hides legacy option. I think it may be better to do separate PR for the actual engine removal.

huguesv (approved):
The _stream.LegacyDebugger event is probably less useful as time goes on as fewer people may be accidentally be using ptvsd on remote python environment, so feel free to merge as is unless you think it should be put back in.

PR #6268, opened 2020-10-01
Formatting options

MikhailArkhipov:
On top of https://github.com/microsoft/PTVS/pull/6267

- Add formatter selection to Tools | Options | Text Editor | Python | Formatting
- Handle formatter seelction
- Disable `Format Selection` when formatter does not support it
- If formatter is not installed, retry formatting after it gets installed
- Use `IVsUIShell` to show message boxes


Python/Product/PythonTools/PythonTools/Editor/Formatting/PythonFormatCommandHandler.cs
```diff
@@ -195,100 +177,47 @@ string[] extraArgs
...
-
-            // TODO: if not configured, tell the user? for now use default
-            if (string.IsNullOrEmpty(formatterId)) {
-                formatterId = defaultFormatterId;
-            }
-
-            if (!string.IsNullOrEmpty(formatterId)) {
-                formatter = _formattingProviders.SingleOrDefault(p => string.Compare(p.Value.Identifier, formatterId, StringComparison.OrdinalIgnoreCase) == 0)?.Value;
-            }
+            var formatterId = _site.GetPythonToolsService().FormattingOptions.Formatter;
```
huguesv, line 216: Please let's keep doing the workspace/project check first, then if not set, get the global default.
That's still useful even though it requires manual editing of json/pyproj at this time.

MikhailArkhipov: Question is - what might store the option there. The file is JSON inside `.vs` hidden folder. So in order to have something there, we'd somehow need to store information there. Since there is no code that would be storing data, I decided not to read it. Technically in VS formatting is in `Tools | Options | Text Editor | Python | Formatting` (as it is with other languages) so setting is always global - otherwise each time use opened `Tools | Options` the setting would come different depending what workspace or project is active.

huguesv: While there is no UI for editing the workspace settings in PythonSettings.json, we still have settings. The experience is bad compared to VS code which has a UI on top of settings, but that's what we have so far. User is expected to hand edit that file right now. For things like unit test framework, search paths, formatter, and interpreter (that last one has a UI -- the python toolbar writes that one).

Having the ability to set the formatter per workspace/project is important. Hopefully we can invest in a UI later, but until then, it's better if the user can edit the setting in the json editor than not being to control it other than global setting.

MikhailArkhipov: Oh, I see, it is `hidden` but documented file... :-)

huguesv: Correct. One of the reasons workspace experience is currently inferior to projects.

Example of documentation: https://docs.microsoft.com/en-us/visualstudio/python/unit-testing-python-in-visual-studio?view=vs-2019#configure-testing-for-python-without-a-project

MikhailArkhipov: OK, restored it.

huguesv (approved):

MikhailArkhipov:
Bad merge. Will reopen from another branch.

PR #6272, opened 2020-10-02
added cancellation support to PTVS.

heejaechang:
currently, it is using a hack to add the support since ILanguageClient doesn't provide a way for us to inject our CancellationStrategy at the right point.

but except the very first a few min once folder is opened, it should work as expected.

### DIFF
--- a/Python/Product/PythonTools/PythonTools.csproj
+++ b/Python/Product/PythonTools/PythonTools.csproj
@@ -198,9 +198,7 @@
       <Private>True</Private>
       <IncludeInVSIX>True</IncludeInVSIX>
     </Reference>
-    <Reference Include="StreamJsonRpc">
-      <HintPath>$(PackagesPath)\StreamJsonRpc\lib\netstandard2.0\StreamJsonRpc.dll</HintPath>
-    </Reference>
+    <Reference Include="StreamJsonRpc" />
     <Reference Include="System" />
     <Reference Include="System.Collections.Immutable" />
     <Reference Include="System.ComponentModel.Composition" />
--- a/Python/Product/PythonTools/PythonTools.csproj
+++ b/Python/Product/PythonTools/PythonTools.csproj
@@ -198,7 +198,6 @@
       <Private>True</Private>
       <IncludeInVSIX>True</IncludeInVSIX>
     </Reference>
-    <Reference Include="StreamJsonRpc" />
     <Reference Include="System" />
     <Reference Include="System.Collections.Immutable" />
     <Reference Include="System.ComponentModel.Composition" />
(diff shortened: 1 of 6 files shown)


heejaechang:
@huguesv can you take a look?

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -209,5 +216,62 @@ public class PythonAnalysisSettings {
...
+                this._jsonRpc = jsonRpc ?? throw new ArgumentNullException(nameof(jsonRpc));
+
+                this._cancellationFolderPath = Path.Combine(Path.GetTempPath(), "python-languageserver-cancellation", this._folderName);
+
+                this._cancellationStrategy = this._jsonRpc.CancellationStrategy;
+                this._jsonRpc.Disconnected += OnDisconnected;
+
+                try {
+                    Directory.CreateDirectory(this._cancellationFolderPath);
+                } catch {
```
huguesv, line 238: please change these catch all to use `IsCriticalException` extension method like it is done at https://github.com/microsoft/PTVS/blob/63992ffc2d5e8453df0af4896ad3128275ec38f8/Python/Product/PythonTools/PythonTools/PythonFeedbackDiagnosticFileProvider.cs#L74

Python/Product/PythonTools/PythonTools.csproj
```diff
@@ -198,6 +198,9 @@
       <Private>True</Private>
       <IncludeInVSIX>True</IncludeInVSIX>
     </Reference>
+    <Reference Include="StreamJsonRpc">
```
huguesv, line 201: please search for StreamJsonRpc in `Build\Common.Build.CSharp.targets` 
you should be able to change `net46` there to `netstandard2.0` and undo this change, as well as undo your change in PreBuild.ps1

huguesv:
@heejaechang the rest looks good, thanks for doing this!

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
heejaechang: thank you. made not catch critical exceptions.

Python/Product/PythonTools/PythonTools.csproj
heejaechang: ah, thank you. I guess I missed this one when I searched for StreamJsonRpc string in the folder.

heejaechang:
@huguesv addressed your feedback.

huguesv (approved):

heejaechang:
@huguesv I can't check this in. I don't have permission.

PR #6275, opened 2020-10-06
Pylance settings page

MikhailArkhipov:
- Pylance settings page.
- Text Editor | Advanced page.
- Plumb setting changes to LSP.
- Don't fire 'changed' event when settings didn't change.
- Allow workspace override for selected options.

image

image


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -158,24 +163,22 @@ public Task<TResult> InvokeWithParameterObjectAsync<TResult>(string targetName,
...
             var settings = new Settings {
                 python = new Settings.PythonSettings {
                     pythonPath = _clientContext.InterpreterConfiguration.InterpreterPath,
                     venvPath = string.Empty,
                     analysis = new Settings.PythonSettings.PythonAnalysisSettings {
-                        logLevel = "log",
-                        autoSearchPaths = false,
-                        diagnosticMode = "openFilesOnly",
-                        diagnosticSeverityOverrides = new Dictionary<string, string>(),
-                        extraPaths = _clientContext.SearchPaths.ToArray(),
```
huguesv, line 176: We shouldn't remove support for `extraPaths`. What was there for it looked correct.

huguesv:
Rename the tree view item on the left from "Diagnostics" to "Analysis" ?

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
MikhailArkhipov: Yes, I need to look into few more (and probably expose them), settings are scattered between Pylance, Pyright and core extension.

MikhailArkhipov: Added more options

Python/Product/PythonTools/PythonTools/Options/PythonAnalysisOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+                changed = true;
+            }
+
+            var typeCheckingMode = _service.LoadString(nameof(TypeCheckingMode), Category) ?? PythonLanguageClient.TypeCheckingMode.Basic;
+            if (TypeCheckingMode != typeCheckingMode) {
+                TypeCheckingMode = typeCheckingMode;
+                changed = true;
+            }
+
+            var typeshedPaths = _service.LoadString(nameof(TypeshedPaths), Category)?.Split(';');
```
huguesv, line 73: If it's a multi-line text box, users may be tempted to put newlines in there. Should we split on `';\r\n'` for extra convenience?

huguesv: oh wrong place

MikhailArkhipov: Yes, makes sense. I probably also need to add some text to the page wrt 'use line breaks or semicolon to separate paths'. Long term I am going to open task to implement better UI like

image

huguesv (approved):
If you haven't already, double check that the tab order is correct in both pages, since that is usually tricky when using the table control.

If we don't have a task for watching for workspace settings changed, we should create one.

Python/Product/PythonTools/PythonTools/Options/PythonAnalysisOptions.cs
MikhailArkhipov: Done.

PR #6279, opened 2020-10-09
Add Python 3.9 support

MikhailArkhipov:


huguesv (approved):

bschnurr (approved):

LazyDodo:
Is there any chance this will still make it in for 16.9? we recently updated blender to python 3.9 and the lack of 3.9 support makes debugging a whole lot less pleasant than it used to be :) 

it seemed to have landed in only the `pylance` branch, but not the `16.9` one so i'm not feeling overly hopeful here.

PR #6283, opened 2020-10-13
Telemetry forwarding

MikhailArkhipov:
- Move telemetry recorder to `Common`.
- Make parameters string/string for simplicity.
- Handle `python/telemetry`.
- Handle exception as `FaultEvent`.
- Update tests.

Tests probably need to be moved from `CookiecutterTests` project. We should probably create `CommonTests` at some point. Also, may try to build faulting Pylance to try exception reporting.


huguesv:
We already have telemetry logging functionality in PythonTools project, via `IPythonToolsLogger` service (defined in VSCommon/Logging). We should be able to use that.

The code in Cookiecutter was written as standalone source base from the rest of PTVS, so it got its own implementation.

MikhailArkhipov:
Perhaps instead replace `IPythonToolsLogger` by the common set? It looks about the same except it has `PythonLogEvent` which are not applicable anymore since analyzer is gone? Basically wrap common class and export the wrapper.

PR #6286, opened 2020-10-14
Forward telemetry from LS to VS

MikhailArkhipov:
Using PythonTools telemetry logger


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientCustomTarget.cs
```diff
@@ -25,18 +27,54 @@
...
+            }
+
+            Trace.WriteLine(telemetry.ToString());
+            var te = telemetry.ToObject<PylanceTelemetryEvent>();
+            if (te == null) {
+                return;
+            }
+
+            if (te.Exception == null) {
+                _logger.LogEvent(PythonLogEvent.LanguageServer, te.Properties);
```
huguesv, line 75: Is EventName also present in te.Properties? Otherwise, it is lost in this case, and maybe we need to add it to the Properties dictionary?

MikhailArkhipov: Yes

huguesv (approved):

PR #6297, opened 2020-10-21
Project support

MikhailArkhipov:
Covers basic scenarios with projects

- Added deployment of an empty targets file
- Added additional warning for Python 2 since gold bar does not appear in all cases
- Added basic templates to Pylance project (not sure if we want all existing ones)
- Added template for configuration file

TODO: Support for multiple projects with different Python versions via configuration files. Opened separate task.


Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -43,9 +43,13 @@ public sealed class DebugAdapterLauncher : IAdapterLauncher {
         public void Initialize(IDebugAdapterHostContext context) => _adapterHostContext = context ?? throw new ArgumentNullException(nameof(context));
 
         public ITargetHostProcess LaunchAdapter(IAdapterLaunchInfo launchInfo, ITargetHostInterop targetInterop) {
+            //if(!CheckPythonVersion(launchInfo.LaunchJson)) {
```
huguesv, line 46: Instead of changing this here, you should update DebugLaunchHelper.CreateDebugTargetInfo, because that's shared by all debuggers launch (debugpy and mixed-mode). There's already a check for < 2.6 over there, it can be changed to < 3.0.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -240,5 +241,16 @@ public sealed class DebugAdapterLauncher : IAdapterLauncher {
                 Marshal.FreeHGlobal(argPointer);
             }
         }
+
+        private static bool CheckPythonVersion(string adapterLaunchJson) {
```
huguesv, line 245: Should be able to remove based on my prior comment.

Common/Product/SharedProject/Utilities.cs
```diff
@@ -344,7 +332,7 @@ public static class Utilities {
                     unknown1 = QueryInterfaceIUnknown(obj1);
                     unknown2 = QueryInterfaceIUnknown(obj2);
 
-                    isSame = IntPtr.Equals(unknown1, unknown2);
+                    isSame = Equals(unknown1, unknown2);
```
huguesv, line 335: Is that right? Which `Equals` is that going to call, this class' `Equals`? If unsure I'd leave it as is, unless you know it was causing a problem.

Python/Product/ProjectWizards/MsPythonConfigWizard.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            if (string.IsNullOrEmpty(projectPath)) {
+                return;
+            }
+
+            var fileName = Path.Combine(projectPath, _configFileName);
+            if (File.Exists(fileName)) {
+                MessageBox.Show(Strings.ConfigFileAlreadyExists, Strings.ProductTitle, MessageBoxButtons.OK, MessageBoxIcon.Error, MessageBoxDefaultButton.Button1);
+            } else {
+                File.WriteAllText(fileName,
```
huguesv, line 54: This might throw if for example if folder is not writeable by user, is it handled by the IDE or do we have to take care of it?

MikhailArkhipov: I guess IDE since I copied it from EditorConfig wizard. Wizards are third party extensions so IDE should be try-catching all calls.

Common/Product/SharedProject/Utilities.cs
MikhailArkhipov: OK. That was C# suggestion.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
MikhailArkhipov: Cool, that is what I have been looking for!

huguesv (approved):

PR #6370, opened 2021-01-22
run PyLint and MyPy at ProjectHome instead of .pyproj directory

### LINKED ISSUE #6368: Wrong working directory for PyLint and MyPy
Targets PythonRunPyLintCommand and PythonRunMypyCommand use WorkingDirectory set to $(MSBuildProjectDirectory) that is wrong if ProjectHome property set to something other than '.'.

That results in errors like: 
```
************* Module access.py
access.py(1,0): warning F0001: No module named access.py [F:fatal]
```

Correct value for WorkingDirectory is $(QualifiedProjectHome).

AdamYoblick:
Fixes #6368


bschnurr:
did you look for other commands using MSBuildProjectDirectory?

Python/Product/BuildTasks/Microsoft.PythonTools.targets
```diff
@@ -168,7 +168,7 @@ permissions and limitations under the License.
     <CreatePythonCommandItem Target="pylint"
                              TargetType="module"
                              Arguments="&quot;--msg-template={abspath}({line},{column}): {category} {msg_id}: {msg} [{C}:{symbol}]&quot; -r n @(Compile, ' ')"
-                             WorkingDirectory="$(MSBuildProjectDirectory)"
+                             WorkingDirectory="$(QualifiedProjectHome)"
```
heejaechang, line 171: what is QualifedProjecthome actually represent?

bschnurr: defined above. looks like it adds a ending backslash or the projectHome dir

AdamYoblick: It's the `ProjectHome` of the pyproj. When you open from existing code, `ProjectHome` gets set to something other than `.`. But `MSBuildProjectDirectory` always points to the directory containing your pyproj, even if the actual code files live somewhere else.

So that's why this is currently broken when creating a project from existing code, since the `MSBuildProjectDirectory` will have no python files underneath, which causes pylint and mypy to blow up when trying to run them.

AdamYoblick: Here's where it gets set, on line 135:

```
    <QualifiedProjectHome Condition="'$(ProjectHome)' != '' and '$(ProjectHome)' != '.'">$([System.IO.Path]::GetFullPath($([System.IO.Path]::Combine($(MSBuildProjectDirectory), $(ProjectHome)))))</QualifiedProjectHome>
```

AdamYoblick:
@bschnurr , no, I didn't look for any other instances of MSBuildProjectDirectory. This is a case where that value is not valid, but there are many cases where it is. I'd rather just fix one bug at a time and not scope creep on this, since this is an easy fix. 😄

bschnurr (approved):

PR #6337, opened 2021-01-12
Customize variable presentation using tools options page

AdamYoblick:
Fixes #6334
Fixes #6342 

Definitely open to UI changes or approvals if we need, I'm not familiar with the process.

Here's what it looks like:

image

image

It also fixes the immediate window from showing too much information:

image

PR #6372, opened 2021-01-25
Use global debug options instead of project-specific ones

### LINKED ISSUE #6369: Failed to launch debug adapter in romote debugging with debugpy.
Python Environment:
image

**Steps to Reproduce**
1.Typing code in .py and set breakpoint in the code.
```
input('>')
print('hello')
```
2.Install "debugpy" in the current python environment.
3.In command prompt(right-click the python environment and select "Open Command Prompt Here…"), execute:
python -m debugpy --listen 127.0.0.1:4022 C:\Users\vwazho\source\repos\PythonApplication1\PythonApplication1\PythonApplication1.py
4. Launch VS and Debug > Attach to Process, in dialog:
		a. Select connection type with Python remote(ptvsd)
		b. Connection target with 127.0.0.1:4022
		c. In Available processes list, select the appropriate python.exe, click "Attach"
5. Type enter in the console
6.Check that it stops at the breakpoint
**Expected behavior**
remote attach successfully with no errors

**Actual behavior**
error dialog pop up when attaching.
image

AdamYoblick:
Fixes #6369 

This is a fix for a regression I introduced when reading variable presentation options from the options page. The bug occurs when attaching to a remote debugpy process instead of launching the debugger directly, which is a case I did not test.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
+++ b/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
@@ -203,12 +203,16 @@ namespace Microsoft.PythonTools.Debugger {
             launchJson.DebugStdLib = debugService.DebugStdLib;
             launchJson.ShowReturnValue = debugService.ShowFunctionReturnValue;
 
-            try {
-                var adapterLaunchInfoJson = JObject.Parse(adapterLaunchInfo.LaunchJson);
-                AddVariablePresentationOptions(adapterLaunchInfoJson, launchJson);
-            } catch (JsonReaderException) {
-                // this should never happen
-                Debug.Fail("adapterLaunchInfo is not valid json");
+            // adapterLaunchInfo.LaunchJson can be null when attaching to a remote process
+            if (adapterLaunchInfo.LaunchJson != null) {
+
+                try {
+                    var adapterLaunchInfoJson = JObject.Parse(adapterLaunchInfo.LaunchJson);
+                    AddVariablePresentationOptions(adapterLaunchInfoJson, launchJson);
+                } catch (JsonReaderException) {
+                    // this should never happen
+                    Debug.Fail("adapterLaunchInfo is not valid json");
+                }
             }
 
             var excludePTVSInstallDirectory = new PathRule() {
--- a/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
+++ b/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
@@ -203,17 +203,12 @@ namespace Microsoft.PythonTools.Debugger {
             launchJson.DebugStdLib = debugService.DebugStdLib;
             launchJson.ShowReturnValue = debugService.ShowFunctionReturnValue;
 
-            // adapterLaunchInfo.LaunchJson can be null when attaching to a remote process
-            if (adapterLaunchInfo.LaunchJson != null) {
-
-                try {
-                    var adapterLaunchInfoJson = JObject.Parse(adapterLaunchInfo.LaunchJson);
-                    AddVariablePresentationOptions(adapterLaunchInfoJson, launchJson);
-                } catch (JsonReaderException) {
-                    // this should never happen
-                    Debug.Fail("adapterLaunchInfo is not valid json");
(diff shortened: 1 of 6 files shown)


Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs
```diff
@@ -203,12 +203,16 @@ public sealed class DebugAdapterLauncher : IAdapterLauncher {
...
             launchJson.ShowReturnValue = debugService.ShowFunctionReturnValue;
 
-            try {
-                var adapterLaunchInfoJson = JObject.Parse(adapterLaunchInfo.LaunchJson);
-                AddVariablePresentationOptions(adapterLaunchInfoJson, launchJson);
-            } catch (JsonReaderException) {
-                // this should never happen
-                Debug.Fail("adapterLaunchInfo is not valid json");
+            // adapterLaunchInfo.LaunchJson can be null when attaching to a remote process
+            if (adapterLaunchInfo.LaunchJson != null) {
```
heejaechang, line 207: so what is AddDebuggerOptions for?

heejaechang: so, if adapterLanguageInfo.launchJson is null, when this is called (https://github.com/microsoft/PTVS/blob/b5e38ce254348acce692b4cbe05c449df2c28c2b/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterLauncher.cs#L71), what do we expect?

int19h: https://github.com/microsoft/PTVS/blob/b5e38ce254348acce692b4cbe05c449df2c28c2b/Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugInfo.cs#L25-L105

Note that there are two different debug info classes there - one for launch, and one for attach. But they have a common base class, and `VariablePresentation` is there. So it seems like it should be available regardless of how the session gets started.

AdamYoblick: Thanks, makes sense. I guess what confuses me is the two different ways that you can get these options. The one I used in https://github.com/microsoft/PTVS/pull/6337/files seems to be the `PythonToolsService`, and those options get packed into the adapterLaunchInfo.LaunchJson...but these other options are being read from the global `IPythonDebugOptionsService`.

It doesn't make sense to me to have two different paths to the same options. Should I just always be using the global one?

AdamYoblick: It seems that I can just add some fields to the `IPythonDebugOptionsService` to expose the variable presentation stuff that already exists on the `DebuggerOptions` class. But this means that lots of the other code I added in the previous PR can go away, if we're just always going to be reading these options from the global options service.

int19h: Hm, good point. This one is really meant for properties that are project-specific; for global properties, I guess it can just read it directly, since there's no ambiguity over which project it comes from.

@karthiknadig might have some thoughts about this.

karthiknadig: The two ways were there to handle project specific and global option. It should be fine to read non-project specific properties directly from global location. 

PS: This part is also a bit messy because of the way we had to handle `ptvsd4` vs `debugpy` at the same time.

AdamYoblick: ok thank you everyone, I have a clear path now. Global makes sense to me too, since these options are not project specific. And since they are moving to global in vscode too, I think that's a good approach.

AdamYoblick:
Ok this all seems to be working. I tested both when launching the debugger directly through VS as well as attaching to a debugpy process, and both respect the variable presentation options from the Tools -> Options page. So this is ready for review again :)

int19h (approved):

bschnurr (approved):

heejaechang (approved):

AdamYoblick:
Thanks everyone!

PR #6377, opened 2021-01-28
Update loclayout.proj to reliably generate relative paths

zooba:


bschnurr:
Per email conversation. Locale team will fix things on their side.

PR #6388, opened 2021-02-04
Added azure-pipelines.yml to replace the classic AzDO pipeline

AdamYoblick:
This yaml will drive our CI and PR builds once we create a new AzDO pipeline that consumes it. Right now, the file is not being consumed, but it must be in the repo before it can be used by AzDO 😄 

This yaml should be functionally equivalent to the old classic build definition. The only difference is that both CI and PR builds are enabled, and PR builds only perform a subset of the build steps, for speed.

The yaml build was tested at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4442296&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9 and seems to be working just fine.

### DIFF
--- a/Build/16.0/nuget.config
+++ b/Build/16.0/nuget.config
@@ -1,10 +1,10 @@
 <configuration>
   <packageSources>
     <clear />
-    <!--<add key="PLS" value="https://pvsc.blob.core.windows.net/vs-python-ls/index.json" protocolVersion="3" />-->
+    <add key="PLS" value="https://pvsc.blob.core.windows.net/vs-python-ls/index.json" protocolVersion="3" />
     <add key="VS Editor" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vssdk/nuget/v3/index.json" protocolVersion="3" />
     <add key="VS Editor Impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vs-impl/nuget/v3/index.json" protocolVersion="3" />
     <add key="VSIDEPublicFeed" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/VSIDEPublicFeed/nuget/v3/index.json" protocolVersion="3" />
-    <!--<add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />-->
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
   </packageSources>
 </configuration>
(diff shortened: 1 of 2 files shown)


Build/16.0/nuget.config
```diff
@@ -1,5 +1,6 @@
 <configuration>
   <packageSources>
+    <clear />
```
bschnurr, line 3: 👍

azure-pipelines.yml
```diff
@@ -0,0 +1,235 @@
...
+    SourceFolder: '$(Build.BinariesDirectory)\raw\setup\swix'
+    Contents: '**'
+  continueOnError: true
+
+# Build the product
+- task: MSBuild@1
+  displayName: 'Build product'
+  inputs:
+    solution: Python/Product/dirs.proj
+    msbuildVersion: 16.0
```
bschnurr, line 122: should we make `msbuildVersion` a variable too?

azure-pipelines.yml
```diff
@@ -0,0 +1,235 @@
...
+    inputs:
+      signType: 'Test' # TODO: Change this to $(SignType) once Real signing is approved for this AzDO project ID
+      feedSource: 'https://devdiv.pkgs.visualstudio.com/DefaultCollection/_packaging/MicroBuildToolset/nuget/v3/index.json'
+
+  - task: MicroBuildLocalizationPlugin@3
+    displayName: 'Install localization plugin'
+    inputs:
+      type: $(LocType)
+      languages: 'VS'
+      feedSource: 'https://mseng.pkgs.visualstudio.com/DefaultCollection/_packaging/MicroBuildToolset/nuget/v3/index.json'
```
bschnurr, line 100: this loc task says its referenced by a mircobuild task further down. did you find a doc on microbuild yaml?

azure-pipelines.yml
AdamYoblick: It's only referenced once and this is how it was in the classic pipeline, so I'm fine to leave it honestly. It's an easy change if we ever have to change it. 😄

azure-pipelines.yml
AdamYoblick: I don't understand the question. I just created yaml from the existing steps, I didn't change anything here.

azure-pipelines.yml
```diff
@@ -0,0 +1,235 @@
...
+    msbuildVersion: 16.0
+    platform: 'AnyCPU'
+    configuration: '$(BuildConfiguration)'
+
+# Build the installer
+- task: MSBuild@1
+  displayName: 'Build installer'
+  inputs:
+    solution: Python/Setup/dirs.proj
+    msbuildVersion: 16.0
```
bschnurr, line 131: here was the second. but ok

bschnurr (approved):

azure-pipelines.yml
AdamYoblick: ah you're right, it's referenced twice. My mistake. Sure I can make it a variable then 😄

PR #6392, opened 2021-02-08
enable real signing

AdamYoblick:


AdamYoblick:
Going to merge this in because we need to get real signing working again in order to insert into VS.

PR #6416, opened 2021-03-11
Merge master into pylance feature branch

rchiodo:
Moving the feature/pylance to be up to date with the master branch


debug.log
```diff
@@ -0,0 +1 @@
+[0223/161757.992:ERROR:registration_protocol_win.cc(102)] CreateFile: The system cannot find the file specified. (0x2)
```
bschnurr, line 1: looks like this should be excluding. but i'm guessing its already in Master

bschnurr (approved):

debug.log
rchiodo: I don't believe I generated this myself so I think it's already there. Yeah it is.

On a subsequent submit, I can remove it. 

Thanks.

PR #6448, opened 2021-03-16
Fix all the guids to be the same everywhere

rchiodo:
Fix problems with tool windows and menus not working


Python/Product/Pylance/source.extension.vsixmanifest
```diff
@@ -9,7 +9,7 @@
     <Icon>PythonProject.ico</Icon>
     <PreviewImage>PythonProjectBig.ico</PreviewImage>
   </Metadata>
-  <Installation SystemComponent="false" AllUsers="true" InstalledByMsi="false">
+  <Installation SystemComponent="true" AllUsers="true" InstalledByMsi="true">
```
rchiodo, line 12: This is really related to this change but it's kind of in the same spirit. I'm moving the pylance feature branch to be the same as the old stuff until we decide that the extension is going to be outside of VS.

int19h (approved):

PR #6455, opened 2021-03-23
Fix project paths not being used by pylance

rchiodo:
Fixes: https://github.com/microsoft/ptvs-pylance/issues/170

When opening a file that references another, pylance can't find the other file. 

The reason being the relative directory is not part of the extraPaths.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PylanceLanguageServer.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PylanceLanguageServer.cs
@@ -51,6 +51,7 @@ namespace Microsoft.PythonTools.LanguageServerClient {
             var isDebugging = IsDebugging();
             var debugArgs = isDebugging ? GetDebugArguments() : string.Empty;
             var serverFilePath = isDebugging ? GetDebugServerLocation() : GetServerLocation();
+            var debuggerExtra = isDebugging ? "--verbose" : string.Empty;
 
             if (!File.Exists(serverFilePath)) {
                 MessageBox.ShowErrorMessage(_site, Strings.LanguageClientPylanceNotFound);
@@ -69,7 +70,7 @@ namespace Microsoft.PythonTools.LanguageServerClient {
                 RedirectStandardOutput = true,
                 UseShellExecute = false,
                 CreateNoWindow = true,
-                Arguments = $"{debugArgs} \"{serverFilePath}\" -- --stdio --cancellationReceive=file:{this.CancellationFolderName} --verbose",
+                Arguments = $"{debugArgs} \"{serverFilePath}\" -- --stdio --cancellationReceive=file:{this.CancellationFolderName} {debuggerExtra}",
             };
 
             var process = new Process {
@@ -77,7 +78,7 @@ namespace Microsoft.PythonTools.LanguageServerClient {
             };
 
             if (process.Start()) {
-                if (PylanceLanguageServer.IsDebugging()) {
+                if (isDebugging) {
                     // During debugging give us time to attach
                     await Task.Delay(5000);
                 }
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PylanceLanguageServer.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PylanceLanguageServer.cs
@@ -65,17 +65,33 @@ namespace Microsoft.PythonTools.LanguageServerClient {
                 FileName = nodePath,
                 WorkingDirectory = serverFolderPath,
                 RedirectStandardInput = true,
+                RedirectStandardError = true,
                 RedirectStandardOutput = true,
                 UseShellExecute = false,
                 CreateNoWindow = true,
-                Arguments = $"{debugArgs} \"{serverFilePath}\" -- --stdio --cancellationReceive=file:{this.CancellationFolderName}",
+                Arguments = $"{debugArgs} \"{serverFilePath}\" -- --stdio --cancellationReceive=file:{this.CancellationFolderName} --verbose",
             };
 
             var process = new Process {
                 StartInfo = info
             };
 
-            return process.Start() ? new Connection(process.StandardOutput.BaseStream, process.StandardInput.BaseStream) : null;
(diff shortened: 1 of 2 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/PylanceLanguageServer.cs
```diff
@@ -65,17 +65,33 @@ internal sealed class PylanceLanguageServer {
                 FileName = nodePath,
                 WorkingDirectory = serverFolderPath,
                 RedirectStandardInput = true,
+                RedirectStandardError = true,
                 RedirectStandardOutput = true,
                 UseShellExecute = false,
                 CreateNoWindow = true,
-                Arguments = $"{debugArgs} \"{serverFilePath}\" -- --stdio --cancellationReceive=file:{this.CancellationFolderName}",
+                Arguments = $"{debugArgs} \"{serverFilePath}\" -- --stdio --cancellationReceive=file:{this.CancellationFolderName} --verbose",
```
rchiodo, line 72: Oh I should probably remove the --verbose. Or only add it when debugging

bschnurr: yup good call

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -188,6 +197,13 @@ public Task<TResult> InvokeWithParameterObjectAsync<TResult>(string targetName,
                 PythonConstants.ExtraPathsSetting, null, Site, PythonWorkspaceContextProvider.Workspace, out _)?.Split(';')
                 ?? _analysisOptions.ExtraPaths;
 
+            // Extra paths should include the project paths as well.
+            if (this.ProjectContextProvider != null) {
+                var directories = from n in this.ProjectContextProvider.ProjectNodes
+                                  select n.BaseURI.Directory;
+                extraPaths = extraPaths != null ? extraPaths.Concat(directories).ToArray() : directories.ToArray();
```
bschnurr, line 204: @jakebailey would know more about extraPaths.

I'm thinking we shouldn't need to set this.

jakebailey: This is required if these folders have to be import roots.

Better would be to simulate multi-root (if these folders really are distinct), but IDK what VS's status is on doing that.

rchiodo: This was his recommendation actually.

rchiodo: Otherwise it starts with just the root to the solution and doesn't find modules in the root or a project

rchiodo: How does pylance see multiroot?

rchiodo: Is there a new language server for each?

jakebailey: I think this just needs an example project to show what you mean, maybe. Typically we don't want to blanket add each dir, but if these are folders containing the code that people edit and should be import roots, sure.

jakebailey: Pylance supports multi-root natively, so only one LS process, but you have to have all of the calls for multi-root in LSP working.

bschnurr: so just the base project root.. or something the user sets in the project settings as "root"

rchiodo: In a solution there are multiple projects though. Each one of these needs to be parsed as a separate 'workspace' by Pylance, which I what I believe you're saying Jake?

rchiodo: Here's a solution with multiple projects

image

Each 'project' here is like a workspace folder in VS code. This change is putting it into the 'extraPaths' setting. Is there another way to tell the LS about the workspace folders?

rchiodo: The projects in the example above are:
- PythonApplication_Extra
- PythonApplication_Extra2
- PythonApplication2

Each of these has a different folder for its code. They may or may not be under the solution.

In that specific example, the solution is here:
C:\Users\aku91\source\repos\PythonApplication2

While the projects end up:
C:\Users\aku91\source\repos\PythonApplication2\PythonApplication_Extra
C:\Users\aku91\source\repos\PythonApplication2\PythonApplication_Extra2
C:\Users\aku91\source\repos\PythonApplication2\PythonApplication2

Although they don't have to be relative to the solution.

rchiodo: I assume that you're talking about this notification instead?

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_workspaceFolders

bschnurr: yes, I believe this would be the better way to handle multiple projects, by using multiple workspaceFolders, instead of extrapaths.

jakebailey: Yes, this seems much more like multi-root. See also https://microsoft.github.io/language-server-protocol/specification#initialize which has the folders field you could use.

rchiodo: Okay thanks. I'm going to try getting that to work instead.

PR #6460, opened 2021-03-26
Another way to support project directories, use workspaceFolder change events.

rchiodo:
The difficult part of this was getting VS to call our 'registerCapabilities' overload instead of theirs as the StreamJsonRPC stuff does things based on which target registered first.

That's what the reflection code is for, it reverses the order so that our CustomMessageTarget is first and then the VSSDK target is second.


Python/Product/PythonTools/PythonTools/LanguageServerClient/PylanceLanguageServer.cs
```diff
@@ -65,17 +72,41 @@ internal sealed class PylanceLanguageServer {
...
+                // Write to output if we can't launch for some reason
+                if (process.HasExited) {
+                    string output = process.StandardError.ReadToEnd();
+                    var outputWindow = OutputWindowRedirector.GetGeneral(_site);
+                    outputWindow.WriteLine(output);
+                } else {
+                    // Otherwise create a connection where we wrap the stdin stream so that we can intercept all messages
+                    return new Connection(
+                        process.StandardOutput.BaseStream, 
+                        new StreamIntercepter(process.StandardInput.BaseStream, _serverSendHandler, (a) => { }));
```
rchiodo, line 101: This allows us to intercept the 'initialize' message and add parameters to it.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -153,6 +171,38 @@ internal sealed class PythonLanguageClient : ILanguageClient, ILanguageClientCus
             _rpc.CancellationStrategy = new CustomCancellationStrategy(_server.CancellationFolderName, _rpc);
             _rpc.AllowModificationWhileListening = false;
 
+            // We also need to switch the order on handlers for all of the rpc targets. Until
+            // the VSSDK gives us a way to do this, use reflection.
+            try {
```
rchiodo, line 175: This is where the internal ```targetRequestMethodToClrMethodMap``` that the json rpc object has is reversed. I had to use reflection to accomplish this.

Basically if you look here:https://github.com/microsoft/vs-streamjsonrpc/blob/640dca7d211154f93ee9771d43af2c5452a00587/src/StreamJsonRpc/Reflection/TargetMethod.cs#L44

You can see how when attempting to find a target method it finds the first item in the candidate list. 

Since the CustomMessageTarget is always the second item in the list (if it registers for the same methods as the VSSDK), then I had to reverse the list order.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -64,11 +68,14 @@ internal sealed class PythonLanguageClient : ILanguageClient, ILanguageClientCus
         public JoinableTaskContext JoinableTaskContext;
 
         private readonly DisposableBag _disposables;
-        private IPythonLanguageClientContext _clientContext;
+        private IPythonLanguageClientContext[] _clientContexts;
```
rchiodo, line 71: There used to be one of these for the whole solution. The first project. 

Now there are multiple and we send updates for each.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -226,20 +284,104 @@ public Task<TResult> InvokeWithParameterObjectAsync<TResult>(string targetName,
...
+                try {
+                    // If this is the initialize method, add the workspace folders capability
+                    if (message.Value<string>("method") == "initialize") {
+                        if (message.TryGetValue("params", out JToken messageParams)) {
+                            var capabilities = messageParams["capabilities"];
+                            if (capabilities != null) {
+                                if (capabilities["workspace"] == null) {
+                                    capabilities["workspace"] = JToken.FromObject(new { });
+                                }
+                                capabilities["workspace"]["workspaceFolders"] = true;
```
rchiodo, line 318: This is where the 'extra' capability is added. This allows pylance to support workspace folders

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -226,20 +284,104 @@ public Task<TResult> InvokeWithParameterObjectAsync<TResult>(string targetName,
...
 
-            return new PythonLanguageClientContextGlobal(OptionsService);
+        private void OnSolutionClosing() {
+            if (_workspaceFoldersSupported && IsInitialized && _sentInitialWorkspaceFolders) {
+                JoinableTaskContext.Factory.RunAsync(async () => {
+                    // If workspace folders are supported, then send our workspace folders
+                    var folders = from n in this.ProjectContextProvider.ProjectNodes
+                                  select new WorkspaceFolder { uri = new System.Uri(n.BaseURI.Directory), name = n.Name };
+                    if (folders.Any()) {
+                        await InvokeDidChangeWorkspaceFoldersAsync(new WorkspaceFolder[0], folders.ToArray());
```
rchiodo, line 339: In these handlers is where the workspace folder change events are sent.

bschnurr:
wow quite the maneuvering to get this to work.

bschnurr:
So in Open Project mode we can have multiple projects that will each have their own workspace in Pylance. 
What happens in open folder mode, I'm assuming just one workspace still, right?

Python/Product/PythonTools/PythonTools.csproj
```diff
@@ -181,10 +182,10 @@
     </Reference>
     <Reference Include="Microsoft.VisualStudio.Workspace" />
     <Reference Include="Microsoft.VisualStudio.Workspace.VSIntegration.Contracts" />
-    <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed">
-      <HintPath>$(PackagesPath)\Newtonsoft.Json\lib\net45\Newtonsoft.Json.dll</HintPath>
+    <Reference Include="Newtonsoft.Json, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed">
+      <HintPath>$(PackagesPath)\Microsoft.VSSDK.BuildTools\tools\vssdk\bin\lib\Newtonsoft.Json.dll</HintPath>
       <Private>True</Private>
-      <IncludeInVSIX>True</IncludeInVSIX>
+      <IncludeInVSIX>False</IncludeInVSIX>
```
bschnurr, line 188: did this fix the newtonsoft warnings?

bschnurr (approved):

jakebailey:
One of the requirements for multi-root is that the client can query setting for each workspace (provide a URI to the client then query settings for it, kinda like Resource in the core ext); does this handle that? Or is this still using the hacked settings pushed in for VS?

Python/Product/PythonTools/PythonTools.csproj
rchiodo: Sorry but I don't recall. I did this since we don't need to ship it newtonsoft as it's part of VS already.

rchiodo:
> One of the requirements for multi-root is that the client can query setting for each workspace (provide a URI to the client then query settings for it, kinda like Resource in the core ext); does this handle that? Or is this still using the hacked settings pushed in for VS?

AFAIK this will send new settings for each folder to pylance (so each can have say their own interpreter). I _think_ it's working but I didn't really worry too much about it. I can fix that later if it doesn't really work.

This is more about just getting the projects recognized as workspace folders and setting up the ability to listen to registerCapabilities (which I'm going to need for file watcher stuff later)

rchiodo:
> So in Open Project mode we can have multiple projects that will each have their own workspace in Pylance.
> What happens in open folder mode, I'm assuming just one workspace still, right?

In open folder mode it should be a single settings/workspace combination. I can't test it right now to be sure (in the middle of debugging something else), but that's the intent.

jakebailey:
> > One of the requirements for multi-root is that the client can query setting for each workspace (provide a URI to the client then query settings for it, kinda like Resource in the core ext); does this handle that? Or is this still using the hacked settings pushed in for VS?
> 
> AFAIK this will send new settings for each folder to pylance (so each can have say their own interpreter). I _think_ it's working but I didn't really worry too much about it. I can fix that later if it doesn't really work.
> 
> This is more about just getting the projects recognized as workspace folders and setting up the ability to listen to registerCapabilities (which I'm going to need for file watcher stuff later)

Sure. AFAIK we hacked pylance to support VS, but didn't assume multi-root would work (and therefore any settings pushed in are global to all workspaces). Maybe I'm wrong, though. That behavior is technically wrong, and I would think the spec would still allow push-style configs that we could work with.

I don't really have enough knowledge about VS or C# to be able to approve or comment on this PR; the method seems fine from the POV that it's unlikely VS will natively support this for a while. I'm sure Bill or Pavel know better what sort of hacks are condoned/not condoned.

int19h (approved):

PR #6461, opened 2021-03-29
Draft: Move pylance to come from npm

rchiodo:
This is a draft at the moment. Not sure how to try out the setup authoring just yet.


Common/Product/SharedProject/CommonPackage.cs
```diff
@@ -83,20 +83,24 @@ public abstract class CommonPackage : AsyncPackage, IOleComponent {
         internal static object CommandsLock => _commandsLock;
 
         protected override void Dispose(bool disposing) {
-            _uiThread.MustBeCalledFromUIThreadOrThrow();
-            try {
-                if (_componentID != 0) {
-                    _compMgr.FRevokeComponent(_componentID);
-                    _componentID = 0;
-                }
+            if (_uiThread != null)
```
int19h, line 86: This change feels wrong - now `Dispose()` is callable from any thread, but it won't actually dispose all the things that need to be disposed. If it's getting called from a background thread somewhere, shouldn't the caller be fixed?

rchiodo: Yeah it was caused by me missing some stuff. UI thread is actuall null in some cases. I'll remove the invokeasync

rchiodo:
Submitting this to see if this will run with the token on a subsequent submission

PR #6467, opened 2021-04-03
Implement support for watching files

rchiodo:
For: https://github.com/microsoft/ptvs-pylance/issues/96

Implemented support for filewatchers with either the workspace object or the System.IO.FileWatcher class (for solutions).

Also fixed a bug when opening a single project that confuses the 'workspace' for pylance (rootPath should not be set when using workspaceFolders)

Added a UI test to verify pylance is working.

I'd like to do more testing after this to verify all scenarios are working, but this is the base file watching support.

### DIFF
--- a/Common/Product/SharedProject/CommonProjectNodeProperties.cs
+++ b/Common/Product/SharedProject/CommonProjectNodeProperties.cs
@@ -28,9 +28,11 @@ namespace Microsoft.VisualStudioTools.Project {
     [ClassInterface(ClassInterfaceType.AutoDual)]
     public class CommonProjectNodeProperties : ProjectNodeProperties, IVsCfgBrowseObject, VSLangProj.ProjectProperties {
         private OAProjectConfigurationProperties _activeCfgSettings;
+        private string __idStr;
 
         internal CommonProjectNodeProperties(ProjectNode node)
             : base(node) {
+            __idStr = node.ID.ToString();
         }
 
         #region properties
@@ -427,7 +429,7 @@ namespace Microsoft.VisualStudioTools.Project {
 
         [Browsable(false)]
         public string __id {
-            get { throw new NotImplementedException(); }
+            get => __idStr;
         }
 
         [Browsable(false)]
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
@@ -25,6 +25,10 @@ namespace Microsoft.PythonTools.LanguageServerClient.FileWatcher {
             // Ignore some common directories
             _matcher.AddExclude("**/.vs/**/*.*");
 
+            // Ignore files that end with ~ or TMP
+            _matcher.AddExclude("**/*~");
+            _matcher.AddExclude("**/*TMP");
+
             // Depending upon if this is a workspace or a solution, listen to different change events.
             if (workspaceService != null && workspaceService.CurrentWorkspace != null) {
                 workspaceService.CurrentWorkspace.GetService<IFileWatcherService>().OnFileSystemChanged += OnFileChanged;
@@ -35,7 +39,9 @@ namespace Microsoft.PythonTools.LanguageServerClient.FileWatcher {
                     _solutionWatcher = new System.IO.FileSystemWatcher();
                     _solutionWatcher.Path = Path.GetDirectoryName(dte.Solution.FileName);
                     _solutionWatcher.IncludeSubdirectories = true;
-                    _solutionWatcher.NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.CreationTime | NotifyFilters.DirectoryName | NotifyFilters.FileName;
+                    _solutionWatcher.NotifyFilter = 
+                        NotifyFilters.LastWrite | NotifyFilters.CreationTime | 
+                        NotifyFilters.DirectoryName | NotifyFilters.FileName | NotifyFilters.Size;
                     _solutionWatcher.Changed += OnFileChanged_Sync;
                     _solutionWatcher.Created += OnFileChanged_Sync;
                     _solutionWatcher.Deleted += OnFileChanged_Sync;
@@ -81,7 +87,10 @@ namespace Microsoft.PythonTools.LanguageServerClient.FileWatcher {
                     // Send out the event to the language server
                     var renamedArgs = e as System.IO.RenamedEventArgs;
                     var didChangeParams = new DidChangeWatchedFilesParams();
-                    if (renamedArgs == null) {
+
+                    // Visual Studio actually does a rename when saving. The rename is from a file ending with '~'
+                    if (renamedArgs == null || renamedArgs.OldFullPath.EndsWith("~")) {
+                        renamedArgs = null;
                         didChangeParams.Changes = new FileEvent[] { new FileEvent() };
                         didChangeParams.Changes[0].Uri = new Uri(e.FullPath);
 
@@ -95,6 +104,10 @@ namespace Microsoft.PythonTools.LanguageServerClient.FileWatcher {
                             case WatcherChangeTypes.Changed:
                                 didChangeParams.Changes[0].FileChangeType = FileChangeType.Changed;
                                 break;
+                            case WatcherChangeTypes.Renamed:
+                                didChangeParams.Changes[0].FileChangeType = FileChangeType.Changed;
+                                break;
+
                             default:
                                 didChangeParams.Changes = Array.Empty<FileEvent>();
                                 break;
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
@@ -1,0 +1,157 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.FileSystemGlobbing;
+using Microsoft.PythonTools.Interpreter;
+using Microsoft.VisualStudio.LanguageServer.Protocol;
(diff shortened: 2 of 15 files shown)


Common/Product/SharedProject/CommonProjectNodeProperties.cs
```diff
@@ -427,7 +429,7 @@ internal CommonProjectNodeProperties(ProjectNode node)
 
         [Browsable(false)]
         public string __id {
-            get { throw new NotImplementedException(); }
+            get => __idStr;
```
rchiodo, line 432: This fixes a crash I was having when trying to use the Quick find thing at the top of VS.

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
```diff
@@ -0,0 +1,176 @@
...
+using Microsoft.VisualStudio.LanguageServer.Protocol;
+using Microsoft.VisualStudio.Workspace;
+using Microsoft.VisualStudio.Workspace.VSIntegration.Contracts;
+using StreamJsonRpc;
+
+namespace Microsoft.PythonTools.LanguageServerClient.FileWatcher {
+    class Listener : IDisposable {
+        private readonly JsonRpc _rpc;
+        private System.IO.FileSystemWatcher _solutionWatcher;
+        private Microsoft.Extensions.FileSystemGlobbing.Matcher _matcher = new Microsoft.Extensions.FileSystemGlobbing.Matcher(StringComparison.InvariantCultureIgnoreCase);
```
rchiodo, line 18: This object here can use globs to match file names. It's how I'm doing the glob lookup for the files to watch from pylance.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -316,6 +428,14 @@ public Task<TResult> InvokeWithParameterObjectAsync<TResult>(string targetName,
                                     capabilities["workspace"] = JToken.FromObject(new { });
                                 }
                                 capabilities["workspace"]["workspaceFolders"] = true;
+                                capabilities["workspace"]["didChangeWatchedFiles"]["dynamicRegistration"] = true;
+
+                                // Root path and root URI should not be sent. They're deprecated and will
+                                // just confuse pylance with respect to what is the root folder. 
+                                // https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize
+                                // Setting them to empty will make pylance think they're gone.
+                                messageParams["rootPath"] = "";
```
rchiodo, line 437: New integration test found this bug. rootPath being set forces Pylance to pick this as a workspace folder. Then later when the same folder is sent for workspaceFolderChanged, Pylance ends up with two workspaces with the same path. This messes up its internal map and makes it look like the workspace never finishes loading.

These two properties are deprecated and replaced by the workspace folder notification in the LSP so I just removed them.

Python/Tests/Core.UI/IntellisenseTests.cs
```diff
@@ -0,0 +1,87 @@
...
+                        e.Insert(e.Snapshot.Length, "import sys\r\nsys.");
+                        e.Apply();
+                    }
+
+                    // Move cursor
+                    doc.TextView.Caret.MoveTo(new Microsoft.VisualStudio.Text.SnapshotPoint(doc.TextView.TextSnapshot, doc.TextView.TextSnapshot.Length));
+                });
+
+                // Wait until pylance is ready
+                Assert.IsTrue(Microsoft.PythonTools.LanguageServerClient.PythonLanguageClient.ReadyTask.Wait(60000), "Pylance did not start");
```
bschnurr, line 42: in my benchmark tests I wait for a telemetry event named language_server/analysis_complete
https://dev.azure.com/mseng/DSTools/_git/pyls-benchmarks?path=%2Fdriver%2Fdrive.py&version=GBmaster&line=106&lineEnd=107&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

bschnurr: just a suggestion to speed things up.

rchiodo: Ha!, that's exactly what the readytask waits for. I guess I picked the right event to wait for then.

bschnurr: oh. nevermind. i see its using that. nice!

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -228,10 +267,75 @@ public Task<TResult> InvokeWithParameterObjectAsync<TResult>(string targetName,
...
+
+                    var param = new DidOpenTextDocumentParams {
+                        TextDocument = textDocumentItem,
+                    };
+                    param.TextDocument.Text = textBuffer.CurrentSnapshot.GetText();
+
+                    await InvokeTextDocumentDidOpenAsync(param).ConfigureAwait(false);
+                }
+            }
+        }
```
bschnurr, line 304: should we group these tasks in a `Task.WhenAll`

rchiodo: Good idea. I don't see why they can't all go at once.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -283,11 +387,16 @@ public Task<TResult> InvokeWithParameterObjectAsync<TResult>(string targetName,
             await InvokeDidChangeConfigurationAsync(config);
```
bschnurr, line 383: minor. just noticed this `await ` isn't needed. could just return `InvokeDidChangeConfigurationAsync`

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -329,6 +449,15 @@ public Task<TResult> InvokeWithParameterObjectAsync<TResult>(string targetName,
             return data;
         }
 
+        private async Task OnWorkspaceOpening(object sende, EventArgs e) {
+            if (_workspaceFoldersSupported && IsInitialized && _sentInitialWorkspaceFolders && WorkspaceService.CurrentWorkspace != null) {
+                // Send just this workspace folder. Assumption here is that the language client will be destroyed/recreated on
+                // each workspace open
+                var folder = new WorkspaceFolder { uri = new System.Uri(WorkspaceService.CurrentWorkspace.Location), name = WorkspaceService.CurrentWorkspace.GetName() };
+                await InvokeDidChangeWorkspaceFoldersAsync(new WorkspaceFolder[] { folder }, new WorkspaceFolder[0]);
```
bschnurr, line 462: probably more of these extra 'await' that could just be returns.  could inverse the if logic to return a completed task but thats more a styling sugguestion

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -216,10 +249,16 @@ public Task InvokeTextDocumentDidChangeAsync(LSP.DidChangeTextDocumentParams req
         public Task InvokeDidChangeConfigurationAsync(LSP.DidChangeConfigurationParams request)
             => _rpc == null ? Task.CompletedTask : _rpc.NotifyWithParameterObjectAsync("workspace/didChangeConfiguration", request);
```
bschnurr, line 246: does _rpc ever equal null?

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
rchiodo: Actually it is. Well unless I'm misunderstanding how C# tasks work. There's an await at the top of this function. AFAICT that means I have to await the task on exit too? At least when I change this to a return, it gives me a compiler error:

image

bschnurr: i guess because its not an async Task then?

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
rchiodo: It can if the attach hasn't occurred yet.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
rchiodo: No I don't think it's checking that. 

It's if you have an async applied to a function, you cannot return a task. You have to return the type of the task and in this case the Task is of type void, so you cannot return anything.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
rchiodo: Yeah the async means the implicit else on the if statement doesn't need to return a completed task.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
bschnurr: hmm. we might need the project root when searching for a pyright configfile.

eric just fixed a bug related to this. https://github.com/microsoft/pyright/commit/e25b2c71df40f6ef195f77c03d79a4b0f26a6495

bschnurr (approved):

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
rchiodo: workspaceFolderChanged should cause the root to be set.

PR #6486, opened 2021-04-12
Fix problem with running conda activation on the main thread.

rchiodo:
https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1289875?src=WorkItemMention&src-action=artifact_link

Don't run conda activation events on the main thread.

### DIFF
--- a/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
+++ b/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
@@ -228,7 +228,7 @@ namespace Microsoft.PythonTools.Interpreter {
             ForceDiscoverInterpreterFactories();
         }
 
-        private void ForceDiscoverInterpreterFactories() {
+        private async void ForceDiscoverInterpreterFactories() {
             DiscoveryStarted?.Invoke(this, EventArgs.Empty);
 
             // Discover the available interpreters...
@@ -237,7 +237,7 @@ namespace Microsoft.PythonTools.Interpreter {
             var found = new List<PythonInterpreterInformation>();
 
             try {
-                FindCondaEnvironments(found);
+                found.AddRange(await FindCondaEnvironments());
             } catch (ObjectDisposedException) {
                 // We are aborting, so silently return with no results.
                 return;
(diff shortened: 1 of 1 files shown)


rchiodo:
@bschnurr or @int19h can you guys take a look? Thanks

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -300,15 +300,17 @@ internal class CondaInfoResult {
...
+                return await FindCondaEnvironments(mainCondaExePath);
             }
+            return Enumerable.Empty<PythonInterpreterInformation>().ToList().AsReadOnly();
         }
 
-        private IReadOnlyList<PythonInterpreterInformation> FindCondaEnvironments(string condaPath) {
-            var condaInfoResult = _joinableTaskFactory.Run(() => ExecuteCondaInfoAsync(condaPath));
+        private async Task<IReadOnlyList<PythonInterpreterInformation>> FindCondaEnvironments(string condaPath) {
+            // Make sure to run this on a background thread as it can take a long time.
+            var condaInfoResult = await System.Threading.Tasks.Task.Run(() => ExecuteCondaInfoAsync(condaPath));
```
int19h, line 313: As I recall, it used to be necessary to use `JoinableTaskFactory` in VS (in general, not just here) to avoid deadlocks when two tasks are both scheduled to run on the UI thread, but are awaiting each other. Did that change?

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -228,7 +228,7 @@ class CondaEnvironmentFactoryProvider : IPythonInterpreterFactoryProvider, IDisp
             ForceDiscoverInterpreterFactories();
         }
 
-        private void ForceDiscoverInterpreterFactories() {
+        private async void ForceDiscoverInterpreterFactories() {
```
int19h, line 231: `async void` means that any exception thrown in the body of the method is going to crash VS, right? In particular, `DiscoveryStarted` event handlers and `OnInterpreterFactoriesChanged` can no longer throw - but I don't think that's guaranteed. 

IIRC the usual pattern for this kind of stuff was to make it return `Task`, and then the calling code can use the `DoNotWait()` extension method defined somewhere to explicitly set it to fire-and-forget (and handle exceptions if they arise).

rchiodo: Good idea. I'll adjust.

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
rchiodo: It doesn't look like there's two tasks awaiting each other in this case? The main thread is awaiting on the task I kicked off. That thread is not calling back into the main thread AFAICT. 

The reason I moved this to Task.Run instead of JTF.Run. JTF.Run is a synch blocking wait essentially. At least according to the linked bug.

Is there a JTF safe method for kicking off a background thread to do work but not blocking? JTF.RunAsync seems to run the call on the main thread (when I debug it).

rchiodo:
@int19h did you have any more comments? I don't think the JTF.Run is the thing to use. According to the bug logged and debugging the code, JTF.Run will force the conda run onto the main thread. We don't need JTF if we don't need to talk back to main thread so I think Task.Run should be fine.

Additionally Task.Run is used elsewhere in PTVS

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -483,7 +485,7 @@ private sealed class DiscoverOnDispose : IDisposable {
             public void Dispose() {
                 Interlocked.Decrement(ref _provider._ignoreNotifications);
                 if (_forceDiscovery) {
-                    _provider.ForceDiscoverInterpreterFactories();
+                    _provider.ForceDiscoverInterpreterFactories().DoNotWait();
```
bschnurr, line 488: seems strange to run discovery when in dispose. maybe they just want to clear the list?

bschnurr (approved):

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
rchiodo: Yeah not sure why.

PR #6495, opened 2021-04-22
Support control click navigation like MPLS did

rchiodo:
Fixes: https://github.com/microsoft/ptvs-pylance/issues/90

Basically implement the INavigableSymbolSource provider so that control click works.

It uses the textDocument/definition request to find the thing under the cursor.

Here's a picture of the hover:

image


.npmrc
```diff
@@ -0,0 +1,3 @@
+registry=https://pkgs.dev.azure.com/devdiv/_packaging/Pylance/npm/registry/
```
rchiodo, line 1: You need this in order to pull down pylance from the internal feed. After this is set you run the directions supplied here to connect:
https://dev.azure.com/devdiv/DevDiv/_packaging?_a=connect&feed=Pylance

Python/Product/Common/Strings.resx
```diff
@@ -3040,6 +3040,7 @@ Please specify at least one package.</value>
   </data>
   <data name="ConfigFileAlreadyExists" xml:space="preserve">
     <value>Type checker configuration file already exists in this folder.</value>
+  </data>
```
rchiodo, line 3043: This was a bug from the merge from master.

Python/Tests/Core/WorkspaceTestHelper.cs
```diff
@@ -405,6 +407,12 @@ public class MockWorkspaceContext : IPythonWorkspaceContext {
                 throw new NotImplementedException();
             }
 
+            public bool IsTrusted {
+                get => true;
+                set {
+                }
+            }
```
rchiodo, line 414: Another merge problem. Not sure why this didn't come out in the initial build.

rchiodo:
I'm going to log a bug on the VS SDK for them to implement this. This should be part of the normal handling of an LSP

Python/Product/PythonTools/PythonTools/Editor/EditorExtensions.cs
```diff
@@ -64,6 +65,59 @@ internal static class EditorExtensions {
...
+
+            var startSnapshotLine = snapshot.GetLineFromLineNumber(startLine);
+            int startCharacterValue = startCharacter.HasValue ? startCharacter.Value : startSnapshotLine.Length;
+
+            // This should be >, not >=. Lines have one more position than characters, and length gives the number of characters.
+            if (startCharacterValue > startSnapshotLine.Length) {
+                return new SnapshotSpan(snapshot, new Span());
+            }
+
+            var startSnapshotPoint = startSnapshotLine.Start.Add(startCharacterValue);
```
bschnurr, line 99: do we need to worry about negative numbers?

bschnurr (approved):

Python/Product/PythonTools/PythonTools/Editor/EditorExtensions.cs
rchiodo: For startCharacterValue? Could do type change to force it to be positive.

rchiodo: uint instead of int

bschnurr: yes 'startCharacterValue'

rchiodo: Actually uint doesn't work because all of the types everywhere use int. If it was negative it would just be lost in a typecast anyway.

I'll add a check instead.

PR #6497, opened 2021-04-29
Support auto complete in interactive window

rchiodo:
Fixes: https://github.com/microsoft/PTVS-Pylance/issues/76

Implement auto complete in the interactive window.

Most of this code I copied from the VSLanguageServerClient repo because this is the same thing they do for normal python files just with a projection buffer instead of a normal ITextBuffer.

I logged a bug against them to implement projection buffer support.


Python/Product/PythonTools/PythonTools/Commands/CurrentEnvironmentCommand.cs
```diff
@@ -32,29 +32,28 @@
 namespace Microsoft.PythonTools.Commands {
     class CurrentEnvironmentCommand : OleMenuCommand {
         private readonly IServiceProvider _serviceProvider;
-        private readonly EnvironmentSwitcherManager _envSwitchMgr;
-        private readonly IPythonToolsLogger _logger;
+        private EnvironmentSwitcherManager _envSwitchMgr = null;
```
rchiodo, line 35: Refactored this a bit as the GetPythonToolsService was taking a long time in package load.

Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/StreamIntercepter.cs
```diff
@@ -52,8 +52,16 @@ class StreamIntercepter : Stream {
         public override long Seek(long offset, SeekOrigin origin) => baseStream.Seek(offset, origin);
         public override void SetLength(long value) => baseStream.SetLength(value);
         public override void Write(byte[] buffer, int offset, int count) {
-            var writeHandlerResult = writeHandler.Invoke(new StreamData{ bytes = buffer, offset = offset, count = count });
-            baseStream.Write(writeHandlerResult.bytes, writeHandlerResult.offset, writeHandlerResult.count);
+            if (writeHandler != null) {
```
rchiodo, line 55: Refactored the stream intercepting to skip intercepting if already handled the appropriate message. This speeds up intellisense a little bit as we don't deserialize every message twice (once in our stuff and once in the VS SDK bits)

Python/Product/PythonTools/PythonTools/Repl/Completion/AsyncCompletionSource.cs
```diff
@@ -0,0 +1,655 @@
...
+using ServiceProvider = Microsoft.VisualStudio.Shell.ServiceProvider;
+using Microsoft.VisualStudio.LanguageServer.Client;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.PythonTools.Infrastructure;
+using Microsoft.VisualStudio.InteractiveWindow;
+using Microsoft.VisualStudio.Text.Projection;
+
+namespace Microsoft.PythonTools.Repl.Completion {
+    /// <summary>
+    /// This class implements auto complete for the REPL windows
```
rchiodo, line 46: The majority of this code I copied from the VS platform bits. I'm thinking I might want to just fix the bug I logged against them instead.

rchiodo: Going to look into how hard that would be.

rchiodo: Okay never mind. This is better. Their code uses the main TextBuffer all over the place and even has comments saying they don't support projection buffers.

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
```diff
@@ -34,10 +34,10 @@ class Listener : IDisposable {
                 workspaceService.CurrentWorkspace.GetService<IFileWatcherService>().OnFileSystemChanged += OnFileChanged;
                 _root = workspaceService.CurrentWorkspace.Location;
             } else {
-                var dte = (EnvDTE80.DTE2)site.GetService(typeof(EnvDTE.DTE));
-                if (dte != null && dte.Solution != null && dte.Solution.FileName != null) {
+                var path = GetSolutionDirectory(site);
```
bschnurr, line 37: much nicer

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -256,21 +264,27 @@ public Task InvokeDidChangeConfigurationAsync(LSP.DidChangeConfigurationParams r
             await SendDocumentOpensAsync();
         }
 
-        public Task<LSP.CompletionList> InvokeTextDocumentCompletionAsync(LSP.CompletionParams request, CancellationToken cancellationToken = default)
-            => _rpc == null ? Task.FromResult(new LSP.CompletionList()) : _rpc.InvokeWithParameterObjectAsync<LSP.CompletionList>("textDocument/completion", request, cancellationToken);
+        public Task<object> InvokeTextDocumentCompletionAsync(LSP.CompletionParams request, CancellationToken cancellationToken = default)
+            => _rpc == null ? Task.FromResult<object>(null) : _rpc.InvokeWithParameterObjectAsync<object>("textDocument/completion", request, cancellationToken);
```
bschnurr, line 268: how come we dont know the result type anymore?

rchiodo: The result type based on the spec is not LSP.CompletionList. It can be one of LSP.CompletionList or LSP.CompletionItem[] or null.

rchiodo: Exact text of the spec:

```
Response:

result: CompletionItem[] | CompletionList | null. If a CompletionItem[] is provided it is interpreted to be complete. So it is the same as { isIncomplete: false, items }
```

rchiodo: And since C# doesn't support type unions, it has to be an object.

Python/Product/PythonTools/PythonTools/Repl/Completion/AsyncCompletionSourceProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+        public IAsyncCompletionBroker EditorCompletionBroker { get; set; } = null;
+
+        public IAsyncCompletionSource GetOrCreate(ITextView textView) {
+            AsyncCompletionSource source = null;
+
+            // First make sure this is for a REPL text view. One of the buffers has to be a repl buffer
+            var matches = textView.BufferGraph.GetTextBuffers((b) => b.IsReplBuffer());
+            if (matches.Count > 0) {
+                // Then make sure we have a language client to use.
+                var service = ServiceProvider.GetService(typeof(PythonToolsService)) as PythonToolsService;
```
bschnurr, line 52: might want to add a check on ServiceProvider

Python/Product/PythonTools/PythonTools/Repl/Completion/MarkdownUtil.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Microsoft.PythonTools.Repl.Completion {
+    internal static class MarkdownUtil {
+        // This regex matches markdown code blocks. It extracts the content into two groups:
+        // the language name and the code content.
+        private const string CodeBlockRegex = "^```(.*)[\\S\\s]^([\\S\\s]*?)[\\S\\s]^```";
```
bschnurr, line 12: i forget if its possible in C# to pre compile these regex strings

Python/Product/PythonTools/PythonTools/Repl/Completion/AsyncCompletionSourceProvider.cs
rchiodo: MEF would fail if service provder wasn't there. Meaning this class would never get called (because its imports wouldn't be met). Or did you mean check the PythonToolsService to make sure it's returned?

Python/Product/PythonTools/PythonTools/Repl/Completion/MarkdownUtil.cs
rchiodo: Yeah it is. I'll change it.

bschnurr (approved):

PR #6503, opened 2021-05-06
Fix some more tests, disable 27

rchiodo:


Python/Product/PythonTools/PythonTools/Commands/ExecuteInReplCommand.cs
```diff
@@ -211,7 +211,7 @@ internal sealed class ExecuteInReplCommand : Command {
             var eval = (IPythonInteractiveEvaluator)window.InteractiveWindow.Evaluator;
 
             // The interpreter may take some time to startup, do this off the UI thread.
-            await ThreadHelper.JoinableTaskFactory.RunAsync(async () => {
+            await _serviceProvider.GetPythonToolsService().UIThread.InvokeAsync(async () => {
```
bschnurr, line 214: what's the benefit here?

Python/Product/TestAdapter/testlauncher.py
```diff
@@ -177,7 +177,8 @@ def collectwapper():
             notfound = getattr(collector, '_notfound', [])
             if notfound:
                   for arg, exc in notfound: 
-                      line = "(no name {!r} in any of {!r})".format(arg, exc.args[0])
+                      arg_zero = exc[0] if isinstance(exc, list) else exc.args[0]
+                      line = "(no name {!r} in any of {!r})".format(arg, arg_zero)
```
bschnurr, line 181: "exc[0] if isinstance(exc, list) else exc.args[0]"  wow this looks like python

Python/Product/PythonTools/PythonTools/Commands/ExecuteInReplCommand.cs
rchiodo: ThreadHelper library isn't available during unit tests. It was causing exceptions when running.

rchiodo: PythonToolsService.UIThread is using a MockUIThread during testing.

Python/Product/TestAdapter/testlauncher.py
rchiodo: It's a python file? Were you commenting on how different python looks than C#?

bschnurr (approved):

PR #6514, opened 2021-05-20
Get symbols archived for Parsing/Core dlls

rchiodo:
And add 17 theme file


Python/Product/PythonTools/Theme/make_theme_pkgdef.py
```diff
@@ -13,7 +13,7 @@
     'CT_SYSCOLOR': 3,
 }
 
-for VS_VERSION in (15, 16):
+for VS_VERSION in (15, 16, 17):
```
bschnurr, line 16: should we change these to v15, v16, v17

then also update `THEME_PKGDEF`

to make it easier to search in the future?

bschnurr (approved):

Python/Product/PythonTools/Theme/make_theme_pkgdef.py
rchiodo: I don't think that will work as they have to be integer/numbers below where it's comparing the version numbers.

PR #6515, opened 2021-05-20
VS interop should not be included in our VSIX

rchiodo:
This didn't matter when the VS Interop was the same version as was shipped in VS
I upgraded this morning and our VS interop is older and the Solution Explorer broke when our package loads (it was loading the VS interop from our extension folder first)
We don't need to ship this and just consuming the types should be fine.

### DIFF
--- a/Python/Prerequisites/deploy_test_targets.cmd
+++ b/Python/Prerequisites/deploy_test_targets.cmd
@@ -13,12 +13,9 @@ echo This script should be run as an administrator.
 
 set D=%~dp0
 
-rem Guess some directories for Visual Studio 16
+if NOT EXIST "%VSINSTALLDIR%" call :failure
+
 call :docopy "%VSINSTALLDIR%\MSBuild\Microsoft\VisualStudio\v17.0"
-if errorlevel 1 call :docopy "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Preview\MSBuild\Microsoft\VisualStudio\v16.0"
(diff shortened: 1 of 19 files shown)


Python/Prerequisites/deploy_test_targets.cmd
```diff
@@ -38,4 +35,21 @@ if not exist "%TARGET%" mkdir "%TARGET%"
 copy /Y "*.targets" "%TARGET%"
 popd
 
+set TARGET=%VSINSTALLDIR%\MSBuild\Microsoft\VC\v160\Platforms
```
rchiodo, line 38: Sorry forgot I was working on testing this at the same time. This isn't shipping code so should be fine to submit too.

Python/Product/Core/Core.csproj
```diff
@@ -132,11 +131,6 @@
     <None Include="source.extension.vsixmanifest" />
     <Content Include="License_en-US.rtf" />
     <Content Include="$(PackagesPath)\Newtonsoft.Json\lib\net45\Newtonsoft.Json.dll" />
-    <Content Include="$(PackagesPath)\Microsoft.VisualStudio.LanguageServer.Protocol\lib\netstandard2.0\Microsoft.VisualStudio.LanguageServer.Protocol.dll" />
```
bschnurr, line 135: Protocol.dll removed on purpose?

rchiodo: Sorry should have mentioned it too. We don't need to ship it anymore either.

bschnurr (approved):

PR #6516, opened 2021-05-20
Fix task dialog and problem with listener

rchiodo:
Task dialog wasn't setup to work with 64 bit. 

Listener had a bug on environment change.

### DIFF
--- a/Python/Product/VSCommon/Infrastructure/TaskDialog.cs
+++ b/Python/Product/VSCommon/Infrastructure/TaskDialog.cs
@@ -230,10 +230,10 @@ namespace Microsoft.PythonTools.Infrastructure {
                 config.pszFooter = Footer;
                 config.pszVerificationText = VerificationText;
                 config.pfCallback = Callback;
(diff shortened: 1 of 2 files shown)


Python/Product/VSCommon/Infrastructure/TaskDialog.cs
```diff
@@ -509,8 +513,141 @@ private static class NativeMethods {
             internal const int IDYES = 6;
             internal const int IDNO = 7;
             internal const int IDCLOSE = 8;
+            internal const int WM_USER = 0x0400;
+            internal static IntPtr NO_PARENT = IntPtr.Zero;
+
+            [DllImport(ExternDll.User32, CharSet = CharSet.Auto, SetLastError = true)]
```
rchiodo, line 519: I copied this from other code in VS that is also using the task dialog.

rchiodo:
@bschnurr can you sign off on this when you get the chance? I'd like to insert this with my next change to VS

PR #6519, opened 2021-05-24
Support mixed mode debugging in 3.9

rchiodo:
I have yet to test 3.8, but this gets 3.9 working. I hope. This code is super fragile. Seems like there has to be a better way to do this stuff.

If CPython changes anything (which they did in 3.9), none of this will work.

Update: Got 3.8 to work too.


Python/Product/DebuggerHelper/trace.cpp
```diff
@@ -594,4 +594,17 @@ void *EvalFrameFunc(void* f, int throwFlag)
     return nullptr;
 }
 
+typedef void* (*_PyFrameEvalFunction_39)(void*,void*, int);
```
rchiodo, line 597: I should put a comment here saying why this is different.

bschnurr (approved):

Python/Product/DebuggerHelper/trace.cpp
bschnurr: would help, yes

int19h:
These kinds of changes in CPython is why there's a whole declarative framework here to select the correct proxy implementation. There's no way around it - we can't run Python code here (because thread can be stopped in native with GIL held, for example), and so we have to directly interpret the runtime internal data structures. We already use PDBs to determine field offsets, at least, so when they add new fields to existing structs, it keeps working. But if the structure changes fundamentally, there's nothing we can do about that other than add another branch.

So, every major CPython release is potentially breaking, if it touches the internals that we depend on.

rchiodo:
AFAICT, all of this code is really trying to write to the correct locations in the CPython runtime to enable tracing? Could this be done with an API that CPython exports from the dll?

int19h:
The bit that is touched in this particular case, yes - but the same can also happen whenever e.g. dict or str representations change.

Python C API in general is not actually a stable ABI across major version numbers, precisely because they expose struct fields directly like this. There's a subset called "limited API" that doesn't do that, treating everything as opaque handles with accessor functions to manage state; but it's much more restricted, and specifically doesn't cover advanced functionality that we need such as tracing. 

The other problem is that most of Python C API require the caller to hold the global interpreter lock. So in general case, when debugging is paused somewhere in native code that holds the GIL (possibly in Python VM itself), we can't invoke them without deadlocking the process. For example, reading a value from a list or dict via Python requires GIL, because the refcount for the returned object is incremented by the getter. I can't think any way to solve this problem. short of CPython exporting non-refcount-updating versions of all read operations for all the data structures that we need, specifically for us.

PR #6525, opened 2021-05-27
Support the 'workspace/configuration' message from Pylance.

rchiodo:
Change to responding to 'workspace/configuration' so that we can have a different interpreter per project.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
@@ -173,7 +173,76 @@ namespace Microsoft.PythonTools.LanguageServerClient {
             customTarget.WatchedFilesRegistered += WatchedFilesRegistered;
             customTarget.WorkspaceFolderChangeRegistered += OnWorkspaceFolderWatched;
             customTarget.AnalysisComplete += OnAnalysisComplete;
+            customTarget.WorkspaceConfiguration += OnWorkspaceConfiguration;
             await StartAsync.InvokeAsync(this, EventArgs.Empty);
+        }
+
+        private async Task OnWorkspaceConfiguration(object sender, WorkspaceConfiguration.ConfigurationArgs args) {
+            // This is where we send the settings for each client context.
+            // Pylance will send a workspace/configuration request for each workspace
+            // We return the language server settings as a response
+            object[] result = new object[0] { };
+
+            // Return the matching results
+            foreach (var item in args.requestParams.items) {
+                // Find the matching context for the item
+                var context = _clientContexts.Find(c => item.scopeUri != null && PathUtils.IsSamePath(c.RootPath, item.scopeUri.LocalPath));
+                if (context == null)
+                    continue;
+
+                if (context is PythonLanguageClientContextProject) {
+                    // Project interactions are main thread only.
+                    await JoinableTaskContext.Factory.SwitchToMainThreadAsync();
+                }
+
+                object settings = null;
+                if (item.section == "python") {
+                    Debug.Assert(_analysisOptions != null);
+
+                    var extraPaths = UserSettings.GetStringSetting(
+                        PythonConstants.ExtraPathsSetting, null, Site, PythonWorkspaceContextProvider.Workspace, out _)?.Split(';')
+                        ?? _analysisOptions.ExtraPaths;
+
+                    var stubPath = UserSettings.GetStringSetting(
+                        PythonConstants.StubPathSetting, null, Site, PythonWorkspaceContextProvider.Workspace, out _)
+                        ?? _analysisOptions.StubPath;
+
+                    var typeCheckingMode = UserSettings.GetStringSetting(
+                        PythonConstants.TypeCheckingModeSetting, null, Site, PythonWorkspaceContextProvider.Workspace, out _)
+                        ?? _analysisOptions.TypeCheckingMode;
+
+                    var ver3 = new Version(3, 0);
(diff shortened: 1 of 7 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -173,9 +173,78 @@ internal sealed class PythonLanguageClient : ILanguageClient, ILanguageClientCus
...
+                            typeCheckingMode = typeCheckingMode,
+                            useLibraryCodeForTypes = true,
+                            completeFunctionParens = _advancedEditorOptions.CompleteFunctionParens,
+                            autoImportCompletions = _advancedEditorOptions.AutoImportCompletions
+                        }
+                    };
+                }
+
+                // Copy settings in order
+                var resultArray = new object[result.Length + 1];
```
rchiodo, line 238: I'm going to change this to use a list and then convert the list to an array at the end.

jakebailey:
IIRC there's an initialization option that needs to be set to declare that the client supports the call; is that in this PR? I did a look but didn't find it, so wondering if it's working regardless.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -472,6 +485,7 @@ public Task<object> InvokeCommandAsync(LSP.ExecuteCommandParams request, Cancell
                                 }
                                 capabilities["workspace"]["workspaceFolders"] = true;
                                 capabilities["workspace"]["didChangeWatchedFiles"]["dynamicRegistration"] = true;
+                                capabilities["workspace"]["configuration"] = true;
```
rchiodo, line 485: Right here @jakebailey

rchiodo: That's where the setting is turned on.

jakebailey: Thanks, clearly I had the wrong search terms.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -173,9 +173,75 @@ internal sealed class PythonLanguageClient : ILanguageClient, ILanguageClientCus
...
+                    var typeCheckingMode = UserSettings.GetStringSetting(
+                        PythonConstants.TypeCheckingModeSetting, null, Site, PythonWorkspaceContextProvider.Workspace, out _)
+                        ?? _analysisOptions.TypeCheckingMode;
+
+                    var ver3 = new Version(3, 0);
+                    if (context.InterpreterConfiguration.Version < ver3) {
+                        MessageBox.ShowWarningMessage(Site, Strings.WarningPython2NotSupported);
+                    }
+
+                    settings = new LanguageServerSettings.PythonSettings {
```
jakebailey, line 219: Do you know if PYTHONPATH is supported? There's another config that could be added there if so.

int19h (approved):

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -173,9 +173,75 @@ internal sealed class PythonLanguageClient : ILanguageClient, ILanguageClientCus
...
+                        pythonPath = context.InterpreterConfiguration.InterpreterPath,
```
rchiodo, line 220: This one right here? Or did you mean something else?

rchiodo: Did you mean this one?

```_envPYTHONPATH```

Comment (rchiodo) [reply Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs:220]: Yes I think I can add that one too. I'll do _envPYTHONPATH

Comment (rchiodo) [reply Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs:220]: Sorry it doesn't have that information. It just has the name of the 'PYTHONPATH' environment variable but not the actual values.

Comment (rchiodo) [reply Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs:220]: Is it necessary for pylance to work correctly? I think I might be able to run the environment to extract it, but that's a lot more code.

Comment (jakebailey) [reply Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs:220]: No, it's not necessary to run. If VS doesn't provide a way to configure this, then there's no reason to try and fetch it. VSC does, so I piped that down.

File: Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs [hunk 5]
```diff
@@ -298,17 +361,15 @@ public Task<object> InvokeCommandAsync(LSP.ExecuteCommandParams request, Cancell
             }
         }
 
-        private void OnSettingsChanged(object sender, EventArgs e) => SendDidChangeConfigurations().DoNotWait();
+        private void OnSettingsChanged(object sender, EventArgs e) => InvokeDidChangeConfigurationAsync(new LSP.DidChangeConfigurationParams() {
+            Settings = null
```
rchiodo, line 365: I think I should put a comment here indicating why this is null.

Issue #6537, opened 2021-05-28
Cookiecutter Explorer is missing 


ttSpace:
**Describe the bug**

![image](https://user-images.githubusercontent.com/82852072/119957363-98a73900-bfd4-11eb-905e-31d3d9dfc916.png)

**Steps to Reproduce**

Launch VS, select View > Cookiecutter Explorer
![image](https://user-images.githubusercontent.com/82852072/119957861-0ce1dc80-bfd5-11eb-9467-e61ad8bab5be.png)

rchiodo:
Seems this works when debugging but not during the installation.

PR #6553, opened 2021-06-02
Add back in the building of the cookie cutter vsix

rchiodo:
Fixes: #6537 

CookieCutter vsix was not being generated.

### DIFF
--- a/Python/Setup/loclayout.proj
+++ b/Python/Setup/loclayout.proj
@@ -23,7 +23,6 @@
         Microsoft.PythonTools.Debugger.Concord.dll;
         Microsoft.PythonTools.dll;
         Microsoft.PythonTools.EnvironmentsList.dll;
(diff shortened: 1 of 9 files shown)


Python/Setup/loclayout.proj
```diff
@@ -23,6 +23,7 @@
         Microsoft.PythonTools.Debugger.Concord.dll;
         Microsoft.PythonTools.dll;
         Microsoft.PythonTools.EnvironmentsList.dll;
+        Microsoft.PythonTools.LiveShare.dll;
```
rchiodo, line 26: Actually I should remove these. Live share isn't needed anymore with pylance as VS will handle the negotiations for it.

rchiodo:
@bschnurr @int19h can you guys take a look?

bschnurr (approved):

PR #6560, opened 2021-06-07
Fix building with D17

rchiodo:
Seems like references can't be found from the actual build environment. We have to explicitly download all of our references now.

This PR adds the missing modules so that you can build locally.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientCustomTarget.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientCustomTarget.cs
@@ -163,13 +163,17 @@ namespace Microsoft.PythonTools.LanguageServerClient {
 
         [JsonRpcMethod("workspace/configuration")]
         public async Task<object> OnWorkspaceConfiguration(JToken arg) {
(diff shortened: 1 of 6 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClientCustomTarget.cs
```diff
@@ -163,13 +163,17 @@ private sealed class PylanceException : Exception {
 
         [JsonRpcMethod("workspace/configuration")]
         public async Task<object> OnWorkspaceConfiguration(JToken arg) {
-            var reqParams = arg.ToObject<ConfigurationParams>();
-            if (this.WorkspaceConfiguration != null && reqParams != null) {
-                var eventArgs = new ConfigurationArgs { requestParams = reqParams, requestResult = null };
-                await this.WorkspaceConfiguration.InvokeAsync(this, eventArgs);
-                return eventArgs.requestResult;
+            try {
+                var reqParams = arg.ToObject<ConfigurationParams>();
```
rchiodo, line 167: Noticed this on shutdown. It was throwing an exception on conversion.

bschnurr (approved):

PR #6571, opened 2021-06-11
Add conversion step for python.parsing pdbs

rchiodo:
Hopefully this will fix the symbol publish error here:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1331578/

The root cause (or at least one of the causes) was that the pdbs for the Microsoft.Python.Parsing dlls were portable pdbs. The symbol archiver in the build doesn't support portable, only windows pdbs.


bschnurr:
wow nice.

bschnurr (approved):

bschnurr (approved):

PR #6585, opened 2021-06-28
Implement attach with debugpy

rchiodo:
Fixes: https://github.com/microsoft/PTVS-Pylance/issues/152

Support using debugpy to do local attach too. 

I also fixed up the debug repl to work with debugpy, but it seems it doesn't do anything you can't do with the debugger's immediate window. After some discussion with @gramster and @judej, we're likely going to cut the debug repl. I want to submit what I have here because it works but then I'll submit a follow up that just deletes all of this code. This way if we change our mind I can just revert the delete.


rchiodo:
@karthiknadig can you take a look at the attach stuff I did? I believe you did remote attach for debugpy in ptvs?

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterAttachProcess.cs
```diff
@@ -0,0 +1,194 @@
...
+using System.Net.Sockets;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Logging;
+using Microsoft.VisualStudio.Debugger.DebugAdapterHost.Interfaces;
+using Microsoft.VisualStudio.Debugger.Interop;
+
+namespace Microsoft.PythonTools.Debugger {
+    internal sealed class DebugAdapterAttachProcess : ITargetHostProcess, IDisposable {
```
rchiodo, line 30: This used to be called RemoteAdapterAttachProcess, not sure why it didn't pick up the rename.

Python/Product/PythonTools/PythonTools/Debugger/DebuggerHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+using Microsoft.VisualStudio.Debugger.Interop;
+using Microsoft.VisualStudio.InteractiveWindow;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.PythonTools.Debugger {
+    /// <summary>
+    /// Helper class to expose IDebug* variables for use elsewhere (instead of using EnvDTE.Debugger)
+    /// </summary>
+    public class DebuggerHelper: IDebugEventCallback2 {
```
rchiodo, line 20: This entire file should go away once I remove the debug interactive repl. It reimplements all of the PythonProcess/Thread stuff.

Python/Product/PythonTools/PythonTools/Extensions.cs
```diff
@@ -746,15 +746,15 @@ class ShellInitializedNotification : IVsShellPropertyEvents {
             if (debugger == null) {
                 return;
             }
-            AD7Engine engine = AD7Engine.GetEngineForProcess(debugger.CurrentProcess);
-            if (engine != null) {
-                await engine.RefreshThreadFrames(debugger.CurrentThread.ID, ct);
+            //AD7Engine engine = AD7Engine.GetEngineForProcess(debugger.CurrentProcess);
```
rchiodo, line 749: I should remove the commented code here.

rchiodo: Or maybe verify the current debugger is python.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterAttachProcess.cs
rchiodo: @karthiknadig I was hoping this is where you could take a look. This file and the other ones you wrote were modified slightly to support local attach.

rchiodo: Then the AD7Engine class was modified to just be a pass through to the VsDebugAdapterHost engine that we use for launch.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterAttachProcess.cs
```diff
@@ -0,0 +1,203 @@
...
+            Debug.WriteLine($"DebugPy exited:");
+            Exited?.Invoke(this, EventArgs.Empty);
+        }
+
+        public static ITargetHostProcess LocalAttach(DebugAttachInfo debugAttachInfo, IAdapterLaunchInfo launchInfo, ITargetHostInterop targetInterop) {
+            // For the local case, we need to start the process ourselves
+            AD_PROCESS_ID adProcessId = new AD_PROCESS_ID();
+            adProcessId.ProcessIdType = (uint)enum_AD_PROCESS_ID.AD_PROCESS_ID_SYSTEM;
+            adProcessId.dwProcessId = (uint)launchInfo.AttachProcessId;
+            launchInfo.DebugPort.GetProcess(adProcessId, out var process);
```
bschnurr, line 59: does GetProcess return something like success/fail or error?

rchiodo: It returns an HR result. All of the debugger APIs are actually COM calls.

However if you have a port, you always have a process. It's exceptional (hence an exception) if there isn't one.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterAttachProcess.cs
```diff
@@ -0,0 +1,203 @@
...
+using Microsoft.PythonTools.Debugger.DebugAdapter;
+using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Logging;
+using Microsoft.VisualStudio.Debugger.DebugAdapterHost.Interfaces;
+using Microsoft.VisualStudio.Debugger.Interop;
+
+namespace Microsoft.PythonTools.Debugger {
+    internal sealed class DebugAdapterAttachProcess : ITargetHostProcess, IDisposable {
+        private const int _debuggerConnectionTimeout = 60000; // 60 seconds. Can take a long time to connect
+        private DebugAdapterAttachProcessStream _stream;
```
bschnurr, line 34: should we set `_stream` to `null` here?

rchiodo: Default of any reference type should be null:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values

bschnurr (approved):

rchiodo:
@int19h if you have a chance to look at this, that would be great. This is only the start of debugpy for attach though. I still need a way to attach if there's no python interpreter. Investigating using [PyInstaller](https://pyinstaller.readthedocs.io/en/stable/usage.html)

int19h:
I'll take a look tomorrow.

Regarding PyInstaller, I wonder if it'd actually save much space. It's basically just packaging Python itself and those parts of stdlib that are needed to run things, but pydevd has plenty of dependencies. It might be easier to just package a full-fledged local Python install - and it might come in handy later.

We could also consider re-implementing this code path in C++ and drop the Python dependency. The meat of it is [this file](https://github.com/microsoft/debugpy/blob/main/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/add_code_to_python_process.py), and since this is for PTVS, we'd only need to do it for Win32-specific branches. The code in debugpy that invokes it is much simpler.

rchiodo:
> We could also consider re-implementing this code path in C++ and drop the Python dependency. The meat of it is this file, and since this is for PTVS, we'd only need to do it for Win32-specific branches. The code in debugpy that invokes it is much simpler.

That sounds interesting. Can the build for debugpy generate C++ code? I assume so since it's injecting a dll (although maybe the dll was prebuilt).

We'd need both 32 and 64 bit. I'll enter your feedback into the

int19h:
debugpy builds via setup.py, which has some functionality to handle simple C libraries (`distutils.command.build_clib`). I believe this should be sufficient for injection purposes.

I forgot that it's not just the injection, though. The injected code will run using the Python interpreter already loaded by the target process - but it still needs to be able to spawn a separate adapter process, and it needs python.exe for that. There's no reliable way to get it from within the debuggee in all scenarios - in hosted ones (where this feature is more commonly used), `sys.executable` will return the actual .exe of the hosting process, and python.exe might not even be physically present.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterAttachProcess.cs
int19h: If it's an HRESULT return from managed perspective, it should still check - I think there was a `RequireOk` helper method somewhere?

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/DebugAdapterAttachProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// <summary>
+        /// Gets an available port on the loopback interface.
+        /// </summary>
+        private static int GetAvailablePort() {
+            TcpListener l = new TcpListener(IPAddress.Loopback, 0);
+            l.Start();
+            int port = ((IPEndPoint)l.LocalEndpoint).Port;
+            l.Stop();
+            return port;
+        }
```
int19h, line 97: This is not a reliable way to do this, because the port can be taken by some other process as soon as you close the listener socket.

It would be better to implement this using `debugpy --connect` to invert the direction, so that you can just keep the socket open on VS side, and meanwhile tell debugpy where to connect.

Alternatively, when using `--listen`, you can tell debugpy to listen on :0, and get the actual port number by setting the environment variable `DEBUGPY_ADAPTER_ENDPOINTS` to some temp filename. When adapter starts listening for client connections, it'll write a JSON snippet into that file which has the actual host & port being listened on. This is still not ideal, though, because you have to poll for that file in a loop - it was really added mostly for testing purposes where that's okay.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/ProcessExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            process.Exited += ExitAction;
+
+            try {
+                if (process.HasExited) {
+                    return true;
+                }
+
+                if (await Task.WhenAny(tcs.Task, Task.Delay(Timeout.Infinite, cancellationToken)) != tcs.Task) {
+                    process.Exited -= ExitAction;
+                    process.Kill();
```
int19h, line 32: This can throw in many weird corner cases (permissions, zombie processes etc). It seems like the intent of this method is for the caller to not worry about such details, so it should probably catch & swallow exceptions from `Kill()`.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/ProcessExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                    if (cancellationToken.IsCancellationRequested) {
+                        tcs.TrySetCanceled();
+                    } else {
+                        // Timeout is reached so set the result to false.
+                        tcs.TrySetResult(false);
+                    }
+                }
+
+                // Re-await the task so that any exceptions/cancellation are rethrown.
```
int19h, line 42: I'm not quite sure how this is supposed to work. If the returned task is faulted or canceled, the caller will find out as soon as it awaits on it, no? It doesn't seem like re-wrapping it would do anything differently.

Python/Product/PythonTools/PythonToolsPackage.cs
```diff
@@ -469,6 +469,9 @@ protected override object GetAutomationObject(string name)
             // test discoverer and test executor to connect back to VS.
             Environment.SetEnvironmentVariable("_PTVS_PID", Process.GetCurrentProcess().Id.ToString());
 
+            // Setup the debug watcher so we can query for IDebug types
+            var x = DebuggerHelper.Instance;
```
int19h, line 473: This can use `_` to indicate that the returned value is ignored.

But, better yet, since it relies on side effects from the getter, I think `Instance` should become `GetInstance()` instead.

Python/Product/VSCommon/Infrastructure/ComDisposer.cs
```diff
@@ -0,0 +1,155 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.PythonTools.Infrastructure {
+    internal abstract class ComDisposerBase : IDisposable {
+        protected System.Collections.Generic.List<object> ObjectList { get; set; }
```
int19h, line 10: Full qualification shouldn't be needed here.

Python/Product/VSCommon/Infrastructure/ComDisposer.cs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        protected override void DisposeObject(object o) {
+            try {
+                if (o is IntPtr) {
+                    ReleaseComReference((IntPtr)o);
+                }
+
+                ReleaseComObject(o);
+            } catch { System.Diagnostics.Debug.Assert(false, "Could not release COM reference"); } // Dispose should never throw.
```
int19h, line 138: `Debug.Fail()`

Python/Product/VSCommon/Infrastructure/ComDisposer.cs
int19h: Also, I think this should be a set, since otherwise an object added twice will also be disposed twice, which is never the right thing to do.

Python/Product/PythonTools/PythonTools/Debugger/DebugAdapter/ProcessExtensions.cs
rchiodo: The initial WhenAny can exit on the cancellation token. I don't believe that will throw an exception if the cancellation token is canceled? Actually no it will throw a TaskCanceledExceptoin (according to MSDN docs for Task.Delay)

I didn't actually write this code. Copied it from the diag hub team I believe.

I agree the second await seems unnecessary. I'll also add the try/catch around the kill.

rchiodo:
Closing this after talking with Pavel. We may implement debugpy attach a different way.

Issue #6597, opened 2021-07-07
vs code 19 issue


HRIDAY-SHINDE:
**Describe the bug**
[A clear and concise description of the bug]

**Steps to Reproduce**
1.
2.
3.

**Expected behavior**
[A clear and concise description of your expectation]

**Additional context and screenshots**
[Add any other information such as screenshots, error message, or stack trace]

**Sample Code**
```
XXX
```

**Configuration information (If you are providing a diagnostics file (see below), skip this section)**
VS Version: [Can be found under "Help->About Microsoft Visual Studio"]
PTVS version: [Can be found under "Help->About Microsoft Visual Studio->Installed products"]
Python version:
Debugger Type (if applicable): [New debugger or legacy debugger?]

**Diagnostics file**
The diagnostics file contains information such as the Python version, environments, loaded assemblies, configuration options, etc. This file makes it easier for us to reproduce and fix bugs, so we would really appreciate it if you could provide it to us. This file can be generated by going to Tools->Options->Python->Diagnostics.
You can attach the file to this issue or email it to us at ptvshelp@microsoft.com. If you email it, please add a link to this issue.
```
XXX
```
[Window Title]
devenv.exe

[Main Instruction]
An unexpected error occurred

[Content]
Please press Ctrl+C to copy the contents of this dialog and report this error to our issue tracker.

[V] Show details  [Close]

[Expanded Information]
```
Build: 16.10.21111.19

System.InvalidOperationException: Operation cannot be started because pip is not installed in the environment at 'D:\'.
   at Microsoft.PythonTools.Interpreter.PipPackageManager.<AbortIfNotReady>d__27.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PythonTools.Interpreter.PipPackageManager.<InstallAsync>d__39.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PythonTools.Project.TestFrameworkInfoBar.<InstallPyTestAsync>d__4.MoveNext()
[...]

Issue #6582, opened 2021-06-26
Error while opening MVS


iamaryan-ag:
Error while opening MVS

**Steps to Reproduce**
I simply just opened Microsoft Visual Studio via my applications folder

**Expected behavior**
No error

**Additional context and screenshots**
[Window Title]
devenv.exe

[Main Instruction]
An unexpected error occurred

[Content]
Please press Ctrl+C to copy the contents of this dialog and report this error to our issue tracker.

[V] Show details  [Close]
![image](https://user-images.githubusercontent.com/75726601/123502318-24aa9000-d669-11eb-9913-87a8497f4f52.png)

[Expanded Information]
```
Build: 16.10.21111.19

System.UnauthorizedAccessException: Access to the path 'C:\Users\iamar\Documents\My Music' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileSystemEnumerableIterator`1.CommonInit()
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler, Boolean checkHost)
   at System.IO.Directory.EnumerateFiles(String path, String searchPattern, SearchOption searchOption)
   at Microsoft.PythonTools.Interpreter.PythonWorkspaceContext.<EnumerateUserFiles>d__59.MoveNext()
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source)
   at Microsoft.PythonTools.Project.TestFrameworkWorkspaceInfoBar.PythonTestFileFound(Predicate`1 fileFilter)
   at Microsoft.PythonTools.Project.TestFrameworkInfoBar.<CheckAsync>d__2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PythonTools.Project.TestFrameworkWorkspaceInfoBar.<CheckAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #6604, opened 2021-07-13
Handle different exceptions that can occur and cause VS crashes

rchiodo:
Fixes: #6597
Fixes: #6582

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/ConfigureTestFrameworkInfoBar.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/ConfigureTestFrameworkInfoBar.cs
@@ -112,16 +112,21 @@ namespace Microsoft.PythonTools.Project {
         private async Task<bool> InstallPyTestAsync() {
             var packageManagers = _infoBarData.InterpreterOptionsService.GetPackageManagers(_infoBarData.InterpreterFactory);
 
-            foreach (var packageManager in packageManagers) {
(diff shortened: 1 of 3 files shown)


Python/Product/PythonTools/PythonTools/Project/ConfigureTestFrameworkInfoBar.cs
```diff
@@ -112,16 +112,21 @@ internal abstract class TestFrameworkInfoBar : PythonInfoBar {
...
-                    new PackageSpec("pytest"),
-                    new VsPackageManagerUI(Site),
-                    CancellationToken.None
-                );
-
-                if (pytestInstalled) {
-                    return true;
+            try {
+                foreach (var packageManager in packageManagers) {
+                    bool pytestInstalled = await packageManager.InstallAsync(
```
bschnurr, line 117: do we want the catch inside the loop, so that it still continues?

bschnurr (approved):

Python/Product/PythonTools/PythonTools/Project/ConfigureTestFrameworkInfoBar.cs
rchiodo: Yeah why not. It exits on the first one that succeeds. It's more complicated logic though. If it's the only package manager, it needs to output the error. Otherwise succeed. I think I can just save the exception in that case.

Issue #6609, opened 2021-07-16
ILanguageClient is changing - need to update


rchiodo:
On july 23rd, VSLanguageClient is changing to add two new methods to ILanguageClient. We need to update our packages.config to pick up the new bits.

rchiodo:
![image](https://user-images.githubusercontent.com/19672699/125989980-b1cbf049-3871-411c-a15e-638c01573769.png)

PR #6612, opened 2021-07-20
Add support for ILanguageClient new methods

rchiodo:
ILanguageClient is changing (well because it can) and we need to change with it.

Fixes: #6609 

Tested with bits here:
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/results?buildId=4998492&view=ms.vss-build-web.run-extensions-tab

However we need to wait for that submission to go into main (or put this submission in at the same time) before we finish the PR into VS main.

Miguel should provide a branch we can insert into after Wednesday.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/DidChangeWatchedFilesRegistrationOptions.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/DidChangeWatchedFilesRegistrationOptions.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.PythonTools.LanguageServerClient.FileWatcher {
+    /// <summary>
+    /// Class representing the options for registering workspace/didChangeWatchedFiles support.
+    /// </summary>
+    [DataContract]
+    public class DidChangeWatchedFilesRegistrationOptions {
+        /// <summary>
+        /// Gets or sets the watchers that should be registered.
+        /// </summary>
+        [DataMember]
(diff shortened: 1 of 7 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/DidChangeWatchedFilesRegistrationOptions.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
bschnurr, line 1: license header?

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/DidChangeWatchedFilesRegistrationOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Linq;
```
bschnurr, line 3: `using System.Linq;` actually used?

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/DidChangeWatchedFilesRegistrationOptions.cs
rchiodo: Good point. Will fix

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/DidChangeWatchedFilesRegistrationOptions.cs
rchiodo: Should be fixed now

bschnurr (approved):

PR #6643, opened 2021-08-03
Narrator fix. Modifying the InterpreterView to return both Name and InterpreterPath…

bschnurr:
… in the ToString() function so that narrator will say the complete strings showing to the user.

Previously we've dont similar fixes on ProjectView, one downside is now both Name and Path strings are shown in the drop box when collapsed

Method #2 suggestion here:
https://docs.microsoft.com/en-us/archive/blogs/winuiautomation/giving-your-xaml-element-an-accessible-name-part-3-other-interesting-ways
"2. Set an accessible name on a list item"

Internal bug 1362537

image

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
+++ b/Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
@@ -22,7 +22,6 @@ using System.Windows;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
 using Microsoft.PythonTools.Project;
-using Microsoft.PythonTools.Project.ImportWizard;
 
 namespace Microsoft.PythonTools.Environments {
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools/Environments/InterpreterView.cs
```diff
@@ -16,11 +16,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Windows;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
 using Microsoft.PythonTools.Project;
+using Microsoft.PythonTools.Project.ImportWizard;
```
bschnurr, line 25: dont think this is needed. will remove

rchiodo (approved):

PR #6658, opened 2021-08-06
fix for narrator in packages windows.

bschnurr:
Now forcing a selection of label "IsDefaultMessage" after button press, and triggering narrator to update in the code behind so that it reads "This is the default environment for new projects".

fix internal bug 1351252

Actual Result:​
Narrator/NVDA is not announcing the status (This is the default environment for new projects) which is available on screen on selecting 'Make this the default environment for new projects' button.

Expected Result:​
Narrator/NVDA should announce the status as "This is the default environment for new projects" after selecting 'Make this the default environment for new projects' button.

### DIFF
--- a/Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
+++ b/Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
@@ -56,7 +56,8 @@
                                    Visibility="Collapsed"
                                    Margin="6 9 6 6"
                                    Padding="0"
-                                   AutomationProperties.Name="{x:Static l:Resources.EnvironmentPathsExtensionIsDefaultLabel}">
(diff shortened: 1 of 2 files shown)


int19h (approved):

Python/Product/EnvironmentsList/EnvironmentPathsExtension.xaml
```diff
@@ -36,6 +36,7 @@
                                     HorizontalAlignment="Stretch"
                                     Command="{x:Static l:EnvironmentView.MakeGlobalDefault}"
```
rchiodo, line 37: Does the command still fire if you have a click event?

bschnurr: yes!

bschnurr: I was worried by that as well.

PR #6677, opened 2021-08-18
Respond to different screen widths and scaling for AddEnvironment dialog

rchiodo:
Second pass for:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1355277?src=WorkItemMention&src-action=artifact_link

Here's what it looks like at 200% scaling and 1280x800

image

1920x1080 and 200%
image

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
+++ b/Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
@@ -149,6 +149,7 @@
                         </Setter>
                     </Style>
                     <l:LandscapePortraitConverter x:Key="LandscapePortraitConverter" />
+                    <l:ScreenAdjustingConverter x:Key="ScreenAdjustingConverter" />
 
                 </ResourceDictionary>
             </ResourceDictionary.MergedDictionaries>
@@ -162,13 +163,13 @@
                 Text="{x:Static common:Strings.CondaNotFoundError}"
                 TextWrapping="Wrap"/>
         </Grid>
-        <ItemsControl 
+        <ScrollViewer VerticalScrollBarVisibility="Auto">
+
+            <ItemsControl 
             Style="{Binding 
                 RelativeSource={RelativeSource AncestorType=UserControl},
                 Converter={StaticResource LandscapePortraitConverter}, 
                 ConverterParameter=MainGridLandscape|MainGridPortrait}">
-            <ScrollViewer
-                VerticalScrollBarVisibility="Auto">
 
                 <StackPanel>
                     <Label
@@ -181,7 +182,10 @@
                         x:Name="ProjectComboBox"
                         Margin="0 0 0 17"
                         MinHeight="29"
-                        Width="249"
+                        Width="{Binding 
+                            RelativeSource={RelativeSource Mode=Self},
+                            Converter={StaticResource ScreenAdjustingConverter}, 
+                            ConverterParameter=249}"
                         HorizontalAlignment="Left"
                         ItemsSource="{Binding Path=Projects}"
                         SelectedItem="{Binding Path=SelectedProject}"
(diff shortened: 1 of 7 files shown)


Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
```diff
@@ -162,13 +163,13 @@
                 Text="{x:Static common:Strings.CondaNotFoundError}"
                 TextWrapping="Wrap"/>
         </Grid>
-        <ItemsControl 
+        <ScrollViewer VerticalScrollBarVisibility="Auto">
```
rchiodo, line 166: Moved the scroll viewer out of just the 'left' hand side controls. Now it will allow scrolling in those cases where the screen is too thin.

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
```diff
@@ -181,7 +182,10 @@
                         x:Name="ProjectComboBox"
                         Margin="0 0 0 17"
                         MinHeight="29"
-                        Width="249"
+                        Width="{Binding
```
rchiodo, line 185: Some of these comboboxes were ending up too wide in the maximum scaling situation (Screen is only 600px wide in that scenario)

Python/Product/PythonTools/PythonTools/Wpf/Utilities.cs
```diff
@@ -29,5 +31,13 @@ internal static class Utilities {
...
             return msg;
         }
+
+        /// <summary>
+        /// Determines if dialogs should display in landscape or portrait mode
+        /// </summary>
+        /// <remarks>
+        /// Picks portrait mode if screen is too small as well (with assumption scrolling will be used)
+        /// </remarks>
+        public static bool IsLandscape => SystemParameters.PrimaryScreenWidth > SystemParameters.PrimaryScreenHeight && SystemParameters.PrimaryScreenWidth > 800;
```
rchiodo, line 41: Forcing portrait mode when less than 800 pixels

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
bschnurr: i can't remember if there is an auto width based on percentage of parent

rchiodo: Not that I could find. All the examples I found referenced writing a converter

rchiodo:
1200x800 at 150%
image

rchiodo:
1920x1080 at 150%

image

Python/Product/PythonTools/PythonTools/Environments/AddCondaEnvironmentControl.xaml
rchiodo: There is a possibility of using a grid instead. That might work too. Combobox would have to be set to full stretch (and that can be difficult given it doesn't do that by default - you have to override the template)

For example this is one of the comboboxes set to stretch (it only stretches to fill the content)

image

Which must be why Hugue forced the width.

rchiodo: The solution here is probably simpler than generating a new template for the different controls.

bschnurr: here is the example i found. but will go with what you have
```
<Grid >
    <Grid.ColumnDefinitions>
        <ColumnDefinition Width="2*" /><!--This will make any control in this column of grid take 2/5 of total width-->
        <ColumnDefinition Width="3*" /><!--This will make any control in this column of grid take 3/5 of total width-->
    </Grid.ColumnDefinitions>
    <Grid.RowDefinitions>
        <RowDefinition MinHeight="30" />
    </Grid.RowDefinitions>

    <TextBlock Grid.Column="0" Grid.Row="0">Your text block a:</TextBlock>
    <TextBlock Grid.Column="1" Grid.Row="0">Your text block b:</TextBlock>
</Grid>
```

bschnurr (approved):

PR #6709, opened 2021-09-02
Fix/suppress build warnings and move up to VS2022 build tools

AdamYoblick:
- Fixes/suppresses build warnings. Build is all green locally AND in PR build.
- Moves C++ projects up to v143 build tools
- Moves pipeline to a new agent pool that has VS2022 installed
- Moves variable library to `PTVS-Dev17` to allow old pipeline to use the old library
- Add binlog generation and build artifact publishing


AdamYoblick:
@rchiodo I know you aren't doing this work anymore, just added you in case you had time. Thanks 😄

int19h (approved):

AdamYoblick:
Ah crap, looks like the 143 toolset isn't installed on the build pool:

##[error]C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(439,5): Error MSB8020: The build tools for v143 (Platform Toolset = 'v143') cannot be found. To build using the v143 build tools, please install v143 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution".

I'll reach out to VS Eng support to see if they are planning on adding at least the build tools to the pool.

AdamYoblick:
Spoke with VS Eng Support, they are going to roll a new pool in the next couple days, so I'll hold off on rolling back the toolset change. Once the new pool is up, I'll modify the azure-pipelines.yml to point at the new pool and see if this PR passes.

Python/Setup/vsix.proj
```diff
@@ -26,6 +26,5 @@
   </Target>
 
   <Import Project="$(TargetsPath)\Common.Build.Traversal.targets" />
-  <Import Project="$(TargetsPath)\MicroBuild.targets" />
```
bschnurr, line 29: MicroBuild.targets seems important. r you sure you want to get rid of that?

AdamYoblick: This was removed because `Common.Build.targets` already imports it. It was a duplicate import 😄

rchiodo (approved):

int19h (approved):

PeezoSlug (approved):

PR #6719, opened 2021-09-20
Fix component governance bugs

AdamYoblick:
Fix vulnerabilities found by component governance scans

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -3,7 +3,7 @@
         {
             "Component": {
                 "Type": "pip",
-                "pip": { "Name": "debugpy", "Version": "1.0.0rc2" }
+                "pip": { "Name": "debugpy", "Version": "1.2.0" }
             },
             "DevelopmentDependency": false
         }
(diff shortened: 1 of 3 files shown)


cgmanifest.json
```diff
@@ -3,7 +3,7 @@
         {
             "Component": {
                 "Type": "pip",
-                "pip": { "Name": "debugpy", "Version": "1.0.0rc2" }
+                "pip": { "Name": "debugpy", "Version": "1.2.0" }
```
int19h, line 6: This version is almost one year old by now. Should this be 1.4.3?

AdamYoblick: Maybe, but PTVS is using 1.2.0 (https://github.com/microsoft/PTVS/blob/main/Build/debugpy-version.txt) which is why I made the cgmanifest.json match. The cg bug was complaining about the mismatch.

I agree we should move it up, but probably as part of a different PR. Does that sound ok?

int19h (approved):

PR #6742, opened 2021-10-02
Fix cookiecutter venv creation when using python from Microsoft Store

### LINKED ISSUE #6736: Cookie cutter does not generate project
**Describe the bug**
[A clear and concise description of the bug]

**Steps to Reproduce**
1. Create a new Python Classifier project
2. From the cookie cutter window that opens up, choose the sk-learn cookiecutter and click next

**Expected behavior**
A project is created

**Actual behavior**
----- Installing cookiecutter -----
Creating virtual environment at 'C:\<user folder>\AppData\Local\Microsoft\CookiecutterTools\env'
Error: Command '['C:\\<user folder>\\AppData\\Local\\Microsoft\\CookiecutterTools\\env\\Scripts\\python.exe', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 106.
Creating virtual environment without pip at 'C:\<user folder>\AppData\Local\Microsoft\CookiecutterTools\env'
Installing pip into virtual environment at 'C:\<user folder>\AppData\Local\Microsoft\CookiecutterTools\env'
The system cannot find the file specified
A task was canceled.
----- Failed to install cookiecutter -----

Microsoft Visual Studio Enterprise 2022 Preview
Version 17.0.0 Preview 4.1
VisualStudio.17.Preview/17.0.0-pre.4.1+31717.71
Microsoft .NET Framework
Version 4.8.04161

Installed Version: Enterprise

Visual C++ 2022   00476-70000-00000-AA046
Microsoft Visual C++ 2022

ASP.NET and Web Tools 2019   17.0.616.20688
ASP.NET and Web Tools 2019

Azure App Service Tools v3.0.0   17.0.616.20688
Azure App Service Tools v3.0.0

C# Tools   4.0.0-4.21458.2+2bfff7b9348e779628a06b86af04b5239d3a926d
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Cookiecutter   17.0.21243.2
Provides tools for finding, instantiating and customizing templates in cookiecutter format.

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Microsoft MI-Based Debugger   1.0
Provides support for connecting Visual Studio to MI compatible debuggers

Microsoft Visual C++ Wizards   1.0
Microsoft Visual C++ Wizards

Microsoft Visual Studio VC Package   1.0
Microsoft Visual Studio VC Package

NuGet Package Manager   6.0.0
NuGet Package Manager in Visual Studio. For more information about NuGet, visit https://docs.nuget.org/

ProjectServicesPackage Extension   1.0
ProjectServicesPackage Visual Studio Extension Detailed Info

Python - Django support   17.0.21243.2
Provides templates and integration for the Django web framework.

Python - Profiling support   17.0.21243.2
Profiling support for Python projects.

Python - VC Project Support   17.0.21153.2
Provides support for launching C++ projects with Python debugging enabled.

Python with Pylance   17.0.21243.2
Provides IntelliSense, projects, templates, debugging, interactive windows, and other support for Python developers.

TypeScript Tools   17.0.0901.2001
TypeScript Tools for Microsoft Visual Studio

Visual Basic Tools   4.0.0-4.21458.2+2bfff7b9348e779628a06b86af04b5239d3a926d
Visual Basic components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Visual Studio Code Debug Adapter Host Package   1.0
Interop layer for hosting Visual Studio Code debug adapters in Visual Studio

```

### PR DESCRIPTION
**Pull request by AdamYoblick**
Fixes #6736 

See the bug above for a detailed description of what this fix entails.

### DIFF
--- a/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
+++ b/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
@@ -120,8 +120,11 @@ namespace Microsoft.CookiecutterTools.Model {
                 var redirectedPath = result.StandardOutputLines.FirstOrDefault();
                 _redirector.WriteLine(Strings.LookingForRedirectedEnvFound.FormatUI(redirectedPath));
                 return redirectedPath;
-            } catch (ProcessException){
+            } catch (ProcessException p){
                 _redirector.WriteLine(Strings.LookingForRedirectedEnvFailed.FormatUI(path));
+                foreach (var line in p.Result.StandardErrorLines) {
+                    _redirector.WriteLine(line);
+                }
                 throw;
             }
         }
--- a/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
+++ b/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
@@ -35,6 +35,7 @@ namespace Microsoft.CookiecutterTools.Model {
         private string _envInterpreterPath;
         private readonly Redirector _redirector;
 
+
         internal string DefaultBasePath { get; set; } = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
 
 
@@ -49,15 +50,15 @@ namespace Microsoft.CookiecutterTools.Model {
             _interpreter = interpreter;
             var localAppDataFolderPath = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
 
+            _redirector = redirector;
+
             // This is where the env *should* be created, but not necessarily where it will live on disk.
             // See the GetRealPath function for more details
             _expectedEnvFolderPath = Path.Combine(localAppDataFolderPath, "Microsoft", "CookiecutterTools", "env");
 
-            // get the real paths to the env and interpreter, in case they've been redirected
+            // Get the real paths to the env and interpreter, in case they've been redirected
             _envFolderPath = Task.Run(() => GetRealPath(_expectedEnvFolderPath)).Result;
             _envInterpreterPath = GetInterpreterPathFromEnvFolderPath(_envFolderPath);
-
-            _redirector = redirector;
         }
 
         public async Task<bool> IsCookiecutterInstalled() {
@@ -97,16 +98,20 @@ namespace Microsoft.CookiecutterTools.Model {
         // in case it's been redirected.
         private async Task<string> GetRealPath(string path) {
 
-            // The expected path is going to be passed into python as a string, which means backslashes have to be escaped.
-            var expectedPath = path.Replace("\\", "\\\\");
-            var command = $"import os; print(os.path.realpath('{expectedPath}'))";
-            var output = ProcessOutput.Run(
+            // if we can see the path, it hasn't been redirected, so no need to call into python
+            if (Directory.Exists(path) || File.Exists(path)) {
+                return path;
+            }
+
+            // get the redirected path from python
+            var command = $"import os; print(os.path.realpath(r'{ path }'))";
+            var output = ProcessOutput.Run (
                 _interpreter.InterpreterExecutablePath,
                 new[] { "-c", command },
                 null,
                 null,
                 false,
-                _redirector
+                null
             );
 
             var result = await WaitForOutput(_interpreter.InterpreterExecutablePath, output);
@@ -143,7 +148,7 @@ namespace Microsoft.CookiecutterTools.Model {
 
             // If we get here, the environment was created successfully.
             // Update the envFolderPath, in case it's been redirected.
-            _envFolderPath = Task.Run(() => GetRealPath(_expectedEnvFolderPath)).Result;
+            _envFolderPath = await GetRealPath(_expectedEnvFolderPath);
             _envInterpreterPath = GetInterpreterPathFromEnvFolderPath(_envFolderPath);
 
             // install pip in the new environment, wherever it is
--- a/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
+++ b/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
@@ -104,6 +104,7 @@ namespace Microsoft.CookiecutterTools.Model {
             }
 
             // get the redirected path from python
+            _redirector.WriteLine(Strings.LookingForRedirectedEnv.FormatUI(path));
             var command = $"import os; print(os.path.realpath(r'{ path }'))";
             var output = ProcessOutput.Run (
                 _interpreter.InterpreterExecutablePath,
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by AdamYoblick** [comment]
@zooba Steve, I added you to this PR if you don't mind, since you gave me guidance on how to approach a workaround for the redirect. If you don't mind taking a look, I'd appreciate it. Thank you 😄

File: Python/Product/Cookiecutter/Model/CookiecutterClient.cs [hunk 1]
```diff
@@ -30,33 +30,38 @@ namespace Microsoft.CookiecutterTools.Model {
...
             var localAppDataFolderPath = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
-            _envFolderPath = Path.Combine(localAppDataFolderPath, "Microsoft", "CookiecutterTools", "env");
-            _envInterpreterPath = Path.Combine(_envFolderPath, "scripts", "python.exe");
+
+            // This is where the env *should* be created, but not necessarily where it will live on disk.
+            // See the GetRealPath function for more details
+            _expectedEnvFolderPath = Path.Combine(localAppDataFolderPath, "Microsoft", "CookiecutterTools", "env");
+
+            // get the real paths to the env and interpreter, in case they've been redirected
+            _envFolderPath = Task.Run(() => GetRealPath(_expectedEnvFolderPath)).Result;
```
zooba, line 57: Using `.Result` on a Task is equivalent to just doing the operation directly (except it's more likely to cause a deadlock).

To do this properly, you'd start the task running and keep a reference to it on the instance, then `await` it to get the result when it's actually needed.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -72,43 +77,76 @@ class CookiecutterClient : ICookiecutterClient {
...
+        // under %localappdata%, the venv will actually be created under the python install localcache,
+        // and a redirect will be put in place that is only understood by the python.exe used to create
+        // the venv.
+        // Therefore, we have to use the python intepreter to check the real path of the venv,
+        // in case it's been redirected.
+        private async Task<string> GetRealPath(string path) {
+
+            // The expected path is going to be passed into python as a string, which means backslashes have to be escaped.
+            var expectedPath = path.Replace("\\", "\\\\");
+            var command = $"import os; print(os.path.realpath('{expectedPath}'))";
```
zooba, line 102:
```suggestion
            var command = $"import os; print(os.path.realpath(r'{expectedPath}'))";
```

And then don't replace the backslashes

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -72,43 +77,76 @@ class CookiecutterClient : ICookiecutterClient {
...
+                new[] { "-c", command },
                 null,
                 null,
                 false,
                 _redirector
             );
-            await WaitForOutput(_interpreter.InterpreterExecutablePath, output);
+
+            var result = await WaitForOutput(_interpreter.InterpreterExecutablePath, output);
+            return result.StandardOutputLines.FirstOrDefault();
```
zooba, line 113: If the process fails for some reason, `result.StandardErrorLines` will have the reason. It's probably worth grabbing these and logging or throwing them somewhere the user can see what happened.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -72,43 +77,76 @@ class CookiecutterClient : ICookiecutterClient {
...
                 null,
                 null,
                 false,
                 _redirector
             );
             await WaitForOutput(_interpreter.InterpreterExecutablePath, output);
 
+            // If we get here, the environment was created successfully.
+            // Update the envFolderPath, in case it's been redirected.
+            _envFolderPath = Task.Run(() => GetRealPath(_expectedEnvFolderPath)).Result;
```
zooba, line 146: We can `await` here, so should definitely do that rather than blocking on the task.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
```diff
@@ -72,43 +77,76 @@ class CookiecutterClient : ICookiecutterClient {
...
 
-            _redirector.WriteLine(Strings.InstallingCookiecutterCreateEnv.FormatUI(_envFolderPath));
+        // If the global interpreter comes from the Microsoft Store, and you try to create a venv
+        // under %localappdata%, the venv will actually be created under the python install localcache,
+        // and a redirect will be put in place that is only understood by the python.exe used to create
+        // the venv.
+        // Therefore, we have to use the python intepreter to check the real path of the venv,
+        // in case it's been redirected.
+        private async Task<string> GetRealPath(string path) {
+
```
zooba, line 99: Fast-path return if we can see the path already? No point running the extra process in this case

AdamYoblick: Great point, will fix

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
AdamYoblick: @zooba I did what you suggested. It's a bit of code duplication, but I don't see how I can avoid that. Please let me know if this is what you had in mind.

Thanks!

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
AdamYoblick: Done.

Python/Product/Cookiecutter/Model/CookiecutterClient.cs
AdamYoblick: I was able to simplify this logic a bit. The _expectedEnvFolderPath is the only path supplied to the constructor, so instead of having a common task, all the methods just check to see if _envFolderPath has been set already. If it has, we've already checked for redirection, and it's a no-op.

The only exception is the method that creates a new venv, in which case it needs to check for redirection again, even if _envFolderPath is set from before. This is done via a boolean that forces the redirection check (python call) to run again.

AdamYoblick:
@bschnurr @int19h @zooba This PR is now ready for review (for real this time 😄 ). Please take a look whenever you get a chance, no rush. Thanks!

PR #6780, opened 2021-11-09
fix addcondaenvironment window scroll with keyboard, when zoomed

bschnurr:
fixes keyboard arrows not scrolling when zoomed.

Changes:
- fixed by adding a height value to the scroll `Height ="300"`
- added a stackpanel for better alignment.
- reformatted doc

internal issue 1355277
Conda environment_Resize Text: Content present in 'Packages preview' section is not visible at different zooming conditions (150%, 175%, 200%).


AdamYoblick (approved):
For future UI changes, please copy/paste an image of the bugfix so everyone can see it 😄 Even better if you can do a "before/after", but even just "after" is ok.

int19h (approved):

PR #6805, opened 2021-12-01
use environment variables to drive symbol publishing

AdamYoblick:
The reason for this PR is to fix broken symbol publishing, specifically symbol **archiving**. 

When publishing symbols using System.AccessToken instead of a PAT, the symbols do not get archived, and the insertion PR into Visual Studio will fail due to this. When publishing symbols using a PAT, the symbols are available publicly, and archiving is handled automatically. As a workaround, we can always manually archive the symbols, but I'd rather just have this work like it did before, without any additional steps required on our end.

- Created variables used by symbol publishing in shared library at https://devdiv.visualstudio.com/DevDiv/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=381&path=PTVS-Dev17
- Removed similar variables from the PublishSymbols task, since they are no longer needed there


int19h (approved):

AdamYoblick:
Manually ran full build at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5507215&view=results and it's all green. So we should be good.

PR #6823, opened 2021-12-14
Generate and publish a software bill of materials

### LINKED ISSUE #6811: Cybersecurity EO Requirements and Audit
See e-mail from Steve Dower.

The only part we need to worry about right now is the Software Bill of Materials. See https://www.1eswiki.com/wiki/ADO_sbom_Generator for more details. This needs to be added to our AzDO build yaml.

Needs to be done for the following products:
- [x] PTVS
          - #6823 
- [x] Pylance
          - https://github.com/microsoft/pyrx/pull/2059
- [ ] PyDebug(?)

AdamYoblick:
Fixes #6811 

PR build won't generate the BOM, so I ran a full build on this topic branch. BOM was generated and published as a build artifact. See https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5545546&view=artifacts&pathAsName=false&type=publishedArtifacts.

I checked the BOM for correctness and it seems good to me. The only glaring omission I noticed was debugpy not being mentioned anywhere, but maybe that's because it's python with no binaries?


bschnurr:
Found some info here, maybe add a comment to it.
https://eng.ms/docs/initiatives/executive-order/executive-order-requirements/executiveorderoncybersecurity/softwarebillofmaterials

maybe our component governance is also missing debugpy

bschnurr:
maybe see what path component gov uses

https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/custom-sbom-generation-workflows

<html>
<body>


BuildComponentPath | -bc | The folder containing the build components and packages. This folder path is passed as is to the component detection tool to detect packages in this folder. If you use CG tools to detect packages in your code, please provide us with the same path here.
-- | -- | --



</body>
</html>

AdamYoblick:
I believe component governance runs automatically against the source, we don't run it explicitly. The SBOM instructions said to point it to our release before we publish it, so that's what I did. That release is what goes into VS. I think the issue is the SBOM tool needs to crack open VSIXs and scan the contents as well, not sure if it's doing all that. And if it is, I'm not sure if it's including non-binaries, like debugpy.

bschnurr:
just wondering if this would help.

`BuildComponentPath: $(Build.SourcesDirectory)`

AdamYoblick:
No, that's too heavy handed. The BOM is only supposed to be generated against what we ship, not what we have in source control. For example: `In the task config, you must specify the root path of the folder for where you want the SBOM to be generated. We recommend using the same root path location that you publish your build drop to. An example is (System.ArtifactsDirectory)/buildoutput`

The folder that we publish that gets injected into VS is this:

```
  # Upload vsts drop used by Visual Studio insertions
  - task: MicroBuildUploadVstsDropFolder@1
    displayName: 'Upload vsts drop'
    inputs:
      DropFolder: '$(Build.StagingDirectory)\release'
      DropServiceUri: 'https://devdiv.artifacts.visualstudio.com/DefaultCollection'
      VSDropServiceUri: 'https://vsdrop.corp.microsoft.com/file/v1'
      AccessToken: '$(System.AccessToken)'
    continueOnError: true```

That's why I used `$(Build.StagingDirectory)\release`

bschnurr (approved):

PR #6848, opened 2022-01-11
Auto-update pylance and debugpy dependencies when running PreBuild.ps1

### LINKED ISSUE #6840: Auto update Pylance and Debugpy dependencies as part of build
# Pylance
From a high level, here's what the logic flow will look like:
- Pylance version in package.json will drive all npm installs, as it normally does.
- This version will be set to "latest" by default, which will always install the latest version.
- PreBuild.ps1 (which is always called before building) will be used to install pylance, in both local and CI builds.

In CI builds, there are some additional steps:
- A `PylanceVersion` runtime parameter will be added to the build yaml. It will default to "latest", and will be settable at build queue time.
- Before installing pylance, we will run a script that will overwrite the version string in package.json with the version provided by the runtime parameter. This will allow the user to roll back versions.
- After installing pylance, we will tag the build with the version that was installed for easy tracking.
    - Tagging the build DOES WORK, but AzDO throws an exception for some reason. No idea why. Following up with their team and commenting out the tagging code for now.

Tasks:
- [x] Set `@pylance/pylance` version to "latest" in package.json (locally)
- [x] Verify latest version is installed when calling `npm install --save` (locally)
- [x] Verify we can get the installed version. (locally)
          - The powershell to run the command and print JUST the version string looks like this:
```powershell
$output = & npm ls @pylance/pylance
$version = $output[1] -split "@" | Select-Object -Last 1
$version
```
- [x] Modify the build yaml to no longer npm install pylance (since we now do it in the prebuild.ps1) (azdo)
- [x] Add a build step that tags the build with the pylance version that was installed
          - From powershell, this looks like `Write-Host "##vso[build.addbuildtag]$pylanceVersion"`
- [x] At this point, the build should now be installing latest pylance on every run and should be tagging the build with the version used. TEST NOW and verify this works.
          - Confirmed that latest pylance is being installed (https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5599427&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=1a5cc010-7735-550a-9d76-c0b745122dab)
          - Tagging the build DOES WORK, but AzDO throws an exception for some reason. No idea why. Following up with their team and commenting out the tagging code for now.
- [x] Write a script that will overwrite the version in the specified package.json with the specified pylance version (locally).
- [x] Add a `PylanceVersion` runtime parameter to the build pipeline. The parameter should default to "latest" but should be settable at build queue time. (azdo)
- [x] Add a build step that uses the script to overwrite the version in package.json based on the runtime parameter (azdo).
- [x] At this point, the build should accept a pylance version string when manually queueing. TEST NOW with "2021.11.0" and verify this works, and make sure the build is tagged correctly.

# DebugPy
From a high level, here's what the logic flow will look like:
- Debugpy version in `.\build\debugpy-version.txt` will continue to drive the debugpy install, as it currently does.
- This version will be set to "latest" by default, which will always install the latest version.
- PreBuild.ps1 (which is always called before building) will be used to install debugpy, in both local and CI builds (which is the current logic).

In CI builds, there are some additional steps:
- A `Debugpy` runtime parameter will be added to the build yaml. It will default to "latest", and will be settable at build queue time.
- Before installing debugpy, we will run a script that will overwrite the version string in debugpy-version.txt with the version provided by the runtime parameter. This will allow the user to roll back versions.
- After installing debugpy, we will tag the build with the version that was installed for easy tracking.

Tasks:
- [x] Modify `.\build\install_debugpy.py` get the latest version of debugpy if the specified version is "latest". Then proceed with normal downloading.
- [x] Add a build step that tags the build with the debugpy version that was installed.
- [x] At this point, the latest debugpy should be downloading and installing, and the build should be tagged. TEST NOW and verify this works.
- [x] Add a `DebugpyVersion` runtime parameter to the build pipeline. The parameter should default to "latest" but should be settable at build queue time.
- [ ] At this point, the build should accept a debugpy version string when manually queueing. TEST NOW with "1.4.3" and verify this works, and make sure the build is tagged correctly.

AdamYoblick:
Fixes #6840

This change does two main things from a high level:
1. Defaults pylance and debugpy to "latest" and automatically installs the latest available versions when running PreBuild.ps1
2. Adds runtime parameters to the build yaml (and pipes them through to PreBuild.ps1) so we can roll back to older versions (if we need to) when manually queueing a build.


Build/install_debugpy.py
```diff
@@ -1,64 +1,64 @@
...
 import json
 import os
 import urllib.request as url_lib
 import zipfile
 
-DEBUGGER_PACKAGE = "debugpy"
+# if this import fails, run PreBuild.ps1, which will pip install packaging
+from packaging.version import parse as version_parser
+
 DEBUGGER_PYTHON_VERSIONS = ("cp35", "cp36", "cp37", "cp38")
```
int19h, line 13: `cp39`?

int19h (approved):

Build/install_debugpy.py
AdamYoblick: Fixed

AdamYoblick:
I manually tested the rollback mechanism and it seems to be working just fine for both pylance and debugpy. Going to merge this in.

PR #6850, opened 2022-01-14
re-enable build tagging with a fix

AdamYoblick:


AdamYoblick:
Ran the build manually to make sure the tagging is now working, and it is. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5632133&view=results

PR #6902, opened 2022-03-04
Fix python version retrieval in conda envs for python 3.10 and higher

### LINKED ISSUE #6749: The Python version is display wrong when use the latest python to create Conda environment.
**Environment**
image

**Steps to Reproduce**
1.  Create a Python project
2.  Right click Python Environments -> Add Environment -> Conda environment
3.  Choose the latest python version to create
image
4. Click Create button
5. Observes

**Expected behavior**
The environment is created successfully and display without aby error.

The Python version is display wrong when use the latest python to create Conda environment.
It displayed Python0.0, it should display Python 3.10
image

AdamYoblick:
Fixes #6749

Retrieving python version information from a conda environment with python >= 3.10 has been fixed, shown below:

image

Fixing this bug surfaced two new issues with both pylance and debugpy not working with these new environments. I've created new issues to track these:
- https://github.com/microsoft/PTVS/issues/6963
- https://github.com/microsoft/PTVS/issues/6964


AdamYoblick:
@bschnurr - Pylance doesn't seem to be working correctly for a conda environment with python 3.10 installed, but it does work for a conda environment with python 3.9.7 installed. See here:

Conda with Python 3.9.7:
image

Conda with Python 3.10:
image

Can you think of any reasons why this might be?

AdamYoblick:
I spoke with @int19h and Karthik about the debugger not working with a conda environment containing python 3.10. This is due to the following bug in conda: https://github.com/conda/conda/issues/11072.

Karthik mentioned that we could run the debugger using `conda run -n <env> --no-capture-output --live-stream python <debugpy>` to get around this issue, so I'm going to try to do that and update this PR when I do.

Python/Product/PythonTools/PythonTools/Debugger/DebugLaunchHelper.cs
```diff
@@ -158,9 +158,25 @@ static class DebugLaunchHelper {
...
+            // See https://github.com/conda/conda/issues/11072
+            // The workaround is to run this instead:
+            // conda run -n <env> --no-capture-output --live-stream python <debugpy>
+            //if (CondaUtils.IsCondaEnvironment(Path.GetDirectoryName(config.GetInterpreterPath())) &&
+            //    config.Interpreter.Version > new Version(3, 10)) {
+            //
+            //    var condaProvider = provider.GetComponentModel().GetService<CondaEnvironmentFactoryProvider>();
+            //    var condaExe = condaProvider
+            //
+            //    //DoConda310Workaround(provider, jsonObj);
```
bschnurr, line 171: is this all commented out on purpose?

AdamYoblick:
Ah that was stuff I meant to remove. I'll remove it :)
________________________________

Python/Product/PythonTools/PythonTools/Debugger/DebugLaunchHelper.cs
AdamYoblick: Ah that was stuff I meant to remove. I'll remove it :)

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -300,36 +336,60 @@ internal class CondaInfoResult {
...
+                // if the folder doesn't exist, skip it
+                if (string.IsNullOrEmpty(folder) || !Directory.Exists(folder)) {
+                    continue;
+                }
+
+                // skip the root to avoid duplicate entries, root is
                 // discovered by CPythonInterpreterFactoryProvider already.
-                // Older versions of `conda info` used to not return the root.
-                return condaInfoResult.EnvironmentFolders
-                    .AsParallel()
```
bschnurr, line 319: reason for not doing it in parallel? i'd double check the history for a reason why they added parallel

AdamYoblick: Couple of reasons I did it this way.

1. The large chained linq statement is much harder to debug than vanilla code. I'm not a fan of this.
2. I'm not a fan of calling a function that does actual work (CreateEnvironmentInfo) nested in a Select statement. Again, harder to read, harder to debug.
3. The amount of performance we get back here is kind of negligible. I tested this with 10 conda environments and the results came back very quickly with no parallelism. The user is not going to have hundreds or thousands of conda environments.

int19h: I would suggest breaking down the LINQ statement into several steps, with a named variable capturing each step - this way it's still high-level (i.e. the code focuses on *what* is being done, not *how* it's being done), and variable names serve as comments describing each step. For debugging, you can then break on each step and see what the current set of candidates is.

BTW the C# debugger in VS lets you set breakpoints on arbitrary subexpressions, although the UI is a bit fiddly. But it can be very useful when debugging call chains like these.

AdamYoblick: That's exactly what I did, in addition to removing the parallelization because it honestly isn't needed for the number of environments there are.
I refactored the code to put the terminal cases first, inverted the conditionals to include early returns to make the code more readable. The logic is the same.

AdamYoblick: And just because the tools let you debug subexpressions doesn't justify code like this. This is messy. The expression is too long. This is an example of using linq because you can, not because you should. It doesn't simplify anything, you've got a select doing actual work, etc...

In the refactored code, you can very easily set breakpoints and see if the environment folder doesn't exist, for example.

All my opinion, of course.

int19h: I don't think there's any fundamental difference between the same number of expressions vs statements, aside from tooling support. Subjectively, clarity with an explicit loop is lower than LINQ sequence comprehension to me for the same reason why if/goto has less clarity than a loop - it's less structured and less abstract.

All that aside, it's kind of a moot question anyway since the code now needs to use `await`, which precludes (stock) LINQ.

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -343,7 +403,7 @@ internal class CondaInfoResult {
             }
 
             var arch = CPythonInterpreterFactoryProvider.ArchitectureFromExe(interpreterPath);
-            var version = CPythonInterpreterFactoryProvider.VersionFromSysVersionInfo(interpreterPath);
+            var version = await Task.Run(() => ExecuteCondaListAsync(condaPath, name));
```
int19h, line 406: It seems that `ExecudeCondaListAsync` is always wrapped in `Task.Run` to ensure that it runs on a background thread. It might be better to move that logic into the method itself - basically just use `ConfigureAwait(false)` for all its awaits.

AdamYoblick: I believe there's only one await in that function. So you're saying just `await ExecuteCondaListAsync`, then use ConfigureAwait(false) inside the method?

int19h: Yep, that should do the trick.

Side note: we should probably get into the habit of using `ConfigureAwait(false)` in general, except for those places where a postback to the original thread is actually needed. The default in .NET is really unfortunate - it brings a significant perf hit and increases possibility of deadlocks. But, this might be tricky for existing code, where it might not be clear whether any given code block that follows an `await` implicitly relies on it being the UI thread or not. So it might be best to do this piecemeal whenever we significantly modify the code, and have to ask those questions anyway.

int19h (approved):

Issue #6636, opened 2021-08-03
The activated environment does not changed to bold when change active environment.


linette-zyy:
**Environment**
![image](https://user-images.githubusercontent.com/59192491/127968239-49f3cc40-2cfd-4e9d-ac67-9a6cfc6f55ca.png)

**Steps to Reproduce**
1. Add more than one Python environments
2. Right-click none-bold python environment and select "Activate environment"

**Expected behavior**
The choice is saved with the project and the environment is bold.

**Actual result**
The activated environment does not changed to bold.
![image](https://user-images.githubusercontent.com/59192491/127968514-5aa698c9-e74b-4f5b-bb92-4661f5b03327.png)

PR #6913, opened 2022-03-09
Fixing Issue #6636

StellaHuang95:
This commit adds BoldActiveEnvironment() to the end of ActiveInterpreter setter to bold the active environment properly.

I did a few tests to verify the change:

1. Switch from one environment to another:
Animation

2. The global default interpreter will be activated after deleting all environments.
Animation2

3. Add a new environment to the list and activate it.
Animation3

4. Delete all the envs so only the global default remains. Then add a new one. 
Animation4

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
@@ -278,6 +278,7 @@ namespace Microsoft.PythonTools.Project {
                 if (_active != oldActive || oldActive == null) {
                     ActiveInterpreterChanged?.Invoke(this, EventArgs.Empty);
                 }
+                BoldActiveEnvironment();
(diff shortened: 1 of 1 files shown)


AdamYoblick:
Can you also please show this working when adding a new environment to an already existing list of environments?

StellaHuang95:
@AdamYoblick Sure, I just did.

AdamYoblick:
Oh nevermind, I thought adding a new environment automatically activated it. It only seems to do that when you only have the global default in the list, then add a new one. Can you please verify that case? Delete all the envs so only the global default remains. Then add a new one. The new one should be activated and bolded. If that works, then we should be good. 😄

StellaHuang95:
@AdamYoblick Of course! I just added the new test and looks like it's working.

AdamYoblick (approved):

PR #6918, opened 2022-03-15
try npm authenticate instead of manually modifying the .npmrc file

AdamYoblick:


int19h (approved):

AdamYoblick:
Both PR build and manual run of full build from topic branch have passed the Restore Packages step, so this seems to be working. Going to merge it in.

PR #6928, opened 2022-03-17
Fix internal bug #1489308

StellaHuang95:
This PR fixed internal bug #1489308 where focus is not clearly visible on remove packages button.

Currently the background color of the button will change and become more visible when focus is on it.
Animation2


AdamYoblick (approved):

AdamYoblick:
This looks good to me, although I worry about the contrast difference when coloring the background of the remove button. I wouldn't worry about it now though, just keep in mind that they might come back and tell us we need it to be more noticeable, but don't block the PR because of that. Maybe we'll get lucky 😄

PR #6929, opened 2022-03-18
Fix VS crash when creating virtual environment

### LINKED ISSUE #6904: VS crashes while creating a virtual environment.
**Describe the bug**

image

**Steps to Reproduce**

1. Create a Python Application project.
2. Create virtual environment.

**Expected behavior**

The virtual environment was successfully installed.

VS crashes while creating a virtual environment.

creating_virtualenv

Here is the .NET exception I am pulling from the Event Manager.
<details>
<p>

```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.NullReferenceException
   at Microsoft.PythonTools.LanguageServerClient.FileWatcher.Listener+<OnFileChanged>d__12.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.PythonTools.LanguageServerClient.FileWatcher.Listener+<OnFileChanged_Sync>d__11.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()

```

</p>
</details>

AdamYoblick:
Fixes #6904 

This bug seems to be due to a race condition on the _rpc instance in the file listener. Since all calls using the rpc are async, the _rpc_Disconnected callback (which sets _rpc to null) can be triggered at any time during or between these calls.

The crash happens in two different ways, depending on the timing:

1. The _rpc is set to null in between rpc calls. This causes a NullReferenceException the next time the rpc is attempted to be used. 
     * The solution for this is to check for null before using it, every time. While this doesn't technically eliminate the race condition (since we're not using a mutex), the chance of an NRE is very small.
3. The _rpc is disposed during an rpc call. This causes the rpc call to throw a ConnectionLostException, which was unhandled. 
     * The solution for this is to catch all ConnectionLostExceptions and ignore them, since there's nothing to be done on a failure.

After adding these checks, the env creation completes with no additional errors. I also tried modifying an existing file (and saving), and adding a new file, to make sure the file listener was picking up those events, and it is.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
@@ -114,11 +114,14 @@ namespace Microsoft.PythonTools.LanguageServerClient.FileWatcher {
                 return;
             }
 
-            // If the rpc has been disconnected, the _rpc_Disconnected callback will set _rpc to null.
-            // So check for null now, and also check before every rpc call to avoid a null reference exception
+            // _rpc can be disposed and set to null at any time, since all the calls using it are asynchronous.
+            // If the rpc has already been set to null, we're done.
+            // Otherwise, store a local copy of the rpc to avoid having to check for null every time it's used.
+            // Then, when using it, catch any expected exceptions to avoid a crash.
             if (_rpc is null) {
                 return;
             }
+            var rpc = _rpc;
 
             // Create something to match with
             var item = new InMemoryDirectoryInfo(_root, new string[] { e.FullPath });
@@ -168,9 +171,7 @@ namespace Microsoft.PythonTools.LanguageServerClient.FileWatcher {
 
                 if (didChangeParams.Changes.Any()) {
                     try {
-                        if (_rpc != null) {
-                            await _rpc.NotifyWithParameterObjectAsync(Methods.WorkspaceDidChangeWatchedFiles.Name, didChangeParams);
-                        }
+                        await rpc.NotifyWithParameterObjectAsync(Methods.WorkspaceDidChangeWatchedFiles.Name, didChangeParams);
                         if (renamedArgs != null) {
                             var textDocumentIdentifier = new TextDocumentIdentifier();
                             textDocumentIdentifier.Uri = new Uri(renamedArgs.OldFullPath);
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
```diff
@@ -108,72 +108,91 @@ class Listener : IDisposable {
             await OnFileChanged(sender, e);
         }
         private async Task OnFileChanged(object sender, System.IO.FileSystemEventArgs e) {
-            // Skip directory change events and when rpc has ben disconnected
-            if (!e.IsDirectoryChanged() && _rpc != null) {
```
AdamYoblick, line 112: Ugh sorry about the formatting showing up as one big change. I should have put in the logic change as its own commit, then the early returns as another one so it would be easier to see the diff.

The only changes are:

1. Make sure the _rpc isn't null when you try to use it
2. Catch all ConnectionLostExceptions from _rpc calls and ignore them.

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
```diff
@@ -108,72 +108,91 @@ class Listener : IDisposable {
...
-                    } else {
-                        // file renamed
-                        var deleteEvent = new FileEvent();
-                        deleteEvent.FileChangeType = FileChangeType.Deleted;
-                        deleteEvent.Uri = new Uri(renamedArgs.OldFullPath);
 
-                        var createEvent = new FileEvent();
-                        createEvent.FileChangeType = FileChangeType.Created;
-                        createEvent.Uri = new Uri(renamedArgs.FullPath);
+            // Skip directory change events
```
AdamYoblick, line 112: Ugh sorry about the formatting showing up as one big change. I should have put in the logic change as its own commit, then the early returns as another one so it would be easier to see the diff.

The only changes are:

1. Make sure the _rpc isn't null when you try to use it
2. Catch all ConnectionLostExceptions from _rpc calls and ignore them.

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
```diff
@@ -108,72 +108,91 @@ class Listener : IDisposable {
...
+            if (e.IsDirectoryChanged()) {
+                return;
+            }
 
-                        didChangeParams.Changes = new FileEvent[] { deleteEvent, createEvent };
+            // If the rpc has been disconnected, the _rpc_Disconnected callback will set _rpc to null.
+            // So check for null now, and also check before every rpc call to avoid a null reference exception
+            if (_rpc is null) {
```
int19h, line 119: I think it would be better to copy it into a local - that way there's no assignment race, and you only need to worry about caching the exception from `Dispose`.

int19h (approved):

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
AdamYoblick: So you're saying avoid the null check altogether and just see what exception gets thrown when trying to use a disposed rpc, and catch that along with the one I already caught?

int19h: It can be null to begin with, so I don't see any way to avoid the check. But if you assign to a local instead, it could be just a single check on that local.

PR #6938, opened 2022-03-24
Fix crashes due to rpc connection being lost

AdamYoblick:
I fixed these previously in the Listener but I was able to consistently repro the same issues in the PythonLanguageClient, so these need to be caught there too.

I know there's a better way to generically handle this exception for all these different rpc calls, but that code ended up really messy so I opted to just duplicate the try catch since there's only a few of them. If anyone has any ideas on how to wrap each of the calls without it being super ugly, I'm all ears 😄

### DIFF
--- a/Python/Product/Common/Infrastructure/JsonRpcWrapper.cs
+++ b/Python/Product/Common/Infrastructure/JsonRpcWrapper.cs
@@ -1,0 +1,60 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PythonTools.Infrastructure;
+using StreamJsonRpc;
+
+namespace Microsoft.PythonTools.Common.Infrastructure {
+    internal class JsonRpcWrapper {
+
+        private readonly JsonRpc _rpc;
+
+        // list of exception types to ignore
+        private static readonly Type[] _exceptionsToIgnore = {
+            typeof(ObjectDisposedException),
+            typeof(ConnectionLostException)
+        };
+
+        public JsonRpcWrapper(JsonRpc rpc) {
+            _rpc = rpc;
+        }
+
+        public Task NotifyWithParameterObjectAsync(string targetName, object argument = null) {
+
+            if (_rpc is null) {
+                return Task.CompletedTask;
+            }
+
+            // if the underlying rpc connection was disposed, or the connection was lost, ignore the errors
+            return _rpc.NotifyWithParameterObjectAsync(targetName, argument)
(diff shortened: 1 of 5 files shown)


AdamYoblick:
I know there's a better way to generically handle this exception for all these different rpc calls, but I that code ended up really messy so I opted to just duplicate the try catch since there's only a few of them. If anyone has any ideas on how to wrap each of the calls without it being super ugly, I'm all ears 😄

Python/Product/PythonTools/PythonTools/LanguageServerClient/FileWatcher/Listener.cs
```diff
@@ -114,14 +114,11 @@ class Listener : IDisposable {
...
             }
 
-            // _rpc can be disposed and set to null at any time, since all the calls using it are asynchronous.
-            // If the rpc has already been set to null, we're done.
-            // Otherwise, store a local copy of the rpc to avoid having to check for null every time it's used.
-            // Then, when using it, catch any expected exceptions to avoid a crash.
-            if (_rpc is null) {
+            // _rpc can be disposed and/or set to null at any time, since all the calls using it are asynchronous.
+            // Check for null and disposed now, and every time it's used.
+            if (_rpc is null || _rpc.IsDisposed) {
```
int19h, line 119: Since this code is entered via `async void`, it can potentially run concurrently (not just asynchronously) on a different thread relative to the one that does `Dispose()` and `=null`, so this check still has a race condition between left and right-hand side. This can be mitigated by copying it to a local before checking, or by using nullable bools:
```cs
if (_rpc?.IsDisposed != false)  // null or true
```
But also, looking at the rest of the class, it looks like it might be possible to just make `_rpc` readonly, and drop the null checks altogether?

AdamYoblick: Ahh, good catch. I didn't think about the rpc being set to null before the second check on that conditional, but yeah that's definitely true.

`_rpc` gets set to null in the `_rpc_Disconnected` callback, so I don't think it can be readonly, just because of that. The reason I added the null checks back in was because that's the existing pattern in the PythonLanguageClient. But I agree, the code is much better to read the other way. I'll add the local back in.

AdamYoblick:
Refactored the code a bit so all the exception handling takes place in the new `JsonRpcWrapper` class. This prevents the race condition where `_rpc` is set to null while we're using it, and eliminates the duplicated silence exception code.

Python/Product/Common/Infrastructure/TaskExtensions.cs
```diff
@@ -147,7 +150,34 @@ static class TaskExtensions {
...
+
+        /// <summary>
+        /// Silently handles a collection of exception types
+        /// </summary>
+        public static Task<U> SilenceExceptions<U>(this Task<U> task, IEnumerable<Type> exceptionTypes) {
+            return task.ContinueWith(t => {
+                try {
+                    return t.Result;
+                } catch (AggregateException ex) {
+                    ex.Handle(e => exceptionTypes.Any(f => e.GetType().Equals(f)));
```
int19h, line 179: `f.IsAssignableTo(e)` would reflect the semantics of `catch` more closely and be less brittle, esp. if the library starts throwing a private derived class.

Python/Product/Common/Infrastructure/JsonRpcWrapper.cs
```diff
@@ -0,0 +1,60 @@
...
+            typeof(ConnectionLostException)
+        };
+
+        public JsonRpcWrapper(JsonRpc rpc) {
+            _rpc = rpc;
+        }
+
+        public Task NotifyWithParameterObjectAsync(string targetName, object argument = null) {
+
+            if (_rpc is null) {
```
int19h, line 40: Given that this is `readonly`, it seems that this check would never actually be hit?

Python/Product/Common/Infrastructure/TaskExtensions.cs
AdamYoblick: I couldn't find `IsAssignableTo` on the type class, but I found `IsAssignableFrom` so that should work in the same way. 😄

Python/Product/Common/Infrastructure/JsonRpcWrapper.cs
AdamYoblick: Well since it's a timing issue, it's possible that the _rpc was null when the wrapper was created, and null was passed into the constructor. So that's why I do the check

int19h (approved):

Issue #6942, opened 2022-03-29
Failed to set startup file when create new project from existing python code.


linette-zyy:
**Environment**
![image](https://user-images.githubusercontent.com/59192491/160564488-cbcbbdb0-ded2-488e-9109-15a8bfec1235.png)

**Steps to Reproduce**
1. Launch VS and select File > new > Project > Python Application named PythonApplication18
2. Right click PythonApplication18 and add new item named module1.py
3. Set the PythonApplication18.py as Start File and save all
![image](https://user-images.githubusercontent.com/59192491/160565933-0adb1ee9-ed9f-4e87-9b4b-aa877022d508.png)
4. File > new > Project > From Existing Python code
5. In the wizard that appears,  
a. Set the path to PythonApplication18,  
![image](https://user-images.githubusercontent.com/59192491/160569804-0b6a15ce-3419-4b24-8243-d84ef8246650.png)
b. Set a filter for file types 
c. Specify any search paths that your project requires 
d. Select Next 
6. In the next dialog, select the PythonApplication18.py as startup file for my project and select Next 
![image](https://user-images.githubusercontent.com/59192491/160567891-463d0efd-c0eb-47ee-ae3a-1436a7c6f933.png)
7. Select the location in which to save the project file and click Finish

**Expected behavior**
PythonApplication18.py is set as Start File

**Actual behavior**
module1.py is the Start File
![image](https://user-images.githubusercontent.com/59192491/160569185-44c9f6b6-487f-4eec-b043-05513f621450.png)

PR #6950, opened 2022-04-05
Fix startup file settings when creating new project from existing python code

StellaHuang95:
This PR fixes issue #6942 

Test1: set startup project to module2
Animation2 - startup module2

Test2: set startup project to PythonApplication4
Animation2 - startup python


Python/Product/PythonTools/PythonTools/Project/ImportWizard/ImportSettings.cs
```diff
@@ -204,6 +204,8 @@ internal class ImportSettings : DependencyObject {
                 _isAutoGeneratedProjectPath = true;
             }
 
+            var startup = this.StartupFile;
```
int19h, line 207: Could you add a comment here explaining the need for a workaround? Otherwise this might get accidentally reverted in some future refactoring.

Python/Product/PythonTools/PythonTools/Project/ImportWizard/ImportSettings.cs
```diff
@@ -216,7 +218,7 @@ internal class ImportSettings : DependencyObject {
                 }
             }
 
-            StartupFile = SelectDefaultStartupFile(fileList, StartupFile);
```
int19h, line 219: The difference between `this.StartupFile` above vs `StartupFile` here caught my eye. There's no actual difference, of course - but when reading the code, due to the distinction, one might assume that there is one, and it's important enough to highlight. Let's keep them consistent for clarity.

StellaHuang95: @int19h Just updated `this.StartupFile` to `StartupFile` to be consistent.

Python/Product/PythonTools/PythonTools/Project/ImportWizard/ImportSettings.cs
StellaHuang95: Sure, just added.

int19h (approved):

Issue #6785, opened 2021-11-16
When vs2022 edits python formatting Chinese strings are garbled


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/vs2022-编辑python进行格式化的时候中文字符串会乱码/1577322)._

---
When vs2022 edits python formatting Chinese strings are garbled
I did the path operation when I edited the python code with vs2022 The file name contains Chinese
Then I formatted parts of the python file using CTRL-K, CTRL-D Chinese, and the characters in the python file became garbled

---
#### Feedback Bot on 11/10/2021, 00:07 AM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

---
### Original Solutions
(no solutions)

AdamYoblick:
Are you able to provide a repro for this please? Maybe a snippet or a file demonstrating the issue?

judej:
Closing old issue. If this is still a problem, please reopen with the information requested. thanks

zzzmfps:
Version: VS 2022 17.1.1
Formatter: All (autopep8, yapf, black)

It works fine only if formatters are not needed at all (statements are already formatted). Otherwise unformatted lines will be garbled.

```
list1 = ['1', '2']  # normal without Chinese characters
list2 = ['一', '二']  # works fine if statement is already formatted

list3 = ['一','二']  # if misses a whitespace, for example
# formatter will garble the statement and even produce a syntax error:
# list3 = ['涓€', '浜?]
```

AdamYoblick:
@StellaHuang95 This might be related to the encoding issue you're working on

StellaHuang95:
Running the formatter from command line seems to work. It might be a problem with how we handle the patch from black.

PR #6965, opened 2022-04-20
Fix formatter not formatting special characters

StellaHuang95:
This PR fixes issue #6785 

Formatters in PTVS were not able to format special characters such as Chinese and Germany characters correctly. Not sure what encoding/decoding is used on PTVS/formatter end, which might be the cause, so pass in UTF-8 to decode patch returned from formatters.

Test 1:
formatter1

Test 2 with Black:
Animation2 - formatter1

Test 3 with autopep8:
Animation2 - pep8

Test 4 with yapf:
Animation2 - y


int19h (approved):

AdamYoblick:
Great job!

AdamYoblick (approved):

PR #6984, opened 2022-04-27
Fix profiling standalone script error

### LINKED ISSUE #6769: Profiling:Error prompted the parameter cannot be null when it is set in profiling standalone script.
**Environment**
image

**Steps to Reproduce**
1.Create python application project named PythonApplication1 and typing code in it.
2.Debug->Launch Python Profiling …
3.Download new python environment under the custom path
3.Checked the option "Standalone script", In part of script Option: 
a. Python Interpreter: the new python environment under the custom path
b. Script: PythonApplication1.py
c. Working Directory: enter your working directory
image
4.Click "Start"

**Expected behavior**
Start without any errors

Error popup window
image

StellaHuang95:
fixes #6769 

The issue was caused by two reasons:

- The `Interpreter Path` is entered after `Python Interpreter`, so the interpreter properties including id, name, and path never get updated. Manually set the properties in `Interpreter Path` setter to handle to ordering issue.

- The `CanSpecifyInterpreterPath` is not set correctly after constructing the custom python interpreter, so `InterpreterPath` is always returned as `null` from `GetTarget()`

Some tests:

1. Custom interpreter
Animation2 - profiling others

2. Existing interpreter
Animation2 - profiling python3 8

3. Open project
Animation2 - open project

### DIFF
--- a/Python/Product/Profiling/Profiling/CustomPythonInterpreterView.cs
+++ b/Python/Product/Profiling/Profiling/CustomPythonInterpreterView.cs
@@ -1,0 +1,28 @@
+// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+namespace Microsoft.PythonTools.Profiling {
+    class CustomPythonInterpreterView : PythonInterpreterView {
+        public CustomPythonInterpreterView() :
+            base(Strings.LaunchProfiling_OtherInterpreter, "", null) {
+        }
+        public CustomPythonInterpreterView(string path) :
+            base(Strings.LaunchProfiling_OtherInterpreter, "", path) {
+        }
+
+    }
+}
+ 
--- a/Python/Product/Profiling/Profiling/CustomPythonInterpreterView.cs
+++ b/Python/Product/Profiling/Profiling/CustomPythonInterpreterView.cs
@@ -16,10 +16,7 @@
 
 namespace Microsoft.PythonTools.Profiling {
     class CustomPythonInterpreterView : PythonInterpreterView {
-        public CustomPythonInterpreterView() :
-            base(Strings.LaunchProfiling_OtherInterpreter, "", null) {
-        }
-        public CustomPythonInterpreterView(string path) :
+        public CustomPythonInterpreterView(string path = null) :
             base(Strings.LaunchProfiling_OtherInterpreter, "", path) {
         }
 
--- a/Python/Product/Profiling/Profiling/StandaloneTargetView.cs
+++ b/Python/Product/Profiling/Profiling/StandaloneTargetView.cs
@@ -165,7 +165,7 @@ namespace Microsoft.PythonTools.Profiling {
                 if (_interpreterPath != value) {
                     _interpreterPath = value;
                     if (Interpreter.Name == Strings.LaunchProfiling_OtherInterpreter && Interpreter.Path == null) {
-                        Interpreter = new PythonInterpreterView(Strings.LaunchProfiling_OtherInterpreter, "Global|"+ Strings.LaunchProfiling_OtherInterpreter, _interpreterPath);
+                        Interpreter = new PythonInterpreterView(Strings.LaunchProfiling_OtherInterpreter, Strings.LaunchProfiling_OtherInterpreter, _interpreterPath);
                     }
                     OnPropertyChanged("InterpreterPath");
                 }
--- a/Python/Product/Profiling/Profiling/StandaloneTargetView.cs
+++ b/Python/Product/Profiling/Profiling/StandaloneTargetView.cs
@@ -29,6 +29,7 @@ namespace Microsoft.PythonTools.Profiling {
     /// </summary>
     sealed class StandaloneTargetView : INotifyPropertyChanged {
         private readonly ReadOnlyCollection<PythonInterpreterView> _availableInterpreters;
+        private readonly PythonInterpreterView _customInterpreter;
 
         private PythonInterpreterView _interpreter;
         private string _interpreterPath;
@@ -55,7 +56,7 @@ namespace Microsoft.PythonTools.Profiling {
                 )
             ).ToList();
 
-            var _customInterpreter = new CustomPythonInterpreterView();
+            _customInterpreter = new CustomPythonInterpreterView();
             availableInterpreters.Add(_customInterpreter);
             _availableInterpreters = new ReadOnlyCollection<PythonInterpreterView>(availableInterpreters);
 
--- a/Python/Product/Profiling/Profiling/StandaloneTargetView.cs
+++ b/Python/Product/Profiling/Profiling/StandaloneTargetView.cs
@@ -56,7 +56,7 @@ namespace Microsoft.PythonTools.Profiling {
                 )
             ).ToList();
 
-            _customInterpreter = new PythonInterpreterView(Strings.LaunchProfiling_OtherInterpreter, "", null);
+            _customInterpreter = new CustomPythonInterpreterView();
             availableInterpreters.Add(_customInterpreter);
             _availableInterpreters = new ReadOnlyCollection<PythonInterpreterView>(availableInterpreters);
 
@@ -148,7 +148,7 @@ namespace Microsoft.PythonTools.Profiling {
                 if (_interpreter != value) {
                     _interpreter = value ?? _customInterpreter;
                     OnPropertyChanged("Interpreter");
-                    CanSpecifyInterpreterPath = (_interpreter == _customInterpreter);
+                    CanSpecifyInterpreterPath = (_interpreter is CustomPythonInterpreterView);
                 }
             }
         }
@@ -164,9 +164,8 @@ namespace Microsoft.PythonTools.Profiling {
             set {
                 if (_interpreterPath != value) {
                     _interpreterPath = value;
-                    if (Interpreter.Name == Strings.LaunchProfiling_OtherInterpreter && Interpreter.Path == null) {
-                        Interpreter = new PythonInterpreterView(Strings.LaunchProfiling_OtherInterpreter, Strings.LaunchProfiling_OtherInterpreter, _interpreterPath);
-                        CanSpecifyInterpreterPath = (_interpreter.Name == _customInterpreter.Name);
+                    if (Interpreter is CustomPythonInterpreterView) {
+                        Interpreter = new CustomPythonInterpreterView(_interpreterPath);
                     }
                     OnPropertyChanged("InterpreterPath");
                 }
--- a/Python/Product/Profiling/Profiling/StandaloneTargetView.cs
+++ b/Python/Product/Profiling/Profiling/StandaloneTargetView.cs
@@ -108,7 +108,7 @@ namespace Microsoft.PythonTools.Profiling {
             if (IsValid) {
                 return new StandaloneTarget {
                     PythonInterpreter = CanSpecifyInterpreterPath ? null : Interpreter.GetInterpreter(),
-                    InterpreterPath = CanSpecifyInterpreterPath ? InterpreterPath : null,
+                    InterpreterPath = CanSpecifyInterpreterPath ? null : InterpreterPath,
                     Script = ScriptPath ?? string.Empty,
                     WorkingDirectory = WorkingDirectory ?? string.Empty,
                     Arguments = Arguments ?? string.Empty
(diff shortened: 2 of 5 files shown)


Python/Product/Profiling/Profiling/StandaloneTargetView.cs
```diff
@@ -164,6 +164,9 @@ public StandaloneTargetView(IServiceProvider serviceProvider, StandaloneTarget t
             set {
                 if (_interpreterPath != value) {
                     _interpreterPath = value;
+                    if (Interpreter.Name == Strings.LaunchProfiling_OtherInterpreter && Interpreter.Path == null) {
+                        Interpreter = new PythonInterpreterView(Strings.LaunchProfiling_OtherInterpreter, "Global|"+ Strings.LaunchProfiling_OtherInterpreter, _interpreterPath);
```
AdamYoblick, line 168: Where did this `Global|` string come from?

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -286,13 +286,16 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
 
         private static void ProfileStandaloneTarget(SessionNode session, StandaloneTarget runTarget, bool openReport) {
             LaunchConfiguration config;
-            if (runTarget.PythonInterpreter != null) {
+            if (runTarget.PythonInterpreter != null && runTarget.PythonInterpreter.Id != "Global|Other...") {
```
AdamYoblick, line 289: Anything in `Strings.` gets localized, which means `Other` might not be the right thing to compare against. You should be using `Strings.LaunchProfiling_OtherInterpreter` for the second part of the string, like you did in `StandaloneTargetView.cs`

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
AdamYoblick: I would consider making this special string combination a static readonly string somewhere, since you use it more than once.

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -286,13 +286,16 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
...
                 var registry = session._serviceProvider.GetComponentModel().GetService<IInterpreterRegistryService>();
                 var interpreter = registry.FindConfiguration(runTarget.PythonInterpreter.Id);
                 if (interpreter == null) {
                     return;
                 }
                 config = new LaunchConfiguration(interpreter);
+            } else if (runTarget.PythonInterpreter.Id == "Global|Other...") {
```
AdamYoblick, line 296: Same as above. Hardcoded strings are usually bad practice. This should be reading from a static string somewhere, and `Other...` should come from the localized strings.

Python/Product/Profiling/PythonProfilingPackage.cs
AdamYoblick, line 289: I would actually nest these conditionals because they are checking two different things, and the final `else` is really only supposed to run if the interpreter is null. So it's technically correct, but it's easy to introduce a bug here.

Something like this:

```csharp
if (runTarget.PythonInterpreter != null) {
  if (runTarget.PythonInterpreter.Id != "Global|Other...") {
    ...
  } else {
    ...
  }
else {
  config = new LaunchConfiguration(null);
}
```

AdamYoblick: (and again, `"Global|Other..."` should not be hardcoded here, it should be checking against a static string)

AdamYoblick (changes requested):
See previous comments about hardcoded strings 😄

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
StellaHuang95: Good point, I used `Global` here because that's what all the other existing interpreters have in their names. Others also use it to search for corresponding `InterpreterFactoryProvider` here: https://github.com/microsoft/PTVS/blob/84add9c25ff26c1f39ab998400a8a66fb16ee60f/Python/Product/VSInterpreters/Interpreter/InterpreterRegistryService.cs#L304

But I can delete it since we are not searching by id.

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
```diff
@@ -108,7 +108,7 @@ public StandaloneTargetView(IServiceProvider serviceProvider, StandaloneTarget t
             if (IsValid) {
                 return new StandaloneTarget {
                     PythonInterpreter = CanSpecifyInterpreterPath ? null : Interpreter.GetInterpreter(),
-                    InterpreterPath = CanSpecifyInterpreterPath ? InterpreterPath : null,
+                    InterpreterPath = CanSpecifyInterpreterPath ? null : InterpreterPath,
```
int19h, line 111: It feels like the original code here intended either one or the other property to be null, depending on whether `CanSpecifyInterpreterPath` is set. Is that changing by design?

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
```diff
@@ -164,6 +164,9 @@ public StandaloneTargetView(IServiceProvider serviceProvider, StandaloneTarget t
             set {
                 if (_interpreterPath != value) {
                     _interpreterPath = value;
+                    if (Interpreter.Name == Strings.LaunchProfiling_OtherInterpreter && Interpreter.Path == null) {
```
int19h, line 167: I don't think interpreter names are intended to be unique, so they shouldn't be used for checks like these. It's what `Id` is supposed to be for, but you don't always get that here...

So, I think it would be better to somehow mark the `PythonInterpreterView` for the "Other" case. The most straightforward way to model this would be to derive `CustomPythonInterpreterView` - then you can use that with `is` to check. You could then also replace this:
```
_customInterpreter = new PythonInterpreterView(Strings.LaunchProfiling_OtherInterpreter, "", null);
```
with just:
```
_customInterpreter = new CustomPythonInterpreterView();
```
and move the argument values into the constructor for `CustomPythonInterpreterView`.

int19h: In fact, better yet, you could move any logic that's conditional on whether it's a regular or a custom interpreter into the interpreter view class, making them virtual methods. For example, instead of the conditional in `ProfileStandaloneTarget`, you can have `virtual GetInterpreterConfiguration()` in `PythonInterpreterView`, and override it in `CustomPythonInterpreterView`.

StellaHuang95: Hi @int19h , I understand you want a separate class for the custom python interpreter, but I might still need to call the `CustomPythonInterpreterView` setter inside of the `_interpreterpath` setter to set the properties. Or do you have better suggestions? Thanks.

int19h: Since the property values don't have to be computed from context, you can just move them to the base constructor call:
```cs
class CustomPythonInterpreterView : PythonInterpreterView {
    CustomPythonInterpreterView():
        base(Strings.LaunchProfiling_OtherInterpreter, "", null) {
    }
}
```

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
StellaHuang95: Hi @int19h, the `InterpreterPath` is not used anywhere in the original code, because existing python interpreters just use id to find the configuration. However, the custom interpreter constructs `VisualStudioInterpreterConfiguration` using `runTarget.InterpreterPath`, that's why I swapped them so that actual values can be returned.

Python/Product/Profiling/PythonProfilingPackage.cs
```diff
@@ -287,12 +287,17 @@ protected override int CreateToolWindow(ref Guid toolWindowType, int id)
         private static void ProfileStandaloneTarget(SessionNode session, StandaloneTarget runTarget, bool openReport) {
             LaunchConfiguration config;
             if (runTarget.PythonInterpreter != null) {
-                var registry = session._serviceProvider.GetComponentModel().GetService<IInterpreterRegistryService>();
-                var interpreter = registry.FindConfiguration(runTarget.PythonInterpreter.Id);
-                if (interpreter == null) {
-                    return;
+                if (runTarget.PythonInterpreter.Id == Strings.LaunchProfiling_OtherInterpreter) {
```
StellaHuang95, line 290: @int19h I don't think I can use `is CustomPythonInterpreterView` here to distinguish custom vs regular interpreters. The `runTarget` is a `StandaloneTarget` object, and the `runTarget.PythonInterpreter` is a `PythonInterpreter` project. So I have to compare `id` here. Do you see better options? :)

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
int19h: Looking at the docstrings for the properties involved, I'm still confused about this. `CanSpecifyInterpreterPath` is basically whether the textbox for that is enabled in the UI or not, right? So if it is `true`, shouldn't we be returning the target with `InterpreterPath` set from the property (which in turn gets it from the textbox)? In which case it would also make sense to have `PythonInterpreter` set to null, as it was before the change.

This would then allow `ProfileStandaloneTarget` to distinguish `PythonInterpreter` objects based solely on `Interpreter == null`. In retrospect, I think that was actually the original intent of this code as it was written. Perhaps the real bug here all along was that `PythonInterpreter` wasn't being set correctly in this case?

StellaHuang95: Your comments make sense to me. But I don't think the bug was `PythonInterpreter` wasn't being set correctly. It only has one attribute, which is `id`. `ProfileStandaloneTarget` won't be able to config the custom interpreter with only the `id`, it needs the `InterpreterPath` returned from `GetTarget()`. But I do agree `CanSpecifyInterpreterPath`  is a bit confusing here, the good thing is that it's not breaking anything. Maybe I can just git rid of this check `InterpreterPath = CanSpecifyInterpreterPath ? null : InterpreterPath` here and just return `InterpreterPath` since it's not affecting anything? Thanks again for taking a look Pavel!!

int19h: I meant that it was improperly set to a `PythonInterpreter` object with null `Id`, rather than to a null `PythonInterpreter` reference. Note that in `ProfileStandaloneTarget`, there's a check for that null, and a different branch is taken in that case: 
```cs
            LaunchConfiguration config;
            if (runTarget.PythonInterpreter != null) {
                var registry = session._serviceProvider.GetComponentModel().GetService<IInterpreterRegistryService>();
                var interpreter = registry.FindConfiguration(runTarget.PythonInterpreter.Id);
                if (interpreter == null) {
                    return;
                }
                config = new LaunchConfiguration(interpreter);
            } else {
                config = new LaunchConfiguration(null);
            }

            config.InterpreterPath = runTarget.InterpreterPath;
            config.ScriptName = runTarget.Script;
            config.ScriptArguments = runTarget.Arguments;
            config.WorkingDirectory = runTarget.WorkingDirectory;

            RunProfiler(session, config, openReport);
```

So, if `runTarget.Interpreter` is null, it would create a blank config, and then set its `InterpreterPath` properly. But for some reason, that null branch is never taken anymore.

StellaHuang95: Yeah that makes sense to me, I modified the code a little bit so it's working as expected now.

Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
```diff
@@ -25,7 +25,7 @@
 namespace Microsoft.PythonTools.Interpreter {
     public class LaunchConfigurationUtils {
         public static Dictionary<string, string> GetFullEnvironment(LaunchConfiguration config, IServiceProvider serviceProvider, UIThreadBase uiThread) {
-            if (config.Interpreter == null) {
+            if (config.Interpreter == null && config.InterpreterPath == null) {
```
int19h, line 28: Might be better to split this into two separate checks, so that you can throw a different exception using `nameof(InterpreterPath)` for the message.

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
```diff
@@ -166,6 +166,7 @@ public StandaloneTargetView(IServiceProvider serviceProvider, StandaloneTarget t
                     _interpreterPath = value;
                     if (Interpreter.Name == Strings.LaunchProfiling_OtherInterpreter && Interpreter.Path == null) {
                         Interpreter = new PythonInterpreterView(Strings.LaunchProfiling_OtherInterpreter, Strings.LaunchProfiling_OtherInterpreter, _interpreterPath);
+                        CanSpecifyInterpreterPath = (_interpreter.Name == _customInterpreter.Name);
```
int19h, line 169: In the setter for `Interpreter`, a similar check is written as `(_interpreter == _customInterpreter)`. I think the intent was to reuse `_customInterpreter` object in this scenario, rather than creating a new view every time.

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
StellaHuang95: @int19h Just updated the PR with `CustomPythonInterpreterView` 😄.

Python/Product/Profiling/Profiling/CustomPythonInterpreterView.cs
```diff
@@ -0,0 +1,28 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+namespace Microsoft.PythonTools.Profiling {
+    class CustomPythonInterpreterView : PythonInterpreterView {
+        public CustomPythonInterpreterView() :
+            base(Strings.LaunchProfiling_OtherInterpreter, "", null) {
+        }
+        public CustomPythonInterpreterView(string path) :
```
int19h, line 22: Nit: you can collapse these two ctors into a single one with a default value: `string path = null`.

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
```diff
@@ -56,7 +56,7 @@ sealed class StandaloneTargetView : INotifyPropertyChanged {
                 )
             ).ToList();
 
-            _customInterpreter = new PythonInterpreterView(Strings.LaunchProfiling_OtherInterpreter, "", null);
+            _customInterpreter = new CustomPythonInterpreterView();
```
int19h, line 59: Can you check if we're still using `_customInterpreter` anywhere other than this method? If not, it should become a local.

int19h (approved):

Python/Product/Profiling/Profiling/StandaloneTargetView.cs
StellaHuang95: It's only used inside this class and have been changed to a local variable.

PR #6990, opened 2022-05-05
Add python script to detect python architecture

### LINKED ISSUE #6775: Profiling: Error prompted , Python install from Microsoft Store.
**Describe the bug**

image

OS : Windows 11 Enterprise 21H2
Python install from Microsoft Store

**Steps to Reproduce**
1. Create python application project named PythonApplication2 and typing code in it.
2. Debug->Launch Python Profiling …
3. Checked the option "Open project" and select the current project in drop-down list.
4. Click "Start".

**Expected behavior**

Start without any errors.

image

StellaHuang95:
fixes #6775 

The issue was that PTVS tries to get the architecture of the python interpreter and use that to determine which dll to use for profiling, but it's getting an access denied when trying to call `kernel32.GetBinaryTypeW` when pointing at the windows store install. So I added extra call to get Python print out its architecture.

Note that the python script I added can only distinguish between 32bit and 64bit, but it can't distinguish between ARM64 and AMD64.

I have tested with 32bit python from python.org, 64bit python from python.org, python downloaded from Windows Store, python 3.6, 3.7, 3.8, 3.9, 3.10.  Hope that can capture all the scenarios.
Animation2 - pyarch

### DIFF
--- a/Python/Product/Profiling/Profiling/ProfiledProcess.cs
+++ b/Python/Product/Profiling/Profiling/ProfiledProcess.cs
@@ -30,7 +30,7 @@ namespace Microsoft.PythonTools.Profiling {
         private readonly ProcessorArchitecture _arch;
         private readonly Process _process;
         private readonly PythonToolsService _pyService;
-        public string pyArch { get; set; }
+        private string _pyArch;
         public ProfiledProcess(PythonToolsService pyService, string exe, string args, string dir, Dictionary<string, string> envVars) {
 
             string pythonInstallDir = Path.GetDirectoryName(PythonToolsInstallPath.GetFile("VsPyProf.dll", typeof(ProfiledProcess).Assembly));
@@ -38,12 +38,12 @@ namespace Microsoft.PythonTools.Profiling {
             var arg = ProcessOutput.QuoteSingleArgument(Path.Combine(pythonInstallDir, "getArch.py"));
             getPyArch(exe, arg);
 
-            if (pyArch == "64") {
+            if (_pyArch == "64") {
                 _arch = ProcessorArchitecture.Amd64;
-            } else if (pyArch == "32") {
+            } else if (_pyArch == "32") {
                 _arch = ProcessorArchitecture.X86;
             } else {
-                throw new InvalidOperationException(Strings.UnsupportedArchitecture.FormatUI(_arch));
+                throw new InvalidOperationException(Strings.UnsupportedArchitecture.FormatUI(_pyArch));
             }
 
             dir = PathUtils.TrimEndSeparator(dir);
@@ -113,7 +113,7 @@ namespace Microsoft.PythonTools.Profiling {
 
         void Process_OutputDataReceived(object sender, DataReceivedEventArgs e) {
             if (!string.IsNullOrEmpty(e.Data)) {
-                pyArch = e.Data;
+                _pyArch = e.Data;
             }  
         }
 
--- a/Python/Product/Profiling/Profiling/ProfiledProcess.cs
+++ b/Python/Product/Profiling/Profiling/ProfiledProcess.cs
@@ -38,12 +38,12 @@ namespace Microsoft.PythonTools.Profiling {
             var arg = ProcessOutput.QuoteSingleArgument(Path.Combine(pythonInstallDir, "getArch.py"));
             getPyArch(exe, arg);
 
-            if (_pyArch == "64") {
-                _arch = ProcessorArchitecture.Amd64;
-            } else if (_pyArch == "32") {
+            if (_pyArch.EndsWith("-arm64")) {
+                throw new InvalidOperationException(Strings.UnsupportedArchitecture.FormatUI(_pyArch));  
+            } else if (_pyArch.EndsWith("-32")) {
                 _arch = ProcessorArchitecture.X86;
             } else {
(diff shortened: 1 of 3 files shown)


Python/Product/Profiling/Profiling/ProfiledProcess.cs
```diff
@@ -30,11 +30,20 @@ sealed class ProfiledProcess : IDisposable {
         private readonly ProcessorArchitecture _arch;
         private readonly Process _process;
         private readonly PythonToolsService _pyService;
-
+        public string pyArch { get; set; }
```
AdamYoblick, line 33: C# properties should be PascalCase. Please capitalize this.

AdamYoblick: Also, if no other classes are using this, it should be private instead of public.

Python/Product/Profiling/Profiling/ProfiledProcess.cs
```diff
@@ -82,6 +89,34 @@ sealed class ProfiledProcess : IDisposable {
             _process.StartInfo = processInfo;
         }
 
+        public void getPyArch(string exe, string arg) {
```
AdamYoblick, line 92: C# methods should be PascalCase. Please capitalize this.

Python/Product/Profiling/Profiling/ProfiledProcess.cs
```diff
@@ -30,11 +30,20 @@ sealed class ProfiledProcess : IDisposable {
...
+
+            var arg = ProcessOutput.QuoteSingleArgument(Path.Combine(pythonInstallDir, "getArch.py"));
+            getPyArch(exe, arg);
+
+            if (pyArch == "64") {
+                _arch = ProcessorArchitecture.Amd64;
+            } else if (pyArch == "32") {
+                _arch = ProcessorArchitecture.X86;
+            } else {
+                throw new InvalidOperationException(Strings.UnsupportedArchitecture.FormatUI(_arch));
```
AdamYoblick, line 46: This error message doesn't make sense. It's supposed to be printing the arch that is unsupported, but you're printing the value of `_arch`, which hasn't been assigned if this code path is taken.

I believe you want to print `pyArch` here instead.

Python/Product/Profiling/Profiling/ProfiledProcess.cs
StellaHuang95: @AdamYoblick This value is also set in `Process_OutputDataReceived`, I can't find a way to have `Process_OutputDataReceived` return `e.Data` so I have to set `pyArch` there.

AdamYoblick: Ah that's fine, then don't make it readonly. Just make it a private field. No need to be public because it's not used by a different class, and no need to be a property because you're really not getting any value from the getter nor the setter.

Python/Product/Profiling/Profiling/ProfiledProcess.cs
```diff
@@ -82,6 +89,34 @@ sealed class ProfiledProcess : IDisposable {
...
+                    Arguments = arg,
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true,
+                    RedirectStandardError = true,
+                    CreateNoWindow = true
+                },
+                EnableRaisingEvents = true
+            };
+
+            process.ErrorDataReceived += Process_OutputDataReceived;
```
AdamYoblick, line 105: If you're not going to be doing anything complicated with the output, you can simplify the handlers by adding them inline, without the need for the callback. Like this:

```csharp
process.ErrorDataReceived += (s, e) => pyArch = e.Data;
process.OutputDataReceived += (s, e) => pyArch = e.Data;
```

Up to you :)

AdamYoblick: @zooba I believe you wanted us to determine arch in a different way. I had Stella just move ahead with what she had so she wouldn't be blocked. Can you please let us know how you would prefer arch to be checked in getArch.py? Thanks 😄

Python/Product/Profiling/getArch.py
```diff
@@ -0,0 +1,3 @@
+﻿import struct
+
+print(struct.calcsize("P")*8)
```
AdamYoblick, line 3: @zooba I had Stella move ahead with this just so she wouldn't be blocked, but I know this isn't exactly what you wanted. I figured we could fix this up in the PR, so here we are. Can you please remind us how you'd prefer this to be done here? Thanks!

Python/Product/Profiling/Profiling/ProfiledProcess.cs
StellaHuang95: ok, done.

Python/Product/Profiling/getArch.py
zooba: Check for `sys.winver.endswith('-32')` or `sys.winver.endswith('-arm64')`.

AdamYoblick: And it looks like amd64 will just return the major.minor. I ran `sys.winver` on my 64-bit python install and I got `3.9`. So those cases should be enough to determine the arch of the python exe

AdamYoblick (approved):

PR #7002, opened 2022-05-11
Fix compile errors in old VS 16.11 branch

AdamYoblick:
This branch doesn't currently compile, these changes fix that.


StellaHuang95 (approved):

AdamYoblick:
The Dev17 PR build check is probably not going to pass for this, since this is 16.11. The change is very trivial though, and it compiles locally for both me and Stella, so I'm going to merge this in so the branch compiles.

PR #7010, opened 2022-05-19
Fix VS crash when user's Python version is not in PythonLanguageVersion enum

### LINKED ISSUE #7009: VS crashes when python interpreter version is not in PythonLanguageVersion enum
This came from a watson crash, tracked at https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1541522

An InvalidOperationException is thrown when comparing the version of a python interpreter on disk to an enum of all the python versions we support. The reported version of "4.5.5" is definitely not in the enum, so we get:

- Python\Product\Common\Parsing\PythonLanguageVersion.cs -> throws InvalidOperationException
- Python\Product\VSInterpreters\Interpreter\AstPythonInterpreterFactory.cs -> catches the above exception and throws an ArgumentException, which is unhandled and crashes VS

A more graceful way to fix this would be:

If ToLanguageVersion() throws an exception:
1. Replace the ArgumentException with a Debug.Assert so we still see the problem when debugging
2. Set LanguageVersion to LanguageVersion.None, and handle that case from the caller.

For example, in the code that is enumerating all the environments to display them, just ignore the ones where version = PythonVersion.None, because those are bad.
Maybe print something to the VS output window so the user understands why their environment is missing?

AdamYoblick:
Fixes #7009

To repro this, I installed python 3.11 beta from python.org. Then I commented out `V311` from the `PythonLanguageVersion` enum and verified that `ToLanguageVersion()` returned `PythonLanguageVersion.None`, which is what we would expect if the version is not supported. This breakpoint gets hit when the installed environments are being enumerated (which is used to populate the environments window).

Then I added python 3.11 as an existing environment and activated it. I saw the environment in the project just fine. I got a warning in the debugger output about breakpoints might not hit, but they did. Profiling didn't work, but the error is "graceful"

image

So I think this fix is viable. VS no longer crashes, I'm able to add 3.11 as an existing environment, as well as a virtual environment and debug with both.

I still don't know why the version was being reported as 4.5.5 in the watson crash, but I believe our code can handle this edge case now.

### DIFF
--- a/Python/Product/Common/Parsing/PythonLanguageVersion.cs
+++ b/Python/Product/Common/Parsing/PythonLanguageVersion.cs
@@ -59,8 +59,9 @@ namespace Microsoft.PythonTools.Common.Parsing {
             if (Enum.IsDefined(typeof(PythonLanguageVersion), value)) {
                 return (PythonLanguageVersion)value;
             }
-            throw new InvalidOperationException("Unsupported Python version: {0}".FormatInvariant(version));
+            else {
+                return PythonLanguageVersion.None;
+            }
         }
-
     }
 }
--- a/Python/Product/Common/Parsing/PythonLanguageVersion.cs
+++ b/Python/Product/Common/Parsing/PythonLanguageVersion.cs
@@ -15,6 +15,7 @@
 // permissions and limitations under the License.
 
 using System;
+using System.Diagnostics;
 using Microsoft.PythonTools.Common.Core;
 using Microsoft.PythonTools.Common.Core.Extensions;
 
@@ -60,6 +61,7 @@ namespace Microsoft.PythonTools.Common.Parsing {
                 return (PythonLanguageVersion)value;
             }
             else {
+                Trace.WriteLine(Strings.PythonVersionNotSupportedTraceText.FormatUI(version));
                 return PythonLanguageVersion.None;
             }
         }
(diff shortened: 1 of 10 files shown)


Python/Product/Common/Parsing/PythonLanguageVersion.cs
```diff
@@ -59,8 +59,9 @@ public static bool IsImplicitNamespacePackagesSupported(this PythonLanguageVersi
             if (Enum.IsDefined(typeof(PythonLanguageVersion), value)) {
                 return (PythonLanguageVersion)value;
             }
-            throw new InvalidOperationException("Unsupported Python version: {0}".FormatInvariant(version));
+            else {
+                return PythonLanguageVersion.None;
```
int19h, line 63: Might be worth adding a `Trace.WriteLine` or something like that here.

int19h (approved):

Python/Product/Common/Parsing/PythonLanguageVersion.cs
AdamYoblick: What's the difference between `Debug.WriteLine` and `Trace.WriteLine`? I know the debug one only runs when debugging, but does Trace write to the VS output window?

int19h: The only difference is that `Debug` is `[Conditional("DEBUG")]`, and `Trace` is not. Both write to whatever registered trace listeners there are. Under VS, both would go to the Output window.

PR #6978, opened 2022-04-26
Don't try to show conda environments that do not have python installed

AdamYoblick:
Fixes #6977 

This bug only happens when the user manually installs a conda environment, but doesn't install python that environment. This fix makes it so conda environments that do not have python installed do not show up anywhere. Installing python in the environment and reloading the solution will cause the environment to show up as normal.

PR #7019, opened 2022-05-26
fix VS crash with bad conda environments

### LINKED ISSUE #7013: VS crash in CondaEnvironmentFactoryProvider (Watson)
This is from a watson bug at https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1540179

AdamYoblick:
Fixes #7013

I was unable to repro this bug, and Watson didn't really give me any more information besides the line that causes the exception. So I just made sure we don't allow the conda environment (nor its associated configuration) to be null. This will prevent the NRE from being thrown.


int19h (approved):

debonte:
I dug into one of the dump files and confirmed that the list you're getting back from `FindCondaEnvironments` contains a null `PythonInterpreterInformation` object. It doesn't look like that's possible anymore. My guess is that this was already fixed by https://github.com/microsoft/PTVS/pull/6978.

PR #7023, opened 2022-06-02
Fix solution load when project contains windows store python environment

AdamYoblick:
When trying to open a solution containing a windows store python environment, we get an `Access is denied` when trying to run the python.exe within it. There's no good workaround for this that I know of. This prevents the project from loading in the solution explorer, and the user has to manually reload it.

So this change catches the `Win32Exception` thrown when this happens and defaults to no search paths, similar to what happens if the `get_search_paths.py` file isn't found. This causes the project to load correctly.

### DIFF
--- a/Python/Product/Common/Core/OS/ProcessServices.cs
+++ b/Python/Product/Common/Core/OS/ProcessServices.cs
@@ -15,6 +15,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


Python/Product/Common/Core/OS/ProcessServices.cs
```diff
@@ -39,18 +40,23 @@ public sealed class ProcessServices : IProcessServices {
...
-                } catch (IOException) { } catch (OperationCanceledException) { }
+                    try {
+                        output = await process.StandardOutput.ReadToEndAsync();
+                        await process.WaitForExitAsync(30000, cancellationToken);
+                    } catch (IOException) { } catch (OperationCanceledException) { }
 
+                    return output;
+                }
+            // Handle the case when trying to call python from the Microsoft store and we get access denied
+            } catch (Win32Exception) {
```
int19h, line 59: This will also silence errors when trying to run various tools (e.g. when right-clicking the project node in Solution Explorer and using the "Python" submenu), making them seemingly succeed with nothing in the output to indicate what went wrong. It might be better to have a separate helper wrapper that does this, and use it on a case by case basis where needed.

AdamYoblick: Ah you're right. I only saw two references to this but I didn't realize it's implementing an interface method, so my mistake.

In that case, the easiest thing to do would probably be to catch a Win32Exception in `PythonLibraryPath.GetSearchPathsFromInterpreterAsync()`, where this method is called from. Thoughts?

int19h: It is a fairly common pattern for us to invoke Python to gather some information from it, so there might be a few other places where it'd need to be handled - just not all of them. If it's really just this one, then yeah, no point in making a wrapper just to call it once.

AdamYoblick: I agree, we call into python all over the place. That probably needs to be refactored into common code. But I think that change is beyond the scope of this fix, as I just noticed this fix today when messing with another bug. 😄

int19h (approved):

PR #7053, opened 2022-06-21
Output the error when git is not installed or not on the PATH

### LINKED ISSUE #6872: Cannot use Cookiecutter to load template.
**Describe the bug**

image

**Steps to Reproduce**

1. Open Cookiecutter Explorer.
2. Load the following template: https://github.com/brettcannon/python-azure-web-app-cookiecutter 

**Expected behavior**

The Template loaded successfully.

Error cloning the repository.
BTW, load the template from the "Recommended" section, you can repro it.

**Output info**
```
----- Cloning template 'https://github.com/brettcannon/python-azure-web-app-cookiecutter' -----
The system cannot find the file specified
A task was canceled.
----- Failed to clone template 'https://github.com/brettcannon/python-azure-web-app-cookiecutter' -----
```

StellaHuang95:
fixes #6872 

Print out the error when user fails to clone Cookiecutter templates because git is not installed or not on the PATH.

Here's what happens after the fix:

When git is installed and on the PATH:
Animation2 - with git

When it's not installed or on the PATH:
Animation2 - without git

### DIFF
--- a/Python/Product/Cookiecutter/Model/GitClient.cs
+++ b/Python/Product/Cookiecutter/Model/GitClient.cs
@@ -55,11 +55,11 @@ namespace Microsoft.CookiecutterTools.Model {
             var arguments = new string[] { "clone", repoUrl };
             try {
                 // check if it's actually git
-                if(!canRunGit(_gitExeFilePath)) {
+                if(!CanRunGit(_gitExeFilePath)) {
                     throw new GitRunException(Strings.GitFailedToRunError);
                 }
-            } catch (Exception) {
-                throw new GitRunException(Strings.GitFailedToRunError);
+            } catch (Exception ex) {
+                throw new GitRunException(Strings.GitFailedToRunError, ex);
             }
             using (var output = ProcessOutput.Run(_gitExeFilePath, arguments, targetParentFolderPath, GetEnvironment(), false, redirector)) {
                 await output;
@@ -91,8 +91,8 @@ namespace Microsoft.CookiecutterTools.Model {
             }
         }
 
-        public bool canRunGit(string exe) {
-            var process = new Process {
+        private static bool CanRunGit(string exe) {
+            using (var process = new Process {
                 StartInfo = new ProcessStartInfo {
                     FileName = exe,
                     Arguments = "--version",
@@ -101,16 +101,16 @@ namespace Microsoft.CookiecutterTools.Model {
                     RedirectStandardError = true,
                     CreateNoWindow = true,
                 }
-            };
+            }) {
+                process.Start();
+                process.WaitForExit();
 
-            process.Start();
-            process.WaitForExit();
-            
-            var output = process.StandardOutput.ReadLine();
+                var output = process.StandardOutput.ReadLine();
 
-            if (output != null && output.StartsWith("git version")) {
-                return true;
-            }
+                if (output != null && output.StartsWith("git version")) {
+                    return true;
+                }
+            }   
             return false;
         }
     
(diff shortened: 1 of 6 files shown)


Python/Product/Cookiecutter/Model/GitClient.cs
```diff
@@ -52,6 +52,9 @@ class GitClient : IGitClient {
             }
 
             var arguments = new string[] { "clone", repoUrl };
+            if (!ExistsOnPath(_gitExeFilePath)) {
+                throw new FileNotFoundException();
```
AdamYoblick: As Pavel mentioned offline, it would be better to run `git verion` to make sure git can actually run. No need to check the path.

Also, I believe VS now comes with git built in, so using assuming git.exe is on the path no longer seems necessary, we can just use the one that comes with VS. But if you don't want to make that change, I'm fine with just trying to run git.

Python/Product/Cookiecutter/ViewModel/CookiecutterViewModel.cs
```diff
@@ -843,6 +843,9 @@ class CookiecutterViewModel : INotifyPropertyChanged {
                     CloningStatus = OperationStatus.Failed;
 
                     _outputWindow.WriteErrorLine(ex.Message);
+                    if (ex.GetType().Name == "FileNotFoundException") {
```
AdamYoblick: No need to do this extra work. Just throw an exception and include the error string within it, like this: `throw new Exception(Strings.GitNotFoundError)`. Then like 845 will print the error message you need and you can get rid of this entire code block you added.

AdamYoblick (approved):

Python/Product/Cookiecutter/Model/GitClient.cs
```diff
@@ -82,6 +91,29 @@ class GitClient : IGitClient {
             }
         }
 
+        public bool canRunGit(string exe) {
```
AdamYoblick, line 94: Methods should be in PascalCase. Please rename to `CanRunGit`.

AdamYoblick: Also there's probably no need for this method to be `public`. Please make it `private`. It can also be `static` because it doesn't interact with any instance members.

Python/Product/Cookiecutter/Model/GitClient.cs
```diff
@@ -52,8 +53,13 @@ class GitClient : IGitClient {
...
             var arguments = new string[] { "clone", repoUrl };
-            if (!ExistsOnPath(_gitExeFilePath)) {
-                throw new FileNotFoundException();
+            try {
+                // check if it's actually git
+                if(!canRunGit(_gitExeFilePath)) {
+                    throw new GitRunException(Strings.GitFailedToRunError);
+                }
+            } catch (Exception) {
+                throw new GitRunException(Strings.GitFailedToRunError);
```
int19h, line 62: This should include the original exception as the inner exception for the one that it throws. This way, if it escapes and gets reported in a log, the full context is available.

Python/Product/Cookiecutter/Model/GitClient.cs
```diff
@@ -83,25 +89,31 @@ class GitClient : IGitClient {
...
-            if (File.Exists(fileName)) {
-                return Path.GetFullPath(fileName);
-            }
-            var values = Environment.GetEnvironmentVariable("PATH");
-            foreach (var path in values.Split(Path.PathSeparator)) {
-                var fullPath = Path.Combine(path, fileName);
-                if (File.Exists(fullPath))
-                    return fullPath;
+        public bool canRunGit(string exe) {
+            var process = new Process {
```
int19h, line 95: `Process` is disposable, so this needs a `using` variable declaration or block to ensure cleanup.

int19h (approved):

PR #7066, opened 2022-06-30
Add localization support to Options -> Python -> Debugging -> variable group categories.

### LINKED ISSUE #6727: Python tools -> options variable group categories are not localized
**Describe the bug**
Python tools -> options variable group categories are not localized

**Steps to Reproduce**
1. Install VS2022 int preview (https://aka.ms/vs/17/intpreview/vs_enterprise.exe)
2. Install Python workload and another language pack (spanish, for example)
3. Start VS and change the language (Tools -> Options -> International Settings)
4. Restart VS
5. Re-browse to Tools -> Options -> International Settings
6. Notice the strings in the dropdown are not localized:

image

**Expected behavior**
The values in the dropdown should be localized

image

StellaHuang95:
fixes #6727

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Options/PythonDebuggingOptionsControl.cs
+++ b/Python/Product/PythonTools/PythonTools/Options/PythonDebuggingOptionsControl.cs
@@ -22,13 +22,18 @@ namespace Microsoft.PythonTools.Options {
     public partial class PythonDebuggingOptionsControl : UserControl {
         public PythonDebuggingOptionsControl() {
             InitializeComponent();
+            var presentationModes = new List<KeyValuePair<PresentationMode, string>>() {
+                new KeyValuePair<PresentationMode, string>(PresentationMode.Group, Strings.VariablePresentation_Group),
+                new KeyValuePair<PresentationMode, string>(PresentationMode.Hide, Strings.VariablePresentation_Hide),
+                new KeyValuePair<PresentationMode, string>(PresentationMode.Inline, Strings.VariablePresentation_Inline),
+            };
 
             // set allowed values for the variable presentation options
             var varPresComboBoxes = new List<ComboBox> { _showVariablesClassComboBox, _showVariablesFunctionComboBox, _showVariablesProtectedComboBox, _showVariablesSpecialComboBox };
-            foreach (var varPresComboBox in varPresComboBoxes) {
-                varPresComboBox.Items.Add(Strings.VariablePresentation_Group);
-                varPresComboBox.Items.Add(Strings.VariablePresentation_Hide);
-                varPresComboBox.Items.Add(Strings.VariablePresentation_Inline);
+            foreach(var varPresComboBox in varPresComboBoxes) {
+                varPresComboBox.DataSource = new BindingSource(presentationModes, null);
+                varPresComboBox.ValueMember = "Key";
+                varPresComboBox.DisplayMember = "Value";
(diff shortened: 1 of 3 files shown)


Python/Product/PythonTools/PythonTools/Options/PythonDebuggingOptionsControl.cs
```diff
@@ -40,10 +42,47 @@ public partial class PythonDebuggingOptionsControl : UserControl {
...
+                _showVariablesSpecialComboBox.SelectedIndex = 0;
+            } else if (pyService.DebuggerOptions.VariablePresentationForSpecial == PresentationMode.Hide) {
+                _showVariablesSpecialComboBox.SelectedIndex = 1;
+            } else if (pyService.DebuggerOptions.VariablePresentationForSpecial == PresentationMode.Inline) {
+                _showVariablesSpecialComboBox.SelectedIndex = 2;
+            } else {
+                _showVariablesSpecialComboBox.SelectedIndex = 0;
+            }
+
+
```
int19h, line 85: To avoid repetition here and below, it might be better to use WinForms data binding instead. If you create a mapping like so:
```cs
var presentationModes = new Dictionary<PresentationMode, string> {
   { PresentationMode.Group, Strings.VariablePresentation_Group },
    ...
};
```
Then you can bind every combobox in the same exact way:
```cs
foreach (var comboBox in new[] { _showVariablesClassComboBox, ...  }) {
  comboBox.DataSource = presentationModes;
  comboBox.ValueMember = "Key";
  comboBox.DisplayMember = "Value";
}
```
And use `SelectedValue` to get/set the selection for each:
```cs
_showVariablesClassComboBox.SelectedValue = pyService.DebuggerOptions.VariablePresentationForClasses;
...
```

StellaHuang95: ok thank you! That makes the code much cleaner. I will update that. 😊

StellaHuang95: @int19h Hi Pavel, assigning dictionary to `comboBox.DataSource` is giving me "Complex DataBinding accepts as a data source either an IList or an IListSource" error. So I had to convert the dictionary to a `BindingSource` to make it work.

StellaHuang95: @int19h oh wait, I forgot to switch the list of KeyValuePair back to dictionary. Let me change that real quick!

int19h (approved):

StellaHuang95:
> Hi Stella, Did you test this to make sure the variable presentation still works in the debugger?

@AdamYoblick Hi Adam, yes I tested to make sure debugger behaves the same as before I made all the changes, but not necessarily the same as [#6337](https://github.com/microsoft/PTVS/pull/6337). For example, in your old PR, you were seeing variables in the Auto window, but in the latest VS there're only variables in the Locals window.

I can show you the Locals windows before and after I made my changes:

1. All Group, before vs after
image

2. All Inline, before vs after
image

3. All Hide, before vs after
image

The behaviors are the same before and after the changes, but I do notice that setting all the variables to "Hide" didn't really hide all variables, is that a separate issue, or that's just how debugger is supposed to behave?

AdamYoblick (approved):

Issue #7043, opened 2022-06-09
The result of Test discovery was incorrect.


ttSpace:
**Describe the bug**

![image](https://user-images.githubusercontent.com/82852072/172789728-4ab27026-11fb-493a-9ff6-4f55fe312567.png)

**Steps to Reproduce**

1. Create Python application project.
2. Right-click project, select Add > New Item …, then select "Python Unit Test" followed by "Add".
3. Add test_1.py file with code
```
import unittest

class Test_test_1(unittest.TestCase):
    def test_A(self):
        self.fail("Not implemented")

if __name__ == '__main__':
    unittest.main()

```
4. Set Test framework as "unittest" in python project properties page.
5. Open TE with the Test > Windows > Test Explorer menu command

**Expected behavior**

The result of Test discovery was correct.

**Additional context and screenshots**

The result of Test discovery was incorrect.
There is only one test but it shows that two were found.

![MicrosoftTeams-image](https://user-images.githubusercontent.com/82852072/172789972-8778f8c8-571b-4c3c-850f-50ef7fff1d2d.png)

StellaHuang95:
Might have to do with parallel test discovery that was introduced in preview 2. Filed an internal ticket and can be tracked here: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1560000

PR #7086, opened 2022-07-19
Fix over-discovery issue for unittest and pytest.

StellaHuang95:
fixes https://github.com/microsoft/PTVS/issues/7043

The VS testing team has recently changed test discovery such that on machines with multiple CPU cores, it can discover tests from different `sources` in parallel inside different test host processes. We have been over discovering tests because our adapter would unnecessarily (re)discover tests in all the .py files in each process. So for example, if we have 3 sources and 2 tests files, our adapter is going to return 3 * 2 tests discovered in total.

The fix is to just have our test discoverer get the list of all tests in the test root directory regardless, and then manually filter based on the test container/source.

I ran some tests for different scenarios and it's working as expected.

- Unittest, project view
unittest-project

- Unittest, folder view
unittest-folder

- Pytest, project view
pytest - project

- Pytest, folder view
pytest - folder

The only issue is that test explorer will throw an exception if no test is discovered and returned in certain test containers, but it is on the Test Explorer side.
no test available


Python/Product/TestAdapter.Executor/Pytest/PytestTestDiscoverer.cs
```diff
@@ -59,16 +59,16 @@ ITestCaseDiscoverySink discoverySink
...
 
             var env = InitializeEnvironment(sources, settings);
             var outputFilePath = Path.GetTempFileName();
             var arguments = GetArguments(sources, settings, outputFilePath);
 
-            LogInfo("cd " + settings.WorkingDirectory);
-            LogInfo("set " + settings.PathEnv + "=" + env[settings.PathEnv]);
-            LogInfo($"{settings.InterpreterPath} {string.Join(" ", arguments)}");
+            //LogInfo("cd " + settings.WorkingDirectory);
+            //LogInfo("set " + settings.PathEnv + "=" + env[settings.PathEnv]);
```
bschnurr, line 70: i think we should add back logs

StellaHuang95: @bschnurr But that will be printed out for every source, is that ok? 10 test containers will print 10 times.

Python/Product/TestAdapter.Executor/UnitTest/UnitTestTestDiscoverer.cs
```diff
@@ -138,16 +138,22 @@ ITestCaseDiscoverySink discoverySink
...
-            arguments.Add("discover");
-            arguments.Add("unittest");
-            arguments.Add("--output-file");
-            arguments.Add(outputfilename);
-            //Note unittest specific options go after this separator
-            arguments.Add("--");
-            arguments.Add(settings.UnitTestRootDir);
-            arguments.Add(settings.UnitTestPattern);
-
+            foreach (var source in sources.ToArray()) {
```
bschnurr, line 141: if we are calling 'unitTest' for every source file. shouldn't we pass in the source file?

Python/Product/TestAdapter.Executor/Pytest/PytestTestDiscoverer.cs
```diff
@@ -43,7 +43,7 @@ public class PytestTestDiscoverer : PythonTestDiscoverer {
             
         }
 
-        public override void DiscoverTests(
+        public override void DiscoverTestsForProject(
```
bschnurr, line 46: you tested both 'open folder' and 'open project' ?

Python/Product/TestAdapter.Executor/Pytest/PytestTestDiscoverer.cs
```diff
@@ -59,16 +59,16 @@ ITestCaseDiscoverySink discoverySink
...
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
-            var workspaceText = settings.IsWorkspace ? Strings.WorkspaceText : Strings.ProjectText;
-            LogInfo(Strings.PythonTestDiscovererStartedMessage.FormatUI(PythonConstants.PytestText, settings.ProjectName, workspaceText, settings.DiscoveryWaitTimeInSeconds));
+            //var workspaceText = settings.IsWorkspace ? Strings.WorkspaceText : Strings.ProjectText;
+            //LogInfo(Strings.PythonTestDiscovererStartedMessage.FormatUI(PythonConstants.PytestText, settings.ProjectName, workspaceText, settings.DiscoveryWaitTimeInSeconds));
 
             var env = InitializeEnvironment(sources, settings);
             var outputFilePath = Path.GetTempFileName();
             var arguments = GetArguments(sources, settings, outputFilePath);
```
bschnurr, line 67: dont we want to change 'GetArguments' so that it only looks for tests in a single source file now?

Python/Product/TestAdapter.Executor/Pytest/PytestTestDiscoverer.cs
bschnurr: maybe instead of "info' there is a 'verbose' or diagnostic flag

Python/Product/TestAdapter.Executor/UnitTest/UnitTestTestDiscoverer.cs
StellaHuang95: @bschnurr yes you are right, there's only 1 element in the `sources.ToArray()`, so instead of a for loop, maybe I should just change to `sources.ToArray()[0]`.

Python/Product/TestAdapter.Executor/Pytest/PytestTestDiscoverer.cs
StellaHuang95: @bschnurr We might be able to do that for pytest so that it only discovers 1 test at a time, but for unittest the argument has to be a directory, can't be a path to a single file.

Python/Product/TestAdapter.Executor/UnitTest/UnitTestTestDiscoverer.cs
```diff
@@ -101,7 +101,7 @@ ITestCaseDiscoverySink discoverySink
 
                 foreach (var tc in testcases) {
                     // Note: Test Explorer will show a key not found exception if we use a source path that doesn't match a test container's source.
-                    if (settings.TestContainerSources.TryGetValue(tc.CodeFilePath, out _)) {
+                    if (tc.CodeFilePath == sources.ToArray()[0] && settings.TestContainerSources.TryGetValue(tc.CodeFilePath, out _)) {
```
AdamYoblick, line 104: Same comment as above. Consider making a little cleaner using linq.

Python/Product/TestAdapter.Executor/UnitTest/UnitTestTestDiscoverer.cs
AdamYoblick, line 141: You don't need call `ToArray()` on an `IEnumerable<string>` to enumerate it. Consider removing this.

AdamYoblick: And yes, if there's always guaranteed to be one element, you can get rid of the loop and only use the single element.

Python/Product/TestAdapter.Executor/UnitTest/UnitTestTestDiscoverer.cs
```diff
@@ -138,16 +138,22 @@ ITestCaseDiscoverySink discoverySink
...
-            arguments.Add(outputfilename);
-            //Note unittest specific options go after this separator
-            arguments.Add("--");
-            arguments.Add(settings.UnitTestRootDir);
-            arguments.Add(settings.UnitTestPattern);
-
+            foreach (var source in sources.ToArray()) {
+                var testRootDir = PathUtils.GetAbsoluteDirectoryPath(settings.ProjectHome, settings.UnitTestRootDir);
+                // only return valid arguments when source is a subpath of test root directory
+                if (PathUtils.IsSamePath(PathUtils.GetParent(source), testRootDir)) {
```
AdamYoblick, line 144: This will only work if the source is one directory under the testRootDir. Try putting tests in a deeper folder and I think this will fail.

You might need to use `PathUtils.IsSubpathOf(testRootDir, PathUtils.GetParent(source))` instead

AdamYoblick: I spoke to Stella offline about this and this is an existing bug, even without her changes. She's going to try to fix it as part of this PR because it's so closely related to what this PR is trying to fix. 

But if it becomes too much feature-creep, we'll just create a new issue for it and it can get done later.

PR #7096, opened 2022-07-27
Add environment name input validation when creating a new virtual environment

### LINKED ISSUE #7089: Visual Studio crashes when adding a new Python virtual environment and typing a slash in the `Name` field
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Visual-Studio-crashes-when-adding-a-new-/10093909)._

---
I wanted to add a Python virtual environment to my project under a subdirectory called `.venv`, so that I can add the `.venv` to my `.gitignore` file. I attempted to add a new virtual environment called `.venv\my_virtual_environment_v3.10.5`, but Visual Studio crashes as soon as I type the slash. See screenshot:

---
#### Feedback Bot on 7/13/2022, 01:36 AM: 

(private comment, text removed)
#### Ed Peguillan III on 7/13/2022, 01:40 AM: 

(private comment, text removed)

---
### Original Solutions
(no solutions)

AdamYoblick:
Fixes #7089

The bug reported was an infinite loop that causes a stack overflow when a backslash was present in the environment name. However, this was just one possible bad outcome - there are many others. The real problem is that we weren't doing any input validation on the environment name, which allowed the user to do bad things (like crash Visual Studio).

This PR does the following
- Enable validation on the env name text box
- Display validation errors to the user
- Disable the `Create` button if there are validation errors, enable it if there are not

Here's what it looks like:

EnvNameValidation


Python/Product/PythonTools/PythonTools/Environments/AddVirtualEnvironmentControl.xaml
```diff
@@ -142,10 +142,13 @@
                                     Converter={StaticResource ScreenAdjustingConverter}, 
                                     ConverterParameter=249}"
                                 HorizontalAlignment="Left"
-                                Text="{Binding Path=VirtualEnvName,Mode=TwoWay,UpdateSourceTrigger=PropertyChanged,ValidatesOnNotifyDataErrors=False}"
                                 AutomationProperties.AutomationId="EnvName"
                                 AutomationProperties.IsRequiredForForm="True"
-                                AutomationProperties.LabeledBy="{Binding ElementName=EnvNameLabel}"/>
```
StellaHuang95, line 148: What does the changes in the file do?

StellaHuang95 (approved):

Python/Product/PythonTools/PythonTools/Environments/AddVirtualEnvironmentControl.xaml
AdamYoblick: Binding modes are TwoWay by default, so I removed that. The main problem here is that `ValidatesOnNotifyDataErrors` was set to false, meaning this textbox won't validate when the codebehind sets errors on the `VirtualEnvName`.

This defaults to true, so I removed it. Now any errors set on `VirtualEnvName` will show up in the UI next to this textBox.

bschnurr (approved):

AdamYoblick:
Here's some more background on these changes:

There are a few different ways to perform validation in WPF, one of which is to implement the `INotifyDataErrorInfo` interface. In our case, the `EnvironmentViewBase` class already implemented this interface, so we had very little work to do. We just had to make sure that:

- The textBox (which is bound to the `VirtualEnvName` property) had `ValidatesOnNotifyDataErrors` set to true (which is the default)
- The view performs validation logic on `VirtualEnvName` whenever that property value changes
- If the validation fails, the view should call SetError (and pass in the `VirtualEnvName` property), which invokes the `ErrorsChanged` event.
  - Since the textbox is bound to that property, it will display validation errors when any exist for that property
- Make sure `IsAcceptEnabled` is set to false if `HasErrors` (from the `INotifyDataErrorInfo` interface)  is true
  - This enables/disables the Create button based on the presence of errors, because `IsEnabled` on the button is bound to `IsAcceptEnabled`

See Some good docs on WPF data binding are at https://docs.microsoft.com/en-us/dotnet/desktop/wpf/data/?view=netdesktop-6.0. Specifically, https://docs.microsoft.com/en-us/dotnet/desktop/wpf/data/?view=netdesktop-6.0#data-validation.

PR #7140, opened 2022-08-24
Look for package named 'python' when getting conda envinronment version

### LINKED ISSUE #7137: Visual Studio 17.3.0 can detect the wrong Python version in Conda environment if another package contains the word "Python"
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Visual-Studio-1730-can-detect-the-wron/10118876)._

---
[severity:I'm unable to use this version] [regression] [worked-in:Whatever was before 17.3.0]
Visual Studio 17.3.0 thinks TEST's Python version is 3.8, as expected:
`conda create  --yes --name TEST --channel conda-forge python=3.8`

&nbsp;

Visual Studio 17.3.0 thinks Python's version  is 1.0.4 and the debugger refuses to launch because it doesn't support that version of Python, etc. `1.0.4` is the version of **msgpack-python**:
`conda install --yes --name TEST --channel conda-forge msgpack-python`

&nbsp;

It is my guess that something has changed where whatever parses a Conda environment just searches for the first (alphabetically sorted) package that contains the word 'python' and uses that package's version as the version of Python. My guess is based on the fact that `msgpack-python` sorts before `python`.
This issue only started happening after updating to 17.3.0

---
#### Feedback Bot on 8/15/2022, 07:19 PM: 

(private comment, text removed)

---
### Original Solutions
(no solutions)

AdamYoblick:
Fixes #7137

This fixes a bug where the version being returned for a Conda python environment was wrong. It was taking the first package that had "python" in the name, not doing an exact match on "python". Works like a charm now.

### DIFF
--- a/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
+++ b/Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
@@ -313,9 +313,18 @@ namespace Microsoft.PythonTools.Interpreter {
                 try {
                     // deserialize the json into an object and return the version
                     var result = JsonConvert.DeserializeObject<CondaListResult>(json);
-                    if (result != null && result.Any()) {
-                        version = new Version(result.First().Version);
+                    if (result == null) {
+                        return version;
                     }
+
+                    // look for a package named "python"
+                    var pythonResult = result.FirstOrDefault(x => x.Name == "python");
+                    if (pythonResult == null) {
+                        return version;
+                    }
+
+                    // get the version from this package
+                    version = new Version(pythonResult.Version);
                     
                 } catch (JsonException ex) {
                     Debug.WriteLine("Failed to parse: {0}".FormatInvariant(ex.Message));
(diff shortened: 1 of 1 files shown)


Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
```diff
@@ -313,9 +313,18 @@ class CondaEnvironmentFactoryProvider : IPythonInterpreterFactoryProvider, IDisp
...
                     // deserialize the json into an object and return the version
                     var result = JsonConvert.DeserializeObject<CondaListResult>(json);
-                    if (result != null && result.Any()) {
-                        version = new Version(result.First().Version);
+                    if (result == null) {
+                        return version;
                     }
+
+                    // look for a package named "python"
+                    var pythonResult = result.FirstOrDefault(x => x.Name == "python");
```
bschnurr, line 321: should we lowercase `x.Name`?

StellaHuang95 (approved):

Python/Product/VSInterpreters/Interpreter/CondaEnvironmentFactoryProvider.cs
AdamYoblick: Nah, C# properties are supposed to be PascalCase by convention. The `JsonProperty` attribute handles changing the case appropriately during serialization.

PR #7141, opened 2022-08-25
Fix binskim compliance issues

AdamYoblick:
This PR addresses the binskim compliance issues at https://devdiv.visualstudio.com/DevDiv/_queries/query/?wiql=%20%20%20%20%20SELECT%20ID%2CSeverity%2CState%2C%5BAssigned%20To%5D%2CTitle%20FROM%20WorkItem%20WHERE%20Tags%20Contains%27TSA%239718578-Python%20VS%20IDE%20PTVS-BinSkim-52695775%27%20%20%20%20%20

- For ControlFlowGuardCheck, `/guard:cf` needs to be passed on both the compiler and linker command lines. I saw a dropdown option in VS project properties for the compiler, but I didn't see one for the linker, so I added it explicitly.

Compiler:
image

Linker:
image

- For EnableSpectreMigrations, just set a compiler flag:
image

However, you also need to install C++ libs that support Spectre Mitigrations so this flag can select the appropriate libs to use. This is done through the VS installer, and you will get a build error if you don't have them. I've added the required steps to the wiki at https://github.com/microsoft/PTVS/wiki/Build-and-Debug-Instructions-for-PTVS. The steps are:

- Install the following individual components (see https://aka.ms/Ofhn4c for more info):
   - MSVC v143 - VS 2022 C++ x64/x86 Spectre-mitigated libs (Latest)
   - C++ ATL for latest v143 build tools with Spectre Mitigations (x86 & x64)
   - C++ MFC for latest v143 build tools with Spectre Mitigations (x86 & x64)

### DIFF
--- a/Python/Product/DebuggerHelper/DebuggerHelper.vcxproj
+++ b/Python/Product/DebuggerHelper/DebuggerHelper.vcxproj
@@ -104,7 +104,7 @@
       <EnableCOMDATFolding>false</EnableCOMDATFolding>
       <LinkTimeCodeGeneration>Default</LinkTimeCodeGeneration>
       <OptimizeReferences>false</OptimizeReferences>
-      <AdditionalOptions Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">/GUARD:CF %(AdditionalOptions)</AdditionalOptions>
+      <AdditionalOptions Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">/guard:cf %(AdditionalOptions)</AdditionalOptions>
     </Link>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Release'">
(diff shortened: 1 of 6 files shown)


Python/Product/DebuggerHelper/DebuggerHelper.vcxproj
```diff
@@ -60,6 +60,7 @@
   </PropertyGroup>
   <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <PlatformToolset>v143</PlatformToolset>
+    <SpectreMitigation>Spectre</SpectreMitigation>
```
int19h, line 63: This one is only set for `Debug|x64` per property group condition - it needs to be defined for other configurations, too, or better yet, in an unconditional `<PropertyGroup>` so that it applies to all of them.

int19h: (and ditto for all the other properties)

AdamYoblick: Ahh you are right. I only added them for Debug, and for Debug there is only a single platform for each of these projects. But for release, there is both x64 and x86. I will fix this.

AdamYoblick: Ok this is fixed in the latest PR update.

int19h (approved):

PR #7144, opened 2022-08-30
Fix CodeQL static analysis compliance warnings

AdamYoblick:
Fix semmle warnings at https://devdiv.visualstudio.com/DevDiv/_queries?tempQueryId=bdca2043-7ac3-4421-93d8-88acb6ee3405

Most have to do with checking the revoked cert list when creating web requests, but there is also a missing null check.


Python/Product/DebuggerHelper/trace.cpp
```diff
@@ -481,6 +481,9 @@ static void TraceLine(void* frame) {
 
     void* f_code = ReadField<void*>(frame, fieldOffsets.PyFrameObject.f_code);
     void* co_filename = ReadField<void*>(f_code, fieldOffsets.PyCodeObject.co_filename);
+    if (co_filename == NULL) {
```
int19h, line 484: `nullptr` is preferred over `NULL`.

int19h (approved):

PR #7145, opened 2022-08-31
Implement directory-based test containers 

StellaHuang95:
fixes https://github.com/microsoft/PTVS/issues/7043

The VS testing team has recently changed test discovery such that on machines with multiple CPU cores, it can discover tests from different sources in parallel inside different test host processes. We have been over discovering tests because our adapter would unnecessarily (re)discover tests in all the .py files in each process. So for example, if we have 3 sources and 2 tests files, our adapter is going to return 3 * 2 tests discovered in total.

The fix is to just supply a directory as the ITestContainer.Source from ITestContainerDiscoverer. The ITestDiscoverer will be called with the directory that we supply and it can discover tests in this directory when that happens.

- Potential follow up changes that can be made in the future:

If we are discovering at the folder level (esp. for open folder where the folder is always the root folder), the incremental discovery could be slow. For example, if only one python test file has changed since the previous discovery then you would still be discovering all files under the folder.

Although this is no different than what we were doing before, it would be nice to somehow "remember" / persist the tests in .py files that haven't changed so that incremental discovery can be faster...

It could be a follow up change since the current switch to directory based would be no worse than what our adapter was already doing.

The only main difference from the previous test explorer is the way the tests are displayed in the Test Explorer's hierarchy. See more details [here](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1560000#:~:text=Another%20alternative%20is,land%20until%2017.4...).

I ran some tests for different scenarios and it's working as expected. Test scenarios including project view, folder view, changing search path, adding/removing/modifying testing files.

- Unittest, project view
unittest project view

- Unittest, folder view
unittest folder view

- Pytest project view
pytest project view

- Pytest, folder view
pytest folder view


AdamYoblick:
The PR build will keep failing because you are using an attribute that was added to the most recent version of Visual Studio, and the agent pool that runs our builds has an older version. I've reached out to the DevDiv Fundamentals team to ask about a preview-pool or something similar. This PR will have to sit until they have a solution :)

shyamnamboodiripad:
@StellaHuang95 The Test Platform APIs (including the above attribute) can also be consumed via NuGet packages. If you switch your projects to consume these via NuGet, you wouldn't need a matching version of VS installed in your CI for build - you would only need a matching VS version when you test the changes.

For example, the VS 17.4 version of MS.TP.OM.dll (which should contain the above attribute) should be available in this NuGet package https://www.nuget.org/packages/Microsoft.TestPlatform.ObjectModel/17.4.0-preview-20221003-03. Looks like other NuGet packages are already being referenced in this repo via this file - https://github.com/microsoft/PTVS/blob/main/Build/17.0/packages.config.

Tagging @nohwnd in case there are any other gotchas that we should be aware of regarding the Test Platform NuGet packages.

Python/Product/TestAdapter/TestAdapter.csproj
```diff
@@ -49,9 +49,13 @@
     <PlatformTarget>AnyCPU</PlatformTarget>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Microsoft.VisualStudio.TestPlatform.ObjectModel">
+    <!-- <Reference Include="Microsoft.VisualStudio.TestPlatform.ObjectModel">
       <HintPath>$(DevEnvDir)\Extensions\TestPlatform\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll</HintPath>
       <Private>False</Private>
+    </Reference> -->
+    <Reference Include="Microsoft.TestPlatform.ObjectModel">
```
shyamnamboodiripad, line 56: Looks like the reference may be missing `.VisualStudio` in its name? (The name of the package is different from the name of the contained binary 😢)...

Also, you may be able to skip the hint path entirely now (similar to other references like MS.VS.Shell.* below which are also referenced via pacakges.config...

StellaHuang95: Thank you for catching this😊. I just made the changes.

shyamnamboodiripad: Btw looks like some test projects below use the old hint path from VS. Would they need similar changes as the current project to avoid build breaks?

StellaHuang95: You are absolutely right, just realized that as well:)

StellaHuang95:
The build finally passed.🥲Can I get a review whenever you get a chance? @AdamYoblick @bschnurr Thank you!!

Python/Product/TestAdapter.Executor/TestAdapter.Executor.csproj
```diff
@@ -58,8 +58,12 @@
...
       <IncludeInVSIX>False</IncludeInVSIX>
     </Reference>
-    <Reference Include="Microsoft.VisualStudio.TestPlatform.ObjectModel">
-      <HintPath>$(DevEnvDir)\CommonExtensions\Microsoft\TestWindow\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll</HintPath>
+    <!-- <Reference Include="Microsoft.VisualStudio.TestPlatform.ObjectModel">
+      <HintPath>$(DevEnvDir)\Extensions\TestPlatform\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll</HintPath>
+      <Private>False</Private>
+    </Reference> -->
+    <Reference Include="Microsoft.TestPlatform.ObjectModel">
+      <HintPath>$(PackagesPath)\Microsoft.TestPlatform.ObjectModel.17.4.0-preview-20221003-03\lib\net462\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll</HintPath>
```
bschnurr, line 66: do we still need these hint paths? or was that just for testing locally with a custom testplatform lib?

Build/17.0/packages.config
```diff
@@ -84,4 +84,5 @@
   <package id="System.Threading.Tasks.Extensions" version="4.5.4"/>
   <package id="Microsoft.VisualStudio.RpcContracts" version="17.0.50-preview-0002-0005" />
   <package id="Microsoft.DiaSymReader.Pdb2Pdb" version="1.1.0-beta2-21181-01" />
+  <package id="Microsoft.TestPlatform.ObjectModel" version="17.4.0-preview-20221003-03" />
```
bschnurr, line 87: should we wait for a none preview version with the same changes?

StellaHuang95: This is what's given to me by the test explorer team. I ran the tests locally and haven't found any issues. Are you worried the preview version is unstable in some way?

shyamnamboodiripad: The new APIs present in the above NuGet package should be available in preview builds of VS 2022 v17.4. The package simply makes it so that you can build without having to have a matching version of VS installed in your CI. But when testing locally, everything should work so long as you test against a recent 17.4 preview build.

bschnurr (approved):

PR #7201, opened 2022-10-25
Allow '%', '&', '#' in file names in Python projects

davkean:
These characters are actually valid in file names, and the restriction is being dropped removed from Solution Explorer which previously blocked this. After that block has been removed, there is a further block inside Python tools that prevents files, folders and projects to be renamed to contain these characters.

Note: Unlike other projects, Python had already allowed "&" despite the error message not being updated to reflect that.

See: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/431541

### DIFF
--- a/Common/Product/SharedProject/ProjectResources.resx
+++ b/Common/Product/SharedProject/ProjectResources.resx
@@ -199,7 +199,7 @@ Files and folders cannot be:
 - Empty strings 
 - System reserved names, including 'CON', 'AUX', PRN', 'COM1' or 'LPT2' 
 - contain only '.'
-- have any of the following characters: / ? : &amp;  \ * " &lt; &gt; | # %</value>
+- have any of the following characters: / ? : \ * " &lt; &gt; |</value>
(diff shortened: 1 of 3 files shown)


AdamYoblick (approved):

davkean:
@AdamYoblick Could you merge these changes? Can you also confirm that this will make 17.5?

AdamYoblick:
@davkean I just merged them in. We insert every build from main into vs main, so this will flow in. I'll track it to make sure it gets all the way :)

AdamYoblick:
PTVS CI build is running at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6962523&view=results. I'll reply back once there's a VS insertion.

davkean:
@AdamYoblick Thanks!

AdamYoblick:
@davkean This change has been merged into vs main 😄

davkean:
@AdamYoblick awesome!

PR #7212, opened 2022-11-01
Ensure PreBuild always installs latest pylance

AdamYoblick:
When trying to pull the latest pylance into the public feed, I noticed that just running `npm install` without deleting the `node_modules` folder didn't install the latest pylance if there was one already there.

So we just delete the `node_modules` folder, if it exists, before calling `npm install`

### DIFF
--- a/Build/PreBuild.ps1
+++ b/Build/PreBuild.ps1
@@ -85,8 +85,14 @@ try {
         $packageJson | ConvertTo-Json -depth 8 | Set-Content $packageJsonFile
     }
 
-    # force install latest pylance, always grab latest even if a lower version is already installed
-    npm install --force
+    # delete pylance install folder to make sure we always get latest
+    $nodeModulesPath = Join-Path $buildroot "node_modules"
+    if (Test-Path -Path $nodeModulesPath) {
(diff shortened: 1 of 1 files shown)


Build/PreBuild.ps1
```diff
@@ -86,8 +85,8 @@ try {
         $packageJson | ConvertTo-Json -depth 8 | Set-Content $packageJsonFile
     }
 
-    # install pylance
-    npm install
+    # force install latest pylance, always grab latest even if a lower version is already installed
+    npm install --force
```
StellaHuang95, line 89: So what this force install does is to omit the .\build\nuget locals all -clear step? Is my understanding correct?

StellaHuang95 (approved):

Build/PreBuild.ps1
AdamYoblick: No, clearing the nuget cache is for nuget only. This is specifically for npm only :)

AdamYoblick: This way fails the build because the warning from npm writes to the error stream. So instead, my latest change checks for the `node_modules` folder and deletes it if it exists

StellaHuang95 (approved):

Issue #7260, opened 2022-12-07
VS-2022 "Python/Native Debugging" not working after being brought back from missing


flexorx:
After a clean install VS-2022 and "returning back" "Python/Native Debugging", as suggested [here](https://github.com/microsoft/PTVS/issues/6712) with copying files from `.../v16/...` to `.../v17/...`, the debugging feature doesn't work in Python code, and the breakpoints won't be triggered in Python code at all. 

**Steps to Reproduce**
1. Make sure to remove all Pythons from Windows. That is to make sure we are testing everything against same vanilla Python.
2. Install Python 3.6 x64 with Web Installer, all default locations, but check to install debug symbols and everything else in the "Additional" settings. We use 3.6 here to avoid these things with venv occurring from 3.8 and onwards.
3. Fresh-install VS-2017, VS-2019 and VS-2022. DO NOT install Python with Visual Studio: this is to guarantee that all three different VS are tested against exact same Python, which we installed previously.
4. Fix VS-2022 "Python/Native Debugging" missing, as described in https://github.com/microsoft/PTVS/issues/6712.
5. Clone my projects showing the issue: https://github.com/flexorx/simple_cpp_python. These are 3 identical projects, but each set to the right VS target.
6. Change the paths in projects to the right ones for your Python 3.6 installed location, i.e. `c:\Users\x1\AppData\Local\Programs\Python\Python36\include\` to your location, and `c:\Users\x1\AppData\Local\Programs\Python\Python36\libs\` to yours.
7. See the video attached below, follow how the breakpoints are set, and go through all three projects to verify that the Python breakpoint is hit in VS-2017 and VS-2019, but not in VS-2022.

![StrangeThing2](https://user-images.githubusercontent.com/49031207/206269314-54865818-108e-47c5-8b6f-8eb210342173.gif)

**Expected behavior**

Debugging with breakpoints work both in Python and in C++ in "Python/Native Debugging" in VS-2022.

**Sample Code**

https://github.com/flexorx/simple_cpp_python

LazyDodo:
> Here's a great repro which uses Python 3.6, which means that python 3.10+ is not the cause of this.
> https://github.com/microsoft/PTVS/issues/7260

disclaimer: total gut feeling but given I was in this area recently for 3.10 support i have a good feeling this is correct

3.6 isn't supported, according to
```
        private Int32Proxy GetTracingPossible(PythonRuntimeInfo pyrtInfo, DkmProcess process) =>
[...]

PR #7413, opened 2023-02-23
Restore missing mixed-mode debugging option

### LINKED ISSUE #6712: The option "Python/Native Debugging" is missing.
**Environment**
image

**Steps to Reproduce**
1. Create new project: Visual C++ > Windows Desktop > Windows Console Application
2. Open the property page
3. Click Debugging tab
4. Click to open the drop-down list 
4. Observes

**Expected behavior**
The option "Python/Native Debugging" should exist in the list.
image

**Actual result**
The option "Python/Native Debugging" is missing.
image

AdamYoblick:
Fixes #6712

We had a remaining hardcoded `v160` install path for the `VCDebugLauncher` vsix that was causing files to be installed to the wrong location. These files were xaml files that add UI to Visual Studio, and our targets file was trying to include these files from the `v170` location, if they exist.

The problem is that they never exist because they go to the wrong place, which means the mixed mode debugging option is always missing from VS2022. This fixes that problem.

Unfortunately, this problem doesn't repro when building locally and deploying to the exp hive, it's a "VS Setup Only" problem. So I can't really test to make sure that this works before merging in this PR. The steps to test are:
1. Merge in this PR
2. Wait for the CI build to finish from the main branch
3. Wait for the VS insertion PR to trigger after the CI build
4. Wait for the VS insertion PR to generate a VS installer
5. Install VS using the installer from step 4, make sure to select the python workload, then do the steps mentioned in https://github.com/microsoft/PTVS/issues/6712 to verify that the missing debugging option is now present.

There is a related problem (https://github.com/microsoft/PTVS/issues/7260) which I will be looking at next.

### DIFF
--- a/Python/Product/VCDebugLauncher/17.0/source.extension.vsixmanifest
+++ b/Python/Product/VCDebugLauncher/17.0/source.extension.vsixmanifest
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
-        <Identity Id="PythonVCDebugLauncher.5AF7E179-1406-4804-9267-93F3BD7C7422" Version="17.0.0" Language="en-US" Publisher="Microsoft Corporation" />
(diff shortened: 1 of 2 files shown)


int19h (approved):

StellaHuang95 (approved):

AdamYoblick:
Tested with the installer provided in the VS insertion PR at https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/454017

Everything looks good:

image

PR #7422, opened 2023-03-01
Remove broken task that we don't own

AdamYoblick:


StellaHuang95 (approved):
What does this task do?

AdamYoblick:
Indexes the code for internal code searching across multiple products. The problem is that someone asked me to add this a long time ago and the step is broken and blocking the build. I've searched through all my email and chats and I can't seem to find who asked me to add it, so I'm removing it :)

PR #7489, opened 2023-03-30
Fix asynchronous environment discovery and selection

### LINKED ISSUE #7444: Anaconda/Conda environments are broken in VS2022, works in VS2019
From e-mail, repro steps:

1. Install Anaconda
2. Add a new conda environment, set as default for new projects

image

3. Restart VS. Create a new python application.
4. Anaconda environment is not loaded, Python 3.9 is set as default:

image

5. Wait for about 30 seconds, the anaconda environment is loaded as shown below. However, it is not set to default because VS has already set it to Python 3.9:

image

AdamYoblick:
Fixes #7444

There are some bugs around environment loading and selection when the environments are discovered asynchronously, like with conda environments. There are lots of edge cases that need to be tested. See comments below for each scenario.

Specifically, conda environments weren't being shown in solution explorer because the currently selected environment was the global default, and the existing conditional was not behaving properly in that case. We were also showing an infobar for "missing conda environment" before they had a chance to be discovered.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
@@ -90,6 +90,7 @@ namespace Microsoft.PythonTools.Project {
         private readonly PythonNotSupportedInfoBar _pythonVersionNotSupportedInfoBar;
 
         private readonly SemaphoreSlim _recreatingAnalyzer = new SemaphoreSlim(1);
+        private bool _isRefreshingInterpreters = false;
 
         public event EventHandler LanguageServerInterpreterChanged;
 
@@ -224,36 +225,32 @@ namespace Microsoft.PythonTools.Project {
                 Debug.Assert(this.FileName != null);
                 var oldActive = _active;
 
+                // stop listening for installed files changed
                 var oldPms = _activePackageManagers;
                 _activePackageManagers = null;
-
                 foreach (var pm in oldPms.MaybeEnumerate()) {
                     pm.InstalledFilesChanged -= PackageManager_InstalledFilesChanged;
                 }
 
                 lock (_validFactories) {
-                    if (_validFactories.Count == 0) {
-                        // No factories, so we must use the global default.
+                    // if there are no valid factories,
+                    // or the specified factory isn't in the valid list,
+                    // use the global default
+                    if (_validFactories.Count == 0 || value == null || !_validFactories.Contains(value.Configuration.Id)) {
                         _active = null;
-                    } else if (value == null || !_validFactories.Contains(value.Configuration.Id)) {
-                        // Choose a factory and make it our default.
-                        // TODO: We should have better ordering than this...
-                        var compModel = Site.GetComponentModel();
-
-                        _active = InterpreterRegistry.FindInterpreter(
-                                _validFactories.ToList().OrderBy(f => f).LastOrDefault()
-                        );
                     } else {
                         _active = value;
                     }
                 }
 
+                // start listening for package changes on the active interpreter again
                 _activePackageManagers = InterpreterOptions.GetPackageManagers(_active).ToArray();
                 foreach (var pm in _activePackageManagers) {
                     pm.InstalledFilesChanged += PackageManager_InstalledFilesChanged;
                     pm.EnableNotifications();
                 }
 
+                // update the InterpreterId element in the pyproj with the new active interpreter
                 if (_active != oldActive) {
                     if (_active != null) {
                         BuildProject.SetProperty(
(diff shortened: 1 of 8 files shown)


AdamYoblick:
# Case 1 - Conda environment found (project view)

To repro:

1. Install miniconda
2. Create a new python application
3. Add a new conda environment to the project (If you see an error about openssl missing, that's a conda bug. See https://github.com/conda/conda/issues/11795#issuecomment-1461824323 for a workaround)
4. Save the project and close the solution
5. Re-open the solution
6. The previously selected conda environment should be present under `Python Environments` in the solution explorer, and it should be activated and bolded.

It looks like the bug manifests here in a few ways. If there are no other environments present in the list, the global default seems to be added and activated. Then, when the conda environment is detected, it's added to the solution explorer and the global default is removed. Since the conda environment is the only one in the list, it gets selected.

However, if there's another env in the list (like a virtual env), this environment is present AND activated while waiting for the conda env to load. This is a problem since the `<InterpreterId>` element is updated in the pyproj, which means the conda environment will not be activated correctly, even though it IS detected and added to the solution explorer.

The fix for this should be never to activate an environment unless it's the one specified in the pyproj (or it's the global default).

AdamYoblick:
# Case 2 - Conda environment not found (project view)

To repro:

1. Complete all repro steps from previous case above
2. Remove conda environment from machine using conda command prompt (`conda remove -n ENV_NAME --all`)
3. Open python project with reference to this now-deleted environment.

## Scenario 1 - Only conda environment in list
## Scenario 2 - Virtual environment also in list

Python/Product/VSInterpreters/Interpreter/InterpreterRegistryService.cs
```diff
@@ -72,12 +75,21 @@ sealed class InterpreterRegistryService : IInterpreterRegistryService, IDisposab
...
         private void EnsureFactoryChangesWatched() {
+
             if (!_factoryChangesWatched) {
                 BeginSuppressInterpretersChangedEvent();
                 try {
                     foreach (var provider in GetProviders()) {
                         provider.InterpreterFactoriesChanged += Provider_InterpreterFactoriesChanged;
+
+                        // if the provider is async, listen for the completed event
+                        if (provider is IPythonInterpreterFactoryProviderAsync asyncProvider) {
```
AdamYoblick, line 89: I did this in case we add more async providers in the future. I could have just checked if the provider is the conda one (since it's the only async one we have) but I figured this was slightly better.

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -204,6 +202,16 @@ internal partial class PythonProjectNode :
             Site.GetUIThread().Invoke(() => RefreshInterpreters());
         }
 
+        // Called once all async interpreter factories have finished discovering interpreters
+        private void OnInterpreterDiscoveryCompleted(object sender, EventArgs e) {
+            if (!_asyncInfoBarCheckTriggered) {
+                _asyncInfoBarCheckTriggered = true;
+
+                // Check for any missing environments and show info bars for them
+                _condaEnvCreateInfoBar.CheckAsync().HandleAllExceptions(Site, typeof(PythonProjectNode)).DoNotWait();
```
bschnurr, line 211: what about open folder mode?

should be just recheck all infobars? there is a conda one too `_condaEnvCreateInfoBar`

AdamYoblick: I checked open folder mode, but there is no PythonProject node in that case. We don't have a Python Environments node, we just use the normal environment list that's in VS and you can select the one you want from the dropdown when you are going to run the program. Since these bugs manifest in the project node, I don't believe these bugs repro when opening in folder view. At least, I couldn't repro them. 😄

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -229,36 +237,32 @@ internal partial class PythonProjectNode :
...
                 lock (_validFactories) {
-                    if (_validFactories.Count == 0) {
-                        // No factories, so we must use the global default.
+                    // if there are no valid factories,
+                    // or the specified factory isn't in the valid list,
+                    // use the global default
+                    if (_validFactories.Count == 0 || value == null || !_validFactories.Contains(value.Configuration.Id)) {
                         _active = null;
-                    } else if (value == null || !_validFactories.Contains(value.Configuration.Id)) {
-                        // Choose a factory and make it our default.
```
StellaHuang95, line 244: The change you made makes more sense to me, I don't get why previously we just randomly return the last item in the ordered env list and make it active.

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -777,76 +780,104 @@ internal partial class PythonProjectNode :
...
+                var pythonEnvironmentsNode = _interpretersContainer;
+                if (pythonEnvironmentsNode == null) {
+                    return;
                 }
-            }
 
-            foreach (var id in InvalidInterpreterIds) {
-                if (!RemoveFirst(remaining, n => n._absentId == id)) {
-                    node.AddChild(InterpretersNode.CreateAbsentInterpreterNode(this, id));
+                // clear out all interpreter nodes since we're going to re-add them
```
StellaHuang95, line 811: why are all interpreter nodes removed and added again? didn't follow this part.

AdamYoblick: Because interpreters have changed. The logic before was too convoluted. It's much easier to just clear the list in the solution explorer, and repopulated it with all the interpreters that were found, plus dummy nodes for interpreters that are referenced but no longer found, plus a global interpreter if appropriate.

This logic is much easier to follow, there's no checking what's already there, no collisions. Simple. And the performance difference is negligible because we're just updating nodes in solution explorer, and the number of interpreters is never in the hundreds or anything like that.

Python/Product/VSInterpreters/Interpreter/InterpreterRegistryService.cs
```diff
@@ -86,6 +98,21 @@ sealed class InterpreterRegistryService : IInterpreterRegistryService, IDisposab
...
+            // Since we know how many async providers we have, keep track of how many times this callback is hit.
+            // Once the number of calls == the number of providers, we know all async interpreter discovery is done.
+            _asyncInterpreterDiscoveryCompletedCount++;
+
+            if (_asyncInterpreterDiscoveryCompletedCount < _asyncProviderCount) {
+                return;
+            }
+
+            // if we get here, interpreter discovery is finished
+            InterpreterDiscoveryCompleted?.Invoke(this, EventArgs.Empty);
```
StellaHuang95, line 113: So the `_condaEnvCreateInfoBar.CheckAsync()` is called when the `InterpreterDiscoveryCompleted` event is raised, and there are two scenarios where this event is raised:
1. When `EnsureInitialized()` is called on project load
2. When the `InterpretersChanged` event calls `EnsureFactoryChangesWatched()`, this is where you make sure all the async calls are done before checking the info bar.

Is my understanding correct?

StellaHuang95 (approved):

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
bschnurr: Ok. I thought because it was also calling _condaEnvCreateInfoBar in open folder that it should refresh on conda environments detected

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -777,76 +780,104 @@ internal partial class PythonProjectNode :
...
+                var interpreterNodes = pythonEnvironmentsNode.AllChildren.OfType<InterpretersNode>().ToList();
+                interpreterNodes.ForEach(pythonEnvironmentsNode.RemoveChild);
+
+                // if we have no interpreter factories, and the active interpreter is the global default,
+                // add a node for it
+                if (InterpreterFactories.Count() == 0 && IsActiveInterpreterGlobalDefault && ActiveInterpreter.IsRunnable()) {
```
int19h, line 817: Might want to use `Any()` here to avoid enumerating past the first one.

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -777,76 +780,104 @@ internal partial class PythonProjectNode :
...
+                // This can include virtual environments that have been deleted, interpreters that have been uninstalled,
+                // or conda environments that are still being discovered asynchronously.
+                foreach (var id in InvalidInterpreterIds) {
+                    pythonEnvironmentsNode.AddChild(InterpretersNode.CreateAbsentInterpreterNode(this, id));
                 }
-                foreach (var child in expandAfter) {
-                    child.ExpandItem(EXPANDFLAGS.EXPF_ExpandFolder);
+
+                // Handle the expand/collapse state of the container node, plus each interpreter node
+                // TODO: Is this still needed? We cleared out all the previous nodes so the previous expanded info is gone?
```
int19h, line 855: Yeah, I think this won't do anything useful now. If you want to preserve the expansion state on refresh it needs to be captured before clearing the nodes out.

Python/Product/VSInterpreters/Interpreter/InterpreterRegistryService.cs
```diff
@@ -86,6 +98,21 @@ sealed class InterpreterRegistryService : IInterpreterRegistryService, IDisposab
             }
         }
 
+        // Called when a single async interpreter factory provider finishes discovering interpreters
+        private void Provider_AsyncInterpreterDiscoveryCompleted(object sender, EventArgs e) {
+
+            // Since we know how many async providers we have, keep track of how many times this callback is hit.
+            // Once the number of calls == the number of providers, we know all async interpreter discovery is done.
+            _asyncInterpreterDiscoveryCompletedCount++;
```
int19h, line 106: Are these all guaranteed to come on the same thread? It might be worth adding a comment to that effect on `InterpreterDiscoveryCompleted` if so, but if not, this needs to be an atomic increment-and-check.

int19h (approved):

Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
AdamYoblick: I decided to just expand the python environments node and keep all the child nodes collapsed.

Python/Product/VSInterpreters/Interpreter/InterpreterRegistryService.cs
AdamYoblick: Added a locking mechanism 😄

Issue #7402, opened 2023-02-22
Inconsistent Commenting/Uncommenting Behavior Between VS and VS Code 


StellaHuang95:
The second part of the fix to #6849 

Etienne said he will fix their implementation in 17.6 Preview 2 or 3. I submitted a PR on my end to ship my language config with TextMake.

PR: https://dev.azure.com/devdiv/DevDiv/_git/VS-Platform/pullrequest/453429

Once he makes his change, I will fix the issue on the PTVS side.

StellaHuang95:
The last PR I did was intended to have PTVS provide the same language-configuration.json that VS Code Python uses. It didn't fix anything in particular in that PR. However, it's causing some issues with automatic brace completion in VS, especially for nested ones.

I haven't figure out the cause yet, so undo the changes to have more time to investigate.
https://devdiv.visualstudio.com/DevDiv/_git/VS-Platform/pullrequest/456414

PR #7533, opened 2023-04-26
Fix Inconsistent Commenting/Uncommenting Behavior Between VS and VS Code

StellaHuang95:
fixes https://github.com/microsoft/PTVS/issues/7402

Uncommenting (Ctrl+K, Ctrl+U) in VS only removes the #, but commenting (Ctrl+K, Ctrl+C) in VS Code adds # plus a space. If you comment in VS Code and then uncomment in VS you're left with an unexpected space at the start of the line. So change the commenting behavior in VS to add the extra space to be consistent with VS Code.

Now the comment/uncomment behavior looks like this:
commment uncomment


AdamYoblick (approved):

LozmindaSnr:
Still unfixed Visual Studio 2022 version 17.10 June 2024.

StellaHuang95:
@LozmindaSnr Not sure what you meant by unfix. I can't repro the issue in the latest VS. Can you provide more detailed repro steps and behaviors?

LozmindaSnr:
Quoting the OP:  "_So change the commenting behaviour in VS to add the extra space to be consistent with VS Code._" (Inline with some coding standards and other editors, in these brackets is my extra comment)
``` c++
// This is easier to read
//Than this in Visual Studio 2022
```
If you comment a line of code in  Visual Studio 2022 version 17.10 you still don't get a space after the "//" . It's no an option to change (that I can find), the 'bug' doesn't appear to be fixed.
Perhaps I've made an error ?

StellaHuang95:
How did you comment the line? The fix is intended to remove the space when using the Edit -> Advanced -> Uncomment Selection option, without affecting the behavior of comments.

dominicpalmer:
I agree with @LozmindaSnr that there's an underyling issue here which still isn't fixed. I don't think it relates to PTVS specifically. This issue is with _commenting_ text, not with uncommenting.

Consider this line of text: `This is some example text`

### Uncommenting Behaviour (Correct)
- Let's say the line has previously been commented, and it's now `// This is some example text`. On uncommenting it, VS will correctly convert it back to `This is some example text`.
- Let's say the line has previously been _manually_ commented to `//This is some example text` (no space between `//` and the start of the text). VS correctly uncomments this back to `This is some example text`.

### Commenting Behaviour (Incorrect)
  - On commenting the line, we get `//This is some example text`, _not_ `// This is some example text` as with VSCode (and other editors).  **The issue is that commenting a line does not insert a space between the `//` and the text.**

### Impact
Now lets imagine the line of text is already commented: `// This is some example text`. If we uncomment then comment again, we won't be left with the same text as we started with. The cycle is:

1. Original: `// This is some example text`
2. Uncommented: `This is some example text`
3. Commented again: `//This is some example text`

LozmindaSnr:
So I followed the chain back to here (apologies if it's not quite in the right order..)
https://github.com/microsoft/PTVS/issues/7624
https://developercommunity.visualstudio.com/t/insert-space-before-comment/10340193
https://github.com/microsoft/PTVS/issues/7624
https://developercommunity.visualstudio.com/t/insert-space-before-comment/10340193?q=xps
[And then there's this stack overflow question](https://stackoverflow.com/questions/28567114/how-to-get-visual-studio-quick-comments-to-add-a-space-after-the-double-slash-to)

StellaHuang95:
@LozmindaSnr @dominicpalmer I am trying to understand your scenario. We are PTVS (Python Tools for Visual Studio), and only Python-related issues are within the scope of our repository. The issue you're posting here involves comment symbols `//`, which are not used for comments in Python. In Python, comments are denoted using the `#` symbol. Therefore, the language in which this issue occurs is important because it may fall under the responsibility of another team. Could you please confirm the programming language you are referring to?

dominicpalmer:
@StellaHuang95 For me it's C#. The scenario I describe is essentially that commenting doesn't add a space between // and the text.

I think you're right that this is an issue for another team, but I'm not sure which one.

LozmindaSnr:
According to [this SO post ](https://stackoverflow.com/questions/28567114/how-to-get-visual-studio-quick-comments-to-add-a-space-after-the-double-slash-to)it has been fixed, but it's in the next release, I'm guessing 17.11. 
In our defence even when one googles "space after comment in C++  Visual Studio 2022" the trail ends up here. This could be a function of [google's degerated](https://theweek.com/tech/why-google-search-results-have-gotten-worse) performance rather than anything else. Folks are coming here to post, because that's where other links have sent them.
If this is the wrong dept. apologies and if it is fixed, marvellous !!

Issue #7378, opened 2023-02-06
Conda env is missing after updating python latest version


linette-zyy:
**Environment**
![image](https://user-images.githubusercontent.com/59192491/216937634-3eed8490-4661-456a-94ca-cec853cf37c6.png)

**Steps to Reproduce**
1.Launch VS and create a project
2.Create a conda env
![image](https://user-images.githubusercontent.com/59192491/216937959-e73d8c1b-142c-48e8-94ab-fd3ddd173e96.png)
![image](https://user-images.githubusercontent.com/59192491/216937885-31ae0892-5afa-4b49-af3f-c2e4a5b86fd8.png)
3.Create conda env successfully
4.Upgrade python version 3.7-->3.11

**Actual behavior**
The conda env display missing
![condaenvmissing](https://user-images.githubusercontent.com/59192491/216938498-f83f48fb-f4fd-4361-9a35-c548ccbd7470.gif)

StellaHuang95:
After the fix, the missing label will still show up for a second as it's waiting for re-discovery, then the proper environment will show up.

PR #7603, opened 2023-05-11
Fix conda environment missing after python install

StellaHuang95:
fixes https://github.com/microsoft/PTVS/issues/7378

The issue was that the `IInterpreterRegistryService` doesn't trigger an interpreter discovery after installing/updating python package, so signal the `CondaEnvironmentFactoryProvider` to discover conda interpreter factories after install to fix the issue.

The missing label is gone now.
conda env

Also the environment is showing up in environment pane correctly.
conda env


Python/Product/VSInterpreters/Interpreter/IInterpreterRegistryService.cs
```diff
@@ -84,5 +84,11 @@ public interface IInterpreterRegistryService {
         /// <see cref="BeginSuppressInterpretersChangedEvent"/>.
         /// </summary>
         void EndSuppressInterpretersChangedEvent();
+
+        /// <summary>
+        /// Discovers Conda interpreter factories after updating or installing 
+        /// a Python package.
+        /// </summary>
+        void DiscoverCondaInterpreterFactoriesAfterPythonInstall();
```
AdamYoblick, line 92: This method name is too specific. Yes, you are calling it after a python install, but you could potentially call it for other reasons as well. Consider renaming this to `RefreshCondaInterpreterFactories()` or something similar.

StellaHuang95: @AdamYoblick I've renamed it with what you suggested and tested it again.😊

AdamYoblick (approved):

Issue #7485, opened 2023-03-29
Excessive memory usage in Environments View


int19h:
MemWatson reports Mgd_PythonTools_Microsoft.PythonTools.EnvironmentsList!Microsoft.PythonTools.EnvironmentsList.PipPackageView as the culprit. The problem is that the database of available packages on PyPI is loaded entirely into memory, and furthermore this happens separately for each environment, so as users switch between environments those in-memory databases end up consuming hundreds of megabytes of RAM.

int19h:
From discussions:

> The low-hanging fruit here is making sure that the installable packages list gets cleared as users switch between envs; there's really no reason to keep the list of installable packages in memory for each environment that user has inspected, even once they moved away from it. This should be fairly straightforward, and it would keep the total live count of PipPackageView at the same level, preventing the tool pane from ballooning beyond ~350 Mb (but, of course, this will still get worse as PyPI package count grows). I'm working on this fix; I suggest that we ship that by itself first and see how it affects the numbers out in the wild. 

> Then we can optimize the code to cut down on amount of memory used even without persistent duplicates - this should help with GC. E.g. a bunch of temporary arrays get created to return results from async calls; we can use async enumerables now. But I'm not convinced that this is a long-term solution. Given the sheer amount of data here, I think we need to seriously consider moving to SQLite to cache local package data and then query that from the UI as needed, instead of having the entire view model in memory. This would solve the problem long-term regardless of how large PyPI gets. 

This issue is tracking the first part.

int19h:
Original issue: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1460022

bschnurr:
When changing between environments we weren't always clearing the list of available pip packages in the previous view.

Ideally, we wouldn't load the entire list and create views for them.. but for now just removing the excess memory do to holding on to multiple lists.

PR #7652, opened 2023-06-26
Fix Excessive memory usage in Environments View

bschnurr:
When a new environment was selected we were creating a pip package list around 400k items but we're clearning the previous list.

fixe: https://github.com/microsoft/PTVS/issues/7485

### DIFF
--- a/Python/Product/EnvironmentsList/PipExtension.xaml.cs
+++ b/Python/Product/EnvironmentsList/PipExtension.xaml.cs
@@ -109,6 +109,16 @@ namespace Microsoft.PythonTools.EnvironmentsList {
                     }
                     Subcontext.DataContext = new PipEnvironmentView(view, _provider);
                 }
+            } else {
+                // Data context being deleted clean up old Pip info
+                // see ToolWindow.xaml.cs  _extensions.Clear();
+                var oldView = e.OldValue as EnvironmentView;
+                if (oldView != null) {
+                    var current = Subcontext.DataContext as PipEnvironmentView;
+                    if (current != null) {
+                        current.Dispose();
+                    }
+                }
             }
         }
 
@@ -321,6 +331,15 @@ namespace Microsoft.PythonTools.EnvironmentsList {
             _provider.IsPipInstalledChanged -= PipExtensionProvider_IsPipInstalledChanged;
             _provider.InstalledPackagesChanged -= PipExtensionProvider_InstalledPackagesChanged;
             _installableViewRefreshTimer.Dispose();
+            lock (_installed) {
+                _installed.Clear();
+            }
+            lock (_installableFiltered) {
+                _installableFiltered.Clear();
+            }
+            lock (_installable) { 
+                _installable.Clear();
+            }
         }
 
         public EnvironmentView EnvironmentView {
--- a/Python/Product/VSInterpreters/PackageManager/PipPackageCache.cs
+++ b/Python/Product/VSInterpreters/PackageManager/PipPackageCache.cs
@@ -160,7 +160,7 @@ namespace Microsoft.PythonTools.Interpreter {
                     }
                 }
 
-                return _cache.Values.Select(p => p.Clone()).ToList();
+                return _cache.Values.ToList();
             } finally {
                 _cacheLock.Release();
             }


Python/Product/EnvironmentsList/PipExtension.xaml.cs
```diff
@@ -321,6 +331,15 @@ PipExtensionProvider provider
...
             _installableViewRefreshTimer.Dispose();
+            lock (_installed) {
+                _installed.Clear();
+            }
+            lock (_installableFiltered) {
+                _installableFiltered.Clear();
+            }
+            lock (_installable) { 
+                _installable.Clear();
+            }
```
bschnurr, line 342: clear out our lists when disposed.

Python/Product/EnvironmentsList/PipExtension.xaml.cs
```diff
@@ -109,6 +109,16 @@ internal sealed partial class PipExtension : UserControl, ICanFocus {
...
+            } else {
+                // Data context being deleted clean up old Pip info
+                // see ToolWindow.xaml.cs  _extensions.Clear();
+                var oldView = e.OldValue as EnvironmentView;
+                if (oldView != null) {
+                    var current = Subcontext.DataContext as PipEnvironmentView;
+                    if (current != null) {
+                        current.Dispose();
+                    }
+                }
```
bschnurr, line 121: make sure to call disposed when the extensions list is cleared

Python/Product/VSInterpreters/PackageManager/PipPackageCache.cs
```diff
@@ -160,7 +161,7 @@ string cachePath
                     }
                 }
 
-                return _cache.Values.Select(p => p.Clone()).ToList();
+                return _cache.Values.ToList().AsReadOnly();
```
bschnurr, line 164: slight change. I dont believe we need to clone. since 'ToList' is already making a copy.

tested in a playground
https://dotnetfiddle.net/Widget/6U9bNG

StellaHuang95 (approved):

Python/Product/VSInterpreters/PackageManager/PipPackageCache.cs
gramster: Does ReadfOnlyCollection not implement IList? Wondering why the return type needs changing here.

bschnurr: Ok. I think I can remove all these interface changes because in the end I'm still calling ToList(), which makes a copy anyways.

bschnurr: Maybe in the future this should be lazy and just be IEnumerable.

bschnurr: removed

Python/Product/EnvironmentsList/PipExtension.xaml.cs
```diff
@@ -109,6 +109,16 @@ internal sealed partial class PipExtension : UserControl, ICanFocus {
                     }
                     Subcontext.DataContext = new PipEnvironmentView(view, _provider);
                 }
+            } else {
+                // Data context being deleted clean up old Pip info
```
bschnurr, line 113:
```suggestion
                // When our datacontext is being deleted, clean up old PipEnvironmentView
```

Python/Product/EnvironmentsList/PipExtension.xaml.cs
```diff
@@ -109,6 +109,16 @@ internal sealed partial class PipExtension : UserControl, ICanFocus {
...
+                // see ToolWindow.xaml.cs  _extensions.Clear();
+                var oldView = e.OldValue as EnvironmentView;
```
AdamYoblick, line 115: Nitpick: This can be a little cleaner if you use the `is` keyword instead of `as`. This will attempt the cast and and avoid you having to check for null:

For example,

```csharp
if (e.OldValue is EnvironmentView oldView) {
    if (Subcontext.DataContext is PipEnvironmentView current) {
        current.Dispose();
    }
}
```

AdamYoblick (approved):

AdamYoblick (approved):

PR #7653, opened 2023-06-27
Update lsp protocol package and fix compile errors

AdamYoblick:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1823538

[CompletionItem.TextEdit](https://devdiv.visualstudio.com/DevDiv/_git/VSLanguageServerClient/pullrequest/462902?path=/src/product/Protocol/LanguageServer.Protocol/CompletionItem.cs&_a=files) has been changed from 
 
```C#
public TextEdit? TextEdit
```
to
```C#
public SumType<TextEdit, InsertReplaceEdit>? TextEdit
```
so I needed to fix up the code a little to account for this breaking change.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Repl/Completion/AsyncCompletionSource.cs
+++ b/Python/Product/PythonTools/PythonTools/Repl/Completion/AsyncCompletionSource.cs
@@ -40,6 +40,7 @@ using Microsoft.VisualStudio.Shell;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.VisualStudio.InteractiveWindow;
 using Microsoft.VisualStudio.Text.Projection;
+using Microsoft.VisualStudio.LanguageServer.Protocol;
 
 namespace Microsoft.PythonTools.Repl.Completion {
     /// <summary>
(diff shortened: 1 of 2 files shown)


Python/Product/PythonTools/PythonTools/Repl/Completion/AsyncCompletionSource.cs
```diff
@@ -249,13 +250,16 @@ internal class AsyncCompletionSource : IAsyncCompletionUniversalSource {
...
                         }
 
-                        if (protocolItem.TextEdit != null) {
-                            Utilities.ApplyTextEdit(protocolItem.TextEdit, triggerLocation.Snapshot, buffer);
-                        } else if (protocolItem.InsertText != null) {
-                            buffer.Replace(session.ApplicableToSpan.GetSpan(buffer.CurrentSnapshot), protocolItem.InsertText);
-                        } else if (protocolItem.Label != null) {
-                            buffer.Replace(session.ApplicableToSpan.GetSpan(buffer.CurrentSnapshot), protocolItem.Label);
-                        }
+                        if (protocolItem.TextEdit != null && protocolItem.TextEdit.HasValue) {
```
ying1104, line 253: what need to check both protocolItem.TextEdit != null and protocolItem.TextEdit.HasValue here? Doesn't not null mean it has value?

StellaHuang95, line 253: what need to check both protocolItem.TextEdit != null and protocolItem.TextEdit.HasValue here? Doesn't not null mean it has value?

AdamYoblick: Yep, we don't have to check both. I'll fix 😄

StellaHuang95 (approved):

PR #7670, opened 2023-07-07
Fix node path not found issue

### LINKED ISSUE #7660: An error pops up when create a project
**Environment**
image

**Steps to Reproduce**
1.Launch VS and create a python project

**Expected behavior**
No error pops up

**Actual behavior**
An error pops up and IntelliSense & error list don't work
image
IntelliSense_fail
image

StellaHuang95:
Fixes #7660


PranavSenthilnathan (approved):

AdamYoblick:
Just some background on the repro: This bug does repro in the public preview version of VS, but does NOT repro in IntPreview. I'm assuming there's some sort of registry check going on in that version on the VS side? @PranavSenthilnathan , can you confirm this?

In any case, I will be testing this change in the public preview version, since that's where the repro occurs. If it works there, I'll test it in int preview to be sure, and if it works there, we'll merge it in and get a VS insertion PR going asap.

AdamYoblick (approved):

Issue #7774, opened 2023-12-04
Implement round robin for new issues


StellaHuang95:
Similar to https://github.com/microsoft/pylance-release/blob/main/.github/workflows/issues.yml

PR #7776, opened 2023-12-04
Implement round robin for new issues

StellaHuang95:
Fixes https://github.com/microsoft/PTVS/issues/7774


.github/workflows/issues.yml
```diff
@@ -4,22 +4,76 @@ permissions:
...
-          ignore-if-assigned: true
-
-      - name: If labeled, remove triage
-        uses: andymckay/labeler@master
-        if: ${{ steps.no-labels.outcome == 'skipped' && join(github.event.issue.labels.*.name, ',') != 'needs triage' }}
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          result-encoding: string
+          script: |
+            const issue = await github.rest.issues.get({
+              issue_number: 7774,
```
AdamYoblick, line 23: is this issue number supposed to be hardcoded here (and in other places below)?

AdamYoblick (approved):
Approved with comments

.github/workflows/issues.yml
StellaHuang95: Yes that hardcoded issue is used to store the last person assigned an issue. Whomever is assigned this issue is the last person to get an incoming issue.

PR #7838, opened 2024-02-26
[Draft] Fix PTVS Regressions

StellaHuang95:


Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/StreamIntercepter.cs
```diff
@@ -44,10 +44,17 @@ class StreamIntercepter : Stream {
...
-            readHandler.Invoke(args);
-            return result;
+            var bytesRead = baseStream.Read(buffer, offset, count);
+            var args = new StreamData { bytes = buffer, offset = offset, count = bytesRead };
+            var newBytesRead = readHandler.Invoke(args);
+
+            if (newBytesRead != bytesRead) {
+                byte[] newBuffer = new byte[newBytesRead];
+                Array.Copy(args.bytes, args.offset, newBuffer, 0, newBytesRead);
+                buffer = newBuffer;
```
rchiodo, line 54: This is likely why it doesn't work. Modifying the buffer has no effect. You have to copy into this buffer, not change what it is pointing to.

rchiodo: Instead of allocating the 'newBuffer', just copy into the original buffer at the offset provided.

rchiodo: Maybe the offset isn't being set correctly. Or it could be the buffer isn't being cleared after the read updates it. 

Did you notice your read handler getting hit more than once?

Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/StreamIntercepter.cs
```diff
@@ -44,10 +44,15 @@ class StreamIntercepter : Stream {
...
         public override int Read(byte[] buffer, int offset, int count) {
-            var result = baseStream.Read(buffer, offset, count);
-            var args = new StreamData { bytes = buffer, offset = offset, count = result };
-            readHandler.Invoke(args);
-            return result;
+            var bytesRead = baseStream.Read(buffer, offset, count);
+            var args = new StreamData { bytes = buffer, offset = offset, count = bytesRead };
+            var newBytesRead = readHandler.Invoke(args);
+
+            if (newBytesRead != bytesRead) {
```
rchiodo, line 51: I think this copy is redundant. You're already copying to the buffer passed in inside the read handler.

Issue #7835, opened 2024-02-21
Enable pulling pre-release pylance into PTVS builds


StellaHuang95:
**Goal:**
Create vsix that uses the latest prerelease Pylance bits for CTI's non-sign off testing.

**Implementation:**
Set up some configurable build options that can pull pylance from the internal feed instead of the stable one, create and upload vsix as a build artifact and CTI should be able to test that way. 

** Progress **

Bundling into VS
- [x] Create a VS bootstrapper with our payload injected (see commented step in PTVS pipeline)
- [x] Publish the bootstrapper as a build artifact
- [x] Make sure CTI can consume the bootstrapper 

The bootstrapper work is complete with https://github.com/microsoft/PTVS/pull/7867.
It can be found in the pipeline artifact called "bootstrapper".

Getting pre-release pylance
- [x] Add the prerelease flag to PreBuild.ps1. This is ONLY used for tagging the build. The version is still used to download pylance.
- [x] Tag the build with "Pylance-Stable" for non-prerelease builds only
- [x] Change release pipeline trigger, add "Pylance-Stable" tag filter
- [x] Change pylance release.yml to launch PTVS pipeline on ALL releases, pass in pre-release flag when appropriate

StellaHuang95:
**Suggestion from Erik:**

"Regarding publishing Pylance prerelease builds to npm, I think you said that Adam had suggested having two separate feeds (stable and prerelease). But I think we can achieve the same thing with tags. If you look at the raw feed json (see links below), you can see a list of dist-tags at the top which point to the latest version that was published with each tag.
 
The DevDiv feed has "latest" (which I believe is the latest version published with no tag, aka stable), "pre", and "dev" (which seems to be old/dead), whereas the msft_consumption feed just has "latest" because we've never published prerelease builds there.
 
I think if we started publishing prerelease builds using the "pre" tag, PTVS could update its logic to pull either "latest" or latest "pre" and accomplish the same thing with less work. At least it would be less work on Pylance's side.

DevDiv: [devdiv.pkgs.visualstudio.com/_packaging/Pylance/npm/registry/@pylance/pylance](https://devdiv.pkgs.visualstudio.com/_packaging/Pylance/npm/registry/@pylance/pylance)
[...]

PR #7867, opened 2024-04-08
Create and publish bootstrapper for our test team

AdamYoblick:
Related to https://github.com/microsoft/PTVS/issues/7835

The point of the bootstrapper is to avoid having to do a VS insertion for our test team to consume our bits. I'm assuming the team can access our build artifacts, I'm waiting to hear back from them.

I ran a full build from this PR branch and the bootloader artifacts can be found at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9371793&view=artifacts&pathAsName=false&type=publishedArtifacts


AdamYoblick:
I tested installing PTVS through the bootstrapper and it did work. It just took forever to download, probably because the files are coming from vsdrop.

PR #7873, opened 2024-04-15
remove publish symbols task so we can get rid of pat

### LINKED ISSUE #7870: Publish Symbols - secret removal
When we publish symbols in PTVS, we do it using a PAT (https://github.com/microsoft/PTVS/blob/main/Build/templates/publish_symbols.yml#L27)
This process is described in the docs at https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/azure-artifacts/symbol-service

I removed both Symbol environment variables from the PTVS-Dev17 library at https://devdiv.visualstudio.com/DevDiv/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=381&path=PTVS-Dev17, including the secret PAT.

We should be able to remove the accountName and pat and just publish to the local org (devdiv) instead of symweb. Steve suggested removing the step completely because we also run `MicroBuildArchiveSymbols@4` which supposedly does the same thing.

So remove the publish symbols task and try to insert into VS. If VS doesn't complain about missing symbols, then we're good.

AdamYoblick:
Fixes #7870

I created an insertion PR at https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/544016 and waiting for the Symbol Check task to be done.
The Verify Symbols job is all green at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9417375&view=results.

### DIFF
--- a/Build/templates/publish_symbols.yml
+++ b/Build/templates/publish_symbols.yml
@@ -23,32 +23,10 @@ steps:
         VsPyProf*.dll
       TargetFolder: '$(Build.ArtifactStagingDirectory)/symbols'
 
-  # Publish Symbols
-  # This task uses two environment variables from the PTVS-Dev17 library (ArtifactServices.Symbol.AccountName and ArtifactServices.Symbol.PAT)
-  # If this step is failing due to authentication errors, the PAT is most likely expired. To regenerate the PAT:
-  # 1. Go to https://dev.azure.com/microsoft/_usersSettings/tokens (it has to be here in order for the PAT organization to be microsoft, which is required)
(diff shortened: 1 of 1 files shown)


StellaHuang95 (approved):

StellaHuang95:
so you will wait to see if all the insertion pr checks all pass before merging this PR right?

AdamYoblick:
> so you will wait to see if all the insertion pr checks all pass before merging this PR right?

No, the symbols check step is all I needed to pass, and it did. 😄

gramster (approved):
LGTM

PR #7879, opened 2024-04-24
Set ceval.tracing_possible flag for all subinterpreters on Python 3.9.

int19h:


StellaHuang95:
What does this `ceval.tracing_possible` flag do?

int19h:
It is an internal flag in the Python interpreter that is set when a tracing callback (which is responsible for things like stepping and breakpoints) is registered. If it is not set, the bytecode interpreter inner loop takes the fast path which is not going to invoke the callback, hence stepping/breakpoints not working.

For Python <3.9, this was a global flag shared by all subinterpreters in the process. In 3.9, it became a per-interpreter flag, but the code was only setting it for the first subinterpreter. I believe that ArcGIS is using multiple subinterpreters, and that is why they are affected by the issue (going by their version history, it started showing up for them when they switched to Python 3.9 for their hosted interpreter).

StellaHuang95 (approved):

PR #7886, opened 2024-04-29
Enable ingesting preview builds of pylance

AdamYoblick:
I've changed the logic in PreBuild.ps1 to no longer rely on npm's behavior when specifying "latest" for the version. "Latest" only seems to pull latest stable, which is what we want most of the time. But if there's a stable hotfix, "latest" won't pull it, which is bad.

Here's how the new logic works:

- If the user specifies a specific pylance version, we try to install it. In this case, `pylanceReleaseType` is ignored.
- If the user specifies "latest", we look for the latest version that matches the specified `pylanceReleaseType`. We can tell by the patch number.
- Finally, we tag the build with either `Pylance Stable` or `Pylance Preview`. 
  - This new tag will be used to control triggering of our release pipeline. We will only want to trigger releases (which insert into VS) on builds that are tagged with `Pylance Stable`.

The release pipeline change isn't visible in this PR because it's done directly in azdo. That pipeline is at https://devdiv.visualstudio.com/DevDiv/_releaseDefinition?definitionId=2762&_a=environments-editor-preview

I tested this by running the full build from my PR branch two times.
- The first run was with `stable` selected as the pylance release type. That build is at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9501352&view=results. The tagging and logic seems correct.
- The second run was with `preview` selected as the pylance release type. That build is at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9501353&view=results. The tagging and logic seems correct.

I still need to modify the pylance release process to pass in the appropriate release type once this PR is merged and to launch the PTVS pipeline on every pylance release (stable, hotfix, and preview)

### DIFF
--- a/Build/PreBuild.ps1
+++ b/Build/PreBuild.ps1
@@ -16,8 +16,16 @@ param (
     [string] $outdir, 
     
     # The version of pylance we should download, defaults to "latest"
+    # If "latest" is specified, the script will find the latest release of type $pylanceReleaseType (defaults to "stable")
+    # If an explicit version is specified, the script will always look for that specific version
     [Parameter()]
     [string] $pylanceVersion = "latest", 
+
+    # The type of pylance release we should download, defaults to "stable".
+    # This input is ignored if an explicit version is specified in $pylanceVersion.
+    [Parameter()]
+    [ValidateSet("stable", "pre-release")]
+    [string] $pylanceReleaseType = "stable", 
     
     # The version of debugpy we should download, defaults to "latest"
     [Parameter()]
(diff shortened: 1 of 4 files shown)


StellaHuang95:
On the pylance side, was the changed done to push all stable, hot fix stable and prerelease to the msft_consumption feed?

AdamYoblick:
> On the pylance side, was the changed done to push all stable, hot fix stable and prerelease to the msft_consumption feed?

Actually, no change was needed here. Erik told me we already do this, and I verified it myself as well. If you run the following command from your repo, you can see all the stable, hotfix, and preview versions of pylance: `npm view @pylance/pylance versions`

StellaHuang95:
> > On the pylance side, was the changed done to push all stable, hot fix stable and prerelease to the msft_consumption feed?
> 
> Actually, no change was needed here. Erik told me we already do this, and I verified it myself as well. If you run the following command from your repo, you can see all the stable, hotfix, and preview versions of pylance: `npm view @pylance/pylance versions`

I see, but when the pylance release pipeline triggers PTVS build pipeline to run, it will provide the argument that specifies `stable` or `prerelease` right? 

The end result is that the PTVS build pipeline will run for all new Pylance release, but the PTVS release pipeline will only be triggered by `stable` Pylance, is my understanding correct?

AdamYoblick:
> > > On the pylance side, was the changed done to push all stable, hot fix stable and prerelease to the msft_consumption feed?
> > 
> > 
> > Actually, no change was needed here. Erik told me we already do this, and I verified it myself as well. If you run the following command from your repo, you can see all the stable, hotfix, and preview versions of pylance: `npm view @pylance/pylance versions`
> 
> I see, but when the pylance release pipeline triggers PTVS build pipeline to run, it will provide the argument that specifies `stable` or `prerelease` right?
> 
> The end result is that the PTVS build pipeline will run for all new Pylance release, but the PTVS release pipeline will only be triggered by `stable` Pylance, is my understanding correct?

Correct. 😄

Build/PreBuild.ps1
```diff
@@ -70,10 +78,42 @@ $outdir = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPat
...
+        $versions = npm view @pylance/pylance versions --json | ConvertFrom-Json
+        [array]::Reverse($versions)
+
+        # Find the highest version with an appropriate patch number.
+        # Stable releases have patch numbers < 100, while preview releases have patch numbers >= 100.
+        foreach ($version in $versions) {
+            $patchVersion = $version.Split(".")[2]
+
+            if ($patchVersion -lt 100 -and $pylanceReleaseType -eq "stable") {
+                $pylanceVersion = $version
```
StellaHuang95, line 105: I don't much about Powershell, does `Split` give you a string or number? Do you need to convert it to a number before the comparison?

AdamYoblick: It's a normal string split. You give it a delimiter and it splits a string based on that delimiter. Returns an array of the split string, with the delimiter removed.

In this case, we're splitting a version number that looks like this M.m.p and taking the 3rd element from the split, which is the patch number.

AdamYoblick: I wasn't sure if I had to convert the element to an integer before the comparison, so I tried it out. Turns out you don't have to convert it, it does it automatically.

AdamYoblick: I will add in the conversion though, just so it's clear what the intent it 😄

StellaHuang95: Sounds great, thanks for doing this:)

StellaHuang95 (approved):

PR #7893, opened 2024-05-07
Remove weekly schedule, add some test files

AdamYoblick:
- Remove weekly schedule which was only there to grab the latest pylance release (since pylance release pipeline now chains into the ptvs pipeline)
- Add some files used for testing the run-pipeline rest api


StellaHuang95:
Is this to test that the pylance release pipeline can actually kick off PTVS pipeline?

AdamYoblick:
> Is this to test that the pylance release pipeline can actually kick off PTVS pipeline?

I already tested it through azdo using this yaml file and a test pipeline. The test run is at https://devdiv.visualstudio.com/DevDiv/_build?definitionId=21947
I left the test files in so we could run more random azdo tests if we want to without having to run the whole PTVS build.

The meat of this change is just to remove the weekly schedule since it's no longer needed.

StellaHuang95 (approved):

PR #7936, opened 2024-07-01
Fix attach in 3.6 and above when breakpoint is in an existing frame

rchiodo:
When walking callstacks, the old code would look for `EvalFrameFunc` (or `EvalFrameFunc_39`) on the callstack to identify where we entered a new frame (new function). 

This works for launch because we inject a new frame eval func on every interpreter. During attach, this only works if we happen to have run our injection code before the frame is created. If the user is inside a frame that then causes a breakpoint, `EvalFrameFunc` isn't in the callstack because the frame was entered before the debugger changed the default eval function.

The fix here is to instead use the default eval func to detect where new frames are. This causes some difficulties though. The default eval func looks like so:

```c
PyAPI_FUNC(PyObject *) _PyEval_EvalFrameDefault(PyThreadState *tstate, PyFrameObject *f, int exc);
```

When we walk the callstack and find this function call there we need to figure out the `PyFrameObject` that was passed to the function. This might seem easy (eval the `f`) parameter, but because of how C works, the `f` parameter is actually passed in a register when this function is called. By the time we're walking the callstack, that register has been reused for something else so the `f` parameter is not findable (especially in release/optimized code).

The workaround implemented was to use the `PyThreadState.frame` object to get the current frame. That however is a global variable on the thread and is therefore changed for every `_PyEval_EvalFrameDefault` call. Meaning if you have a callstack with more than one frame, only the topmost frame can be found with `PyThreadState.frame`. But a `frame` also has a `f_back` pointer that points to the previous frame on the callstack, so we can use that to find the frame based on the depth of the callstack.

I used the StackWalkFilter in PTVS to compute these back locations (as that's the only spot where we know the depth) and then using Concord's DkmDataItem class, stored the frame pointer in the DkmStackWalkFrame we create so we can use it later to evaluate local variables and such.


rchiodo:
Added some people that might have to work on this in the future.

azure-pipelines.yml
```diff
@@ -89,6 +89,8 @@ extends:
     pool:
       name: VSEngSS-MicroBuild2022-1ES
       os: windows
+    settings:
+      skipBuildTagsForGitHubPullRequests: true
```
AdamYoblick, line 93: Just wanted to comment on why this line is needed. See https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/features/sdlanalysis/sourceanalysisstage and scroll all the way down.

The 1ESPT can't tag builds that are for forks of github repos. So this line disables the tagging.

AdamYoblick (approved):

AdamYoblick (approved):

PR #7937, opened 2024-07-01
Add native example

rchiodo:
Adding an examples directory for future use

### DIFF
--- a/Examples/PythonNative/PythonNative.cpp
+++ b/Examples/PythonNative/PythonNative.cpp
@@ -2,6 +2,7 @@
 #include <Python.h>
 #include <Windows.h>
 #include <iostream>
+#include <filesystem>
 
 DWORD WINAPI runner(LPVOID lpParam)
 {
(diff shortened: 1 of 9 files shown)


Examples/PythonNative/PythonNative.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <iostream>
+
+DWORD WINAPI runner(LPVOID lpParam)
+{
+    // Runs the python runner that will listen on the named pipe.
+    Py_Initialize();
+
+    std::cout << "Python version: " << PY_VERSION << std::endl;
+
+    char startFile[] = "C:\\Users\\rchiodo\\source\\repos\\PythonNativeSeparateThread\\runner.py";
```
rchiodo, line 13: This could be relative. I'll change that.

Examples/PythonNative/PythonNative.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    PyObject* startObj = Py_BuildValue("s", startFile);
+    FILE* file = _Py_fopen_obj(startObj, "r+");
```
AdamYoblick, line 16: According to the docs at https://docs.python.org/3/c-api/veryhigh.html#c.PyRun_SimpleFile: On Windows, fp should be opened as binary mode (e.g. fopen(filename, "rb")). Otherwise, Python may not handle script file with LF line ending correctly.

rchiodo: Thanks. I'll change it.

rchiodo:
@AdamYoblick did you have any more feedback?

AdamYoblick:
> @AdamYoblick did you have any more feedback?

Looks good to me :shipit:

AdamYoblick (approved):

Issue #7959, opened 2024-07-22
Error executing test by pytest


kenichiice:
**Describe the bug**
When running tests using pytest 8.3.1 on Visual Studio 2022, the following error occurs.

Note that this error does not occur when pytest 8.0.2 is used.

```
Traceback (most recent call last):
  File "C:\PROGRAM FILES\MICROSOFT VISUAL STUDIO\2022\COMMUNITY\COMMON7\IDE\EXTENSIONS\MICROSOFT\PYTHON\CORE\testlauncher.py", line 189, in <module>
    main()
  File "C:\PROGRAM FILES\MICROSOFT VISUAL STUDIO\2022\COMMUNITY\COMMON7\IDE\EXTENSIONS\MICROSOFT\PYTHON\CORE\testlauncher.py", line 30, in main
    run(testRunner, coverage_file, test_file, args)
  File "C:\PROGRAM FILES\MICROSOFT VISUAL STUDIO\2022\COMMUNITY\COMMON7\IDE\EXTENSIONS\MICROSOFT\PYTHON\CORE\testlauncher.py", line 119, in run
    patch_translate_non_printable()
  File "C:\PROGRAM FILES\MICROSOFT VISUAL STUDIO\2022\COMMUNITY\COMMON7\IDE\EXTENSIONS\MICROSOFT\PYTHON\CORE\testlauncher.py", line 136, in patch_translate_non_printable
    translate_non_printable =  getattr(_pytest.compat, "_translate_non_printable")
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module '_pytest.compat' has no attribute '_translate_non_printable'
```

The `_translate_non_printable` that is causing the error seems to have been removed from the pytest code.

* https://github.com/pytest-dev/pytest/commit/3ba40954007d56b0f0cef251660c127fe6540137
* https://github.com/pytest-dev/pytest/discussions/12222

**Steps to Reproduce**
1. Create a Python project on VS 2022 that uses pytest as the testing framework.
2. Install the latest pytest (currently 8.3.1) in the Python environment of the Python project.
3. Add any test code to the Python project.
4. Run the test from VS's Test Explorer.

**Expected behavior**
No error should occur.

**Sample Code**
For example, the following test code will reproduce the error.
```
def test_failure():
    assert 1 == 1
```

**Configuration information (If you are providing a diagnostics file (see below), skip this section)**
[...]

PR #5644, opened 2019-08-21
pytest with parameterize values that include '::' breaking RunAll due…

bschnurr:
… to pytestId not found Fix #5635
Fix #5621

- fixes certain parameterised tests Id's that contain characters like "::" and "\n", This was causing RunAll to fail when a single pytestId wasn't found by pytest itself
- fixes Test Explorer displaying of parameteried tests by replacing "." and "::" wich TE uses to delimit on.
- Catch all fix for RunAll case to change the Pytest erorr handling of testcase not found to still continue with running the remaing tests, while still printing the same error

PR #7960, opened 2024-07-24
Remove the pytestId patch

StellaHuang95:
Fixes https://github.com/microsoft/PTVS/issues/7959

The patch was originally added to address the issue where `::` was not supported in pytest parameterized arguments. See https://github.com/microsoft/PTVS/pull/5644

I discovered that pytest now supports `::` in parameterized arguments. See https://github.com/pytest-dev/pytest/pull/9642.

In this PR, I removed the patch that replaced `::` with an alternative syntax compatible with pytest. You may notice that the patch also patched some other special characters such as `.`, `\r`, etc., I have verified that both test discovery and test execution via the Test Explorer work as expected for test names containing those characters.

**Test Discovery:**
image

**Test Execution:**
image

Although tests run as expected, the test hierarchy displayed in the Test Explorer seems wrong for tests with `::`, I will send it to the TE team.

### DIFF
--- a/Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/pytest/_discovery.py
+++ b/Python/Product/TestAdapter.Executor/PythonFiles/testing_tools/adapter/pytest/_discovery.py
@@ -9,27 +9,6 @@ import pytest
 from .. import util, discovery
 from ._pytest_item import parse_item
 
-#note: this must match testlauncher.py
-def patch_translate_non_printable():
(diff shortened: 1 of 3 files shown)


StellaHuang95:
I filed a bug about the fully qualified name here: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2172395

AdamYoblick (approved):
LGTM! 😄

PR #8019, opened 2024-10-07
update VSSDK.BuildTools package for newer versions of VS int preview

AdamYoblick:
I built locally and I was able to get past the previous build errors. Just make sure you run `powershell .\Build\PreBuild.ps1` from the repo root after you pull this change, since it's a new package version and it needs to be downloaded before building.


StellaHuang95 (approved):
Are the package version numbers in the packages.config automatically generated?

AdamYoblick:
> Are the package version numbers in the packages.config automatically generated?

No, none of them are. They are all manually set. I grabbed this one from the feed, I just looked at the highest version for that package.

bschnurr (approved):

PR #8033, opened 2024-10-21
Fix missing imports due to pylance thinking open documents are under default workspace

bschnurr:
- inject workspace definitions into the "initialize" call to pylance so that there is no chance of a textdocument/didopen call triggering the creation of a dfeault workspace before getfolders is called to setup the workspace.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
@@ -243,9 +243,9 @@ namespace Microsoft.PythonTools.LanguageServerClient {
         }
 
         private Task TriggerWorkspaceUpdateConfig() {
+            Debug.WriteLine("Settings Changed");
             return InvokeDidChangeConfigurationAsync(new LSP.DidChangeConfigurationParams() {
-                // If we pass null settings and workspace.configuration is supported, Pylance will ask
-                // us for per workspace configuration settings. Otherwise we can send
+                // Pylance will ask us for per workspace configuration settings. We can send
                 // default workspace settings here.
                 Settings = GetSettings()
             });
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
@@ -590,14 +590,14 @@ namespace Microsoft.PythonTools.LanguageServerClient {
             return Tuple.Create(data, true);
         }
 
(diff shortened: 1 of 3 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -593,24 +590,35 @@ private Tuple<StreamData, bool> OnSendToServer(StreamData data) {
             return Tuple.Create(data, true);
         }
 
-        private async Task OnWorkspaceOpening(object sende, EventArgs e) {
+        private IEnumerable<WorkspaceFolder> GetFolders() {
```
rchiodo, line 593: Could you have this always return a list? Every place you call it you're turning it into a list.

rchiodo (approved):

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -138,7 +135,7 @@ public async Task<Connection> ActivateAsync(CancellationToken token) {
             await JoinableTaskContext.Factory.SwitchToMainThreadAsync();
             CreateClientContexts();
 
-            _deferredSettingsChangedTimer = new Timer(OnDeferredSettingsChanged, state: null, Timeout.Infinite, Timeout.Infinite);
```
bschnurr, line 141: just replacing this function with a lamda to cut down on fuctions

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -241,8 +238,16 @@ private async Task OnWorkspaceFolders(object sender, WorkspaceFolders.WorkspaceF
...
-            OnWorkspaceFolderWatched(this, EventArgs.Empty);
+            await TriggerWorkspaceUpdateConfig();
+        }
+
+        private Task TriggerWorkspaceUpdateConfig() {
+            Debug.WriteLine("Settings Changed");
+            return InvokeDidChangeConfigurationAsync(new LSP.DidChangeConfigurationParams() {
+                // Pylance will ask us for per workspace configuration settings. We can send
+                // default workspace settings here.
+                Settings = GetSettings()
```
bschnurr, line 249: now include actual settings which get applied to any default workspace

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
bschnurr: true. such a small list. no reason to make it enumerable

PR #8043, opened 2024-10-28
fix for microsoft store failing to create virtual environment due to …

bschnurr:
…outdated virtualenv

- also fix for when env folder is missing a "lib" folder (could be due to improper env path)
- fix for repl rootpath (which is null), when iterating over workspace contexts


Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
```diff
@@ -162,6 +190,13 @@ private static async Task ContinueCreate(IServiceProvider provider, IPythonInter
                 }
             }
 
+            try {
+                if (!useVEnv) {
```
rchiodo, line 194: I think we might need to ask the user here? We're running something on their behalf in their global python.

Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
```diff
@@ -148,9 +148,37 @@ public static async Task<IPythonInterpreterFactory> CreateAndAddFactory(
...
+                        output.Show();
+                    }
+                }
+            }
+        }
+
         private static async Task ContinueCreate(IServiceProvider provider, IPythonInterpreterFactory factory, string path, bool useVEnv, Redirector output) {
             path = PathUtils.TrimEndSeparator(path);
-            var name = Path.GetFileName(path);
+            var name =Path.GetFileName(path);
```
AdamYoblick, line 181: nit: space after the equals

AdamYoblick (approved):

Python/Product/PythonTools/PythonTools/Environments/VirtualEnv.cs
bschnurr: ah true. i think there is an example of this.. i'll try to find where we do something similar

bschnurr: I think i'll try to detect this error case instead and just print this command for the user to run themselves

rchiodo (approved):

PR #8044, opened 2024-10-29
Add another example of interop between C++ and Python

rchiodo:
This is a simpler example than what I already wrote for interoping between C++ and Python. This example calls C++ code from Python.


Python/Product/Debugger.Concord/Proxies/Structs/PyFrameObject.cs
```diff
@@ -68,29 +68,26 @@ public PyFrameObject(DkmProcess process, ulong address)
...
+                name = "_PyEval_EvalFrameDefault";
             }
-
+            addr = pythonInfo.DLLs.Python.GetFunctionAddress(name);
             if (addr == 0) {
                 return false;
             }
 
-            return frame.InstructionAddress.IsInSameFunction(process.CreateNativeInstructionAddress(addr));
+            var addressMatch = frame.InstructionAddress.IsInSameFunction(process.CreateNativeInstructionAddress(addr));
```
rchiodo, line 81: I wasn't getting address matching even when the native frame was the actual `_PyEval_EvalFrameDefault` frame. Maybe something is messed up with the pdb because the RVAs were off.

gramster (approved):

bschnurr (approved):

PR #8048, opened 2024-11-01
Remove non 3.9 code from Mixed mode debugger

rchiodo:
This removes code that was only necessary to debug 3.8 and earlier. Should make life simpler going forward as we add support for 3.10, 11, 12, and 13


Python/Product/Common/Parsing/PythonLanguageVersion.cs
```diff
@@ -43,6 +43,8 @@ public enum PythonLanguageVersion {
         V39 = 0x0309,
         V310 = 0x030a,
         V311 = 0x030b,
+        V312 = 0x030c,
+        V313 = 0x030d,
```
StellaHuang95, line 47: Any Python versions not in this list are assigned `PythonLanguageVersion.None`, which is used to trigger a warning message that says `xxx version is not fully supported`, in the notification bar or pop-up windows when user switches the active environment. With this change, the warning will not appear for versions 3.12 and 3.13. Do we want to remove this warning before these versions are fully supported?

rchiodo: I don't think that's necessary. We're not going to ship these changes until we support 3.12/3.13 so I don't think it matters. We can always fix the warning before we ship.

StellaHuang95 (approved):

Python/Product/Common/Parsing/PythonLanguageVersion.cs
StellaHuang95: Sorry I am confused, I thought you said you're going to hold off on merging these changes until 3.12 are fully supported?

rchiodo: I said ship, but I assumed PTVS didn't auto ship. We have to purposefully insert it, right?

rchiodo: I don't want to pile up a whole bunch of changes in a single branch and then merge it all at once. That would be a huge pain to review.

StellaHuang95: If there's a new change merged into `main` later, the insertion PR created that inserts changes into `VS main` will contain this change as well.

rchiodo: I don't foresee us having to ship something else before mixed mode debugging or 3.12 support, but I suppose that could be possible. We could just fix the warning if that's the case.

AdamYoblick:
Yeah PTVS doesn’t release manually. Every change in main inserts into VS main. We can always abandon the PRs though, we don’t have to merge them.
________________________________

Python/Product/Common/Parsing/PythonLanguageVersion.cs
StellaHuang95: yeah it could be as small as a bugfix made into PTVS main

rchiodo: This is the warning right?

```cs
                        if (version.ToLanguageVersion() == PythonLanguageVersion.None) {
                            Utility.MessageBox.ShowWarningMessage(Site, Strings.PythonVersionNotSupportedInfoBarText.FormatUI(_active.Configuration.Description));
```

So now it won't show anymore? I think that's fine. 3.12 and 3.13 are actually supported for some stuff, just not mixed mode or profiling at the moment. It was allowing 3.10 and 3.11 through already and they weren't supported for mixed mode either.

rchiodo:
Maybe I should merge all of this mixed mode stuff into a branch then if we're worried about making other change at the same time. Bill is working on fixing a microsoft store problem. I don't think we'd want to hold that up until mixed mode works

Python/Product/Common/Parsing/PythonLanguageVersion.cs
StellaHuang95: Yeah we did mention that `some features might not work` in the warning just to let users know that it's not fully supported, but having a separate branch for mixed mode stuff sounds good!

PR #8056, opened 2024-11-07
Get 3.11 to work in mixed mode

rchiodo:
This gets the two test apps we have in the [examples](https://github.com/microsoft/PTVS/tree/main/Examples) folder working with 3.11.

Breakpoints, callstacks, locals, and stepping all work. Probably need more testing around different types of variables, but that shouldn't be 3.11 specific.

### DIFF
--- a/Examples/PythonNative/PythonNative.vcxproj
+++ b/Examples/PythonNative/PythonNative.vcxproj
@@ -104,14 +104,14 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <AdditionalIncludeDirectories>$(LOCALAPPDATA)\Programs\Python\Python310\include;$(LOCALAPPDATA)\Programs\Python\Python311\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(LOCALAPPDATA)\Programs\Python\Python311\include;$(LOCALAPPDATA)\Programs\Python\Python310\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <LanguageStandard>stdcpp17</LanguageStandard>
       <LanguageStandard_C>stdc17</LanguageStandard_C>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
-      <AdditionalLibraryDirectories>$(LOCALAPPDATA)\Programs\Python\Python310\libs;$(LOCALAPPDATA)\Programs\Python\Python311\libs;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalLibraryDirectories>$(LOCALAPPDATA)\Programs\Python\Python311\libs;$(LOCALAPPDATA)\Programs\Python\Python310\libs;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
     </Link>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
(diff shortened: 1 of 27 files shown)


Python/Product/DebuggerHelper/trace.cpp
```diff
@@ -51,10 +51,10 @@ struct {
         int64_t ob_size;
     } PyVarObject;
     struct {
-        int64_t f_back, f_code, f_globals, f_locals, f_lineno;
+        int64_t f_back, f_code, f_globals, f_locals, f_lineno, f_frame;
     } PyFrameObject;
     struct {
-        int64_t co_varnames, co_filename, co_name;
```
rchiodo, line 57: co_varnames wasn't used thankfully in the trace code. So I just removed it. It doesn't exist anymore in 3.11 and would have been a pain to add support for here.

Examples/PythonNative/PythonNative.vcxproj
```diff
@@ -104,14 +104,14 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <AdditionalIncludeDirectories>$(LOCALAPPDATA)\Programs\Python\Python310\include;$(LOCALAPPDATA)\Programs\Python\Python311\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(LOCALAPPDATA)\Programs\Python\Python311\include;$(LOCALAPPDATA)\Programs\Python\Python310\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
heejaechang, line 107: this line seems just ordering change? not actual content change?

rchiodo: Yeah it just made this example default to using 3.11 first.

Python/Product/Debugger.Concord/DebuggerOptions.cs
```diff
@@ -20,6 +20,9 @@
 namespace Microsoft.PythonTools.Debugger.Concord {
     public static class DebuggerOptions {
         // These are intentionally not implemented as auto-properties to enable easily changing them at runtime, including when stopped in native code.
+        //
+        // These show up in commands in the debugger watch window if you set the registry value:
+        // HKCU/Software/Microsoft/PythonTools/Debugger PythonDeveloper: DWORD = 1
```
debonte, line 25: Is the space between Debugger and PythonDebugger intentional? Not supposed to be a slash?

Python/Product/Debugger.Concord/ModuleManager.cs
```diff
@@ -160,6 +165,27 @@ public static void PyCode_New(DkmThread thread, ulong frameBase, ulong vframe, u
                 }.SendLower(process);
             }
 
+            public static void _PyCode_New(DkmThread thread, ulong frameBase, ulong vframe, ulong returnAddress) {
```
debonte, line 168: Does this function name make sense if you know what's going on here?

Python/Product/Debugger.Concord/Proxies/StructProxy.cs
```diff
@@ -20,6 +20,7 @@
 using Microsoft.Dia;
 using Microsoft.PythonTools.Common.Parsing;
 using Microsoft.VisualStudio.Debugger;
+using Microsoft.VisualStudio.Debugger.Interop;
```
debonte, line 23: Is this `using` necessary? Seems weird that code was deleted, but a using was added?

Python/Product/Debugger.Concord/Proxies/Structs/PyCodeObject310.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
```
debonte, line 1: Add copyright comments in the new files?

Python/Product/Debugger.Concord/DebuggerOptions.cs
rchiodo: Yes it's supposed to identify the key and the value. I could be more explicit.

Python/Product/Debugger.Concord/ModuleManager.cs
rchiodo: Yeah it's the name of a function in the python dll. It has to match the name.

Python/Product/Debugger.Concord/Proxies/StructProxy.cs
rchiodo: Oh that's probably a mistake. I can just remove that.

Python/Product/Debugger.Concord/Proxies/Structs/PyCodeObject310.cs
rchiodo: Oh yeah, good idea. Didn't realize that.

AdamYoblick (approved):

Issue #7923, opened 2024-06-05
Anaconda displays incorrectly in environment list.


linette-zyy:
**Describe the bug**
![image](https://github.com/microsoft/PTVS/assets/59192491/1872022c-ee37-48f3-ab53-84c88040462d)

**Steps to Reproduce**
1. Install Anaconda (downloads from https://www.anaconda.com/download/)
2. Create Python Application project
3. Right-click on Python Environments node in SE and select "View All Python Environments"
4. Check the environment list

**Expected behavior**
It should display complete version.

**Additional context and screenshots**
Anaconda version is Python 3.11.7
![image](https://github.com/microsoft/PTVS/assets/59192491/2e2f7c5b-a26c-4ad6-8015-cfcf872745d3)
But it displays Python 3.1
![image](https://github.com/microsoft/PTVS/assets/59192491/36151ce6-6dec-4c3b-aa7c-469e57b556a4)

I also install Python 3.12 from python.org. It displays correct.
![image](https://github.com/microsoft/PTVS/assets/59192491/f37bf1c8-2e51-4b3c-9ef9-1c4885f4183f)

StellaHuang95:
Might be related: https://github.com/microsoft/PTVS/issues/7901#issuecomment-2155832284

StellaHuang95:
@linette-zyy Does this still repro? How did you add the conda environment to your Python project? Could you show more detailed steps or a git that repros the issue?

StellaHuang95:
Never mind I was able to repro it now. I had to uninstall the python 3.12 from python.org and register Anaconda as my default Python. ![Image](https://github.com/user-attachments/assets/e1c2664d-77d7-4e16-9347-455df3dcf831)

PR #8057, opened 2024-11-07
Fix wrong Conda Python version being displayed in the UI

StellaHuang95:
Fixes https://github.com/microsoft/PTVS/issues/7923

PTVS find Python installations by checking the registry entries. The python.org python has some subkeys that PTVS uses to get the version number, while conda python does not.
image

Fix the issue by extracting the first 4 characters of the `tag` if python version is > 3.10, and the first 3 characters otherwise.


rchiodo (commented):
Maybe a regex would be better?

Something like so:

```
\d+\.\d+\.*\d*
```

Which matches 3.13, 3.13.0, 3.2, etc.

StellaHuang95:
> Maybe a regex would be better?
> 
> Something like so:
> 
> ```
> \d+\.\d+\.*\d*
> ```
> 
> Which matches 3.13, 3.13.0, 3.2, etc.

@rchiodo sure, this should give me just the major.minor version from the version string

rchiodo (approved):

PR #8063, opened 2024-11-13
Support for 312 mixed mode debugging

rchiodo:
This should get 3.12 mixed mode debugging working. I tested with the two mixed mode examples we have with stepping and adding some extra types as variables. 

3.12 made a rather difficult change to handle. No longer could we just WriteProcessMemory the `use_tracing` flag on the PyInterpreter. `use_tracing` was replaced with sys.monitoring. Using sys tracing required a function call now.

I first attempted to do this on startup and func eval the `PyEval_SetTraceAllThreads` function but I couldn't find the right spot to actually eval this that wouldn't throw an exception. 

So instead, the trace helper when it first sees a python eval, it injects the tracing then.

There was also a bunch of changes with other data types, like unicode objects and longs.

### DIFF
--- a/Python/Product/Debugger.Concord/CppExpressionEvaluator.cs
+++ b/Python/Product/Debugger.Concord/CppExpressionEvaluator.cs
@@ -32,7 +32,10 @@ namespace Microsoft.PythonTools.Debugger.Concord {
         private readonly DkmStackWalkFrame _nativeFrame;
         private readonly DkmInspectionContext _cppInspectionContext;
 
-        public CppExpressionEvaluator(DkmInspectionContext inspectionContext, DkmStackWalkFrame stackFrame) {
+        public CppExpressionEvaluator(DkmInspectionContext inspectionContext, DkmStackWalkFrame stackFrame) 
+            : this(inspectionContext.InspectionSession, inspectionContext.Radix, stackFrame) { 
+        }
+        public CppExpressionEvaluator(DkmInspectionSession inspectionSession, uint radix, DkmStackWalkFrame stackFrame, DkmEvaluationFlags flags = DkmEvaluationFlags.TreatAsExpression | DkmEvaluationFlags.NoSideEffects) {
             _process = stackFrame.Process;
             var thread = stackFrame.Thread;
 
@@ -53,8 +56,8 @@ namespace Microsoft.PythonTools.Debugger.Concord {
                     DkmStackWalkFrameFlags.None, null, stackFrame.Registers, null);
             }
 
-            _cppInspectionContext = DkmInspectionContext.Create(inspectionContext.InspectionSession, _process.GetNativeRuntimeInstance(), thread, Timeout,
-                DkmEvaluationFlags.TreatAsExpression | DkmEvaluationFlags.NoSideEffects, DkmFuncEvalFlags.None, inspectionContext.Radix, CppLanguage, null);
+            _cppInspectionContext = DkmInspectionContext.Create(inspectionSession, _process.GetNativeRuntimeInstance(), thread, Timeout,
+                flags, DkmFuncEvalFlags.None, radix, CppLanguage, null);
         }
 
         public CppExpressionEvaluator(DkmThread thread, ulong frameBase, ulong vframe, DkmEvaluationFlags flags = DkmEvaluationFlags.TreatAsExpression | DkmEvaluationFlags.NoSideEffects) {
@@ -94,8 +97,8 @@ namespace Microsoft.PythonTools.Debugger.Concord {
             return TryEvaluate(GetExpressionForObject(moduleName, typeName, address, tail));
         }
 
-        public string Evaluate(string expr) {
-            var er = TryEvaluate(expr);
+        public string Evaluate(string expr, DkmEvaluationFlags flags = DkmEvaluationFlags.NoSideEffects) {
+            var er = TryEvaluate(expr, flags);
             var ser = er as DkmSuccessEvaluationResult;
             if (ser == null) {
                 throw new CppEvaluationException(er);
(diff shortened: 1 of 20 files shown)


Python/Product/Debugger.Concord/CppExpressionEvaluator.cs
```diff
@@ -32,7 +32,10 @@ internal class CppExpressionEvaluator {
         private readonly DkmStackWalkFrame _nativeFrame;
         private readonly DkmInspectionContext _cppInspectionContext;
 
-        public CppExpressionEvaluator(DkmInspectionContext inspectionContext, DkmStackWalkFrame stackFrame) {
+        public CppExpressionEvaluator(DkmInspectionContext inspectionContext, DkmStackWalkFrame stackFrame)
```
rchiodo, line 35: I added some overloads here that allowed me to run the EE with just a stack frame and to allow side effects (necessary in order to call a function).

Python/Product/Debugger.Concord/Proxies/Structs/PyFrameObject311.cs
```diff
@@ -41,24 +43,46 @@ public PyFrameObject311(DkmProcess process, ulong address)
...
+
+        public override int ComputeLineNumber(DkmInspectionSession inspectionSession, DkmStackWalkFrame frame) {
+            var setLineNumber = GetFieldProxy(_fields.f_lineno).Read();
+            if (setLineNumber == 0) {
+                // We need to use the CppExpressionEvaluator to compute the line number from
+                // our frame object. This function here: https://github.com/python/cpython/blob/46710ca5f263936a2e36fa5d0f140cf9f50b2618/Objects/frameobject.c#L40-L41
+                var evaluator = new CppExpressionEvaluator(inspectionSession, 10, frame, DkmEvaluationFlags.TreatAsExpression);
+                var funcAddr = Process.GetPythonRuntimeInfo().DLLs.Python.GetFunctionAddress("PyFrame_GetLineNumber");
+                var frameAddr = Address; 
+                setLineNumber = evaluator.EvaluateInt32(string.Format("((int (*)(void *)){0})({1})", funcAddr, frameAddr), DkmEvaluationFlags.EnableExtendedSideEffects);
```
rchiodo, line 82: This code is evaling a call to a function pointer. That looks like this : `((int (*)(void *))<func addr>)(<frame ptr addr>)`. It's how you call a function pointer in C/C++.

The function I'm calling is `PyFrame_GetLineNumber`, which if you look in the CPython source, it computes the line number if it's not already set. It's a public API, so it should always be there.

Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs
```diff
@@ -63,5 +67,33 @@ public PointerProxy<PyFrameObject> frame_obj {
...
+        }
+
+        public PointerProxy<PyFrameObject> FindBackFrame() {
+            // Trace.WriteLine("Searching for back frame ...");
+            var frame = previous.TryRead();
+            while (frame != null && !frame.OwnedByThread()) {
+                frame = frame.previous.TryRead();
+            }
+
+            // Make sure this frame_obj is pointing to this frame
```
rchiodo, line 90: I should update this comment to say why this might be the case.

Python/Product/Debugger.Concord/ModuleManager.cs
```diff
@@ -160,6 +165,27 @@ public static void PyCode_New(DkmThread thread, ulong frameBase, ulong vframe, u
                 }.SendLower(process);
             }
 
+            public static void _PyCode_New(DkmThread thread, ulong frameBase, ulong vframe, ulong returnAddress) {
+                var process = thread.Process;
+                var cppEval = new CppExpressionEvaluator(thread, frameBase, vframe);
+
+                var filenamePtr = cppEval.EvaluateUInt64("con->filename");
```
bschnurr: should we check 'filenamePtr " for null?

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -551,7 +551,8 @@ private Tuple<StreamData, bool> OnSendToServer(StreamData data) {
                                 capabilities["workspace"]["configuration"] = true;
                                 
                                 var folders = GetFolders();
-                                Debug.Assert(folders.Any(), "no workspace or projects found");
+                                // Debug.Assert(folders.Any(), "no workspace or projects found");
+                                // This is allowed with say a loose file.
```
bschnurr: i'll look into this bug

Python/Product/Debugger.Concord/ModuleManager.cs
rchiodo: The FromAddress (line after this) would return a null object I believe, so that check should be redundant.

Yeah it has this code here:

```
 public static PyObject FromAddress(DkmProcess process, ulong address) {
     if (address == 0) {
         return null;
     }
```

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
StellaHuang95: That's the assert failure I've been seen, it repros even before opening any loose files. But for some reason I can't get it to repro anymore.

Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs
```diff
@@ -58,5 +67,33 @@ public PointerProxy<PyFrameObject> frame_obj {
...
+            if (Process.GetPythonRuntimeInfo().LanguageVersion <= PythonLanguageVersion.V310) {
+                return true;
+            }
+            var owner = (GetFieldProxy(_fields.owner) as IValueStore).Read();
+            var charOwner = owner.ToString()[0];
+            return (int)charOwner  == FRAME_OWNED_BY_THREAD;
+        }
+
+        public PointerProxy<PyFrameObject> FindBackFrame() {
+            // Trace.WriteLine("Searching for back frame ...");
```
bschnurr, line 84: wonder if we should have a logger setup for enabling traces

bschnurr: maybe helps for tests

bschnurr (approved):

Issue #6774, opened 2021-11-05
The Python installed from Microsoft Store couldn't view installed packages when first use the environment.


linette-zyy:
**Environment**
Build of Visual Studio: 17.1.0 Preview 2.0 [31904.282.main]
Build of Python Package: 17.0.21295.2
OS version: Windows 11 Enterprise 21H2
Python install from Microsoft Store

**Steps to Reproduce**
1. Create a new python application
2. Expand the one Python Environment in "Python Environments" node
3. Check the package list

**Expected behavior**
All the packages should listed
![image](https://user-images.githubusercontent.com/59192491/140449102-613675c3-7637-4a8c-a643-30aebbdbd535.png)

![image](https://user-images.githubusercontent.com/59192491/140449065-b9258e19-e31a-4a3e-b67d-460678cb2069.png)

**Additional context and screenshots**
The Python installed from Microsoft Store couldn't view installed packages when first use the environment.
![image](https://user-images.githubusercontent.com/59192491/140449184-b0b352b6-9a5a-4a7c-bd86-2cf6494c92d0.png)
![image](https://user-images.githubusercontent.com/59192491/140449228-02ac926d-2d34-4dc7-94ed-28374fe768ad.png)

After I restart the VS, this issue is not reproduce.

linette-zyy:
This issue also can reproduce on today's environment.
![image](https://user-images.githubusercontent.com/59192491/141723441-7b9deba7-3add-4d7d-b329-9fe71304a439.png)
![image](https://user-images.githubusercontent.com/59192491/141723427-f757e6d6-4a98-4a25-b031-abaa1bfaf75c.png)
After restart the VS, we can view installed packages.

StellaHuang95:
I can't repro this. @linette-zyy Could you verify this again on the latest build?

linette-zyy:
![image](https://user-images.githubusercontent.com/59192491/172799541-7c47b33f-f8ae-47aa-a545-7402fceef478.png)
@StellaHuang95 In today's build, I can't repro this when install packages(requests) in the 'Packages(PyPI)' search box
![install_from_MS](https://user-images.githubusercontent.com/59192491/172800551-50423c7a-fb6b-4bf1-aedc-91b5f950fabd.gif)
But i can repro this  when install packages(requests) in the command line
![install_from_MS_commandline](https://user-images.githubusercontent.com/59192491/172800960-2f7abdfc-3e89-4544-8e84-202937e16ac1.gif)

[...]

PR #8064, opened 2024-11-13
fix for msft python failing to detect pip packages due to pip needing…

bschnurr:
… to be updated.

in the StandardErrorLines it returns:

"[notice] A new release of pip is available: 23.0.1 -> 24.3.1"
"[notice] To update, run: C:\\Users\\bschnurr\\AppData\\Local\\Microsoft\\WindowsApps\\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\\python.exe -m pip install --upgrade pip"

now we check for no packages found and error strings. if so we just set packages to null and it falls back to directory search to display instead.

fixes https://github.com/microsoft/PTVS/issues/6774

### DIFF
--- a/Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
+++ b/Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
@@ -444,6 +444,15 @@ namespace Microsoft.PythonTools.Interpreter {
                                             .Where(p => p.IsValid)
                                             .OrderBy(p => p.Name)
                                             .ToList();
+
+
+                                        if(packages.Count() == 0 && proc.StandardErrorLines.Any()) {
+                                            Debug.WriteLine("Failed to run pip to collect packages");
+                                            foreach (var line in proc.StandardErrorLines) {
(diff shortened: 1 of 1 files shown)


Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
```diff
@@ -444,6 +444,15 @@ CancellationToken cancellationToken
...
                                             .OrderBy(p => p.Name)
                                             .ToList();
+
+
+                                        if(packages.Count() == 0 && proc.StandardErrorLines.Any()) {
+                                            Debug.WriteLine("Failed to run pip to collect packages");
+                                            foreach (var line in proc.StandardErrorLines) {
+                                                Debug.WriteLine(line);
+                                            }
+                                            packages = null;
```
bschnurr, line 454: below there is a case for null packages.
   if (packages == null) {
                        // Pip failed, so return a directory listing

rchiodo (approved):

PR #8066, opened 2024-11-13
Try fixing the PTVS crash

StellaHuang95:
fixes https://github.com/microsoft/PTVS/issues/8060

I tried multiple ways including what users described in all similar issues but still wasn't able to repro the crash. From reading the call stacks looks like it's trying to call `GetProjectProperty` on a disposed `PythonProjectNode` object. So adding the check to return early if it's disposed. I did manually testing on all different scenarios and it doesn't crash or affect any behaviors. I will ask users to try again once the fix is shipped.


Python/Product/PythonTools/PythonTools/Project/PythonProjectNode.cs
```diff
@@ -1082,7 +1085,7 @@ protected override void Dispose(bool disposing) {
 
                 _recreatingAnalyzer.Dispose();
             }
-
+            _isDisposed = true;
```
rchiodo, line 1088: Maybe it would be better to put this at the top before all the other things are removed? (In case one of them throws an exception.

Then you can also check if it's already disposed or not and throw an exception if it is.

rchiodo (approved):

PR #8067, opened 2024-11-13
Fix for assert that was triggered when a loose python file was open w…

bschnurr:
…ith no folder or project.

pylance will use a default workspace

fixes https://github.com/microsoft/PTVS/issues/8062


StellaHuang95:
Does Pylance in VSCode also uses default workspace if it's loose file?

StellaHuang95 (approved):

rchiodo (approved):

bschnurr:
> Does Pylance in VSCode also use default workspace if it's loose file?

yup. thats how it works in vscode too

PR #8089, opened 2024-11-29
Fix REPL hang

bschnurr:
New "Empty" exception needs to be handled when there is no message.


Python/Product/PythonTools/ptvsd_repl_launcher.py
```diff
@@ -113,7 +115,27 @@ def _run_repl():
     # execute code on the main thread which we can interrupt
     backend.execution_loop()    
 
+def mock_server(host="127.0.0.1", port=57905):
```
rchiodo, line 118: What's this for? Looks like it's for testing?

bschnurr: ya, actually that was a mistake. thanks

rchiodo (approved):

PR #8096, opened 2024-12-05
Make glass tests run as part of a PR

rchiodo:
[Previously](https://github.com/microsoft/PTVS/pull/8087) I added support for running glass tests for PTVS using some python scripts.

This PR uses those same python scripts to run the tests during a PR. 

It was a little bit tricky as it needs the 'debug' version of python (which the setup python task doesn't install), so it had to download the python installer for windows and install it with the debug bits. 

Tests then run using the glass bits from different parts of the VS drop.

### DIFF
--- a/Build/templates/run_tests.yml
+++ b/Build/templates/run_tests.yml
@@ -6,9 +6,10 @@ steps:
       outputVar: InstallPython312
 
   # Debug helper, print all variables
-  - task: printAllVariables@1
-    displayName: 'Print all variables'
-
+  - powershell: |
+      Get-ChildItem Env:
+    displayName: 'Dump all variables'
+    
   # Build the product
   - task: MSBuild@1
     displayName: 'Build product'
--- a/Build/templates/run_tests.yml
+++ b/Build/templates/run_tests.yml
@@ -10,6 +10,10 @@ steps:
       Get-ChildItem Env:
     displayName: 'Dump all variables'
 
+- powershell: |
+    echo "##vso[task.setvariable variable=PYTHON_PATH]$env:POWERSHELL6_INSTALLPYTHON312"
+  displayName: 'Set PYTHON_PATH variable'
+
   # Build the product
   - task: MSBuild@1
     displayName: 'Build product'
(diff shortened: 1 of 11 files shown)


Build/templates/run_tests.yml
```diff
@@ -0,0 +1,104 @@
...
+  - template: ./install_python.yml
+    parameters:
+      pythonVersion: '3.9.13' # Last version with an installer
+  - template: ./install_python.yml
+    parameters:
+      pythonVersion: '3.10.11' # Last version with an installer
+  - template: ./install_python.yml
+    parameters:
+      pythonVersion: '3.11.9' # Last version with an installer
+  - template: ./install_python.yml
```
bschnurr, line 12: whats going on here, we install a bunch a different python versions but only use one?

rchiodo: We install a bunch of python versions so we can test with them all. The glass setup uses these versions to generate tests for each one.

rchiodo: We use the 3.12 version to run the scripts though.

bschnurr: should we use a strategy?

```yml
- job: BuildAndTest
  displayName: Build and Test with Python
  strategy:
    matrix:
      Python3_9:
        pythonVersion: '3.9.13' # Last version with an installer
      Python3_10:
        pythonVersion: '3.10.11' # Last version with an installer
      Python3_11:
        pythonVersion: '3.11.9' # Last version with an installer
      Python3_12:
        pythonVersion: '3.12.7'
        outputVar: InstallPython312
      Python3_13:
        pythonVersion: '3.13.0'
  steps:
```

bschnurr: or something like

```yml
parameters:
  pythonVersions: ['3.9.13', '3.10.11', '3.11.9', '3.12.7', '3.13.0']

jobs:
- template: matrix_job.yml
  parameters:
    versions: ${{ parameters.pythonVersions }}
matrix_job.yml Template
```
```yml

parameters:
  versions: []

jobs:
- job: BuildAndTest
  displayName: Build and Test with Python
  strategy:
    matrix: ${{ convertToMatrix(parameters.versions) }}
  steps:
    - script: |
        echo "Testing Python version ${{ matrix.pythonVersion }}"
      displayName: Test Python Version
```

rchiodo: The test runs all versions at the same time though

rchiodo: Otherwise we'd be duplicating all the drop, build, and setup for each job

rchiodo: Hmm, maybe I could cache the drop and glass folders though.

rchiodo: Then run the tests in parallel for each version

bschnurr (approved):

PR #8104, opened 2024-12-09
add a new weekly pipeline to update the glass binaries from main branch.

bschnurr:
user PRs weren't able to auth

### DIFF
--- a/Build/glass/cache_glass.yml
+++ b/Build/glass/cache_glass.yml
@@ -1,0 +1,41 @@
+schedules:
+  - cron: "0 0 * * 0" # Runs at 00:00 UTC every Sunday
+    displayName: Weekly build
+    branches:
+      include:
+        - main
+    always: true
+
+steps:
+  - task: AzureCLI@2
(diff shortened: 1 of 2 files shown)


Build/glass/cache_glass.yml
```diff
@@ -0,0 +1,41 @@
...
+
+  # Setup the glass test folder
+  - task: PythonScript@0
+    displayName: 'Setup glass test folder'
+    env:
+      SYSTEM_ACCESSTOKEN: $(DropToken)
+    inputs:
+      scriptSource: 'filePath' # 'filePath' | 'inline'. Required. Script source. Default: filePath.
+      scriptPath: $(Build.SourcesDirectory)/Build/setup_glass.py
+      arguments: '--authTokenVariable SYSTEM_ACCESSTOKEN --buildOutput $(Build.BinariesDirectory)\raw\binaries'
```
rchiodo, line 32: This is going to copy the build output from PTVS into this folder too? Maybe setup-glass should be modified to be able to skip that step?

rchiodo: To be fair, this should work regardless because the run_glass.py file also recopies the build output

PR #8109, opened 2024-12-10
Archive symbols from etwtrace package

AdamYoblick:
After bundling the etwtrace package into PTVS, we are getting symbol check failures when trying to insert into VS. These failures can be seen here:

https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/597367

If you click on the failing check and dig down into the failing step at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=10681983&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=f4062fa4-3467-5042-7419-c88746a3e56c, you can see that all the missing symbols are for `*.pyd` files that are part of the etwtrace package.

So these files need to be copied to the same directory as the other symbol files before symbol archiving is performed.

Symbol archiving is not performed for PR builds, so I had to run a full build to test this change. TODO: Add links to pipeline used for testing

### DIFF
--- a/Build/templates/publish_symbols.yml
+++ b/Build/templates/publish_symbols.yml
@@ -31,6 +31,7 @@ steps:
       Contents: |
         **/*.pyd
       TargetFolder: '$(Build.ArtifactStagingDirectory)/symbols'
+      flattenFolders: true
 
   # Archive Symbols
   # When you insert your binaries into VS or ship them to any other client you need to archive your binaries & symbols to MSDL. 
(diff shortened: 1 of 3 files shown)


StellaHuang95 (approved):

rchiodo:
@bschnurr , isn't the cache glass only supposed to run as part of the scheduled run?

rchiodo (approved):

AdamYoblick:
I'm going to close this PR. Not only does the microbuild archive symbols task seem to ignore pyd files, but Steve is working on getting these symbols uploaded to symweb, which means we won't have to do anything like this. I've asked for a temporary exception from the VS team so we can unblock insertions into vs main, we'll see what they say.

PR #8129, opened 2025-01-07
Fix interactive repl hang (#8128)

bschnurr:
* handle new "Empty" exception

* change IsSupathOf to handle null or empty

* update pylance


rchiodo (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "ptvs",
-  "private": true,
-  "devDependencies": {
-    "@pylance/pylance": "2024.11.3"
-  }
+    "name":  "ptvs",
```
rchiodo, line 2: formatting seems off? 

Also shouldn't there be an associated package-lock.json change?

AdamYoblick: We don't use package-lock.json in PTVS. This file gets updated automatically whenever you run PreBuild.ps1 and a new version of pylance is downloaded.

Issue #8090, opened 2024-12-03
Codeql alert: Uncontrolled command line


AdamYoblick:
Details are at https://vnext.s360.msftcloudes.com/blades/security?cid=61d093dc-bbbd-4126-9478-56eeb6938af6&blade=KPI:a0f0ce42-3063-5d3b-3b47-1ff3143abdc9~SLA:3~AssignedTo:All~Forums:All~waves:All~KPI%20Ranking:All~Tab:Summary~_loc:Security&peopleBasedNodes=luabud_team;judej_team;cowebster_team&global=4:a6368cb1-ce51-437c-99ac-aa6d899cc0b7

AdamYoblick:
I requested an exception for this because this is a helper class/method we use to run generic command lines.

AdamYoblick:
I added a comment into the code that should suppress this warning.

PR #8134, opened 2025-01-08
Add codeql suppression

AdamYoblick:
Fixes https://github.com/microsoft/PTVS/issues/8090

CodeQL doesn't allow you to suppress an warning on an entire file or class, it's per line only.

So even though line 251 already has a suppression, the violation at https://liquid.microsoft.com/codeql/issues/38ae4090-21dd-4617-94cf-6f5f77966286 is now occurring on line 256, so we need another suppression.

That's what this change is.

### DIFF
--- a/Build/17.0/packages.config
+++ b/Build/17.0/packages.config
@@ -85,5 +85,4 @@
   <package id="Microsoft.VisualStudio.RpcContracts" version="17.0.50-preview-0002-0005" />
   <package id="Microsoft.DiaSymReader.Pdb2Pdb" version="1.1.0-beta2-21181-01" />
   <package id="Microsoft.TestPlatform.ObjectModel" version="17.4.0-preview-20221003-03" />
-  <package id="Microsoft.DiagnosticsHub.SDK" version="17.13.35618.1" />
 </packages>
--- a/Build/17.0/packages.config
+++ b/Build/17.0/packages.config
@@ -85,4 +85,5 @@
   <package id="Microsoft.VisualStudio.RpcContracts" version="17.0.50-preview-0002-0005" />
   <package id="Microsoft.DiaSymReader.Pdb2Pdb" version="1.1.0-beta2-21181-01" />
   <package id="Microsoft.TestPlatform.ObjectModel" version="17.4.0-preview-20221003-03" />
+  <package id="Microsoft.DiagnosticsHub.SDK" version="17.13.35618.1" />
 </packages>
--- a/Build/nuget.config
+++ b/Build/nuget.config
@@ -6,6 +6,7 @@
     
     <!-- This is the public feed that we should use for both npm and nuget -->
     <add key="msft_consumption" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption/nuget/v3/index.json" />
+    <add key="VS" value="https://devdiv.pkgs.visualstudio.com/_packaging/VS/nuget/v3/index.json" />
   </packageSources>
   <trustedSigners>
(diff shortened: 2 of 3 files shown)


Build/17.0/packages.config
```diff
@@ -85,4 +85,5 @@
   <package id="Microsoft.VisualStudio.RpcContracts" version="17.0.50-preview-0002-0005" />
   <package id="Microsoft.DiaSymReader.Pdb2Pdb" version="1.1.0-beta2-21181-01" />
   <package id="Microsoft.TestPlatform.ObjectModel" version="17.4.0-preview-20221003-03" />
+  <package id="Microsoft.DiagnosticsHub.SDK" version="17.13.35618.1" />
```
rchiodo, line 88: Is this unrelated?

Python/Product/Common/Infrastructure/ProcessOutput.cs
```diff
@@ -253,7 +253,7 @@ public static ProcessOutput Run(
                 psi.Arguments = string.Join(" ",
                     arguments.Where(a => a != null).Select(QuoteSingleArgument));
             } else {
-                psi.Arguments = string.Join(" ", arguments.Where(a => a != null));
+                psi.Arguments = string.Join(" ", arguments.Where(a => a != null)); // CodeQL [SM00406] Code ql is complaining, but this is a utility class that is called with many different file names.
```
bschnurr, line 256: complaining that we dont verify arguments?

Build/17.0/packages.config
```diff
@@ -85,4 +85,5 @@
...
 </packages>
```
StellaHuang95, line 89: Should this be removed from this PR?

Build/17.0/packages.config
AdamYoblick: Shoot yes, this made it up by mistake, alone with the nuget.config. I'll take them out.

Build/nuget.config
```diff
@@ -6,6 +6,7 @@
     
     <!-- This is the public feed that we should use for both npm and nuget -->
     <add key="msft_consumption" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption/nuget/v3/index.json" />
+    <add key="VS" value="https://devdiv.pkgs.visualstudio.com/_packaging/VS/nuget/v3/index.json" />
```
StellaHuang95, line 9: Same here

Build/17.0/packages.config
AdamYoblick: Yep, removing.

Build/nuget.config
AdamYoblick: Removing :)

StellaHuang95 (approved):

Python/Product/Common/Infrastructure/ProcessOutput.cs
AdamYoblick: Yeah so the actual violation is https://liquid.microsoft.com/Web/Object/Read/ScanningToolWarnings/Requirements/CodeQL.SM00406#Zguide. The problem is that we can't possibly validate arguments here, it's a generic helper class to run a command line, we just join the arguments and run. Just red tape.

bschnurr: so alternatively we look at all the usage points to find a set of strings.. but there will still be some dynamic ones. maybe arg templates safer.

bschnurr (approved):

rchiodo (approved):

PR #8140, opened 2025-01-16
Add azdo api script

AdamYoblick:
Added an azdo api script to get build information for a specified build number


rchiodo (approved):

StellaHuang95:
What do you need the build information for?

StellaHuang95 (approved):

AdamYoblick:
> What do you need the build information for?

Needed to find the commit that was in a VS release branch so we can snap our own release branch for a hotfix. The snap has to be from the same commit to ensure we only add the hotfix. The REST API can find the commit quickly without having to perform a manual search.

PR #8142, opened 2025-01-17
Add release yaml to release branch

AdamYoblick:
This file isn't present in this branch because this branch was snapped from 098fa1a0cd1df68c2a16f0a3a6487ed03032dfe5, which is an older commit in main. The release yaml didn't exist at this point in time. So I have to manually bring it over.


AdamYoblick:
The bug doesn't repro in VS 17.8, so this PR is not needed.

PR #8150, opened 2025-01-27
Add PythonProfilderCommandService to support new profiler

StellaHuang95:
**Implementation:**

The `IPythonProfilerCommandService` interface is where the core functionality is. The VS team can use MEF to import this type into their codebase and call `GetCommandArgsFromUserInput` to retrieve a `PythonProfilingCommandArgs`. This object includes properties such as `PythonExePath`, `WorkingDir`, `ScriptPath`, `Args`, and `EnvVars`, basically all the arguments required by the Diagnostics Hub team to run the command that invokes the `etwtrace` package.

When the `GetCommandArgsFromUserInput` in `PythonProfilerCommandService` is called, first, `UserInputDialog.ShowDialog()` displays a user input dialog, allowing users to specify script arguments. Then, `CommandArgumentBuilder.BuildCommandArgsFromTarget()` is called to construct the command arguments from user input and returns the result to the service caller.

**Testing:**

Current testing is performed manually by calling the service from the existing profiling dialog. A breakpoint is set in `PythonProfilingPackages.cs` right after the user opens the Python profiler window to verify that the expected object is returned.
Unit tests will be added later, but for now, the focus is on rapid iteration and validation.

For example,

Here's user input in the dialog
image

Here's the value the service returns
image

### DIFF
--- a/Python/Product/Profiling/Profiling/CommandBuilder.cs
+++ b/Python/Product/Profiling/Profiling/CommandBuilder.cs
@@ -31,11 +31,11 @@ namespace Microsoft.PythonTools.Profiling {
         /// Constructs a <see cref="TargetCommand"/> based on the provided profiling target.
         /// </summary>
         public TargetCommand BuildCommandFromTarget(ProfilingTarget target) {
-            try {
-                if (target == null) {
-                    return null;
-                }
+            if (target == null) {
+                return null;
+            }
 
+            try {
                 var pythonProfilingPackage = PythonProfilingPackage.Instance;
                 var joinableTaskFactory = pythonProfilingPackage.JoinableTaskFactory;
 
@@ -122,7 +122,7 @@ namespace Microsoft.PythonTools.Profiling {
                 PythonExePath = pythonExePath,
                 ScriptPath = scriptPath,
                 WorkingDir = workingDir,
-                Args = null,
+                Args = Array.Empty<string>(),
                 EnvVars = envVars
             };
             return command;
@@ -154,13 +154,18 @@ namespace Microsoft.PythonTools.Profiling {
             config.ScriptArguments = standaloneTarget.Arguments;
             config.WorkingDirectory = standaloneTarget.WorkingDirectory;
 
+            var argsInput = standaloneTarget.Arguments;
+            var parsedArgs = string.IsNullOrWhiteSpace(argsInput)
+                        ? Array.Empty<string>()
+                        : argsInput.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
+
             var envVars = session._serviceProvider.GetPythonToolsService().GetFullEnvironment(config);
 
             return new TargetCommand {
                 PythonExePath = config.GetInterpreterPath(),
                 WorkingDir = standaloneTarget.WorkingDirectory,
                 ScriptPath = standaloneTarget.Script,
-                Args = standaloneTarget.Arguments,
+                Args = parsedArgs,
                 EnvVars = envVars
             };
         }
--- a/Python/Product/Profiling/Profiling/CommandBuilder.cs
+++ b/Python/Product/Profiling/Profiling/CommandBuilder.cs
@@ -1,0 +1,170 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+
+
+namespace Microsoft.PythonTools.Profiling {
+    using System;
+    using System.Diagnostics;
+    using System.IO;
+    using System.Linq;
+    using System.Windows;
+    using Microsoft.PythonTools.Infrastructure;
+    using Microsoft.PythonTools.Interpreter;
+
+    internal class CommandBuilder {
+
+        /// <summary>
+        /// Constructs a <see cref="TargetCommand"/> based on the provided profiling target.
+        /// </summary>
+        public TargetCommand BuildCommandFromTarget(ProfilingTarget target) {
+            try {
+                if (target == null) {
+                    return null;
+                }
+
+                var pythonProfilingPackage = PythonProfilingPackage.Instance;
+                var joinableTaskFactory = pythonProfilingPackage.JoinableTaskFactory;
+
+                TargetCommand command = null;
+
+                joinableTaskFactory.Run(async () => {
+                    await joinableTaskFactory.SwitchToMainThreadAsync();
+
+                    var name = target.GetProfilingName(pythonProfilingPackage, out var save);
+                    var explorer = await pythonProfilingPackage.ShowPerformanceExplorerAsync();
+                    var session = explorer.Sessions.AddTarget(target, name, save);
+
+                    command = SelectCommandBuilder(target, session);
+
+                });
+
+                return command;
+            } catch (Exception ex) {
+                Debug.Fail($"Error building command: {ex.Message}");
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// Generates a <see cref="TargetCommand"/> based on the profiling target and session details.
+        /// </summary>
+        private TargetCommand SelectCommandBuilder(ProfilingTarget target, SessionNode session) {
+            var projectTarget = target.ProjectTarget;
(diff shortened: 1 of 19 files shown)


StellaHuang95:
@rchiodo Could you also take a look when you get a chance? Just to get a second opinion on it before passing it to the VS team.

Python/Product/Profiling/Profiling/CommandBuilder.cs
```diff
@@ -0,0 +1,170 @@
...
+
+                var pythonProfilingPackage = PythonProfilingPackage.Instance;
+                var joinableTaskFactory = pythonProfilingPackage.JoinableTaskFactory;
+
+                TargetCommand command = null;
+
+                joinableTaskFactory.Run(async () => {
+                    await joinableTaskFactory.SwitchToMainThreadAsync();
+
+                    var name = target.GetProfilingName(pythonProfilingPackage, out var save);
```
rchiodo, line 47: Why does this need to be on the main thread?

Python/Product/Profiling/Profiling/IUserInputService.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Microsoft.PythonTools.Profiling {
+
+    /// <summary>
+    /// Represents a command generated for a profiling target.
+    /// </summary>
+    public class TargetCommand {
+        public string PythonExePath { get; set; }
+        public string WorkingDir { get; set; }
+        public string ScriptPath { get; set; }
+        public string Args { get; set; }
```
rchiodo, line 28: Might this be better as an array?

Python/Product/Profiling/Profiling/IUserInputService.cs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /// <summary>
+    /// Defines a service interface for collecting user input to construct a profiling target command.
+    /// </summary>
+    public interface IUserInputService {
+        /// <summary>
+        /// Collects user input via a dialog and converts it into a <see cref="TargetCommand"/>.
+        /// </summary>
+        TargetCommand GetCommandFromUserInput();
```
AdamYoblick, line 39: Typically, types in an interface should be basic types or other interfaces. Since you are returning a concrete type here, the caller has to know about that type, but they can't know about it unless they compile against this file, which they won't have.

So `GetCommandFromUserInput()` should return an `ITargetCommand`, and your `TargetCommand` class can implement it. Then you can MEF export the ITargetCommand just like you do with the IUserInputService.

AdamYoblick: @rchiodo Would you agree with this? This is how I've seen it done in the past, just wanted to make sure I wasn't telling Stella the wrong thing.

Python/Product/Profiling/Profiling/UserInputService.cs
```diff
@@ -0,0 +1,64 @@
...
+    using System.IO;
+    using System.Linq;
+    using System.Windows;
+    using Microsoft.PythonTools.Infrastructure;
+    using Microsoft.PythonTools.Interpreter;
+
+    /// <summary>
+    /// Implements a service to collect user input for profiling and generate a <see cref="TargetCommand"/>.
+    /// </summary>
+    [Export(typeof(IUserInputService))]
```
rchiodo, line 30: The name of this seems weird to me. It feels like calling it IUserInputService is a bit of a misnomer. Like it should mention profiling somehow. It's only for profiling.

`IPythonProfilingCommandBuilder` maybe?

Python/Product/Profiling/Profiling/UserInputService.cs
```diff
@@ -0,0 +1,64 @@
...
+    class UserInputService : IUserInputService {
+        private readonly CommandBuilder _commandBuilder;
+        private readonly UserDialog _userDialog;
+
+        [ImportingConstructor]
```
AdamYoblick, line 35: I was under the impression that you only had to use `ImportingConstructor` when your constructor is importing MEF types that are exported by someone else. I could be wrong here though, I'm not a MEF expert.

Python/Product/Profiling/Profiling/IUserInputService.cs
rchiodo: Yeah good catch. I agree. An interface or a [struct](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct).

Python/Product/Profiling/Profiling/UserInputService.cs
```diff
@@ -0,0 +1,64 @@
...
+        public TargetCommand GetCommandFromUserInput() {
+            try {
+                var pythonProfilingPackage = PythonProfilingPackage.Instance;
+                var targetView = new ProfilingTargetView(pythonProfilingPackage);
+
+                if (_userDialog.ShowDialog(targetView)) {
+                    var target = targetView.GetTarget();
+                    return _commandBuilder.BuildCommandFromTarget(target);
+                }
+                return null;
```
AdamYoblick, line 56: Since both the try and the catch return null as the terminal statement, you can move the `return null` statement to outside of the try/catch.

Python/Product/Profiling/Profiling/CommandBuilder.cs
```diff
@@ -0,0 +1,170 @@
...
+    using Microsoft.PythonTools.Interpreter;
+
+    internal class CommandBuilder {
+
+        /// <summary>
+        /// Constructs a <see cref="TargetCommand"/> based on the provided profiling target.
+        /// </summary>
+        public TargetCommand BuildCommandFromTarget(ProfilingTarget target) {
+            try {
+                if (target == null) {
```
AdamYoblick, line 35: Argument validation is usually done outside of a try/catch. You can move this check up. :)

Python/Product/Profiling/Profiling/CommandBuilder.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (projectTarget != null) {
+                return BuildProjectCommand(projectTarget, session);
+            } else if (standaloneTarget != null) {
+                return BuildStandaloneCommand(standaloneTarget, session);
+            }
+            return null;
+        }
+
+        private TargetCommand BuildProjectCommand(ProjectTarget projectTarget, SessionNode session) {
+            var solution = PythonProfilingPackage.Instance.Solution;
```
AdamYoblick, line 78: Does this work if there is no solution? Like if the user is trying to profile files in folder view instead of project view?

AdamYoblick (approved):

Python/Product/Profiling/Profiling/UserInputService.cs
StellaHuang95: But there's an actual `CommandBuilder` that converts user input to command arguments. How about I call the main service `PythonProfilingCommandArgumentProvider`, which calls `ArgumentInputDialog` to pop dialog to user and return user input, and calls `ArgumentParser` to convert user input to command arguments?

rchiodo: The important part for me is the name indicates:
- What it's used for = in this case profiling
- What it does/returns = some sort of command? Or is it a target profile?

Python/Product/Profiling/Profiling/UserInputService.cs
StellaHuang95: https://learn.microsoft.com/en-us/dotnet/framework/mef/attributed-programming-model-overview-mef#:~:text=To%20fill%20the,declare%20an%20import.

Yeah I think you're right, it's used when there're parameters in the class's constructor that need to be imported(inject dependencies through the constructor), since mine doesn't have any parameters, it should be removed. I will run some tests without this attribute to make sure it still works.

Python/Product/Profiling/Profiling/CommandBuilder.cs
StellaHuang95: The `ShowPerformanceExplorerAsync()` shows a UI element, and from what I remember, any UI related operations need to happen on the main thread(UI thread).

rchiodo: Ah that makes sense. Thanks.

rchiodo (approved):

StellaHuang95:
I’ve finished addressing all the feedback, updated the PR description to reflect the new file and interface naming, and did all the tests again to ensure that the expected values are returned. I’ll let the insertion PR run first since it takes about a day to complete, but please take another look whenever you have a chance tomorrow @AdamYoblick before I merge this into the VS main branch.

The service call works as intended within PTVS. By MEF exporting the service interface and the returned interface, the Diagnostics Hub team should be able to compile without issues using MEF.

When passing this to the Diagnostics Hub team, I will share the following details, but let me know if there’s anything else you want to add. Thanks!

1. The new service will be inserted into the VS main branch, as it does not impact the current user profiling workflow. However, feel free to let us know if you’d prefer it to remain in a topic branch.

2.  Please use the following command to invoke the etwtrace package: `<path to python.exe> -m etwtrace --diaghub -- <path to script.py> <args>`. Reference to `etwtrace` doc is at [etwtrace PyPI page](https://pypi.org/project/etwtrace/). This provides some command-line examples. Steve should also be able to provide more details if needed.

3. The `etwtrace` package is bundled in the latest IntPreview version of Visual Studio with PTVS, which can be found at
`%LocalAppData%\Microsoft\VisualStudio\<Version>\Extensions\Microsoft Corporation\Python - Profiling\<Version>\etwtrace`. Previously, to invoke the `etwtrace` package in PTVS, I had to append the `etwtrace` module path to the existing `PYTHONPATH` in the `environment variables`.

4. To get the command line arguments, use MEF to call `IPythonProfilerCommandService.GetCommandArgsFromUserInput()`. This method should returns an `IPythonProfilingCommandArgs` object that contains all the needed arguments.

PR #8163, opened 2025-02-05
Add a guard on path

bschnurr:
fix for workitem 2345868


rchiodo (approved):
This fixes the crash, but could the interpreter path being null cause a crash somewhere else?

The root cause is the interpreter path being null here:

https://github.com/microsoft/PTVS/blob/9e5040e6f346f796a0e6d2ad492e2f45dd0d299e/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs#L527-L528

Issue #8166, opened 2025-02-11
An error "Profiling support seems to be missing...." when launch python pofiling.


linette-zyy:
**Environment**
![Image](https://github.com/user-attachments/assets/58820599-ea37-4a5e-a768-b52cd91369d7)

**Steps to Reproduce**
1. Create a new Python application project and add "Lib\test\pythoninfo.py" and remove the default .py file
2. Debug> Launch Python Profiling …

**Expected behavior**
It launched successfully without any error.

**Additional context and screenshots**
![Image](https://github.com/user-attachments/assets/62bc080b-595a-4d9e-b6b5-3ba196623bf5)

And I try to repair, this error still pops up.

PR #8167, opened 2025-02-11
Update Diagnostics Hub Package Guid

StellaHuang95:
fixes https://github.com/microsoft/PTVS/issues/8166

The diagnostics hub team just deleted all that code since nothing actually uses `perfpkg` anymore, even Python experience goes through the new profiler, so changing the GUID to be the Diagnostics Hub Package (new profiler)'s GUID.

Verified the existing profiling experience works now.

profilerGUID


StellaHuang95:
Insertion PR is at https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/610219

Issue #8168, opened 2025-02-11
Address Comments From Diagnostics Hub Team


StellaHuang95:
Comments from Nik:

Ok, I sorta have this consumed in a local branch though I need a few changes from your side:

- [x] We need to let VS know `Microsoft.PythonTools.Profiling.dll` exposes a MEF component. Can you add the following line to your `extension.vsix`:
`<Asset Type="Microsoft.VisualStudio.MefComponent" Path="Microsoft.PythonTools.Profiling.dll" />`

- [x] VS dependency probing needs to know where the profiling DLL exists, otherwise the profiler code throws an exception JITing the method that uses the types exposed by the profiling DLL. I think we need to add the following line to `devenv.exe.config` (solved the issue for me), though I'm not sure how extensions do this:
```
<dependentAssembly>
<assemblyIdentity version="17.0.0.0" name="Microsoft.PythonTools.Profiling" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
<codeBase version="17.0.0.0" href="Extensions\Microsoft\Python\Profiling\Microsoft.PythonTools.Profiling.dll" />
</dependentAssembly>
```

- [x] When I call `GetCommandArgsFromUserInput` I'm getting a nullref exception, specifically `PythonProfilingPackage.Instance` is null. I'm not sure where in your package this is set, though maybe we could just change `GetcommandArgsFromUserInput` to be async and then you can query the VS shell for the package instead of using the singleton.

- [x] Could you create a NuGet package that has `Microsoft.PythonTools.Profiling.dll` so we can reference it from our build? Right now I just linked to it from my VS install which won't work in the actual product. (Seems like internal should be fine, and should only need to be a reference assembly (which reduces the update frequency, probably to "once"): https://learn.microsoft.com/en-us/dotnet/standard/assembly/reference-assemblies)

AdamYoblick:
> We need to let VS know Microsoft.PythonTools.Profiling.dll exposes a MEF component. Can you add the following line to your extension.vsix:
`<Asset Type="Microsoft.VisualStudio.MefComponent" Path="Microsoft.PythonTools.Profiling.dll" />`

This work is done with 3ba91af22

I tested it by building locally and I checked the generated file at `.\BuildOutput\Debug17.0\layout\Microsoft.PythonTools.Profiling\extension.vsixmanifest`
The line that was added exactly matches what was requuested.

StellaHuang95:
[...]

PR #8169, opened 2025-02-12
Load PythonProfilingPackage from shell

StellaHuang95:
Addresses the 3rd issue in https://github.com/microsoft/PTVS/issues/8168.

This issue is that when Diagnostics Hub calls `GetCommandArgsFromUserInput` they 're getting a `nullref` exception, specifically `PythonProfilingPackage.Instance` is null. 

This PR changes `GetcommandArgsFromUserInput` to be async and query the VS shell for the package.

### DIFF
--- a/Python/Product/Profiling/Profiling/PythonProfilerCommandService.cs
+++ b/Python/Product/Profiling/Profiling/PythonProfilerCommandService.cs
@@ -18,7 +18,11 @@ namespace Microsoft.PythonTools.Profiling {
     using System;
     using System.ComponentModel.Composition;
     using System.Diagnostics;
+    using System.Threading.Tasks;
     using System.Windows;
+    using Microsoft.VisualStudio.Shell;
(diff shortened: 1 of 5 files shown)


Python/Product/Profiling/Profiling/PythonProfilerCommandService.cs
```diff
@@ -55,5 +63,22 @@ public IPythonProfilingCommandArgs GetCommandArgsFromUserInput() {
 
             return null;
         }
+
+        private async Task<PythonProfilingPackage> GetPythonProfilingPackageAsync() {
```
StellaHuang95, line 67: @AdamYoblick @rchiodo Can both of you give a review? I verified that within PTVS, the package loads correctly and the service returns the expected arguments, but wasn't able to test calling it from outside of PTVS.

rchiodo (approved):

AdamYoblick (approved):

Issue #8121, opened 2025-01-01
Features not working consistently: task list, navigation dropdown menus


AbhimanyuVS:
**Describe the bug**

This issue is related to issues #7151, #7254, and #8114.

When programming in Python with Visual Studio 2022 Community - both standard release and preview release - certain features do not work correctly or consistently. In particular, the task list and the navigation dropdown menus.

The task list was the main issue. For the past 15 years, I have found the Visual Studio task list to be an invaluable tool. However, as I reported in issue #7254 (which was then merged into issue #7151), the task list has not been working properly in VS2022.

In issue #8114, @StellaHuang95 suggested I download and use the preview release of VS2022, so I did that, and the task list started working again... sometimes. Opening VS2022Preview will sometimes load the task list, and other times will simply not.

The same is true for the navigation dropdown menus. Sometimes these load correctly, and other times they simply do not. In fact, this varies between open projects/files.

**Steps to Reproduce**

1. Open VS2022Preview.
2. Check the task list and the navigation dropdowns.

**Expected behavior**

The task list and navigation dropdown menus should always load correctly and consistently.

**Additional context and screenshots**

Screen recording demonstrating this issue (link should remain active until 1 Jan 2026):
https://1drv.ms/v/s!Aux-sta5GHQDi8pXHUNiORv5yMFr5Q?e=ozo4bH

**Configuration information (If you are providing a diagnostics file (see below), skip this section)**

VS Version: 17.13.0 Preview 2.1
PTVS version: 17.0.24323.2
Python version: 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
Debugger Type (if applicable): N/A

**Diagnostics file**

[PythonToolsDiagnostics_20250101031732.log](https://github.com/user-attachments/files/18285284/PythonToolsDiagnostics_20250101031732.log)

AbhimanyuVS:
Hi, this has been pending for a bit. Is there anything I can do to help out with this?

AbhimanyuVS:
@debonte, is there anything I can do to help with this issue?

debonte:
Next time you see this problem occur, can you grab the contents of the Pylance log in the Output pane? See example below, where the log output is quite short because I only just opened the current solution.

![Image](https://github.com/user-attachments/assets/97670050-d09c-4e9f-8633-d608c6492737)

AbhimanyuVS:
@debonte,

[...]

PR #8172, opened 2025-02-14
Fix for missing default settings at startup.

bschnurr:
guard against no context elements on shutdown.  A context is created for the project or for each workspace or for each repl.

We need to send tasklist info when sending default settings with "workspace/didChangeConfiguration".  

alternatively we could pass in the uri into GetSettings() when "TriggerWorkspaceUpdateConfig" is called.

Fix for https://github.com/microsoft/PTVS/issues/8121

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
@@ -367,12 +367,17 @@ namespace Microsoft.PythonTools.LanguageServerClient {
         }
 
         private LanguageServerSettings.PythonSettings GetSettings(Uri scopeUri = null) {
+            // on shutdown return null
+            if (_clientContexts.Count() == 0) {
+                return null;
+            }
             IPythonLanguageClientContext context = null;
             if (scopeUri == null) {
-                // REPL context has null RootPath
+                // REPL context has null RootPath 
                 context = _clientContexts.Find(c => c.RootPath == null);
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -367,12 +367,17 @@ private async Task NotifyWithParametersAsync(string request, object parameters)
...
+            }
             IPythonLanguageClientContext context = null;
             if (scopeUri == null) {
-                // REPL context has null RootPath
+                // REPL context has null RootPath 
                 context = _clientContexts.Find(c => c.RootPath == null);
                 if (context == null) {
-                    return null;
+                    // use first clientcontext as default
+                    context = _clientContexts.First();
```
rchiodo, line 380: Is there a FirstOrNull thing? I wasn't sure if this would throw if there's no items in the list.

rchiodo: Oh it's called 'FirstOrDefault'. Maybe you should use that?

bschnurr: i check count at the top of the function. but ya. that would be safer

debonte (approved):

rchiodo (dismissed):

StellaHuang95 (approved):

rchiodo (approved):

Issue #8201, opened 2025-03-20
Make PTVS build for Dev18


StellaHuang95:
We need to do something similar to https://github.com/microsoft/PTVS/pull/6505.

This involves lots of MSBuild projects and configuration file changes.

rizi:
@StellaHuang95 wouldn't it make sense that the tasks just use the latest installed version of VS/MsBuild?
At the moment the tasks needs to be updated to support new version of VS.
Also within the tasks "use latest version" is quite misleading, because it means the latest version of VS that is supported by the task and not the latest installed version

PR #6505, opened 2021-05-11
Get PTVS building for D17

rchiodo:
This makes PTVS build (and run) with D17.

I'm going to do the setup authoring work separately after I get this to go through.

PR #8202, opened 2025-03-20
Fix issue with creating python project in Dev18

StellaHuang95:
First try to fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2416924

This is just temporary workaround to make the build work for Dev18 before signoff date.

I created another issue to track the long term fix to build PTVS for Dev18: https://github.com/microsoft/PTVS/issues/8201


Python/Setup/swix/Packages/buildcore.swr
```diff
@@ -1,5 +1,15 @@
 
-folder "InstallDir:\MSBuild\Microsoft\VisualStudio\v$(VSTarget)\Python Tools"
+folder "InstallDir:\MSBuild\Microsoft\VisualStudio\v17.0\Python Tools"
```
AdamYoblick, line 2: We should avoid hard coding the paths as much as we can. That's the whole point of the VSTarget variable which is passed into the build command line. 

Can you please try building with this set to 18.0 on the command line first?

StellaHuang95: This value is currently hardcoded in the build pipeline variable, and it's compared against `v$(VisualStudioVersion)` at VS runtime to locate the template path. I chose to hardcode it for now because:

1. Setting the `VSTarget` to `18.0` on command line would cause build failures due to other compilation errors. The work needed to fix PTVS to build properly for dev18 requires huge amount of work. See the work needed. https://github.com/microsoft/PTVS/pull/6505
2. Signoff deadline is tomorrow so I want to make the core features work first. The variable issue can be addressed later, as we have more time before Dev18 ships.
3. VS is using the same source to build dev17 and dev18, changing the value to one of them will break another VS version. For now, the best approach is to place the templates in both folders.

StellaHuang95: By the way, the build is broken, complaining about the nuget pkg folder is not found, do you have any idea what's going on?

StellaHuang95: Oh I think I know what happened, the build and publish nuget package step is part of `Non-PR steps` so it didn't run, but publish artifacts step still think it runs, this should be fixed later.

rchiodo (approved):

PR #8273, opened 2025-08-13
Fix repl hang is python 3.12 and above. the module 'imp' was removed.

bschnurr:
- also updating pylance and debugpy

How i figured it out
forced 'DEBUG' to be true. it opened another terminal window with some extra prints and callstack.
```DEBUG = os.environ.get('DEBUG_REPL') or os.environ.get('_PTVS_DEBUG_REPL')```
fixes internal issue 2527900

### CODE REVIEW

**Comment by bschnurr** [comment]
Oh. i think i should update our tests to run both latest and oldest supported python versions.

File: Python/Product/PythonTools/ptvsd/repl/__init__.py [hunk 1]
```diff
@@ -39,12 +39,19 @@
 import select
 import time
 import struct
-import imp
+import types
+# Add compatibility code:
```
rchiodo, line 43: Maybe this could be a function?

Issue #8277, opened 2025-08-25
Test Result didn't show after executing in TE


Wycc20:
**Environment**

<img width="1838" height="902" alt="Image" src="https://github.com/user-attachments/assets/e24da78f-f719-49bb-9106-01ff3194796f" />

**Steps to Reproduce**
1. Add Unit Test file in a python application.
2. Right click project -> Properties -> Test-> Change test framework to Unittest.
3. Open TE, the test case will be shown. 
4. Click on the "Run All Test In Views".

**Expected behavior**
All test cases will display icons indicating failure or success.

<img width="862" height="347" alt="Image" src="https://github.com/user-attachments/assets/82ebe4ae-efb9-40e4-81c5-d788119984fc" />

**Actual behavior**
All test cases show pre-executed stage.

<img width="1784" height="795" alt="Image" src="https://github.com/user-attachments/assets/4492539e-439d-4f8b-a447-0aae2545e3ac" />

**Additional context and screenshots**

https://github.com/user-attachments/assets/83f8826b-48e6-4470-a466-8ed7a13d387b

StellaHuang95:
@Wycc20 Could you share the logs from Show output -> Tests?

Wycc20:
@StellaHuang95 Please find the log file attached for your review.

[Output-Tests.txt](https://github.com/user-attachments/files/22372970/Output-Tests.txt)

PR #8285, opened 2025-09-18
Fix issue with test explorer

StellaHuang95:
Fixes https://github.com/microsoft/PTVS/issues/8277

The error message:
System.InvalidOperationException: Could not find VS DTE object for process 6620
   at Microsoft.PythonTools.Infrastructure.VisualStudioProxy.GetDTE()
   at Microsoft.PythonTools.TestAdapter.Services.ExecutorService.DetachFromSillyManagedProcess(VisualStudioProxy app, PythonDebugMode debugMode)
   at Microsoft.PythonTools.TestAdapter.UnittestTestExecutor.TestRunner.Run()
Connected to socket

is caused by build pipeline now only supports building ptvs in dev17 environment, but the process is running in dev18. Fixing it by supporting both versions, but we should revert back once pipeline tool supports dev18.


StellaHuang95:
Verified it's working. 
image

Issue #8290, opened 2025-09-22
The option "Python/Native Debugging" is missing.


linette-zyy:
**Environment**
<img width="736" height="549" alt="Image" src="https://github.com/user-attachments/assets/5d68e98e-e542-40a4-8fd9-faafb3a2956f" />

**Steps to Reproduce**

1. Create new project: Visual C++ > Windows Desktop > Windows Console Application
2. Open the property page
3. Click Debugging tab
4. Click to open the drop-down list
5. Observes

**Expected behavior**
The option "Python/Native Debugging" should exist in the list.
![image](https://user-images.githubusercontent.com/59192491/124555522-c43a0080-de69-11eb-977c-f4ab588d835c.png)

**Actual result**
The option "Python/Native Debugging" is missing.
<img width="825" height="561" alt="Image" src="https://github.com/user-attachments/assets/42fbc21e-7dc5-4619-bcf6-292964ec8c05" />

linette-zyy:
@StellaHuang95 This issue still occurred on the latest build.
<img width="739" height="378" alt="Image" src="https://github.com/user-attachments/assets/a2f7ec58-e471-477e-912e-adef3da4ac7e" />

<img width="945" height="618" alt="Image" src="https://github.com/user-attachments/assets/0de9189d-0181-46ef-9757-811eb0d998a0" />

PR #8293, opened 2025-09-22
Fix the issue where the option "Python/Native Debugging" is missing

StellaHuang95:
fixes https://github.com/microsoft/PTVS/issues/8290.

It's caused by the same reason: PTVS is built in dev17 environment but run in dev18 environment.  Before msbuild@1 supports dev18 build environment, hardcode the path for now. See https://github.com/microsoft/PTVS/issues/8201#issue-2936315398


StellaHuang95:
Tested and verified it works: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=12435493&view=artifacts&pathAsName=false&type=publishedArtifacts

rchiodo (approved):
One question, did you create a branch/tag for Dev17 before you started all this Dev18 work?

StellaHuang95:
> One question, did you create a branch/tag for Dev17 before you started all this Dev18 work?

yes I did, just need Graham or someone who has permission to add branch protection rules.

PR #8299, opened 2025-10-02
[Not checking in] Add python314 support

bschnurr:
Here’s a combined, summarized version of **Python 3.14 Debugging Support in PTVS** with the roadmap and current status integrated for clarity (updated after enabling managed safe attach by default, and after adding interpreter-walk inference + timing instrumentation):

---

# Python 3.14 Debugging Support in PTVS

## Overview

PTVS provides full **Python 3.14 debugging support** across all modes:

* Managed-only (debugpy) – safe attach now DEFAULT (opt‑out env var)
* Mixed-mode / native Concord engine (safe attach implemented Phase 1)
* Remote & safe attach (PEP 768) groundwork
* Stepping, breakpoints, evaluation
* Version recognition and backward compatibility (3.5–3.13)

Key change: Leverage CPython 3.14 `_Py_DebugOffsets` (`debugger_support`) for **safe attach** instead of legacy DLL/code injection. Managed path now attempts safe attach first and only falls back to legacy injection if it fails or is explicitly disabled (`PTVS_SAFE_ATTACH_MANAGED_DISABLE=1`).

---

## Debugpy Dependency: Legacy vs Current (Safe Attach)
**Goal:** Clarify how the requirement for `debugpy` changed with the new `safe_attach_loader.py`.

| Aspect | Legacy Attach / Transitional (ptvsd_loader) | Current Safe Attach (safe_attach_loader) |
|--------|---------------------------------------------|------------------------------------------|
| Primary loader file | `ptvsd_loader.py` (sometimes temp bootstrap) | `safe_attach_loader.py` (single canonical) |
| Bundled package | `ptvsd` + (later) `debugpy` directory in VSIX | `debugpy` directory in VSIX (fallback to legacy `ptvsd`) |
| sys.path handling | Loader typically inserted its dir; dynamic scripts inconsistent | Loader always prepends its directory and explicitly adds adjacent `debugpy` dir |
| Environment needs `pip install debugpy` | Not required, but environment install could shadow bundled version | Not required; bundled copy always reachable and preferred |
| Fallback behavior | `import debugpy` then `import ptvsd as debugpy` | Same, but debugpy import nearly deterministic due to path injection |
| Multiple loader variants | Yes (legacy, dynamic connect, bootstrap) | No (single script) |
| Role confusion (listen vs connect) | Possible (dynamic connect + host connect -> timeouts) | Target always listens, host always connects |
| Initial breakpoint logic | Varied; different scripts | Unified: `PTVS_WAIT_FOR_CLIENT=1` and/or `PTVS_DEBUG_BREAK=1` triggers breakpoint |
| Missing loader failure mode | Silent timeout in many cases | Immediate telemetry: site=ScriptBufferWrite msg=loader file not found |
| Need for sidecar port file | Ad-hoc / absent in dynamic scripts | Supported via `PTVS_DEBUG_PORT_FILE` when dynamic port chosen |

### Current Environment Variables (Supported / Simplified)
| Variable | Meaning | Default |
|----------|---------|---------|
| `PTVS_DEBUG_HOST` | Interface to bind listener | `127.0.0.1` |
| `PTVS_DEBUG_PORT` | Port to listen on (`0` = allocate ephemeral) | `0` |
| `PTVS_WAIT_FOR_CLIENT` | Wait for debugger before resuming | `1` |
| `PTVS_DEBUG_BREAK` | Force initial breakpoint (OR with WAIT) | Off |
| `PTVS_DEBUG_PORT_FILE` | Path to write selected port number | None |
| `PTVS_DEBUG_SESSION` | Correlation / session GUID | None |
| `PTVS_SAFE_ATTACH_LOADER_VERBOSE` | Verbose loader logging (`0` disables) | `1` |
| `PTVS_DEBUG_LOG_FILE` | File sink for loader logs (else stderr) | None |
| (Orchestrator) `PTVS_SAFE_ATTACH_MANAGED_DISABLE` | Force legacy injection path | Off |
| (Orchestrator) `PTVS_MIXED_MODE` | Bypass managed safe attach in mixed mode | Off |

### Migration Guidance
1. Remove any docs telling users to manually install `debugpy` just for attach.
2. Refer only to `safe_attach_loader.py`; retire references to `ptvsd_loader.py` except in legacy troubleshooting appendix.
3. Document single listener role: process listens, VS connects.
4. For Break Immediately scenarios: set `PTVS_WAIT_FOR_CLIENT=1` (optionally also `PTVS_DEBUG_BREAK=1` – they both trigger a breakpoint now).
5. Troubleshooting timeouts: check loader log for `listener started` or failures (`listen failed`). No more silent dependency on environment debugpy.

### Failure Differentiation (Now)
| Symptom | Likely Cause | Action |
|---------|--------------|--------|
| `loader file not found` | VSIX deployment missing / not copied yet | Redeploy / ensure `<IncludeInVSIX>true</IncludeInVSIX>` for loader |
| Socket timeout (no connection) | Listener never started (port bind, path import failure) | Enable verbose loader logs; inspect `listen failed` lines |
| Fallback to `ptvsd` | Bundled debugpy corrupt / missing | Repair VSIX; verify debugpy directory present |
| No breakpoint on attach | WAIT & BREAK both unset | Set `PTVS_WAIT_FOR_CLIENT=1` or `PTVS_DEBUG_BREAK=1` |

---

## Recent Managed Safe Attach Failure Analysis (Oct 2025)
**Symptoms:**
* Offsets parse stops after Strict/Extended/Relaxed and reports "no valid layout" even though flexible fallback could succeed.
* Attach UI test (`AttachBreakImmediately`) remains in run mode (no initial break).
* Mixed-mode project launch stays in design mode (`dbgDesignMode`) instead of breaking.
* Orchestrator reports offsets parse failure on 3.14 with header bytes starting `78 64 65 62 75 67 70 79` (cookie OK) and no 512 size tuple near early qwords.

**Root Causes:**
1. Flexible tuple reordering/variable size support added to parser but gated behind `PTVS_SAFE_ATTACH_FLEX_FALLBACK`; env not set in VS host → fallback never executed.
2. Orchestrator still hard-rejects non‑canonical `script_path_size != 512` (even when parser could flag NonCanonical).
3. Loader preference not updated: continues choosing `ptvsd_loader.py` instead of new `safe_attach_loader.py` for listen/connect scenarios → no debugpy session.
4. Option B (dynamic `debugpy.connect`) implemented, but adapter layer not yet passing `PTVS_DEBUGPY_CLIENT_PORT` → connect script path unused.
5. Mixed-mode (Concord) attempts managed safe attach prematurely; native + managed coordination incomplete → session never starts.

**Mitigations Applied / Pending:**
* Parser: flexible fallback implemented & gated (DONE). Need to default-enable or auto-enable upon strict failure (PENDING).
* Loader: `safe_attach_loader.py` added with verbose + initial breakpoint option (DONE). Need orchestrator search order update (PENDING).
* Break Immediately: loader now triggers `debugpy.breakpoint()` when `WAIT` or `PTVS_DEBUG_BREAK=1` (DONE). Adapter must set proper env (PENDING).
* Option B dynamic connect script generation in orchestrator (DONE). Adapter must allocate a port & set `PTVS_DEBUGPY_CLIENT_PORT` (PENDING).
* Mixed-mode guard: should disable managed safe attach until Concord parity (PENDING).
* Non‑canonical size acceptance: orchestrator must accept validated sizes (e.g. 0x300/0x1234) within sane cap (PENDING).

**Action Items:**
1. Auto-enable flex fallback: if strict + extended + relaxed fail AND cookie matches, internally enable flexible scan (ignore env). Add telemetry flag `flexUsed`.
2. Orchestrator size validation: replace hard check `size == 512` with `(size >= 512 && size <= 0x100000)` when `NonCanonicalSize` flagged.
3. Loader selection order: try `safe_attach_loader.py` (connect vs listen mode) → `ptvsd_loader.py` → temp bootstrap.
4. Adapter integration (Option B):
   * Allocate free port prior to orchestrator.
   * Set env: `PTVS_DEBUGPY_CLIENT_PORT`, `PTVS_DEBUGPY_CLIENT_HOST=127.0.0.1`, `PTVS_DEBUG_BREAK=1` (when AttachBreakImmediately), `PTVS_SAFE_ATTACH_FORCE_CONNECT=1`.
   * Switch `_debugInfo` to `DebugTcpAttachInfo` pointing to chosen port.
5. Mixed-mode detection: if native Python/Concord engine requested (or env `PTVS_MIXED_MODE=1`), set `PTVS_SAFE_ATTACH_MANAGED_DISABLE=1` and skip Option B.
6. Telemetry extension: record phases: parseMode (strict|extended|relaxed|flex), sizeCanonical (bool), loaderType (legacy|listen|connect), breakInjected, mixedModeBypass.
7. Add unit tests:
   * Parser: feed synthetic slab with non‑512 size; ensure flex accepted.
   * Orchestrator: simulate parsed NonCanonical → verify acceptance & loader write.
   * Adapter mock: env var propagation leads to dynamic connect script path chosen.
8. Regression test: ensure negative test `OffsetsParser_Fails_On_WrongSize` remains valid (flex gated or uses env disable).
9. Concord parity (phase 1): port flexible parser + NonCanonical acceptance (read-only) before enabling writes.

**Short-Term Fallback Guidance:**
* For failing environments set: `PTVS_SAFE_ATTACH_FLEX_FALLBACK=1`, `PTVS_SAFE_ATTACH_VERBOSE=1` to confirm flex acceptance.
* If attach must succeed immediately prior to adapter changes: disable managed safe attach (`PTVS_SAFE_ATTACH_MANAGED_DISABLE=1`) to revert to legacy injector.

---

## Refactor Progress Snapshot
| Step | Description | Status |
|------|-------------|--------|
| 1 | Introduce `SafeAttachConfig` (env snapshot) | DONE (in use) |
| 2 | Extract `ThreadStateValidator` | DONE (centralized candidate validation) |
| 3 | Move walk inference + traversal to `InterpreterWalkLocator` | DONE (orchestrator slimmed) |
| 4 | `ProcessMemory` helper | DONE (integrated in orchestrator) |
| 5 | `RemoteWritePlan` | Pending |
| 6 | `PhaseTimer` (may stay dropped if timing not needed) | Deferred / reconsider |
| 7 | Unified logging facade | Pending |
| 8 | State machine orchestration | Pending |
| 9 | Unit tests for components | Partial (config + validator + locator; add ProcessMemory & write plan tests) |
| 10 | Telemetry scaffolding | Pending |

Recent Refactor Updates:
* Validator & walk locator extracted; orchestrator sequence: force → cache → infer (if needed) → walk → heuristic (opt‑in) → export (opt‑in).
* Inference + walk isolated for future Concord reuse.
* `ProcessMemory` helper added; orchestrator now uses it for slab read, eval breaker read, script buffer probe.

---

## Loader Script Alignment (Managed Safe Attach)
**Problem:** Current managed safe attach writes the path of `ptvsd_loader.py` (if present) into the target process, or falls back to a temp script that only calls `debugpy.listen(('127.0.0.1',0))`. Neither reliably establishes a VS↔debugpy session:
* `ptvsd_loader.py` only imports symbols (`attach_process`, etc.) and asserts threading state; it does not start a listener nor connect.
* Temp fallback picks an ephemeral port (0) and never communicates the chosen port or waits for a client.
* Result: Safe attach may complete the memory write phase but no debugger connection is made (silent no‑op).

**Required Behavior:** Executed script must (a) start or connect a debug adapter endpoint (debugpy) and (b) coordinate port/session info with VS (deterministic or discoverable). No manual native callback available in pure safe attach path.

**Remediation Plan:**
1. Introduce `safe_attach_loader.py` (new file) with logic:
   * Read env vars: `PTVS_DEBUG_HOST` (default 127.0.0.1), `PTVS_DEBUG_PORT` (explicit port or 0), `PTVS_DEBUG_SESSION` (GUID), `PTVS_WAIT_FOR_CLIENT` (default 1).
   * If PORT unset or 0: choose free port via debugpy.listen; if chosen dynamically write the resolved port to a temp sidecar file path from `PTVS_DEBUG_PORT_FILE`.
   * Call `debugpy.listen((host, port))` then optional `debugpy.wait_for_client()` depending on env.
2. Orchestrator search order for loader path: `safe_attach_loader.py` → existing `ptvsd_loader.py` (legacy) → temp bootstrap.
3. Before writing script path: materialize `safe_attach_loader.py` into extension install dir if missing and set required env vars.
4. (Optional) For legacy ptvsd flows still needing `ptvsd_loader.py`, keep backward compatibility behind env var `PTVS_SAFE_ATTACH_LEGACY_PTvsd=1`.
5. Add post-write verification: if using new loader and dynamic port, poll sidecar file (or small pipe) with timeout; emit telemetry fields `loader=New/Legacy`, `dynPort=<n>`.
6. Update Concord path to mirror search + env semantics (shared constants & env names).
7. Tests:
   * Unit: verify buffer truncation logic with new script name.
   * Integration mock: confirm that dynamic port file gets written and parsed.
   * Negative: missing env vars → script still creates listener with defaults.

**Acceptance for Loader Fix:**
* Deterministic attach (fixed port) when `PTVS_DEBUG_PORT` set.
* Dynamic port discoverable within <500ms via sidecar when port=0.
* No regress in legacy injection path (explicit disable of safe attach still works).
* Telemetry includes loader type and success/failure site for script execution.

---

## Concord Debugger Integration Tasks (Planned)
Goal: Align native / Concord engine safe attach flow with managed implementation to ensure consistent behavior, diagnostics, and telemetry.

| Area | Task | Outcome |
|------|------|---------|
| Offsets Inference | Port `InterpreterWalkLocator` logic (C#) to Concord side or expose via shared assembly loaded by both | Single authoritative inference implementation |
| Candidate Validation | Replace native validation with managed `ThreadStateValidator` parity (field probes, pending flag, script buffer) | Reduced false positives |
| Cache Reuse | Implement cross-engine cache contract (shared key: pid + pyBase + version) with fast probe API | Faster reattach, consistent reuse semantics |
| Memory Abstraction | Introduce `IConcordProcessMemory` adapter mirroring `ProcessMemory` API (ReadPointer, TryReadU32) | Simplifies porting managed helpers |
| Write Plan | Factor write sequence into shared `RemoteWritePlan` (script path, pending flag, eval breaker) with verification stage | Uniform attach side effects and error codes |
| Logging | Adopt unified prefix `[PTVS][SafeAttach][Concord]` and single success/failure summary line | Easier log correlation across engines |
| Telemetry | Emit Attempt/Result events with phase flags (cacheUsed, inferred, walked, heuristicUsed, exportFallback) | Consistent metrics for reliability tracking |
| Heuristic Path | Gate native heuristic behind same env var `PTVS_SAFE_ATTACH_FORCE_HEURISTIC`; remove implicit fallback | Predictable resolution order |
| Export Fallback | Align bypass logic (3.14+ default) and env var overrides (`ALLOW_EXPORT`, `ALLOW_EXPORT_FALLBACK`) | Security surface parity |
| Fault Injection | Add Concord-side hooks to simulate partial write / read failures for resilience tests | Robustness validation |
| Negative Tests | Add controlled test host injecting malformed slabs, truncated script buffer, bogus pointer chains | Confidence against malformed memory |
| Version Gating | Ensure Concord path early-outs for <3.14 to legacy behavior with clear site code | Clear diagnostics |
| Result Object | Mirror `SafeAttachResult` shape (version, disabled, freeThreaded, reused, exportBypassed, site, message) | Simplified UI/report integration |
| Timing (Optional) | Lightweight phase stopwatch (disabled by default) behind `PTVS_SAFE_ATTACH_VERBOSE` | On-demand perf diagnostics |
| Shared Constants | Consolidate magic values (mask 0x20, size=512, cap offsets) into common constants module | Single source of truth |
| Unit Tests | Add Concord unit & integration tests exercising inference, cache miss/hit, failure sites | Test parity |

### Concord Migration Phases
1. Read-only validation: implement inference + validation without performing writes (dry-run mode, asserts parity with managed discovery).  
2. Full write enablement: integrate write plan & post-write verification.  
3. Telemetry + logging unification.  
4. Fault injection & negative scenarios.  
5. Cleanup: remove duplicated native heuristics superseded by deterministic walk.  

### Acceptance Criteria (Concord)
* ≥95% of managed safe attach success scenarios replicate on Concord in CI.
* Export fallback invoked <2% of 3.14+ attaches (only when inference + cache fail and env var allows).
* Zero false positives in negative test suite (malformed slabs / corrupted pointers).
* Unified structured success/failure log lines present exactly once per attempt.
* Telemetry events emitted for 100% of attempts with consistent schema.

---

## Key Insights (What Made It Work)

* Stop assuming interpreter-walk sextuple is adjacent to the 5‑tuple; perform slab‑wide scan.
* Treat all `_Py_DebugOffsets` entries as offsets (not absolute pointers) – compose addresses at runtime.
* Validate interpreter-walk candidates against live memory (runtime → interpreter → thread list) before use.
* Expand DLL regex to match two or three digit minor versions: `python310/311/314.dll`.
* Deterministic resolution order: cache → inferred/parsed walk → (optional) heuristic → (optional) export fallback.
* Strengthen `PyThreadState` validation: probe eval_breaker (4 bytes), pending flag (int32), script path buffer readability.
* Use fixed 32‑bit eval breaker bit mask 0x20 for 3.14+; retain dynamic mask logic only for <3.14.
* Bypass export/TLS path by default for 3.14+ (reduce dependency & surface area).
* Early exit on first validated `tstate`; count candidates for diagnostics.
* Gate heuristic scan behind explicit env var; avoid unnecessary slow speculative scans.
* Revalidate cached `tstate` cheaply (offset-based field probes) before reuse.
* Ensure UTF‑8 loader script buffer is NUL‑terminated and within declared size.
* Add pending flag read probe to eliminate false positives in corrupted / stale memory cases.

---

## Code Simplification Plan (Upcoming Refactors)
Goal: reduce complexity / duplication in `SafeAttachOrchestrator` while preserving diagnostics.

| Area | Current Issue | Planned Simplification | Benefit |
|------|---------------|------------------------|---------|
| Env Var Lookups | Scattered repeated `EnvVarTrue` calls | Centralize into immutable `Config` snapshot object constructed once per attempt | Fewer branches, clearer intent |
| Validation Logic | Mixed inside orchestrator | Extract `ThreadStateValidator.Validate` enum result | Readability & test coverage |
| Walk Inference & Walk | Inline private methods | `InterpreterWalkLocator.Infer` / `Find` | Separation of concerns |
| Pointer Reads | Repeated lambdas | `ProcessMemory` helper | Less boilerplate |
| Write Sequence | Inline procedural | `RemoteWritePlan.Execute` | Easier post-write verification |
| Error Construction | Many early returns | Result discriminator | Cleaner control flow |
| Heuristic Gate | Inline conditions | Strategy chain | Extensibility |
| Test Helpers | Blob builders inline | `OffsetsBlobBuilder` | Reusable fixtures |
| Magic Constants | Literals inline | `SafeAttachConstants` | Self-documenting |

(Other sections unchanged below.)

### DIFF
--- a/Python/Docs/SafeAttachDebugpy.md
+++ b/Python/Docs/SafeAttachDebugpy.md
@@ -1,0 +1,69 @@
+# Safe Attach: Debugpy Requirements (Old vs New)
+
+## Summary
+The safe attach path no longer depends on the target Python environment having `debugpy` installed. The loader always prefers the VSIX-bundled copy and guarantees import by adjusting `sys.path` before attempting the import. This replaces older behavior that sometimes relied on the environment or multiple fallback loaders.
+
+## Previous (Legacy) Attach Pipeline
+- Loader script: `ptvsd_loader.py` (later transitional support for `debugpy`).
+- Search order:
+  1. Environment `sys.path` (user / venv installed `debugpy` could shadow bundled copy).
+  2. VSIX extension install directory (added to `sys.path` by loader).
+  3. Fallback: import `ptvsd` and alias as `debugpy` if `debugpy` import failed.
+- Multiple potential loader files (`ptvsd_loader.py`, dynamic scripts, bootstrap stubs) increased ambiguity.
+- Break / wait flags were varied (older env variable names) and sometimes duplicated logic.
+- Failure modes: Missing loader file or blocked VSIX deployment produced timeouts (no early telemetry for missing loader).
+
+## Current Safe Attach Pipeline
+- Single canonical loader: `safe_attach_loader.py`.
+- Loader path written into `_Py_DebugOffsets` script buffer by orchestrator.
+- Deterministic import path adjustments:
+  - Inserts its own directory at `sys.path[0]`.
+  - If a `debugpy` package directory sits adjacent to the loader (VSIX-bundled), that path is explicitly inserted (ensures vendored copy loads even if environment has none or an older version later on `sys.path`).
+- Import sequence:
+  1. `import debugpy` (expected to succeed with bundled copy).
+  2. Fallback: `import ptvsd as debugpy` (legacy) only if first import fails.
+- No requirement to pre-install `debugpy` in user/venv.
+- Environment variables (simplified):
+  - `PTVS_DEBUG_HOST` (default `127.0.0.1`).
+  - `PTVS_DEBUG_PORT` (`0` = dynamic allocation).
+  - `PTVS_WAIT_FOR_CLIENT` (`1` causes wait + implicit initial breakpoint unless disabled).
+  - `PTVS_DEBUG_BREAK` (forces initial breakpoint; effective OR with wait).
+  - `PTVS_DEBUG_PORT_FILE` (writes chosen port for host discovery if needed).
+  - `PTVS_SAFE_ATTACH_LOADER_VERBOSE` (verbose logging; default on, set to `0` to disable).
+  - `PTVS_DEBUG_LOG_FILE` (optional log sink path).
+- The orchestrator now fails early with telemetry if the canonical loader is missing (tests may still override via `PTVS_SAFE_ATTACH_LOADER_OVERRIDE`).
+- Inline or legacy fallback generation removed by default (can be reintroduced only in test scenarios).
+
+## Behavioral Differences
+| Aspect | Old | New |
+|--------|-----|-----|
+| Loader files | Multiple (`ptvsd_loader.py`, dynamic scripts, bootstrap) | Single (`safe_attach_loader.py`) |
+| Need environment `debugpy` | Optional; environment could shadow | Not required; VSIX copy always injected |
+| Fallback to `ptvsd` | Common when `debugpy` missing | Still present but rarely triggered |
+| sys.path manipulation | Loader inserted its directory; dynamic scripts inconsistent | Always inserts loader dir + adjacent `debugpy` explicitly |
+| Break Immediately | Varied flags; sometimes separate connect/wait roles | Unified via `PTVS_WAIT_FOR_CLIENT` + `PTVS_DEBUG_BREAK` (either triggers breakpoint) |
+| Telemetry on missing loader | Often silent ? timeout | Immediate failure with `ScriptBufferWrite` site + message |
+| Dynamic connect vs listen roles | Both sides could attempt connect (race/timeouts) | Target always listens; host always connects |
+
+## Migration Notes
+- Remove any external install steps instructing users to `pip install debugpy` solely for attach.
+- Update troubleshooting docs: If attach times out now, primary causes are listener creation failure (port bind) or blocked VSIX deployment—no longer missing Python package.
+- For test harnesses that need to force a specific loader path, use `PTVS_SAFE_ATTACH_LOADER_OVERRIDE`.
+
+## Troubleshooting Quick Reference
+| Symptom | Old Likely Cause | New Likely Cause | Action |
+|---------|------------------|------------------|--------|
+| Timeout waiting for debugger | Wrong connect/listen role, missing loader, missing debugpy | Port blocked or listener creation failed | Examine safe attach log for `listen failed` / port reuse; retry with different port |
+| Immediate `loader file not found` failure | Rarely surfaced | VSIX not deployed / file missing | Rebuild & deploy VSIX; verify file presence in extension folder |
+| No breakpoint on attach | Flags mismatch or dynamic script skipped breakpoint | `PTVS_WAIT_FOR_CLIENT=0` and `PTVS_DEBUG_BREAK` unset | Set `PTVS_WAIT_FOR_CLIENT=1` or `PTVS_DEBUG_BREAK=1` |
+| Falls back to ptvsd | Missing modern debugpy | Corrupted or absent bundled debugpy | Redeploy extension; check extension folder integrity |
+
+## Recommended Documentation Updates
+1. Replace legacy instructions referencing `ptvsd_loader.py` with `safe_attach_loader.py`.
+2. Clarify no environment installation of debugpy is required.
+3. Provide the simplified env var table above.
+4. Remove steps about reverse connect scripts unless future reverse mode is implemented.
+5. Add early failure telemetry codes mapping (e.g. `ScriptBufferWrite: loader file not found`).
+
+---
+Last updated: (auto-generated) Safe Attach debugpy requirements consolidation.
--- a/Python/Docs/py314-safe-attach-spec.md
+++ b/Python/Docs/py314-safe-attach-spec.md
@@ -18,6 +18,53 @@ Key change: Leverage CPython 3.14 `_Py_DebugOffsets` (`debugger_support`) for **
 
 ---
 
+## Debugpy Dependency: Legacy vs Current (Safe Attach)
+**Goal:** Clarify how the requirement for `debugpy` changed with the new `safe_attach_loader.py`.
+
+| Aspect | Legacy Attach / Transitional (ptvsd_loader) | Current Safe Attach (safe_attach_loader) |
+|--------|---------------------------------------------|------------------------------------------|
+| Primary loader file | `ptvsd_loader.py` (sometimes temp bootstrap) | `safe_attach_loader.py` (single canonical) |
+| Bundled package | `ptvsd` + (later) `debugpy` directory in VSIX | `debugpy` directory in VSIX (fallback to legacy `ptvsd`) |
+| sys.path handling | Loader typically inserted its dir; dynamic scripts inconsistent | Loader always prepends its directory and explicitly adds adjacent `debugpy` dir |
+| Environment needs `pip install debugpy` | Not required, but environment install could shadow bundled version | Not required; bundled copy always reachable and preferred |
+| Fallback behavior | `import debugpy` then `import ptvsd as debugpy` | Same, but debugpy import nearly deterministic due to path injection |
+| Multiple loader variants | Yes (legacy, dynamic connect, bootstrap) | No (single script) |
+| Role confusion (listen vs connect) | Possible (dynamic connect + host connect -> timeouts) | Target always listens, host always connects |
+| Initial breakpoint logic | Varied; different scripts | Unified: `PTVS_WAIT_FOR_CLIENT=1` and/or `PTVS_DEBUG_BREAK=1` triggers breakpoint |
+| Missing loader failure mode | Silent timeout in many cases | Immediate telemetry: site=ScriptBufferWrite msg=loader file not found |
+| Need for sidecar port file | Ad-hoc / absent in dynamic scripts | Supported via `PTVS_DEBUG_PORT_FILE` when dynamic port chosen |
+
+### Current Environment Variables (Supported / Simplified)
+| Variable | Meaning | Default |
+|----------|---------|---------|
+| `PTVS_DEBUG_HOST` | Interface to bind listener | `127.0.0.1` |
+| `PTVS_DEBUG_PORT` | Port to listen on (`0` = allocate ephemeral) | `0` |
+| `PTVS_WAIT_FOR_CLIENT` | Wait for debugger before resuming | `1` |
+| `PTVS_DEBUG_BREAK` | Force initial breakpoint (OR with WAIT) | Off |
+| `PTVS_DEBUG_PORT_FILE` | Path to write selected port number | None |
+| `PTVS_DEBUG_SESSION` | Correlation / session GUID | None |
+| `PTVS_SAFE_ATTACH_LOADER_VERBOSE` | Verbose loader logging (`0` disables) | `1` |
+| `PTVS_DEBUG_LOG_FILE` | File sink for loader logs (else stderr) | None |
+| (Orchestrator) `PTVS_SAFE_ATTACH_MANAGED_DISABLE` | Force legacy injection path | Off |
+| (Orchestrator) `PTVS_MIXED_MODE` | Bypass managed safe attach in mixed mode | Off |
+
+### Migration Guidance
+1. Remove any docs telling users to manually install `debugpy` just for attach.
+2. Refer only to `safe_attach_loader.py`; retire references to `ptvsd_loader.py` except in legacy troubleshooting appendix.
+3. Document single listener role: process listens, VS connects.
+4. For Break Immediately scenarios: set `PTVS_WAIT_FOR_CLIENT=1` (optionally also `PTVS_DEBUG_BREAK=1` – they both trigger a breakpoint now).
+5. Troubleshooting timeouts: check loader log for `listener started` or failures (`listen failed`). No more silent dependency on environment debugpy.
+
+### Failure Differentiation (Now)
+| Symptom | Likely Cause | Action |
+|---------|--------------|--------|
+| `loader file not found` | VSIX deployment missing / not copied yet | Redeploy / ensure `<IncludeInVSIX>true</IncludeInVSIX>` for loader |
+| Socket timeout (no connection) | Listener never started (port bind, path import failure) | Enable verbose loader logs; inspect `listen failed` lines |
+| Fallback to `ptvsd` | Bundled debugpy corrupt / missing | Repair VSIX; verify debugpy directory present |
+| No breakpoint on attach | WAIT & BREAK both unset | Set `PTVS_WAIT_FOR_CLIENT=1` or `PTVS_DEBUG_BREAK=1` |
+
+---
+
 ## Recent Managed Safe Attach Failure Analysis (Oct 2025)
 **Symptoms:**
 * Offsets parse stops after Strict/Extended/Relaxed and reports "no valid layout" even though flexible fallback could succeed.
--- a/Python/Product/Attacher/ManagedSafeAttach/SafeAttachOrchestrator.cs
+++ b/Python/Product/Attacher/ManagedSafeAttach/SafeAttachOrchestrator.cs
@@ -195,94 +195,67 @@ namespace Microsoft.PythonTools.Debugger.ManagedSafeAttach {
                 ulong pendingFlagAddr = supportBase + parsed.PendingCall;
                 ulong evalBreakerAddr = tstatePtr + parsed.EvalBreaker;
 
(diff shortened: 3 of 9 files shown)


bschnurr:
current logs
```
[PTVS][ManagedSafeAttach][DA] Fixed port prepared host=127.0.0.1 port=54800
[PTVS][ManagedSafeAttach] OffsetsAddr=0x7FFB34897000 size=131072 first32=78 64 65 62 75 67 70 79 C3 00 0E 03 00 00 00 00 00 00 00 00 00 00 00 00 98 D0 04 00 00 00 00 00 cookieOk=True
[PTVS][OffsetsParser] ver=0x030E00C3 rawFlags=0x00
[PTVS][OffsetsParser.Strict] blockIndex=90 byteOff=0x2D0 eb=0x128 rs=0x1EC8 pc=0x0 sp=0x4 size=512
[PTVS][OffsetsParser.Interpreter] Rejected candidate walk metadata rt=0x1 ih=0x100000001 im=0x1 th=0x0 tm=0x0 tn=0x0 condRt=True condIh=False condIm=True condThTm=True condTn=False
[PTVS][ManagedSafeAttach] Parsed offsets ver=0x30E00C3 evalBreakerOff=0x128 supportOff=0x1EC8 pendingOff=0x0 scriptPathOff=0x4 size=512 flexNonCanonical=False
[PTVS][ManagedSafeAttach] Inferred interpreter walk offsets from slab
[PTVS][ManagedSafeAttach] Walk located tstate=0x7FFB348AED88
[PTVS][ManagedSafeAttach] Using PTVS_DEBUG_PORT=54800
[PTVS][ManagedSafeAttach] Loader resolution final: c:\users\bschnurr\appdata\local\microsoft\visualstudio\18.0_ec6d73a1exp\extensions\microsoft corporation\python\18.0.0\safe_attach_loader.py
[PTVS][ManagedSafeAttach] Using wrapper loader C:\Users\bschnurr\AppData\Local\Temp\safe_attach_wrapper_14352.py host=127.0.0.1 port=54800 break=False
[PTVS][ManagedSafeAttach] SUCCESS pid=14352 ver=0x30E00C3 tstate=0x7FFB348AED88 exportBypassed=True size=512 loader=safe_attach_wrapper_14352.py
[PTVS][ManagedSafeAttach][DA] Safe attach success pid=14352 v=3.14
[PTVS][ManagedSafeAttach][DA] Promoted to DebugTcpAttachInfo host=127.0.0.1 port=54800
[PTVS][DebugAdapterRemoteProcess] Begin connect host=127.0.0.1 initialPort=54800 timeout=45000ms portFile=<none>
[PTVS][DebugAdapterRemoteProcess] Initial wait 300ms before first connect attempt
[PTVS][DebugAdapterRemoteProcess] Attempt 1 connecting to 127.0.0.1:54800 elapsed=303ms
Exception thrown: 'System.Net.Sockets.SocketException' in System.dll
[PTVS][DebugAdapterRemoteProcess] Attempt 1 SocketException code=ConnectionRefused elapsed=2402ms
[PTVS][DebugAdapterRemoteProcess] Waiting 80ms before next attempt (elapsed=2405ms)
[PTVS][DebugAdapterRemoteProcess] Attempt 2 connecting to 127.0.0.1:54800 elapsed=2489ms
```

Python/Product/Attacher/PyAttacher.cs
```diff
@@ -299,6 +306,23 @@ private static Process ConfigureAndStartProcess(ProcessStartInfo psi) {
         internal static DebugAttach AttachAD7Worker(int pid, int portNum, Guid debugId, string debugOptions, EventWaitHandle attachDoneEvent = null) {
             var hProcess = NativeMethods.OpenProcess(ProcessAccessFlags.All, false, pid);
             if (hProcess != IntPtr.Zero) {
+                try {
```
rchiodo, line 309: I may be remembering wrong, but this method of attach isn't used anymore. This is the old ad7 stuff. If you follow this up through references, it's not used.

Python/Product/Debugger.Concord/LocalComponent.cs
```diff
@@ -189,6 +191,44 @@ public override void Handle(DkmProcess process) {
 
                 if (PythonDLLs.GetPythonLanguageVersion(moduleInstance) != PythonLanguageVersion.None) {
                     pyrtInfo.DLLs.Python = moduleInstance;
+
```
rchiodo, line 194: I'm not sure why this code is here? It looks like it's trying to do a remote attach when the python dll loads in the process. Aren't you already attached at this point? I'm not sure why we need to run the loader script here.

Also seems like this code should be moved to a separate function.

Python/Product/Debugger.Concord/RemoteAttach314.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using Microsoft.PythonTools.Debugging.Shared.SafeAttach;
+using Microsoft.PythonTools.Debugger.Concord.Proxies.Structs;
+using Microsoft.VisualStudio.Debugger;
+
+namespace Microsoft.PythonTools.Debugger.Concord {
```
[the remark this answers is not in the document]
bschnurr: isnt' the python process already running and we want to attach to it?  none of this is working at this point. it used to work before i tried to safe attach

Python/Tests/DebuggerTests/DebuggerTests.csproj
rchiodo, line 1: I believe this was removed because the debugger tests were using the old ptvsd debugger. There's no longer a ptvsd debugger so I don't think we need these tests.

gramster:
I think I am confused by this, and all that AI generated noise at the start is not helping. It seems this change is conflating a bunch of things that should all be handled separately, and possibly some things that should not be done here at all:

- start using the offsets data structure in the mixed mode debugger, assuming it provides useful data that simplifies the work of supporting mixed mode in future Python versions
- update mixed mode for 3.14

What is the safe attach stuff? How is that related to updating mixed mode for 3.14? I know we want to use 3.14's new attach to process support in debugpy, but is that something we need for mixed-mode with Concord too? And if so, why is that not being done separately? Surely it would be something in debugpy repo? And right now, if it ain't broke in debugpy we shouldn't be fixing it; we should wait on Brett's work.

bschnurr:
yes this is abandoned. the safe attach stuff should proabably go into debugpy

PR #8307, opened 2025-10-20
try to future-proof deploy test targets script to look for what VS ve…

bschnurr:
…rsion its in

update Django targets to use VS version too


StellaHuang95:
Can you run the build pipeline and download the installer from the pipeline build artifact and verify it still works in dev18 before merging this in?

bschnurr:
pipeline run https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=12683308&view=results

PR #8324, opened 2025-11-07
fix Test discovery

bschnurr:
- we filter the results of unittest and pytest discover to make sure they make the source file.. now the property to check is 'CodeFilePath'
- one test was checking for an error when pytest wasn't installed, but it was getting a cached env.. now we make sure we don't cache in that case
- Enhancement: now our test discovery logs show up in test logs

### DIFF
--- a/Python/Tests/TestAdapterTests/TestDiscovererTests.cs
+++ b/Python/Tests/TestAdapterTests/TestDiscovererTests.cs
@@ -19,7 +19,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Runtime.CompilerServices;
 using Microsoft.PythonTools;
 using Microsoft.PythonTools.Common.Parsing;
 using Microsoft.PythonTools.TestAdapter.Pytest;


Python/Product/TestAdapter.Executor/UnitTest/UnitTestTestDiscoverer.cs
```diff
@@ -112,7 +112,7 @@ ITestCaseDiscoverySink discoverySink
 
                 foreach (var tc in testcases) {
                     // Note: Test Explorer will show a key not found exception if we use a source path that doesn't match a test container's source.
-                    if (settings.TestContainerSources.TryGetValue(tc.Source, out _)) {
+                    if (settings.TestContainerSources.TryGetValue(tc.CodeFilePath, out _)) {
```
bschnurr: CodeFilePath fix

Python/Tests/TestAdapterTests/Mocks/MockMessageLogger.cs
```diff
@@ -18,13 +18,19 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace TestAdapterTests.Mocks {
     class MockMessageLogger : IMessageLogger {
+        private readonly TestContext _ctx;
         public readonly List<Tuple<TestMessageLevel, string>> Messages = new List<Tuple<TestMessageLevel, string>>();
 
-        public void SendMessage(TestMessageLevel testMessageLevel, string message) {
-            Messages.Add(new Tuple<TestMessageLevel, string>(testMessageLevel, message));
+        public MockMessageLogger(TestContext ctx) => _ctx = ctx;
```
bschnurr: now we pass in the TestContext so that prints show up in the test logs

Python/Tests/Core/ProcessOutputTests.cs
```diff
@@ -37,39 +37,39 @@ public class ProcessOutputTests {
         [TestMethod, Priority(UnitTestPriority.P1)]
         public void ArgumentQuoting() {
             foreach (var testCase in new[] {
-                new { Source = "Abc", Expected = "Abc" },
+ new { Source = "Abc", Expected = "Abc" },
```
rchiodo: Spacing seems off? Is it supposed to be with no tabs?

Python/Tests/Core/ProcessOutputTests.cs
```diff
@@ -147,13 +148,13 @@ public void ProcessOutputEncoding() {
...
-                false,
-                null,
-                outputEncoding: new UTF8Encoding(false)
+            "cmd.exe",
+            new[] { "/C", "type " + Path.GetFileName(testData) },
+            testDataPath,
+            null,
+            false,
+            null,
+            outputEncoding: new UTF8Encoding(false)
```
gramster: Same here. The new format looks wrong; revert to the old.

Python/Tests/TestAdapterTests/TestExecutorTests.cs
```diff
@@ -211,7 +213,7 @@ public void RunPytestParameterizedAndDiscovery() {
             );
             var discoveryContext = new MockDiscoveryContext(runSettings);
             var discoverySink = new MockTestCaseDiscoverySink();
-            var logger = new MockMessageLogger();
+            var logger = new MockMessageLogger(TestContext); // pass TestContext
```
gramster: The comments here and elsewhere about // pass TestContext seem really redundant.

Python/Tests/Core/ProcessOutputTests.cs
bschnurr: updated

rchiodo (approved):

PR #8332, opened 2025-11-12
- this test broke when containers changed from source files to direct…

bschnurr:
…ories.

### DIFF
--- a/Python/Tests/TestAdapterTests/ProjectInfoTests.cs
+++ b/Python/Tests/TestAdapterTests/ProjectInfoTests.cs
@@ -16,10 +16,9 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


StellaHuang95:
Should the changes in Python/Product/TestAdapter.Executor/Pytest/PytestTestDiscoverer.cs and Python/Product/TestAdapter.Executor/UnitTest/UnitTestTestDiscoverer.cs also be reverted?

bschnurr:
> Should the changes in Python/Product/TestAdapter.Executor/Pytest/PytestTestDiscoverer.cs and Python/Product/TestAdapter.Executor/UnitTest/UnitTestTestDiscoverer.cs also be reverted?

ya. i'm going to hold off checking this in.. until i revert that change and fix the actual test explorer tests

bschnurr:
i'm just going to include this in the fix for other tests

PR #8336, opened 2025-11-20
Fix a watson bug

StellaHuang95:
fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1909676

A Watson report indicates a UI hang in Visual Studio. The UI thread becomes unresponsive for a significant duration during the initialization of the Python Language Server. The PR switches to a background thread to perform the heavy work of starting the process.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/LanguageServer.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/LanguageServer.cs
@@ -62,6 +62,9 @@ namespace Microsoft.PythonTools.LanguageServerClient {
                 return null;
             }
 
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/LanguageServer.cs
```diff
@@ -62,6 +62,9 @@ public async Task<Connection> ActivateAsync() {
                 return null;
             }
 
+            // Switch to a background thread before starting the process
+            await TaskScheduler.Default;
```
rchiodo: Might it have to switch back to the main thread when it's done? Not sure if the caller expects the result on the main thread or not

Issue #8265, opened 2025-07-23
Python>=3.11 is not in the list of language versions.


linette-zyy:
**Describe the bug**
<img width="797" height="359" alt="Image" src="https://github.com/user-attachments/assets/8e3e067e-72ff-42c5-9bf9-485a492bdf5e" />

**Steps to Reproduce**
1. Create a new Python application
2. Right click python environment and open “Add Environment…” command
3. Click Existing environment tab
<img width="1036" height="697" alt="Image" src="https://github.com/user-attachments/assets/421b2f70-7689-4c4b-a62a-fc29c04603ba" />
4. Browse to a non-virtual env, there will auto-detect but let you customize all parameters. Default is to add it to the registry (make it available globally).
<img width="1023" height="700" alt="Image" src="https://github.com/user-attachments/assets/2a4985dd-7990-4fb8-8667-6e274ac8eb26" />
5. Click already added custom environment and choose Configure in Python Environment 

6. Check language version.

**Expected behavior**
It should show correct version.

**Additional context and screenshots**
Python>=3.11 is not in the list of language versions.
<img width="358" height="635" alt="Image" src="https://github.com/user-attachments/assets/ce7c80e8-b219-459c-803b-d34fde620f4f" />

linette-zyy:
On the create Conda Environment window, the python>=3.11 also not list.

<img width="1198" height="704" alt="Image" src="https://github.com/user-attachments/assets/ba3fd951-9997-47fc-8746-81dd34865bc4" />

linette-zyy:
On today's build 18.3.0 Insiders [11311.1.main], python 3.14 is not list
<img width="974" height="509" alt="Image" src="https://github.com/user-attachments/assets/0e4b2c7e-f56e-4e90-8264-2a7a057bc01c" />

PR #8365, opened 2026-01-20
Add supported python versions

StellaHuang95:
fixes https://github.com/microsoft/PTVS/issues/8265, fix the python version list in the UI.


Python/Product/EnvironmentsList/ConfigurationExtension.xaml.cs
```diff
@@ -332,7 +332,11 @@ sealed class ConfigurationEnvironmentView : INotifyPropertyChanged {
             "3.7",
```
rchiodo: Can we remove 3.7 - 3.9? They're out of service. I don't think we can support them.

PR #8414, opened 2026-02-25
Revert "Update pipelines from Dev17 to Dev18"

bschnurr:
Reverts microsoft/PTVS#8400

reverting because i moved some steps around that dont run in PR and turns out they broke.

trying again with all steps turned on


rchiodo:
Why are you reverting it?

StellaHuang95:
Did the build fail with dev18 variables?

StellaHuang95 (approved):

PR #8507, opened 2026-05-28
Fix Watson #1229982: Catch PathTooLongException in PipPackageManager event handlers

StellaHuang95:
## Issue

[Watson #1229982](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1229982) — `CLR_EXCEPTION_System.IO.PathTooLongException` in `PipPackageManager.OnRenamed` (**18 hits**, build 18.4).

### Problem
The pip package watcher is registered on every `site-packages` directory with `IncludeSubdirectories = true`. `RenamedEventArgs.FullPath` / `OldFullPath` (and `FileSystemEventArgs.FullPath`) lazily build and normalize the path on demand. On .NET Framework that normalization enforces `MAX_PATH` (260) — so even the simple **property accessor** throws `PathTooLongException` for any deeply-nested vendored dependency, before our actual logic runs.

## Fix
Read `e.FullPath` / `e.OldFullPath` once into locals inside `try { … } catch (PathTooLongException) { return; }` at the top of `OnRenamed` and `OnChanged`. The downstream logic uses the local copies, so the property accessors are invoked at most once per event.

**Files changed**
- `Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs`

### DIFF
--- a/Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
+++ b/Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
@@ -756,44 +756,39 @@ namespace Microsoft.PythonTools.Interpreter {
             // RenamedEventArgs lazily builds and normalizes the full paths on demand,
             // and on .NET Framework that normalization enforces MAX_PATH (260) and
             // can throw PathTooLongException - even just reading the properties.
-            // Watch sites use IncludeSubdirectories under site-packages so deeply
-            // vendored packages routinely exceed 260 chars.
-            string fullPath, oldFullPath;
+            // The Directory.Exists / ModulePath.IsPythonFile checks below call into
+            // the same Win32 path APIs and can throw for the same reason, so wrap
+            // the whole inspection. Watch sites use IncludeSubdirectories under
+            // site-packages so deeply vendored packages routinely exceed 260 chars.
             try {
-                fullPath = e.FullPath;
-                oldFullPath = e.OldFullPath;
-            } catch (PathTooLongException) {
-                return;
-            }
-
-            if (Directory.Exists(fullPath) ||
-                ModulePath.IsPythonFile(fullPath, false, true, false) ||
-                ModulePath.IsPythonFile(oldFullPath, false, true, false)) {
-                try {
-                    _refreshIsCurrentTrigger.Change(1000, Timeout.Infinite);
-                } catch (ObjectDisposedException) {
+                if (Directory.Exists(e.FullPath) ||
+                    ModulePath.IsPythonFile(e.FullPath, false, true, false) ||
+                    ModulePath.IsPythonFile(e.OldFullPath, false, true, false)) {
+                    try {
+                        _refreshIsCurrentTrigger.Change(1000, Timeout.Infinite);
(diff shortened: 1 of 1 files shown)


Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
```diff
@@ -753,9 +753,22 @@ public void NotifyPackagesChanged() {
...
+            string fullPath, oldFullPath;
+            try {
+                fullPath = e.FullPath;
+                oldFullPath = e.OldFullPath;
+            } catch (PathTooLongException) {
+                return;
+            }
+
+            if (Directory.Exists(fullPath) ||
+                ModulePath.IsPythonFile(fullPath, false, true, false) ||
```
rchiodo: Copilot generated:
[unverified] The Skeptic notes that `Directory.Exists(fullPath)` and `ModulePath.IsPythonFile(fullPath, ...)` could theoretically also throw `PathTooLongException` if a path passes the getter's normalization but hits a different length check in Win32 APIs. In practice this is unlikely (the getter normalization *is* the MAX_PATH enforcement), but if you want belt-and-suspenders safety, consider widening the try/catch to encompass the `if` block as well. Low risk — the current fix addresses the Watson crash site.

[verified]

Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
```diff
@@ -764,9 +777,18 @@ private void OnRenamed(object sender, RenamedEventArgs e) {
         }
```
rchiodo, line 774: Copilot generated:
[unverified] The Skeptic and Architect both note that if this class has `OnCreated` or `OnDeleted` handlers that also access `e.FullPath`, they have the same latent vulnerability. Worth a quick audit of the other `FileSystemWatcher` event subscriptions in this file. Not blocking — Watson only reported hits on `OnRenamed` — but a proactive fix would prevent future Watson reports.

[verified]

Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
StellaHuang95: Addressed in 0ca2bac6f. Widened the try/catch in both OnRenamed and OnChanged to wrap the full `if` block (including `Directory.Exists`, `ModulePath.IsPythonFile`, and `PathUtils.GetFileOrDirectoryName`), since those calls hit the same Win32 path-normalization code as the `FullPath` getter and can throw for the same reason. Belt-and-suspenders as you suggested.

Python/Product/VSInterpreters/PackageManager/PipPackageManager.cs
StellaHuang95: Audited - there are no other handlers to update. `watcher.Created`, `watcher.Deleted`, and `watcher.Changed` all bind to the same `OnChanged` method (see lines 682-685), and `watcher.Renamed` binds to `OnRenamed`. Both are protected by the (now-widened in 0ca2bac6f) try/catch.

rchiodo (approved):
Approved via Review Center.

PR #8511, opened 2026-05-28
Fix Watson #1488201: Surface cookiecutter install failures without crashing VS

StellaHuang95:
## Issue

[Watson #1488201](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1488201) — `CLR_EXCEPTION_Microsoft.CookiecutterTools.Model.ProcessException` from `CookiecutterClient.WaitForOutput` (**12 hits**, builds 17.11.35219.272 – 18.3.11520.95).

### Problem
The cookiecutter feature creates a side venv, installs pip, then `pip install cookiecutter`. If any of those subprocesses fail (no network, corporate proxy blocking PyPI, corrupt interpreter), the subprocess exits non-zero, `WaitForOutput` throws `ProcessException`, and the exception escapes the WPF Dispatcher → crashes VS.

The original code in `CreateCookiecutterEnv` wrapped only the first install step in `try { … } throw new CriticalException(...)`. The outer handler in `CookiecutterViewModel.EnsureCookiecutterIsInstalledAsync` uses `catch (Exception ex) when (!ex.IsCriticalException())` — so the wrapped `CriticalException` **propagated uncaught** anyway, and the other two install paths (`pip install pip`, `pip install cookiecutter`) had no guard at all.

## Fix
- Add helper `WaitForCookiecutterInstallOutput(...)` that wraps `WaitForOutput`, catches `ProcessException`, writes a formatted error + stderr lines to `_redirector`, and re-throws as `InvalidOperationException`. `InvalidOperationException` is non-critical, so the outer handler catches it cleanly → sets status to `Failed`, returns `false`, fires telemetry.
- Route all three install-path `WaitForOutput` calls through the new helper.
- Remove the old `throw new CriticalException(...)` wrap in `CreateCookiecutterEnv` (the comment claiming it was "needed for EnsureCookiecutterIsInstalledAsync to fail properly" was incorrect — it actually crashed VS).
- Add localized string `CookiecutterInstallFailed` for the formatted error message.

**Files changed**
- `Python/Product/Cookiecutter/Model/CookiecutterClient.cs`
- `Python/Product/Cookiecutter/Strings.resx`
- `Python/Product/Cookiecutter/Strings.Designer.cs`

### DIFF
--- a/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
+++ b/Python/Product/Cookiecutter/Model/CookiecutterClient.cs
@@ -65,15 +65,13 @@ namespace Microsoft.CookiecutterTools.Model {
         }
 
         public async Task CreateCookiecutterEnv() {
-
-            // Create a virtual environment using the global interpreter
-            try {
-                await CreateVenvWithoutPipThenInstallPip();
-            } catch (ProcessException ex) {
-                // critical exception is needed for EnsureCookiecutterIsInstalledAsync() to fail properly
-                var errMsg = Strings.InstallingCookiecutterCreateEnvFailed.FormatUI(_expectedEnvFolderPath);
-                throw new CriticalException(errMsg, ex);
-            }
+            // Failures from the underlying install steps now surface as
+            // InvalidOperationException via WaitForCookiecutterInstallOutput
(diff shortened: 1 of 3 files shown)


rchiodo (approved):
Approved via Review Center.

PR #8513, opened 2026-05-28
Fix Watson #1812702: Catch broken-pipe IOException in StreamIntercepter

StellaHuang95:
## Issue

[Watson #1812702](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1812702) — `crash64: CLR_EXCEPTION_System.IO.IOException_8007006d` (`ERROR_BROKEN_PIPE`) in `StreamIntercepter.Write` (**10 hits**, builds 17.4.33213.308 – 17.5.33627.172).

### Problem
`StreamIntercepter` wraps Pylance's stdin / stdout so PTVS can massage the LSP `initialize` message. When Pylance exits or crashes, the next `Write` / `Read` / `Flush` on the pipe throws `IOException("Pipe is broken")`. StreamJsonRpc *does* eventually convert it to a `Disconnected` event, but the throw bubbles synchronously through 30+ frames of awaiter continuations until it escapes to a thread without a handler — categorized as `crash64` because the unhandled exception surfaced natively.

## Fix
Wrap `Flush()`, `Read(...)`, and `Write(...)` in `try { … } catch (IOException) { } catch (ObjectDisposedException) { }`:
- `Read` returns `0` on failure — StreamJsonRpc treats `0` as EOF and cleanly raises `Disconnected`.
- `Flush` and `Write` simply swallow; StreamJsonRpc detects the disconnect on the next read.

**Files changed**
- `Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/StreamIntercepter.cs`

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/StreamIntercepter.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/StreamIntercepter.cs
@@ -69,21 +69,29 @@ namespace Microsoft.PythonTools.LanguageServerClient.StreamHacking {
         public override long Seek(long offset, SeekOrigin origin) => baseStream.Seek(offset, origin);
         public override void SetLength(long value) => baseStream.SetLength(value);
         public override void Write(byte[] buffer, int offset, int count) {
+            // Compute the handler result outside the try so that exceptions from
+            // writeHandler (which is application logic, not pipe I/O) propagate
+            // normally and never silently null out the handler.
+            StreamData payload;
+            bool keepHandler;
+            if (writeHandler != null) {
+                var writeHandlerResult = writeHandler.Invoke(new StreamData { bytes = buffer, offset = offset, count = count });
+                payload = writeHandlerResult.Item1;
+                keepHandler = writeHandlerResult.Item2;
+            } else {
+                payload = new StreamData { bytes = buffer, offset = offset, count = count };
+                keepHandler = true;
+            }
             try {
-                if (writeHandler != null) {
-                    var writeHandlerResult = writeHandler.Invoke(new StreamData { bytes = buffer, offset = offset, count = count });
-                    baseStream.Write(writeHandlerResult.Item1.bytes, writeHandlerResult.Item1.offset, writeHandlerResult.Item1.count);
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/StreamIntercepter.cs
```diff
@@ -42,26 +42,49 @@ public StreamIntercepter(Stream stream, Func<StreamData, Tuple<StreamData, bool>
...
+                // Pipe to Pylance is broken. Returning 0 signals EOF to StreamJsonRpc
+                // which will cleanly raise Disconnected.
+                return 0;
+            } catch (ObjectDisposedException) {
+                return 0;
+            }
             var args = new StreamData { bytes = buffer, offset = offset, count = result };
             readHandler.Invoke(args);
             return result;
         }
```
rchiodo, line 68: Copilot generated:
-80

[unverified] The `try` block wraps both `writeHandler.Invoke(...)` and `baseStream.Write(...)`. If the delegate throws `IOException` for reasons unrelated to the broken pipe (e.g., I/O in its own logic), it would be silently swallowed and `writeHandler` would never be set to `null`. Consider narrowing the try to wrap only the `baseStream.Write(...)` calls:

```csharp
var writeHandlerResult = writeHandler.Invoke(...);
try {
    baseStream.Write(writeHandlerResult.Item1.bytes, ...);
} catch (IOException) { }
```

Both the Skeptic and Architect flagged this independently. While the Advocate correctly notes the handler is a pure data transform (making real-world impact unlikely), the narrower scope would make intent explicit at zero cost.

[verified]

StellaHuang95: Addressed in 7536cff28. Hoisted `writeHandler.Invoke(...)` out of the try block so handler-thrown `IOException`s (application logic) propagate normally; only `baseStream.Write(...)` remains under the catch. The handler-result branch is preserved via locals (`payload`, `keepHandler`) and the `writeHandler = null` step runs after the try.

rchiodo (approved):
Approved via Review Center.

PR #8514, opened 2026-05-28
Fix Watson #2325087: Tolerate ConnectionLostException in InvokeWithParametersAsync

StellaHuang95:
## Issue

[Watson #2325087](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2325087) — `CLR_EXCEPTION_StreamJsonRpc.ConnectionLostException` in `PythonLanguageClient.InvokeWithParametersAsync` (**10 hits**, builds 17.12.35527.113 – 17.12.35707.178).

### Problem
Editor-initiated LSP requests (completion, hover, go-to-definition, references, command execution) all funnel through `InvokeWithParametersAsync` / `NotifyWithParametersAsync`. Neither method had any exception handling. When Pylance crashes mid-request, `_rpc.InvokeWithParameterObjectAsync` throws `ConnectionLostException`, which propagates up through the editor's WPF-dispatched completion / hover provider → crashes VS.

## Fix
- Snapshot `_rpc` into a local; return `null` early if it is already nulled.
- Wrap the `await rpc.InvokeWithParameterObjectAsync<R>(...)` call in `try { … } catch (ConnectionLostException) { return null; } catch (ObjectDisposedException) { return null; }`. The editor LSP integration tolerates a `null` result (it just shows no completions / references) instead of crashing.
- Apply the same shape to `NotifyWithParametersAsync` (no return value — fire-and-forget notifications are dropped).

**Files changed**
- `Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs`

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
@@ -358,13 +358,38 @@ namespace Microsoft.PythonTools.LanguageServerClient {
         private async Task<R> InvokeWithParametersAsync<R>(string request, object parameters, CancellationToken t) where R : class {
             await _readyTcs.Task.ConfigureAwait(false);
 
-            return await _rpc.InvokeWithParameterObjectAsync<R>(request, parameters, t).ConfigureAwait(false);
+            var rpc = _rpc;
+            if (rpc == null) {
+                return null;
+            }
+
+            try {
+                return await rpc.InvokeWithParameterObjectAsync<R>(request, parameters, t).ConfigureAwait(false);
+            } catch (ConnectionLostException) {
+                // Pylance crashed mid-request. Returning null - editor LSP integration
+                // tolerates null (no completions / no references shown) rather than
+                // tearing down VS via the dispatcher.
+                return null;
+            } catch (ObjectDisposedException) {
+                // JsonRpc was disposed mid-request - same shutdown semantic.
+                return null;
+            }
         }
 
         private async Task NotifyWithParametersAsync(string request, object parameters) {
             await _readyTcs.Task.ConfigureAwait(false);
 
-            await _rpc.NotifyWithParameterObjectAsync(request, parameters).ConfigureAwait(false);
+            var rpc = _rpc;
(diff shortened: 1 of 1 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -358,13 +358,38 @@ public Task<object> InvokeCommandAsync(LSP.ExecuteCommandParams request, Cancell
         private async Task<R> InvokeWithParametersAsync<R>(string request, object parameters, CancellationToken t) where R : class {
```
rchiodo, line 358: Copilot generated:
-395

[unverified] The Architect notes that if other call sites in this class use `_rpc` directly without the same snap+guard+catch pattern, they'd have the identical crash vector. Consider a quick audit of the file for other unguarded `_rpc` usages. If ≥3 sites need protection, extracting a `SafeInvokeAsync<R>(Func<JsonRpc, Task<R>>)` helper would reduce repetition and ensure consistent resilience.

[verified]

StellaHuang95: Addressed in 10cfce0bc. Audited the file - found 3 unguarded `_rpc.NotifyWithParameterObjectAsync(...)` sites:
1. `InvokeDidChangeConfigurationAsync` (was line 325)
2. `InvokeDidChangeWorkspaceFoldersAsync` (was line 330)
3. `OnReanalyzeChanged` (was line 511)

All three now route through the existing `NotifyWithParametersAsync` helper, which already has the snap-to-local + null-check + `ConnectionLostException`/`ObjectDisposedException` catch pattern - so it doubles as the `SafeInvokeAsync` you described, without adding a new abstraction. The fire-and-forget call in `OnReanalyzeChanged` uses `.DoNotWait()` to preserve the original semantic.

The only remaining `_rpc.` usages are property assignments inside the initializer block immediately after `_rpc = rpc` (lines 264-266) and so can't hit `ConnectionLostException`.

rchiodo (approved):
Approved via Review Center.

PR #8517, opened 2026-05-28
Fix Watson cluster D (#1497512, #1487204, #1363565): Silence LSP teardown exceptions in DoNotWait

StellaHuang95:
## Issues

Three Watsons, one architectural root cause (**~8,253 hits combined — the loudest signal in the Watson query**):

- [Watson #1497512](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1497512) — `ObjectDisposedException` in `DoNotWait` from `PythonAnalysisOptions.Save` (**8,088 hits**, 17.9 – 17.12)
- [Watson #1487204](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1487204) — `ConnectionLostException` in `DoNotWait` from `set_ActiveInterpreter` (**156 hits**, 17.9 – 17.12)
- [Watson #1363565](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1363565) — `ObjectDisposedException` in `Microsoft.PythonTools.Common.dll!Unknown` (**9 hits**, 16.11)

### Problem
`TaskExtensions.DoNotWait` re-throws faulted-task exceptions back onto the original `SynchronizationContext` (via `ExceptionDispatchInfo.Capture(...).Throw()`). This is the **correct** behavior for genuine bugs — but PTVS uses `.DoNotWait()` extensively on fire-and-forget LSP notify tasks (`_rpc.NotifyWithParameterObjectAsync(...).DoNotWait()`). When the LSP / Pylance pipeline is being torn down (solution close, project unload, Pylance restart), those tasks faithfully fault with `ObjectDisposedException` (JsonRpc disposed) or `ConnectionLostException` (Pylance pipe closed). `DoNotWait` then re-throws on the UI thread → crashes VS.

This pattern affects every `_rpc.*.DoNotWait()` call site, hence the enormous hit count.

## Fix
Introduce a `_silencedExceptionTypes` list containing `ObjectDisposedException` and `StreamJsonRpc.ConnectionLostException`. Before the rethrow in both `ReThrowTaskException` and `DoNotWaitThreadContinuation`, check `IsSilencedException(ex)` and early-return if matched.

Notes:
- `OperationCanceledException` is *not* in the list — canceled tasks transition to `TaskStatus.Canceled` with `IsFaulted == false`, so the existing `if (task.IsFaulted && …)` guard already skips them.
- `InvalidOperationException` and other "real bug" exceptions remain unsilenced, so the existing `TaskExtensionsTests.DoNotWait` test (which asserts `InvalidOperationException` propagation) still passes.
- `StreamJsonRpc.ConnectionLostException` is referenced fully-qualified; `Python/Product/Common/Common.csproj` already references `StreamJsonRpc`.
- This is the only branch with workspace-wide reach — all `.DoNotWait()` call sites (~128) inherit the new silencing behavior. The trade-off is acceptable because both silenced types are inherently shutdown noise; the alternative would be a try/catch at every site.

**Files changed**
- `Python/Product/Common/Infrastructure/TaskExtensions.cs`

### DIFF
--- a/Python/Product/Common/Infrastructure/TaskExtensions.cs
+++ b/Python/Product/Common/Infrastructure/TaskExtensions.cs
@@ -55,6 +55,25 @@ namespace Microsoft.PythonTools.Infrastructure {
             return false;
         }
 
+        // Only silence when every inner exception is a silenced type, so a real
+        // bug riding alongside a teardown exception (e.g. from Task.WhenAll) still
+        // surfaces.
+        private static bool ShouldSilence(AggregateException ae) {
+            if (ae == null) {
+                return false;
+            }
+            var flat = ae.Flatten().InnerExceptions;
+            if (flat.Count == 0) {
+                return false;
+            }
+            for (int i = 0; i < flat.Count; i++) {
+                if (!IsSilencedException(flat[i])) {
+                    return false;
+                }
+            }
+            return true;
+        }
+
         /// <summary>
         /// Suppresses warnings about unawaited tasks and rethrows task exceptions back to the callers synchronization context if it is possible
         /// </summary>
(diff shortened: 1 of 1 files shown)


Python/Product/Common/Infrastructure/TaskExtensions.cs
```diff
@@ -53,13 +83,19 @@ private static void ReThrowTaskException(object state) {
             var task = (Task)state;
             if (task.IsFaulted && task.Exception != null) {
                 var exception = task.Exception.InnerException;
```
rchiodo: Copilot generated:
[unverified] `task.Exception.InnerException` returns only the first inner exception. If a faulted task wraps multiple inners (e.g., from `Task.WhenAll`) where the first is `ObjectDisposedException` (silenced) but a subsequent one is a real bug, the real exception is swallowed. This is pre-existing behavior (the original rethrow had the same limitation), but the silencing filter now extends its impact. Consider iterating `task.Exception.InnerExceptions` and only silencing when *all* are silenced types — or document this trade-off explicitly in the remarks.

[verified]

rchiodo: Copilot generated:
When an exception is silenced, it vanishes with no trace. A lightweight `Trace.TraceInformation` or `Debug.WriteLine` here (and at line 95) would aid future diagnostics ("why isn't my callback running?") without crashing VS. Not blocking, but would help the next person debugging unexpected teardown timing.

[verified]

StellaHuang95: Addressed in e2d5208eb. Added a ShouldSilence(AggregateException) helper that flattens and requires every inner exception to be a silenced type before swallowing. If any inner is non-silenced, we now fall through to the existing rethrow path. The single-exception case (the common one) is unchanged.

StellaHuang95: Addressed in e2d5208eb. Added a Debug.WriteLine in both silenced branches (ReThrowTaskException and DoNotWaitThreadContinuation) that logs the AggregateException type and the first inner's type. Cheap, debug-only, and leaves a breadcrumb in the VS Output window for anyone investigating "why isn't my callback running".

rchiodo (approved):
Approved via Review Center.

PR #8536, opened 2026-06-09
Triage pipeline: scope to Developer Community, enrich diagnostics JSON, fix empty-list + DC-state filtering

StellaHuang95:
## Summary

Five changes to the weekly triage pipeline that pulls customer-reported issues from the internal work-item tracker and posts a summary to `microsoft/PTVS`. Two features the operator asked for, three bug fixes that surfaced when running the pipeline against real data.

| # | Change | Type |
|---|---|---|
| 1 | Only include issues that customers originally reported via Visual Studio Developer Community — exclude internal-only items | Feature |
| 2 | Include the issue title, description, repro steps, and every attached file (not just the PTVS diagnostics log) in the per-issue output | Feature |
| 3 | Delete the per-folder README — keep this folder implementation-only | Cleanup |
| 4 | Don't crash when the week has zero issues to report | Bug fix |
| 5 | Include a couple more types of customer reports + stop discarding items the moment Developer Community auto-closes them | Bug fix |

All five are scoped to `Build/triage/` — nothing else changes.

---

## 1. Only include customer-reported issues

Before this PR, the report included every work item in the team's area path — internal compliance trackers, S360 items, accessibility benchmarks, internal crash buckets, etc. Since the pipeline also posts a public summary to GitHub, that's a real leak risk.

Customer-reported issues that come in through Visual Studio Developer Community arrive with a specific work-item type and a unique field that internal items never have. Filtering by that type gives a clean, sufficient cut: across the last 365 days in this area, 112 of 112 customer-reported items had it, and zero internal items did.

## 2. Richer per-issue output

The per-issue JSON file we publish as the build artifact used to contain only the parsed contents of the `PythonToolsDiagnostics_*.log` attachment. Anyone reading the artifact had to round-trip back to the work-item UI to learn the title, description, repro steps, or open any other attached file (screenshots, crash dumps, project zips, etc.).

The new schema bundles everything an investigator needs in one document:

- **`work_item`** — the basic context: id, title, link, type, state, area, dates, tags, plus the description / repro steps / system info as plain text (HTML stripped).
- **`attachments[]`** — every file attached to the issue, with name, link, size, and (for text-friendly files within a size cap) a short head excerpt of the content.
- **`python_tools_diagnostics`** — the existing structured extraction (VS / PTVS / Python versions, loaded assemblies, last exceptions) kept under a clean namespace.

Companion fix in the sanitizer: it used to scrub only the top-level string fields of the per-issue file, which was fine for the old flat shape but not for the new nested attachments. I added a recursive walker so PII and secret scanning now covers the whole document. Critically, if a secret is detected anywhere inside the diagnostics file, the corresponding work-item file is also flagged so the drop step removes both — previously a secret hidden in an attachment excerpt could still ride along in the build artifact.

## 3. Delete the per-folder README

Operator docs belong outside the repo; this folder should be implementation-only. The pipeline YAML and per-script comment headers already carry the runtime details.

## 4. Don't crash on an empty week

Latent bug, surfaced as soon as commit 1 made "zero items this week" a common case:

> `The property 'Count' cannot be found on this object` — at `post-report-issue.ps1` line 211

The input parser used a PowerShell if-expression to handle the empty-list path:

```powershell
$candidates = if ($null -eq $parsed) { @() } elseif … { @($parsed) }
```

PowerShell quietly collapses single-statement if-expressions whose branch yields an empty array — the variable ends up ``, and the next line tries to read `.Count` on it. Fixed by switching to plain if/elseif/else statements with direct assignment, plus a new test that runs the function against an empty-list file to lock the regression. Same family of bug as #8530.

## 5. Better customer-report coverage + stop dropping auto-closed items

Two more issues showed up on live runs:

**(a) Missing a related work-item type.** Developer Community actually creates two related types when customers report things. The original filter only picked one. Adding the second was a one-line config change. Small minority of items, but real.

**(b) Aggressive "closed" exclusion was hiding almost everything.** The pipeline excluded any item whose state contained the word `Closed`. But Developer Community auto-closes items within hours of submission with states like `DC - Closed - Duplicate`, `DC - Closed - Not Enough Info`, `DC - Closed - Not a Bug`, etc. — all of which still need engineering review. Empirically, 107 of 112 customer items over the last year carried such a state, so the report was returning essentially nothing.

Fixed by excluding only the genuinely-done states (anything with `Fixed` in the name covers the two cases). Everything else flows through to the triage report.

| Window | Before | After |
|---|---|---|
| 7-day  | 0 items | 0 items (only item this week is one truly fixed; correctly excluded) |
| 30-day | 0 items | 13 items needing real triage attention |

---

## Test plan

- `pwsh -File Build/triage/run-tests.ps1` — all four scripts have offline self-tests; all pass on each commit. New coverage added for: filter shape, attachment metadata pass-through, HTML decoding, recursive PII walker (including single-element array preservation), end-to-end secret propagation from a diagnostics attachment back to its work-item file, and the empty-list no-crash path.
- Live pipeline test-runs on the `agent` branch caught the bugs in commits 4 and 5. After commit 5, a 30-day run returns 13 real triage candidates; the typical 7-day run returns 0 because only one customer issue touched this week and it was already fixed.

## Risk

- Pipeline remains read-only — no closes, no labels, no comments back to the work-item tracker.
- The richer per-issue file is published only as an internal build artifact; the public summary issue on GitHub still contains only titles and links — no scrubbed content.
- The sanitizer changes are additive; existing work-item sanitization is unchanged, and diagnostics-side secret detection now flips the work-item flag too so the drop step catches both files.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


StellaHuang95:
Here's an example of the weekly summary list. https://github.com/microsoft/PTVS/issues/8535

Build/triage/parse-diagnostics.ps1
```diff
@@ -53,6 +99,45 @@ function Get-AzdoAuthHeader {
...
+    if ($null -eq $v) { return $Default }
+    return $v
+}
+
+function ConvertTo-PlainText {
+    param([string] $Html)
+    if ([string]::IsNullOrEmpty($Html)) { return '' }
+    # 1. Broad HTML entity decode (NOT a whitelist) — exposes any PII /
+    #    secrets that the original poster may have entity-encoded so the
+    #    downstream sanitizer's regexes can see them.
```
rchiodo: Copilot generated:
📍 Build/triage/parse-diagnostics.ps1:ConvertTo-PlainText
The Skeptic verified by execution that `HtmlDecode`-then-strip ordering destroys legitimate angle-bracketed content: `'use List&lt;int&gt; and Dictionary&lt;K,V&gt;'` becomes `'use List and Dictionary'`. Because entities are decoded to real `<...>` *before* `[regex]::Replace($decoded, '<[^>]+>', '')` runs, generic types and XML/pyproj fragments — exactly the repro detail feature #2 exists to surface — are silently deleted from the artifact. The Advocate defends the decode-first ordering as intentional PII exposure, but both goals can be met: strip tags from the *raw* HTML first, then `HtmlDecode` the remaining text. Root-cause ordering fix needed.

[verified]

Build/triage/parse-diagnostics.ps1
```diff
@@ -145,59 +370,61 @@ function Invoke-Parse {
     $candidates = $raw | ConvertFrom-Json
```
rchiodo, line 477: Copilot generated:
📍 Build/triage/parse-diagnostics.ps1:Invoke-Parse
The Skeptic (structurally confirmed) notes the `PythonToolsDiagnostics_*.log` has no priority in the attachment loop. A work item with ≥10 text attachments ordered before the diagnostics log exhausts `MaxTextAttachments`, records the log as `downloaded=$false`, and `Get-PythonToolsDiagnosticsBlock` returns `present=$false` — silently losing the primary structured extraction that the old code always fetched. Process/guarantee the `PythonToolsDiagnostics_*.log` download before applying the generic per-candidate budget.

[verified]

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [Build/triage/sanitize.ps1:L61](https://github.com/microsoft/ptvs/blob/5d592c272a136d364659027c95aae2d2dec48d20/Build/triage/sanitize.ps1#L61).

Copilot generated:
The Skeptic verified that broad secret heuristics (`connection-string-pw`, matching `Password=`/`Pwd=`/connection strings) now DROP entire triage items rather than redacting. Feature #2 expanded secret scanning from compact structured log fields to full descriptions + repro + up to 10×16KB attachment excerpts, massively enlarging the false-positive surface. A customer writing `Password=hunter2` in a repro flips `sanitization_aborted`, and fetch.yml Step 4 deletes the wi/diag pair — silently shrinking the triage feed with no record of what was suppressed. The in-code comment 'false positives are cheap (just an extra <redacted>)' conflates PII redaction with secret-driven item drop. Consider redacting-in-place for low-confidence patterns in free-text, and dropping only on high-confidence patterns (PEM/AKIA/ghp_/jwt).

[verified]

Build/triage/parse-diagnostics.ps1
StellaHuang95: Fixed in d019940. Reordered `ConvertTo-PlainText` to strip tags from the *raw* HTML first, then `HtmlDecode` the remaining text -- real `<p>`, `<br>`, `</li>` etc. are still stripped, and entity-encoded brackets (`List&lt;int&gt;`, pyproj/XML fragments) now survive into the artifact as `List<int>`. Broad `HtmlDecode` is preserved so the downstream sanitizer still sees PII / secrets hidden behind entity encoding. Added three self-test cases (generics, real-tag-around-encoded-bracket, pyproj snippet) plus an explanatory comment naming the prior bug.

Build/triage/parse-diagnostics.ps1
StellaHuang95: Fixed in d019940. Pulled the PTD-log filename match into a `Test-IsPythonToolsDiagnosticsLog` helper (the regex was already inlined twice), and in the `Invoke-Parse` attachment loop the `PythonToolsDiagnostics_*.log` now (a) bypasses the per-candidate text-attachment budget so it is always downloaded regardless of position, and (b) does not count against `MaxTextAttachments` so its position in the AzDO relation list cannot starve out the structured extraction. Added a simulated-loop self-test that places the PTD log after 10 noisier text attachments and asserts it is still processed (and that the trailing non-PTD text is still correctly budget-skipped).

StellaHuang95:
Fixed in 7befeb7. Reproduced first (`Password=hunter2`, `Pwd=changeme`, `Set Password=test123 is set`, `Pwd=abcdef should not crash`, `Server=...;Password=correct-horse-battery-staple;` — all five flipped `sanitization_aborted` pre-fix), then implemented the abort/redact split you suggested:

- Each entry in `$Script:SecretRegexes` now carries an `action` field. High-confidence patterns (ghp_, github_pat_, gh{s,o,u,r}_, PEM, AKIA, JWT, `AccountKey=`, `?sig=`) stay as `action=abort`. `connection-string-pw` becomes `action=redact` with replacement `$1=<redacted-secret>`, so `Password=hunter2` → `Password=<redacted-secret>` (preserves the prefix so the reader still sees the structure).
- `Remove-PiiFromString` processes abort matches first (flag only) then redact matches (regex replace + optional tracking). Two loops so a generic redact-class span can never eat an overlapping high-confidence match (`Password=ghp_aaa...` still aborts on the ghp_ portion).
- New optional `RedactedHitsRef` ref on `Remove-PiiFromString` / `Remove-PiiFromValue` plus a new `redacted_secrets` field on the sanitized wi for operator visibility — separate from `secret_hits` (still abort-class only). `sanitization_aborted` and fetch.yml Step 4 drop semantics are unchanged.
- Fixed the misleading L61-63 comment that conflated cheap-PII-FPs (redact-in-place) with expensive-secret-FPs (whole-item drop), and updated the script .DESCRIPTION header to document the abort/redact split.

Self-tests added cover: all five repro texts redact-in-place without aborting; five high-confidence patterns (ghp_, PEM, AKIA, JWT, `AccountKey=`) still abort; mixed text (`Password=hunter2 and ghp_...`) still aborts AND scrubs the Password portion; `Sanitize-Candidate` populates `redacted_secrets` without flipping abort; e2e diag-only redact-class hit propagates to `wi.redacted_secrets` without flipping abort and the sanitized diag file on disk no longer contains the original password value. All four `Build/triage/run-tests.ps1` checks pass.

Build/triage/sanitize.ps1
```diff
@@ -143,16 +252,16 @@ function Sanitize-Candidate {
         created_date       = $Candidate.created_date
         comment_count      = $Candidate.comment_count
 
-        title              = Remove-PiiFromString -Text $Candidate.title             -SecretHitsRef $secretsRef
-        description_html   = Remove-PiiFromString -Text $Candidate.description_html  -SecretHitsRef $secretsRef
-        repro_steps_html   = Remove-PiiFromString -Text $Candidate.repro_steps_html  -SecretHitsRef $secretsRef
-        system_info        = Remove-PiiFromString -Text $Candidate.system_info       -SecretHitsRef $secretsRef
+        title              = Remove-PiiFromString -Text $Candidate.title             -SecretHitsRef $secretsRef -RedactedHitsRef $redactedRef
```
rchiodo: Copilot generated:
The Skeptic verified (StrictMode, PS 7.6) that entity-encoded PII survives here while the *same* content is scrubbed in the diag path. `Sanitize-Candidate` scans the raw `description_html`/`repro_steps_html`/`system_info` via `Remove-PiiFromString`, so `alice&#64;example.com` is kept verbatim in `wi-*.json`, whereas `parse-diagnostics.ps1` HtmlDecodes before scanning and correctly redacts it. This PR added the broad `HtmlDecode` to the diag path *specifically* to expose encoded PII/secrets — apply the same reasoning here (e.g. `[System.Net.WebUtility]::HtmlDecode` the HTML fields before `Remove-PiiFromString`, or sanitize the already-decoded `work_item.*` text and drop the raw `*_html` from the artifact). Internal-only artifact, so warning not blocker.

[verified]

Build/triage/sanitize.ps1
```diff
@@ -58,45 +68,97 @@ param(
...
+       pattern = 'AKIA[0-9A-Z]{16}' },
+    @{ name = 'private-key-pem';      action = 'abort';
+       pattern = '-----BEGIN (?:RSA |EC |OPENSSH |DSA |PGP )?PRIVATE KEY-----' },
+    # connection-string-pw: GENERIC pattern that fires on benign customer
+    # wording like `Password=hunter2`, `Set Pwd=changeme in config`, etc.
+    # Redact in place (preserving the `Password=` / `Pwd=` prefix so the
+    # reader still sees the structure) instead of aborting the whole item.
+    @{ name = 'connection-string-pw'; action = 'redact';
+       pattern     = '(?i)(Password|Pwd)\s*=\s*[^;<>"\s]{6,}';
+       replacement = '$1=<redacted-secret>' }
```
rchiodo, line 142: Copilot generated:
`[unverified — fetch.yml Step 4 not in this diff]` Abort-class secrets are detected but never redacted in `$result`, so the literal token (`ghp_…`, PEM, `AKIA…`) is written to `OutDir` on disk. Before this PR the diag file held only compact extracted fields; it now holds up to 10×16KB raw attachment excerpts, greatly widening the window where a real secret is persisted un-redacted. Safety depends entirely on `fetch.yml` Step 4 deleting the pair *before* the artifact upload, which can't be verified here. Consider defense-in-depth: also replace abort-class matches with a marker in place, so the on-disk file never contains the literal secret even if the drop is skipped or reordered.

[verified]

Build/triage/sanitize.ps1
```diff
@@ -221,24 +336,82 @@ function Invoke-Full {
...
+                        }
+                        if ($hasRedact) {
+                            $existingR = @()
+                            if ($wi.PSObject.Properties['redacted_secrets'] -and $wi.redacted_secrets) {
+                                $existingR = @($wi.redacted_secrets)
+                            } elseif (-not $wi.PSObject.Properties['redacted_secrets']) {
+                                # Older wi shape pre-redacted_secrets — add the field.
+                                Add-Member -InputObject $wi -NotePropertyName 'redacted_secrets' -NotePropertyValue @() -Force
+                            }
+                            $wi.redacted_secrets = @(($existingR + $uniqRedact) | Sort-Object -Unique)
```
rchiodo, line 475: Copilot generated:
Orphan-diag fail-open: when an abort-class secret is found in a `diag-*.json` that has no sibling `wi-*.json`, the code only warns ("diag will still be dropped by Step 4 if a sibling appears later") and sets the script-local `$anyAborted`, but the `sanitization_aborted` flag Step 4 actually reads is never written anywhere. A stale/orphan diag containing a real secret would not be dropped. This should fail *closed* — delete the diag immediately or exit non-zero — rather than depend on a wi sibling that may never exist.

[verified]

Build/triage/sanitize.ps1
```diff
@@ -58,45 +68,97 @@ param(
...
 
     $result = $Text
 
-    # 1. Secret detection (no redaction — we abort or flag instead).
+    # 1a. High-confidence secrets: flag for abort (no in-place modification).
+    #     Process these FIRST so a redact-class pattern overlapping the same
+    #     span (e.g. `Password=ghp_aaa...`) doesn't eat the high-confidence
+    #     match before we can detect it.
     foreach ($s in $Script:SecretRegexes) {
+        if ($s.action -ne 'abort') { continue }
```
rchiodo, line 175: Copilot generated:
`$SecretHitsRef.Value += $s.name` on a generic `List[string]` doesn't call `.Add()` — PowerShell re-materializes it as `object[]` and writes it back through the `[ref]`. It works today because downstream only uses `.Count`/`Sort-Object`, but the `List` type is misleading and any future code calling `.Add()` on `$secrets` after the first hit would break. Pre-existing pattern; consider using `.Add()` consistently or typing the field as an array.

[verified]

Build/triage/sanitize.ps1
```diff
@@ -58,45 +68,97 @@ param(
...
+#                the pattern is specific enough that a hit is almost
+#                certainly a real secret.
+#   - 'redact' — lower-confidence pattern that may appear naturally in
+#                customer free-text (e.g. `Password=hunter2` in a repro
+#                description). On a hit we redact in place with the
+#                per-regex `replacement` and let the item flow through
+#                triage. Tracked separately on the sanitized wi as
+#                `redacted_secrets` for operator visibility, without
+#                triggering the abort drop.
+# Feature #2 expanded the scan surface from compact log fields to full
```
rchiodo, line 104: Copilot generated:
The redact-class `connection-string-pw` pattern requires `{6,}` value chars, so a short real credential like `Password=1234` is neither detected nor redacted (Skeptic verified). Acceptable noise tradeoff, but worth a brief comment noting the threshold silently lets short secrets through.

[verified]

Build/triage/parse-diagnostics.ps1
```diff
@@ -106,33 +220,173 @@ function Extract-DiagnosticsFields {
...
+        [int]       $ExcerptBytes
+    )
+    $name = Get-OptionalProp -Object $Attachment -Name 'name'
+    $url  = Get-OptionalProp -Object $Attachment -Name 'url'
+    $size = Get-OptionalProp -Object $Attachment -Name 'size_bytes'
+    $kind = Get-AttachmentKind -Name ([string] $name)
+
+    $entry = [pscustomobject] @{
+        name              = $name
+        url               = $url
```
rchiodo, line 301: Copilot generated:
When AzDO omits `resourceSize` (the `vsfeedback-notes.txt` fixture is exactly this case), `$size` is `$null`, so the pre-download size gate is skipped and `Get-AttachmentEntry` downloads the *entire* file before the post-download `MaxLogBytes` check rejects it. A multi-MB attachment with no `resourceSize` is fully pulled before being discarded, defeating part of the cap's bandwidth purpose. Consider capping bytes read / range-limiting the download so `MaxLogBytes` bounds bandwidth even without `resourceSize`. Low cost-impact (AzDO-bounded).

[verified]

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [Build/triage/query-azdo.ps1:L83](https://github.com/microsoft/ptvs/blob/7befeb76a2ef8ad821c8269816a952242e6d3912/Build/triage/query-azdo.ps1#L83).

Copilot generated:
The Architect structurally confirmed (against the worktree file) that the inline rationale comment near this clause still reads that `Closed` is excluded and is "set by this pipeline when closing." After change #5 both halves are now wrong: the config no longer excludes `Closed`, and the pipeline is read-only (it never closes work items). The justification moved to `config.json`'s `$comment_excludedStates`, but this inline copy was left stale and sits directly above the state clause it (mis)describes. Update or delete it to avoid misleading the next reader.

[verified]

Build/triage/sanitize.ps1
StellaHuang95: Fixed in 51c29f9. Added a `ConvertFrom-HtmlForScan` helper that `HtmlDecode`s `description_html` / `repro_steps_html` / `system_info` *before* `Remove-PiiFromString` runs, mirroring the diag-path treatment. Entity-encoded PII like `alice&#64;example.com` and `Password&#61;hunter2` is now detected and replaced (the on-disk artifact carries the canonical decoded-and-redacted text, never the encoded leak).

I kept the `_html` field names and did *not* strip tags here: keeping `<p>`/`<br>`/etc. preserves document structure for an operator reviewing the artifact, and avoids losing repro detail that happens to contain encoded angle brackets (same reason `ConvertTo-PlainText` does strip-then-decode in the diag path). The helper is null/empty/plain-text safe.

New self-test asserts an encoded email in `description_html`, an encoded `ghp_` in `repro_steps_html`, and an encoded path in `system_info` are all detected (no surviving `&#…;` entities, `<redacted-email>` marker present, `sanitization_aborted=true` from the encoded `ghp_`).

Build/triage/sanitize.ps1
StellaHuang95: Fixed in 51c29f9. Failing closed now: when an abort-class secret hit is detected in a `diag-<id>.json` that has no sibling `wi-<id>.json`, we `Remove-Item` the (already-sanitized but still containing markers/metadata) diag from `-OutDir` immediately. `fetch.yml` Step 4 iterates `wi-*.json` and reads `sanitization_aborted` to decide what to drop, so without a sibling wi file Step 4 would never see this diag — the bytes have to leave disk locally or they'll ride along in the artifact upload.

`$anyAborted` is still set (so `-FailOnSecret` still surfaces the hit), and the warning message now explicitly says "failing closed" instead of the previous "Step 4 will catch it if a sibling appears later" — which, as you noted, was a fail-open assumption.

New e2e self-test covers this: empty candidates list (so `Invoke-Full` emits zero wi files) plus a diag-77777.json containing a `ghp_…`. After `Invoke-Full`, the test asserts `diag-77777.json` is *absent* from `-OutDir`.

Combined with the defense-in-depth redact from the L122 thread, an abort-class hit in the diag-only / orphan path now (a) gets the literal token replaced with a marker, (b) gets the file deleted locally, and (c) sets `$anyAborted` so `-FailOnSecret` still trips downstream.

Build/triage/sanitize.ps1
StellaHuang95: Fixed in 51c29f9. Defense-in-depth implemented: abort-class hits now replace the literal token in place with a `<redacted-secret:<name>>` marker (e.g. `<redacted-secret:github-pat-classic>`) before the function returns. The abort flag is still authoritative for `fetch.yml` Step 4's drop semantics — this just stops the on-disk artifact from carrying the raw `ghp_…` / PEM / `AKIA…` / JWT / `AccountKey=…` in the window between sanitize.ps1 writing the file and Step 4 deleting it. If Step 4 is skipped, reordered, or fails entirely, the artifact upload still doesn't contain the literal secret.

Ordering preserved: abort patterns process first (so a redact-class span overlapping the same text — `Password=ghp_aaa…` — doesn't eat the high-confidence match before we detect it). The marker contains `<`, which the redact-class `connection-string-pw` value class `[^;<>"\s]` excludes, so abort-then-redact never produces a double-replacement; verified with the existing mixed-text test plus a new defense-in-depth test covering ghp_/PEM/AKIA (each must (a) trip the abort flag, (b) lose the literal from output, (c) gain the named marker).

Updated `.DESCRIPTION` and the `$Script:SecretRegexes` header comment to document the new abort = flag + redact behavior so the next reader doesn't conflate it with the redact-class drop-vs-redact split.

Build/triage/sanitize.ps1
StellaHuang95: Fixed in 51c29f9. Added a dedicated comment block next to the `connection-string-pw` entry documenting the `{6,}` floor:

- *Why*: customer free-text repros frequently contain illustrative placeholders (`Password=1234`, `Pwd=abc`, `Password=test`) that aren't real credentials. The 6-char floor trims that noise so we don't flood `redacted_secrets` with marker spam on benign wording.
- *Tradeoff*: a real credential shorter than 6 chars (rare; e.g. `Pwd=abcde`) is silently let through to triage. Accepted.
- *Contrast*: the abort-class patterns above (ghp_/PEM/AKIA/JWT/AccountKey=) have no length floor — their prefix alone makes a hit almost certainly real, so the noise/safety calculus is different and a hit can be high-confidence even at short lengths.

Pure comment-only change; the regex itself is unchanged.

Build/triage/sanitize.ps1
StellaHuang95: Fixed in 51c29f9. Switched both abort and redact loops from `$Ref.Value += $s.name` to `$Ref.Value.Add($s.name) | Out-Null`, so the caller's `Generic.List[string]` backing field is preserved across hits instead of being silently re-materialized as `object[]` on the first `+=`.

New self-test locks the regression down: passes a `List[string]` via `[ref]` to `Remove-PiiFromString` on a string containing two abort-class hits (`ghp_…` and `AKIAIOSFODNN7EXAMPLE`), then asserts (a) `$ref.Value` is still a `List[string]` after the call, and (b) `.Add()` from the caller-side variable still works (`$typedHits.Add('manual-after-call')` doesn't throw). Pre-fix this would fail at (a) — `+=` would have collapsed it to `object[]` after the first hit.

Inline comment added to the abort loop documenting *why* `.Add()` instead of `+=` so the next reader doesn't "tidy" it back.

Build/triage/parse-diagnostics.ps1
StellaHuang95: Fixed in 51c29f9. `Get-AzdoAttachmentBytes` now takes an optional `MaxBytes` parameter; when > 0 it adds an HTTP `Range: bytes=0-<MaxBytes>` header so an attachment with no advertised `resourceSize` can't blow the cap by streaming an unbounded download. `Get-AttachmentEntry` passes `MaxLogBytes` through, so the very case you flagged — `vsfeedback-notes.txt` (or any AzDO relation that omits `resourceSize`) — is now bandwidth-capped at request time, not just disk-write time.

Defense-in-depth on the AzDO side: the post-download `actualSize -gt $MaxLogBytes` check is preserved, so a server that ignores `Range` and 200's the full body still degrades gracefully to the original bytes-on-disk reject behavior. Headers hashtable is cloned (not mutated) so the caller's auth-headers reference is safe.

New parse-diagnostics self-test mocks `Invoke-WebRequest` via a script-scoped function shadow (no real network calls), then asserts:
- `MaxBytes > 0` ⇒ headers contain `Range: bytes=0-<MaxBytes>` and base `Authorization` survives.
- `MaxBytes = 0` (default) ⇒ NO `Range` header (back-compat for any future caller).
- Caller's `$Headers` hashtable is unmutated after the call.

StellaHuang95:
Fixed in 51c29f9. Updated the inline comment above the excluded-states clause in `query-azdo.ps1`. Both halves of the previous text were stale after change #5:
- It claimed `'Closed'` was excluded — but the config now excludes `Fixed` / `Resolved` / `Completed` / `Removed`.
- It claimed the value was "set by this pipeline when closing" — but the pipeline is read-only and never closes work items.

New comment correctly describes substring `NOT CONTAINS` semantics (a single configured value like `Fixed` excludes every state whose name contains it — `DC - Closed - Fixed`, `DC - Fixed Pending Release`, plain `Fixed`), explicitly notes the pipeline is read-only, and points to `config.json:$comment_excludedStates` for the full rationale (which lists every DC lifecycle state and explains the keep/exclude choice with the 95%-auto-closed-by-DC reasoning). Pure docs change.

Build/triage/parse-diagnostics.ps1
```diff
@@ -106,33 +220,207 @@ function Extract-DiagnosticsFields {
...
+    # Pre-download size gate using AzDO's resourceSize when available, so we
+    # don't pull a 50 MB log just to discover it exceeds the cap.
+    if ($null -ne $size -and $size -gt $MaxLogBytes) {
+        $entry.skip_reason = "size $size bytes exceeds cap $MaxLogBytes"
+        return $entry
+    }
+
+    $tmp = $null
+    try {
+        # Pass MaxLogBytes through so Get-AzdoAttachmentBytes range-caps
```
rchiodo, line 324: Copilot generated:
📍 Build/triage/parse-diagnostics.ps1:Get-PythonToolsDiagnosticsBlock
Unresolved regression: `Extract-DiagnosticsFields` runs over `$ptdEntry.content_excerpt` (capped at `ExcerptBytes`=16KB), not the full ≤`MaxLogBytes` download. The Skeptic empirically reproduced this — a 35KB log whose exception block is logged after the assembly dump yields `last_exceptions=1` from the full log but `0` from the 16KB head excerpt. PTVS logs routinely list 100+ assembly lines before the exception section, so this silently empties `last_exceptions` (and can truncate `loaded_assemblies`) in the exact structured-extraction feature #2 advertises. The inline "head-extraction is fine" comment encodes the wrong assumption. The full text is already downloaded then discarded — run the extractor over the full PTD bytes (before excerpting), or give the PTD log an uncapped `ExcerptBytes`, while still capping `content_excerpt` for the JSON.

[verified]

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [Build/triage/README.md:L1](https://github.com/microsoft/ptvs/blob/e560b4be279814a521e469fabbdc8169713946da/Build/triage/README.md#L1).

Copilot generated:
(re: this README deletion)
Unresolved from prior review: deleting this README leaves `fetch.yml:20` (`# Setup prerequisites (one-time, see Build/triage/README.md):`) pointing at a file that no longer exists — a guaranteed dead pointer in the first artifact an operator opens. All four reviewers verified the reference is still live at PR head. The `fetch.yml` header preserves only a subset of setup; the `gh label create azdo-triage-report …` command, the 90-day PAT rotation reminder, the run-fail-grant-rerun Build-Service provisioning sequence, the local dry-run instructions, and the operational/audit notes have no surviving home. Either fix the `fetch.yml` reference and fold the still-load-bearing operator knowledge into the YAML header, or keep a trimmed README.

[verified]

Build/triage/parse-diagnostics.ps1
StellaHuang95: Fixed in ea7018b. Reproduced the regression first (a 35 KB synthetic log with ~24 KB of `Loaded Assembly:` lines followed by a `NullReferenceException` — `last_exceptions=1` on the full text, `0` on the 16 KB head excerpt), then implemented the full-text path you suggested:

- `Get-AttachmentEntry` gained an optional `-FullTextRef [ref]` out-parameter. When the caller passes it, the FULL downloaded text (up to `MaxLogBytes`) is also written there. The JSON-visible `content_excerpt` field stays capped at `ExcerptBytes` — separate channel for callers that need the full bytes for structured extraction without paying the JSON-size cost.
- `Get-PythonToolsDiagnosticsBlock` gained an optional `-PtdFullText` parameter. When provided, `Extract-DiagnosticsFields` runs over that text instead of `content_excerpt`. Falls back to the excerpt for back-compat (test fixtures / older callers that don't capture).
- `Invoke-Parse` wires the two: for the PTD log specifically, it captures `$ptdFullText` via `-FullTextRef`, then passes it to `Get-PythonToolsDiagnosticsBlock`. Non-PTD attachments are unaffected (the budget cap, the excerpt cap, and the on-disk JSON shape all stay the same).
- The bytes are already in `$tmp` so capturing them is free; we're no longer re-throwing the text away before extraction.

New self-test locks the regression down: an in-memory PTD entry whose `content_excerpt` is the 16 KB head and whose tail (provided via `-PtdFullText`) contains the exception block. Asserts (a) excerpt-only still yields 0 exceptions on the same fixture (documents the prior bug so the next reader can't reintroduce it), and (b) full-text extraction recovers the `NullReferenceException`. Existing PTD-block tests continue to pass.

Also corrected the inline "head-extraction is fine" comment that encoded the wrong assumption.

StellaHuang95:
Fixed in ea7018b. You're right — `fetch.yml:20` and `config.json:2` both still pointed at the deleted README. Two parts:

1. **Killed the dead pointers.** `fetch.yml:20` no longer says "see Build/triage/README.md"; `config.json`'s `$comment` now points to the YAML header. `git grep Build/triage/README` is empty.

2. **Folded the still-load-bearing operator knowledge into `fetch.yml`'s header comment** (kept the folder implementation-only per the original PR goal — no restored README):
    - `gh label create azdo-triage-report …` invocation (the post-step does NOT auto-create labels).
    - ~90-day fine-grained-PAT rotation reminder.
    - First-run-fails-grant-rerun Build-Service provisioning sequence (the first run failing on a 401 IS what provisions the identity in AzDO permissions; that surprised the operator and was only documented in the README).
    - Local end-to-end dry-run command sequence, with the `az account get-access-token` bootstrap for `AZDO_ACCESS_TOKEN`.
    - Audit / operational notes — artifact upload contents, idempotency-vs-operator-closed-reports semantics (the bug fixed by `e560b4b`), manual re-runs, PR-safety rationale for `trigger: none` + `pr: none`.

Net: an operator opens `fetch.yml` and finds the same setup checklist + dry-run snippet + operational notes the deleted README used to carry, without having to chase a missing file.

Build/triage/parse-diagnostics.ps1
```diff
@@ -106,33 +225,246 @@ function Extract-DiagnosticsFields {
...
+        present                 = $false
+        reason                  = $Reason
+        source_attachment_name  = $null
+        source_attachment_url   = $null
+        vs_version              = $null
+        ptvs_version            = $null
+        python_version          = $null
+        debugger_type           = $null
+        machine_os              = $null
+        loaded_assemblies       = @()
```
rchiodo, line 380: Copilot generated:
📍 Build/triage/parse-diagnostics.ps1:Get-PythonToolsDiagnosticsBlock
Unresolved prior High finding. `Extract-DiagnosticsFields` still runs over `$ptdEntry.content_excerpt`, which is hard-capped at `ExcerptBytes` (16 KB) in `Get-AttachmentEntry`. The prior Skeptic empirically reproduced this — a ~21–35 KB PTVS log yields `last_exceptions=1` from the full text but `0` from the 16 KB head, because the assembly dump pushes the exception block past the cap (the normal PTVS log layout). The new "head-extraction is fine" comment encodes exactly the assumption that was disproved. The full log is already downloaded (range-capped to 4 MB) and then discarded. Run the extractor over the full PTD bytes before excerpting, or give the PTD log an uncapped extraction input while still capping `content_excerpt` for the JSON. This is a completeness regression in the exact structured-extraction feature this PR advertises.

[verified]

StellaHuang95: Already fixed in [`ea7018b`](https://github.com/microsoft/PTVS/commit/ea7018bc3d11ba86b14594a6c9d96fa38902cd6a) — this comment was anchored to that very commit but the bot didn't see the fix. Current HEAD is `0b06585a`.

The new `Get-PythonToolsDiagnosticsBlock` (parse-diagnostics.ps1:385-432) takes an optional `-PtdFullText` parameter, and `Invoke-Parse` (parse-diagnostics.ps1:476-503) captures the full ≤4 MB download via a new `-FullTextRef [ref]` out-parameter on `Get-AttachmentEntry`. `Extract-DiagnosticsFields` now runs over that full text for the PTD log; the JSON-visible `content_excerpt` stays capped at 16 KB. The misleading "head-extraction is fine" comment is gone and replaced with one that documents the prior bug.

Regression locked in at parse-diagnostics.ps1:662-689: a 35 KB synthetic log with ~24 KB of `Loaded Assembly:` lines followed by a `NullReferenceException` block asserts (a) excerpt-only still yields 0 exceptions on the same fixture (documents the prior bug so it can't be reintroduced) and (b) full-text extraction recovers the exception. All four `Build/triage/run-tests.ps1` self-tests pass.

rchiodo (approved):
Approved via Review Center.

PR #8539, opened 2026-06-10
Add CodeQL analysis workflow and update compliance pipeline schedule comment

bschnurr:


StellaHuang95 (approved):

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,47 @@
+name: CodeQL
+
+on:
```
rchiodo, line 3: Copilot generated:
There is no `schedule:` trigger, so a branch that's quiet for weeks is never re-analyzed against newly-published CodeQL queries for newly-disclosed vulnerability classes. GitHub's default setup includes a weekly `cron` for exactly this. The ADO `always: true` change only partially mitigates this (csharp/python, not JS/TS). Consider adding a weekly `schedule`. (Raised by Advocate and Skeptic.) [unverified]

[verified]

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,47 @@
...
+    branches:
+      - main
+      - release/*
+
+permissions:
+  actions: read
+  contents: read
+  security-events: write
+
+jobs:
```
rchiodo, line 14: Copilot generated:
There is no `concurrency` block, so rapid successive pushes to `main` spawn overlapping full analyses (each up to the 360-min timeout). Adding `concurrency: { group: codeql-${{ github.ref }}, cancel-in-progress: true }` would coalesce superseded runs and save runner minutes. Efficiency only, not correctness. (Advocate and Skeptic.) [unverified]

[verified]

PR #8552, opened 2026-06-16
Fix test target deployment directories

bschnurr:
## Summary
- create missing MSBuild `Python Tools` target directory before copying test targets
- fix delayed expansion in the VC Debug Launcher copy block so it uses the VC target path
- report copy/directory failures and return non-zero on deployment failure

## Validation
- Ran `Python\Prerequisites\deploy_test_targets.cmd` against a temporary fake VS install with `v18.0`; copied 4 Python targets and 3 Win32/3 X64 VC launcher files.

### DIFF
--- a/Python/Prerequisites/deploy_test_targets.cmd
+++ b/Python/Prerequisites/deploy_test_targets.cmd
@@ -1,4 +1,4 @@
-@setlocal
+@setlocal EnableExtensions EnableDelayedExpansion
 @echo off
 rem
 rem The test targets files redirect projects to use the targets deployed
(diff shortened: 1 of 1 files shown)


rchiodo (approved):
Approved via Review Center.

Python/Prerequisites/deploy_test_targets.cmd
```diff
@@ -13,16 +13,31 @@ echo This script should be run as an administrator.
...
 
 rem Dynamically attempt copy for all installed Visual Studio MSBuild version folders (v*.0)
-for /d %%D in ("%VSINSTALLDIR%\MSBuild\Microsoft\VisualStudio\v*.0") do call :docopy "%%~fD"
+set "_DEPLOY_FAILED="
+for /d %%D in ("%VSINSTALLDIR%\MSBuild\Microsoft\VisualStudio\v*.0") do (
+ call :docopy "%%~fD"
+ if errorlevel 1 set "_DEPLOY_FAILED=1"
+)
 
 pause
```
rchiodo, line 32: `pause` runs *before* `if defined _DEPLOY_FAILED exit /B 1`, so any non-interactive/CI caller hangs forever waiting for a keystroke — making the PR's new non-zero exit codes unreachable in exactly the automation context they're meant for. Consider gating it behind interactivity (e.g. `if /I not "%CI%"=="1" pause`). Pre-existing, but it directly undermines this PR's goal.

Python/Prerequisites/deploy_test_targets.cmd
```diff
@@ -39,44 +54,66 @@ if defined _VS_MAJOR set "_VC_TOOLSET_FOLDER=v%_VS_MAJOR%0"
...
-echo from %CD%
-echo to %TARGET%
- if not exist "%TARGET%" mkdir "%TARGET%" >nul 2>&1
- robocopy /E /XO Win32 "%TARGET%\Win32" *.* >nul
- robocopy /E /XO x64 "%TARGET%\X64" *.* >nul
+echo from !CD!
+echo to !TARGET!
+ robocopy /E /XO Win32 "!TARGET!\Win32" *.* >nul
+ if errorlevel 8 (
+ echo [ERROR] Failed to copy Win32 VC Debug Launcher targets to !TARGET!\Win32
```
rchiodo, line 90: An empty source `Win32`/`x64` directory makes `robocopy /E /XO` return 0, which passes the `if errorlevel 8` gate; the script then emits `[WARN] ... missing` with `Win32=0` but still `exit /B 0`. Given the PR's "report failures" intent, a zero-file deploy reported as success is indistinguishable from a real one. Consider tying `_VC_W32_COUNT==0` / `_VC_X64_COUNT==0` to a non-zero result, unless some toolsets legitimately have no launcher files.

rchiodo:
Solid hardening of the deploy script (directory creation, delayed-expansion fix, failure reporting). Two non-blocking warnings worth a look around the `pause`/exit ordering and the empty-source robocopy success case.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

PR #8559, opened 2026-06-19
Prevent conda activation hangs

bschnurr:
## Summary
- avoid joining conda activation through the VS UI thread while constructing launch environments
- add a 30-second timeout for conda activation and kill timed-out activation processes
- avoid holding the activation cache lock while waiting on external activation processes
- do not cache timeout results, so transient activation hangs do not poison the cache
- make the conda activation cache key hash code match its case-insensitive equality semantics
- add regression tests for activation timeout behavior and timeout cache behavior

## Root cause
Conda activation was invoked from GetFullEnvironment through UIThread.InvokeTaskSync with CancellationToken.None. The delegated work launched activate.bat and python.exe to capture the activated environment, then awaited process completion with no timeout while holding the activation cache lock. If activation or the child Python process never exited, devenv.exe could remain stuck in the UI-thread join path indefinitely, and other activation requests could block behind the cache lock.

## Behavior after this change
Activation is still exposed through the existing synchronous environment-construction path, but the external activation wait is moved out of the VS UI-thread invocation path and is bounded by a timeout. On timeout, PTVS logs a Trace warning, kills the activation process, returns an empty activation environment for that attempt, and leaves the cache unmodified so a later activation can retry.

## Tradeoffs / follow-ups
- If GetFullEnvironment is called on the UI thread, the UI may still be blocked until activation completes or the 30-second timeout expires; this is not a full async UI/progress refactor.
- Very slow conda installations may time out and launch without activation-provided environment variables.
- Concurrent cold-cache requests for the same environment may now run duplicate activation attempts because the global cache lock is no longer held during process execution.
- proc.Kill() terminates the started activation process; if child processes survive in some scenarios, process-tree termination may be needed later.
- Timeout diagnostics currently use Trace.TraceWarning only; Activity Log or telemetry may be useful follow-ups.

## Validation
- msbuild Python\\dirs.proj /p:VSTarget=18.0 /p:DeployExtension=true /p:RegisterOutputPackage=true
- vstest.console.exe BuildOutput\\Debug18.0\\raw\\binaries\\PythonToolsTests.dll /Tests:PythonToolsTests.CondaInterpreterFactoryTests.CondaActivationTimesOut,PythonToolsTests.CondaInterpreterFactoryTests.CondaActivationTimeoutIsNotCached /Logger:Console;Verbosity=minimal

### DIFF
--- a/Python/Product/Common/Infrastructure/ProcessOutput.cs
+++ b/Python/Product/Common/Infrastructure/ProcessOutput.cs
@@ -720,6 +720,7 @@ namespace Microsoft.PythonTools.Infrastructure {
             if (_process != null) {
                 bool exited = _process.WaitForExit((int)timeout.TotalMilliseconds);
                 if (exited) {
+                    _process.WaitForExit();
                     // Should have already been called, in which case this is a no-op
                     OnExited(this, EventArgs.Empty);
                 }
--- a/Python/Product/Common/Infrastructure/ProcessOutput.cs
+++ b/Python/Product/Common/Infrastructure/ProcessOutput.cs
@@ -720,7 +720,6 @@ namespace Microsoft.PythonTools.Infrastructure {
             if (_process != null) {
                 bool exited = _process.WaitForExit((int)timeout.TotalMilliseconds);
                 if (exited) {
-                    _process.WaitForExit();
                     // Should have already been called, in which case this is a no-op
                     OnExited(this, EventArgs.Empty);
                 }
--- a/Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
+++ b/Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
@@ -20,11 +20,16 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Logging;
 using Microsoft.VisualStudioTools;
 
 namespace Microsoft.PythonTools.Interpreter {
     public class LaunchConfigurationUtils {
         public static Dictionary<string, string> GetFullEnvironment(LaunchConfiguration config, IServiceProvider serviceProvider, UIThreadBase uiThread) {
+            return GetFullEnvironment(config, serviceProvider, uiThread, null);
+        }
+
+        public static Dictionary<string, string> GetFullEnvironment(LaunchConfiguration config, IServiceProvider serviceProvider, UIThreadBase uiThread, IPythonToolsLogger logger) {
             if (config.Interpreter == null && config.InterpreterPath == null) {
                 throw new ArgumentException("Interpreter was invalid");
             }
@@ -49,7 +54,7 @@ namespace Microsoft.PythonTools.Interpreter {
                 var condaExe = CondaUtils.GetRootCondaExecutablePath(serviceProvider);
                 var prefixPath = config.Interpreter.GetPrefixPath();
                 if (File.Exists(condaExe) && Directory.Exists(prefixPath)) {
-                    var condaEnv = Task.Run(() => CondaUtils.GetActivationEnvironmentVariablesForPrefixAsync(condaExe, prefixPath)).WaitAndUnwrapExceptions();
+                    var condaEnv = Task.Run(() => CondaUtils.GetActivationEnvironmentVariablesForPrefixAsync(condaExe, prefixPath, logger)).WaitAndUnwrapExceptions();
                     baseEnv = PathUtils.MergeEnvironments(baseEnv.AsEnumerable<string, string>(), condaEnv, "Path", pathVar);
                 }
             }
--- a/Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
+++ b/Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
@@ -18,7 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.PythonTools.Infrastructure;
 using Microsoft.VisualStudioTools;
 
@@ -49,7 +49,7 @@ namespace Microsoft.PythonTools.Interpreter {
                 var condaExe = CondaUtils.GetRootCondaExecutablePath(serviceProvider);
                 var prefixPath = config.Interpreter.GetPrefixPath();
                 if (File.Exists(condaExe) && Directory.Exists(prefixPath)) {
-                    var condaEnv = uiThread.InvokeTaskSync(() => CondaUtils.GetActivationEnvironmentVariablesForPrefixAsync(condaExe, prefixPath), CancellationToken.None);
+                    var condaEnv = Task.Run(() => CondaUtils.GetActivationEnvironmentVariablesForPrefixAsync(condaExe, prefixPath)).WaitAndUnwrapExceptions();
                     baseEnv = PathUtils.MergeEnvironments(baseEnv.AsEnumerable<string, string>(), condaEnv, "Path", pathVar);
                 }
             }
--- a/Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
+++ b/Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
@@ -30,6 +30,7 @@ namespace Microsoft.PythonTools.Interpreter {
         }
 
         public static Dictionary<string, string> GetFullEnvironment(LaunchConfiguration config, IServiceProvider serviceProvider, UIThreadBase uiThread, IPythonToolsLogger logger) {
+            // uiThread is retained for API compatibility; conda activation no longer marshals to the UI thread.
             if (config.Interpreter == null && config.InterpreterPath == null) {
                 throw new ArgumentException("Interpreter was invalid");
             }
--- a/Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
+++ b/Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
@@ -50,9 +50,9 @@ namespace Microsoft.PythonTools.Interpreter {
             }
             baseEnv[pathVar] = string.Empty;
 
-            if (CondaUtils.IsCondaEnvironment(config.Interpreter.GetPrefixPath())) {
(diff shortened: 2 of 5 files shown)


Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
```diff
@@ -49,7 +54,7 @@ public static Dictionary<string, string> GetFullEnvironment(LaunchConfiguration
                 var condaExe = CondaUtils.GetRootCondaExecutablePath(serviceProvider);
                 var prefixPath = config.Interpreter.GetPrefixPath();
                 if (File.Exists(condaExe) && Directory.Exists(prefixPath)) {
-                    var condaEnv = uiThread.InvokeTaskSync(() => CondaUtils.GetActivationEnvironmentVariablesForPrefixAsync(condaExe, prefixPath), CancellationToken.None);
+                    var condaEnv = Task.Run(() => CondaUtils.GetActivationEnvironmentVariablesForPrefixAsync(condaExe, prefixPath, logger)).WaitAndUnwrapExceptions();
```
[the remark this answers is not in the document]
bschnurr: Fixed in 8f71ca47b by computing prefixPath with null-conditional access and only running conda activation when an interpreter configuration is available.

Python/Product/VSInterpreters/PackageManager/CondaUtils.cs
```diff
@@ -90,33 +93,86 @@ internal static Task<IEnumerable<KeyValuePair<string, string>>> GetActivationEnv
...
+            if (File.Exists(activateBat)) {
+                var args = prefixPath != null
+                    ? new[] { prefixPath, "&", "python.exe", "-c", PrintEnvironmentCode }
+                    : new[] { "&", "python.exe", "-c", PrintEnvironmentCode };
+
+                using (var proc = ProcessOutput.RunHiddenAndCapture(activateBat, args)) {
+                    var exited = await Task.Run(() => proc.Wait(timeout)).ConfigureAwait(false);
+                    if (!exited) {
+                        Trace.TraceWarning("Conda activation timed out: " + proc.Arguments);
+                        LogActivationTimeout(logger, prefixPath, timeout);
```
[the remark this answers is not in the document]
bschnurr: Fixed in 8f71ca47b by replacing the verbose proc.Arguments trace with a concise timeout warning that includes timeout duration, activate.bat path, prefix, and PID.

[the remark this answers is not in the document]
bschnurr: Follow-up in e3eff07b removes path details from the Trace warning. It now logs only timeout duration and PID, while telemetry remains unchanged.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [Python/Product/VSInterpreters/PackageManager/CondaUtils.cs:L1](https://github.com/microsoft/ptvs/blob/e3eff07b2ac68b79ce9e064d3c663e024e0d0356/Python/Product/VSInterpreters/PackageManager/CondaUtils.cs#L1).

📍 GetActivationEnvironmentVariablesForPrefixUncachedAsync
The success path switched from `await proc` to `await Task.Run(() => proc.Wait(timeout))`. `WaitForExit(Int32)` returning `true` does not guarantee async `OutputDataReceived` handlers have flushed; the not-yet-dispatched JSON env line can be discarded, so `ParseEnvironmentVariables` returns empty and that empty result gets cached (`ShouldCache=true`) for the process lifetime — a silent, sticky loss of activation env vars. Fix: after a successful (`exited==true`) wait, call the parameterless `proc.Wait()` to drain async stdout before reading `StandardOutputLines`/`ExitCode`.

Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
```diff
@@ -18,13 +18,18 @@
...
+using System.Threading.Tasks;
 using Microsoft.PythonTools.Infrastructure;
+using Microsoft.PythonTools.Logging;
 using Microsoft.VisualStudioTools;
 
 namespace Microsoft.PythonTools.Interpreter {
     public class LaunchConfigurationUtils {
         public static Dictionary<string, string> GetFullEnvironment(LaunchConfiguration config, IServiceProvider serviceProvider, UIThreadBase uiThread) {
+            return GetFullEnvironment(config, serviceProvider, uiThread, null);
+        }
```
rchiodo, line 30: 📍 The `uiThread` parameter on `GetFullEnvironment` is now dead — after switching to `Task.Run(...).WaitAndUnwrapExceptions()`, `uiThread` is no longer referenced, yet the signature (and `PythonToolsService`) still threads `UIThread` through. This misleads future readers into assuming UI-thread coordination still happens. Either remove the parameter (if no external caller depends on the public signature) or add a one-line comment noting it's retained for API compatibility and activation no longer marshals to the UI thread.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [Python/Product/VSInterpreters/PackageManager/CondaUtils.cs:L1](https://github.com/microsoft/ptvs/blob/e3eff07b2ac68b79ce9e064d3c663e024e0d0356/Python/Product/VSInterpreters/PackageManager/CondaUtils.cs#L1).

📍 proc.Kill on timeout
`proc.Kill()` terminates only the `activate.bat`/cmd process, not the `& python.exe` child it spawns, so a timeout can leak an orphaned `python.exe`. Acceptable to defer per the PR's stated follow-up, but process-tree termination should be tracked so the timeout path doesn't accumulate orphaned children.

Python/Tests/Core/CondaInterpreterFactoryTests.cs
```diff
@@ -76,6 +81,91 @@ public void CondaWatchEnvironmentsTxtWithCondafolderAndEnvTxt() {
...
+                File.WriteAllText(condaExe, string.Empty);
+                File.WriteAllText(activateBat, "@echo off\r\n:wait\r\ngoto wait\r\n");
+
+                var timedOutEnv = CondaUtils.GetActivationEnvironmentVariablesForPrefixAsync(condaExe, null, TimeSpan.FromMilliseconds(100)).WaitAndUnwrapExceptions();
+                Assert.AreEqual(0, timedOutEnv.Count());
+
+                File.WriteAllText(activateBat, "@echo off\r\necho !!!ENVIRONMENT MARKER!!!\r\necho {\"PTVS_TEST_ENV\":\"success\"}\r\n");
+
+                var env = CondaUtils.GetActivationEnvironmentVariablesForPrefixAsync(condaExe, null, TimeSpan.FromSeconds(5)).WaitAndUnwrapExceptions();
+
```
rchiodo, line 160: 📍 CondaActivationTimeoutIsNotCached
The success-path assertion echoes two tiny lines and exits immediately — output that is almost always already buffered — so it does not exercise the `WaitForExit(timeout)`-drops-output race. The suite can stay green while production intermittently loses (and caches) the environment. Consider a test that emits a large env payload immediately before exit; it would be timing-sensitive, which itself argues for the drain-after-wait fix.

rchiodo:
Main concern: after a successful timed wait, drain async stdout (parameterless `proc.Wait()`) before reading `StandardOutputLines`/`ExitCode`, otherwise a successful activation can intermittently parse empty and cache that empty result. The timeout/cache structure otherwise looks sound.

Python/Product/VSInterpreters/Interpreter/LaunchConfigurationUtils.cs
bschnurr: Fixed in 1d5c13b9 by adding the compatibility note: uiThread is retained for API compatibility and conda activation no longer marshals to the UI thread.

Python/Tests/Core/CondaInterpreterFactoryTests.cs
bschnurr: Fixed in 1d5c13b9. The successful timed wait now drains async stdout before parsing, and the regression test now emits a large JSON payload and verifies it is parsed instead of caching an empty environment.

bschnurr:
@rchiodo addressed the stdout drain concern in 1d5c13b9 by draining after a successful timed wait before reading output/ExitCode. Also strengthened `CondaActivationTimeoutIsNotCached` with a large env payload. Process-tree cleanup is tracked separately in #8560.

Python/Product/Common/Infrastructure/ProcessOutput.cs
```diff
@@ -720,6 +720,7 @@ public bool Wait(TimeSpan timeout) {
             if (_process != null) {
                 bool exited = _process.WaitForExit((int)timeout.TotalMilliseconds);
                 if (exited) {
+                    _process.WaitForExit();
```
rchiodo: 📍 Python/Product/Common/Infrastructure/ProcessOutput.cs:723
The drain fix here is correct and fixes the dropped-stdout race for every `Wait(timeout)` caller. One caveat: the new parameterless `_process.WaitForExit()` blocks until stdout EOF, so if a process exits within the timeout but leaves a grandchild that inherited the redirected stdout handle, this can block unboundedly — reintroducing the timeout-defeating hang class into shared infrastructure. Conda's command spawns no such surviving grandchild, so conda is safe, but consider bounding the drain (or documenting that `Wait(timeout)` no longer guarantees a hard upper bound when stdout is redirected).

rchiodo (approved):
Approved via Review Center.

Python/Product/Common/Infrastructure/ProcessOutput.cs
bschnurr: Fixed in 3ebd1c3f by reverting the shared ProcessOutput.Wait(timeout) drain, so the helper keeps its bounded-wait behavior. The stdout drain remains conda-specific after a successful timed wait, where this activation command is safe.

rchiodo:
Looks good — the hang fix (move conda activation off the UI-thread join path, bound it with a 30s timeout, kill timed-out processes, and skip caching timeout results) is sound and well-tested. The remaining first-pass notes were all non-blocking nits, out-of-scope follow-ups, or premised on a ProcessOutput.cs change that isn't in this PR, so none were posted.

rchiodo (approved):
Approved via Review Center.

rchiodo:
Looks good — the timeout + don't-cache-on-timeout approach is a solid, targeted fix and the regression tests cover the key behaviors. Approving.

rchiodo (approved):
Approved via Review Center.

PR #8563, opened 2026-06-25
Bschnurr/fix pylance project env update

bschnurr:
#### PR Classification
Code cleanup and test robustness improvement.

#### PR Summary
This pull request removes deferred settings update logic in favor of immediate updates and improves test reliability for completion item detection.
- PythonLanguageClient.cs: Removed the deferred settings change timer and related logic, now triggering immediate settings updates; simplified cleanup and removed unnecessary awaits.
- PythonProjectNode.cs: Unsubscribes from CondaInterpreterDiscoveryCompleted during disposal.
- IntellisenseTests.cs: Enhanced the test to robustly wait for the "executable" completion item and improved event handler cleanup.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
@@ -99,12 +99,10 @@ namespace Microsoft.PythonTools.LanguageServerClient {
         private FileWatcher.Listener _fileListener;
         private static TaskCompletionSource<int> _readyTcs = new TaskCompletionSource<int>();
         private bool _loaded = false;
-        private Timer _deferredSettingsChangedTimer;
-        // Set by the dispose action before any other cleanup; consulted by the
-        // deferred timer callback, TriggerWorkspaceUpdateConfig and GetSettings so
-        // a timer tick scheduled before disposal cannot dereference cleared state.
+        // Set by the dispose action before any other cleanup; consulted by
+        // TriggerWorkspaceUpdateConfig and GetSettings so teardown cannot
+        // dereference cleared state.
         private volatile bool _disposed;
-        private const int _defaultSettingsDelayMS = 5000;
 
         public PythonLanguageClient() {
             _disposables = new Common.Core.Disposables.DisposableBag(GetType().Name);
@@ -140,19 +138,6 @@ namespace Microsoft.PythonTools.LanguageServerClient {
             await JoinableTaskContext.Factory.SwitchToMainThreadAsync();
             CreateClientContexts();
 
-            _deferredSettingsChangedTimer = new Timer(state => {
-                // Guard the callback body: a tick scheduled before disposal can fire
-                // after the language client has been torn down.
-                if (_disposed) {
-                    return;
-                }
-                try {
-                    TriggerWorkspaceUpdateConfig();
-                } catch (NullReferenceException) {
-                    // Field was nulled or service disappeared during teardown.
-                } catch (ObjectDisposedException) {
-                }
-            }, state: null, Timeout.Infinite, Timeout.Infinite);
             _analysisOptions = Site.GetPythonToolsService().AnalysisOptions;
             _advancedEditorOptions = Site.GetPythonToolsService().AdvancedEditorOptions;
             _analysisOptions.Changed += OnSettingsChanged;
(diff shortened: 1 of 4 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -514,11 +490,7 @@ private void OnInterpreterChanged(object sender, EventArgs e) {
         }
 
         private void OnSettingsChanged(object sender, EventArgs e) {
-            try {
-                System.Diagnostics.Debug.WriteLine("Settings Changed");
-                _deferredSettingsChangedTimer.Change(_defaultSettingsDelayMS, Timeout.Infinite);
-            } catch (ObjectDisposedException) {
-            }
+            TriggerWorkspaceUpdateConfig().DoNotWait();
         }
```
rchiodo, line 498: 📍 Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs:492
`OnSettingsChanged` now runs `TriggerWorkspaceUpdateConfig()` without the defensive try/catch that the deleted timer callback had. That callback explicitly caught `NullReferenceException`/`ObjectDisposedException` with the comment "Field was nulled or service disappeared during teardown" — meaning this failure was observed in practice. While `TriggerWorkspaceUpdateConfig` self-guards on `_disposed || _rpc == null`, that guard does not stop `GetSettings()` from dereferencing a disposed VS option/service mid-teardown, and the exception could now propagate into VS's options-change dispatch. Consider re-adding the defensive `try/catch (NullReferenceException)/(ObjectDisposedException)` around the call.

Python/Tests/Core.UI/IntellisenseTests.cs
```diff
@@ -44,8 +46,18 @@ public void AutoComplete(PythonVisualStudioApp app) {
...
-                    completionTask.TrySetResult(e.CompletionSession);
+                    session = e.CompletionSession;
+                    itemsUpdatedHandler = (sender, args) => {
+                        items = args.Items.Items.ToArray();
+                        if (items.Any(i => i.InsertText == "executable")) {
+                            executableTask.TrySetResult(items);
+                        }
+                    };
+                    session.ItemsUpdated += itemsUpdatedHandler;
+                    completionTask.TrySetResult(session);
```
rchiodo, line 64: 📍 Python/Tests/Core.UI/IntellisenseTests.cs:49
The handler only subscribes to `session.ItemsUpdated` and waits for a future event containing `executable`. If the completion set already contains `executable` when `CompletionTriggered` fires (fast/cached path) and no further `ItemsUpdated` is raised, `executableTask` never completes and `Wait(60000)` times out. The old synchronous `GetComputedItems(...)` could not miss already-present items. Subscribe-then-check: after subscribing, inspect the current computed items once and `TrySetResult` if `executable` is already present, closing the race in both directions.

rchiodo:
Overall the cleanup is sound. Two non-blocking items worth a look: defensive exception handling lost in OnSettingsChanged during teardown, and a potential race in the updated IntellisenseTests if 'executable' is already present when completion triggers.

rchiodo (approved):
Approved via Review Center.

rchiodo:
Overall this cleanup looks good. Two robustness follow-ups worth a look: (1) the now-immediate settings update lost the defensive try/catch the old timer callback had for teardown, and (2) the new test only waits on ItemsUpdated and can miss the case where 'executable' is already present when completion triggers.

rchiodo (approved):
Approved via Review Center.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
bschnurr: Fixed in 1fa86e18f by restoring the defensive `NullReferenceException`/`ObjectDisposedException` handling around the immediate settings update.

Python/Tests/Core.UI/IntellisenseTests.cs
bschnurr: Fixed in 1fa86e18f by subscribing to `ItemsUpdated` first and then checking `GetComputedItems(...)` immediately, so the test handles both already-present and later-updated completion items.

rchiodo (approved):
Approved via Review Center.

PR #8565, opened 2026-06-30
Fix VSIX payload signing checks

bschnurr:
## Summary
- Authenticode-sign `Newtonsoft.Json.dll` for the Core VSIX payload while skipping strong-name signing
- make the JavaScript item template non-empty so the existing template signing flow signs it instead of skipping it as a zero-byte placeholder
- keep the existing SWIX template source paths and MicroBuild signing flow unchanged

## Validation
- `MSBuild.exe Python\Setup\signlayout.proj /t:_StageTemplateFiles`;_AddTemplateFiles`;ListFiles /p:SignType=none /v:minimal /nologo`
- `MSBuild.exe Python\Setup\swix\Packages\Microsoft.PythonTools.Web.Templates.Vsix.swixproj /t:_GeneratePackage /p:SignType=none /v:minimal /nologo`
- `MSBuild.exe Python\Setup\swix\Packages\Microsoft.PythonTools.Django.Templates.Vsix.swixproj /t:_GeneratePackage /p:SignType=none /p:IncludeDjango=True /v:minimal /nologo`

Fixes microsoft/pyrx#8701

### DIFF
--- a/Python/Setup/signlayout.proj
+++ b/Python/Setup/signlayout.proj
@@ -110,8 +110,6 @@
   <Target Name="_AddTemplateFiles" BeforeTargets="_PreserveUnsigned;ListFiles">
     <ItemGroup>
       <TemplateFilesToSign Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
-      <_EmptyTemplateFiles Include="@(TemplateFilesToSign)"
-                           Condition="'$([System.IO.File]::ReadAllText(%(TemplateFilesToSign.FullPath)))' == ''" />
       <FilesToSign Include="@(TemplateFilesToSign)">
         <Authenticode>3PartyScriptsSHA2</Authenticode>
         <StrongName></StrongName>
@@ -119,8 +117,11 @@
         <UnsignedPath>$(UnsignedOutputPath)templates\%(RecursiveDir)%(Filename)%(Extension)</UnsignedPath>
       </FilesToSign>
     </ItemGroup>
-    <Error Condition="'@(_EmptyTemplateFiles)' != ''"
-           Text="Template JS/TTF files must not be empty because signtool cannot sign zero-byte files and unsigned files fail VS insertion SignCheck: @(_EmptyTemplateFiles)" />
+    <WriteLinesToFile File="$(IntermediateOutputPath)template-files-to-sign.txt"
+                      Lines="@(TemplateFilesToSign->'%(FullPath)')"
+                      Overwrite="true" />
+    <Exec Condition="'@(TemplateFilesToSign)' != ''"
+          Command="powershell -NoProfile -ExecutionPolicy Bypass -Command &quot;$files = Get-Content -LiteralPath '$(IntermediateOutputPath)template-files-to-sign.txt'; $empty = @($files | Where-Object { (Get-Item -LiteralPath $_).Length -eq 0 }); if ($empty.Count -gt 0) { Write-Error ('Template JS/TTF files must not be empty because signtool cannot sign zero-byte files and unsigned files fail VS insertion SignCheck: ' + ($empty -join '; ')); exit 1 }&quot;" />
   </Target>
 
   <Target Name="_PreserveUnsigned" Inputs="%(FilesToSign.FullPath)" Outputs="%(FilesToSign.UnsignedPath)" BeforeTargets="SignFiles">
@@ -165,15 +166,24 @@
 
   <Target Name="_ValidateSignedTemplateFiles"
           Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'real' or '$(SignType)' == 'Test' or '$(SignType)' == 'test'">
+    <PropertyGroup>
+      <_SignToolCommand Condition="'$(_SignToolCommand)' == '' and '$(WindowsSdkBinPath)' != ''">$(WindowsSdkBinPath)signtool.exe</_SignToolCommand>
+      <_SignToolCommand Condition="'$(_SignToolCommand)' == ''">signtool</_SignToolCommand>
+    </PropertyGroup>
     <ItemGroup>
-      <_TemplateFilesToValidate Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
+      <_ScriptTemplateFilesToValidate Include="$(BinariesOutputPath)templates\**\*.js" />
+      <_FontTemplateFilesToValidate Include="$(BinariesOutputPath)templates\**\*.ttf" />
+      <_TemplateFilesToValidate Include="@(_ScriptTemplateFilesToValidate);@(_FontTemplateFilesToValidate)" />
     </ItemGroup>
     <Error Condition="'@(_TemplateFilesToValidate)' == ''"
            Text="No template JS/TTF files were staged for signing." />
-    <WriteLinesToFile File="$(IntermediateOutputPath)template-files-to-validate.txt"
-                      Lines="@(_TemplateFilesToValidate->'%(FullPath)')"
+    <WriteLinesToFile File="$(IntermediateOutputPath)script-template-files-to-validate.txt"
+                      Lines="@(_ScriptTemplateFilesToValidate->'%(FullPath)')"
                       Overwrite="true" />
-    <Exec Command="powershell -NoProfile -ExecutionPolicy Bypass -Command &quot;$files = Get-Content -LiteralPath '$(IntermediateOutputPath)template-files-to-validate.txt'; $unsigned = @($files | Where-Object { -not (Get-AuthenticodeSignature -LiteralPath $_).SignerCertificate }); if ($unsigned.Count -gt 0) { Write-Error ('Template files were not signed: ' + ($unsigned -join '; ')); exit 1 }&quot;" />
+    <Exec Condition="'@(_ScriptTemplateFilesToValidate)' != ''"
+          Command="powershell -NoProfile -ExecutionPolicy Bypass -Command &quot;$files = Get-Content -LiteralPath '$(IntermediateOutputPath)script-template-files-to-validate.txt'; $invalid = @($files | Where-Object { (Get-AuthenticodeSignature -LiteralPath $_).Status -ne 'Valid' }); if ($invalid.Count -gt 0) { Write-Error ('Template JS files were not validly signed: ' + ($invalid -join '; ')); exit 1 }&quot;" />
+    <Exec Condition="'@(_FontTemplateFilesToValidate)' != ''"
+          Command="&quot;$(_SignToolCommand)&quot; verify /pa /all @(_FontTemplateFilesToValidate->'&quot;%(FullPath)&quot;', ' ')" />
   </Target>
 
 
--- a/Python/Setup/signlayout.proj
+++ b/Python/Setup/signlayout.proj
@@ -49,6 +49,9 @@
     <ManagedFiles Include="
         Microsoft.PythonTools.Debugger.VCLauncher.dll;
         " Condition="$(IncludeVCDebugLauncher)" />
+    <ManagedFiles Include="Newtonsoft.Json.dll">
+      <SkipStrongName>true</SkipStrongName>
+    </ManagedFiles>
 
     <UnmanagedFiles Include="
         Microsoft.PythonTools.Debugger.Helper.x64.dll;
@@ -107,10 +110,7 @@
   <Target Name="_AddTemplateFiles" BeforeTargets="_PreserveUnsigned;ListFiles">
     <ItemGroup>
       <TemplateFilesToSign Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
-      <!-- Drop zero-byte files (e.g. the empty 'Add New Item -> JavaScript' template placeholder).
-           signtool cannot map a 0-byte file (CreateFileMapping fails with ERROR_FILE_INVALID 0x3EE),
-           which would otherwise break the signing step. -->
-      <TemplateFilesToSign Remove="@(TemplateFilesToSign)"
+      <_EmptyTemplateFiles Include="@(TemplateFilesToSign)"
                            Condition="'$([System.IO.File]::ReadAllText(%(TemplateFilesToSign.FullPath)))' == ''" />
       <FilesToSign Include="@(TemplateFilesToSign)">
         <Authenticode>3PartyScriptsSHA2</Authenticode>
@@ -119,6 +119,8 @@
         <UnsignedPath>$(UnsignedOutputPath)templates\%(RecursiveDir)%(Filename)%(Extension)</UnsignedPath>
       </FilesToSign>
     </ItemGroup>
+    <Error Condition="'@(_EmptyTemplateFiles)' != ''"
+           Text="Template JS/TTF files must not be empty because signtool cannot sign zero-byte files and unsigned files fail VS insertion SignCheck: @(_EmptyTemplateFiles)" />
   </Target>
 
   <Target Name="_PreserveUnsigned" Inputs="%(FilesToSign.FullPath)" Outputs="%(FilesToSign.UnsignedPath)" BeforeTargets="SignFiles">
@@ -161,9 +163,22 @@
           OverwriteReadOnlyFiles="true" />
   </Target>
 
+  <Target Name="_ValidateSignedTemplateFiles"
+          Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'real' or '$(SignType)' == 'Test' or '$(SignType)' == 'test'">
+    <ItemGroup>
+      <_TemplateFilesToValidate Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
+    </ItemGroup>
+    <Error Condition="'@(_TemplateFilesToValidate)' == ''"
+           Text="No template JS/TTF files were staged for signing." />
+    <WriteLinesToFile File="$(IntermediateOutputPath)template-files-to-validate.txt"
+                      Lines="@(_TemplateFilesToValidate->'%(FullPath)')"
+                      Overwrite="true" />
+    <Exec Command="powershell -NoProfile -ExecutionPolicy Bypass -Command &quot;$files = Get-Content -LiteralPath '$(IntermediateOutputPath)template-files-to-validate.txt'; $unsigned = @($files | Where-Object { -not (Get-AuthenticodeSignature -LiteralPath $_).SignerCertificate }); if ($unsigned.Count -gt 0) { Write-Error ('Template files were not signed: ' + ($unsigned -join '; ')); exit 1 }&quot;" />
+  </Target>
+
 
   <Target Name="_UpdateLayoutWithSignedBinaries"
-          DependsOnTargets="_GetBinariesInLayout;_GetJsFilesInLayout;_CopySignedTemplateFiles"
+          DependsOnTargets="_GetBinariesInLayout;_GetJsFilesInLayout;_ValidateSignedTemplateFiles;_CopySignedTemplateFiles"
           AfterTargets="SignFiles">
     <Copy SourceFiles="%(_BinariesToCopy.FullPath)" DestinationFiles="%(_BinariesToCopy.TargetPath)" />
   </Target>
--- a/Python/Setup/signlayout.proj
+++ b/Python/Setup/signlayout.proj
@@ -110,6 +110,11 @@
   <Target Name="_AddTemplateFiles" BeforeTargets="_PreserveUnsigned;ListFiles">
     <ItemGroup>
       <TemplateFilesToSign Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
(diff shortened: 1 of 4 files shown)


Python/Setup/signlayout.proj
```diff
@@ -161,9 +163,22 @@
...
+      <_TemplateFilesToValidate Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
+    </ItemGroup>
+    <Error Condition="'@(_TemplateFilesToValidate)' == ''"
+           Text="No template JS/TTF files were staged for signing." />
+    <WriteLinesToFile File="$(IntermediateOutputPath)template-files-to-validate.txt"
+                      Lines="@(_TemplateFilesToValidate->'%(FullPath)')"
+                      Overwrite="true" />
+    <Exec Command="powershell -NoProfile -ExecutionPolicy Bypass -Command &quot;$files = Get-Content -LiteralPath '$(IntermediateOutputPath)template-files-to-validate.txt'; $unsigned = @($files | Where-Object { -not (Get-AuthenticodeSignature -LiteralPath $_).SignerCertificate }); if ($unsigned.Count -gt 0) { Write-Error ('Template files were not signed: ' + ($unsigned -join '; ')); exit 1 }&quot;" />
+  </Target>
+
```
rchiodo, line 167: 📍 Python/Setup/signlayout.proj:166
**`_ValidateSignedTemplateFiles` will falsely fail any signed build containing a `.ttf` template.** The Skeptic empirically verified on Windows that `Get-AuthenticodeSignature` never returns a `SignerCertificate` for `.ttf` files (catalog-signed system fonts return `Status=UnknownError`, null cert), whereas it works for `.js`/PE files. Since `_TemplateFilesToValidate` globs `*.ttf`, the `-not (...).SignerCertificate` test is always true for fonts, producing `Template files were not signed: ...ttf` and `exit 1` on every Real/Test build — defeating the PR's goal. Either scope the `SignerCertificate` check to `*.js` and validate `.ttf` via the catalog / `signtool verify /pa /all`, or drop `.ttf` from this check if fonts are signed through a different mechanism. The signing step and the validation step must agree on how a `.ttf` proves it is signed.

Python/Setup/signlayout.proj
```diff
@@ -119,6 +119,8 @@
         <UnsignedPath>$(UnsignedOutputPath)templates\%(RecursiveDir)%(Filename)%(Extension)</UnsignedPath>
       </FilesToSign>
     </ItemGroup>
+    <Error Condition="'@(_EmptyTemplateFiles)' != ''"
+           Text="Template JS/TTF files must not be empty because signtool cannot sign zero-byte files and unsigned files fail VS insertion SignCheck: @(_EmptyTemplateFiles)" />
   </Target>
```
rchiodo, line 125: 📍 Python/Setup/signlayout.proj:121
The new `<Error>` on `_EmptyTemplateFiles` is global and unconditional, but only `JavaScript.js` was de-emptied. The old code *removed* empty placeholders and continued; this now hard-fails the build for any other zero-byte `*.js`/`*.ttf` under `Python\Templates\**`. The original comment noted empty placeholders are a known pattern, so please confirm `JavaScript.js` was the only one (e.g. `git ls-files -z 'Python/Templates/**/*.js' 'Python/Templates/**/*.ttf'` filtered to length 0) before relying on this as a tripwire. The fail-fast intent is good; the risk is purely whether other legitimate empty placeholders exist.

Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
```diff
@@ -8,10 +8,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageSearchPaths Include="$(BinariesOutputPath)templates\Web" />
     <PackageSearchPaths Include="$(SourcePath)" />
```
rchiodo, line 11: 📍 Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj:11
The entire packaging fix hinges on `PackageSearchPaths` being first-match-wins so the signed `$(BinariesOutputPath)templates\Web` copy is preferred over `$(SourcePath)`. Reviewers flagged that SWIX search-path precedence is not determinable from this diff. If SWIX is last-match or merge-all, the unsigned source copy could win (silent no-op fix), re-introducing the exact bug. Please confirm SWIX resolution order; the same applies to the Django swixproj. Note that `_CopySignedTemplateFiles` already copies signed bytes back into the source tree, so the two paths may resolve to identical bytes — worth confirming this change is load-bearing rather than redundant.

Python/Setup/signlayout.proj
```diff
@@ -161,9 +163,22 @@
...
+      <_TemplateFilesToValidate Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
+    </ItemGroup>
+    <Error Condition="'@(_TemplateFilesToValidate)' == ''"
+           Text="No template JS/TTF files were staged for signing." />
+    <WriteLinesToFile File="$(IntermediateOutputPath)template-files-to-validate.txt"
+                      Lines="@(_TemplateFilesToValidate->'%(FullPath)')"
+                      Overwrite="true" />
+    <Exec Command="powershell -NoProfile -ExecutionPolicy Bypass -Command &quot;$files = Get-Content -LiteralPath '$(IntermediateOutputPath)template-files-to-validate.txt'; $unsigned = @($files | Where-Object { -not (Get-AuthenticodeSignature -LiteralPath $_).SignerCertificate }); if ($unsigned.Count -gt 0) { Write-Error ('Template files were not signed: ' + ($unsigned -join '; ')); exit 1 }&quot;" />
+  </Target>
+
```
rchiodo: 📍 Python/Setup/signlayout.proj:178
Validation checks signature *presence*, not *validity*: `Get-AuthenticodeSignature ... .SignerCertificate` is non-null for `HashMismatch`/`NotTrusted` signatures too. It catches the primary failure mode (never signed) but would pass a corrupt/tampered signature. Consider tightening to `.Status -eq 'Valid'`, weighing the stricter cert-trust requirement this places on the build agent.

rchiodo:
Core fix direction is reasonable, but please reconcile how `.ttf` files are signed vs. how the new validation target proves they are signed — as written the validation looks like it will hard-fail signed builds. Also confirm SWIX search-path precedence so the packaging change is actually load-bearing.

bschnurr:
Rethought and simplified in f428eb0e3.

Final PR diff is now intentionally small:

- `Newtonsoft.Json.dll` is still Authenticode-signed but skips strong-name signing.
- `JavaScript.js` is no longer zero-byte, so the existing template signing flow can include it instead of skipping it.

I removed the extra custom JS/TTF signature validation and reverted the SWIX project path changes. The existing setup flow already stages template JS/TTF files to `$(BinariesOutputPath)templates`, signs them, and copies the signed files back to `Python\Templates` before SWIX packaging runs, so changing SWIX `SourcePath` was the wrong layer.

Validated locally with VS MSBuild:

```powershell
MSBuild.exe Python\Setup\signlayout.proj /t:_StageTemplateFiles`;_AddTemplateFiles`;ListFiles /p:SignType=none /v:minimal /nologo
MSBuild.exe Python\Setup\swix\Packages\Microsoft.PythonTools.Web.Templates.Vsix.swixproj /t:_GeneratePackage /p:SignType=none /v:minimal /nologo
MSBuild.exe Python\Setup\swix\Packages\Microsoft.PythonTools.Django.Templates.Vsix.swixproj /t:_GeneratePackage /p:SignType=none /p:IncludeDjango=True /v:minimal /nologo
```

Python/Setup/signlayout.proj
bschnurr: Superseded by f428eb0e3. I removed `_ValidateSignedTemplateFiles` entirely, so `.ttf` files are no longer checked with `Get-AuthenticodeSignature` or any new custom validation. The PR now relies on the existing MicroBuild signing pipeline instead of adding a second validation path.

Python/Setup/signlayout.proj
bschnurr: Simplified in f428eb0e3. I checked the tracked template JS/TTF files and `JavaScript.js` was the zero-byte file causing this path. The existing zero-byte skip remains unchanged; the PR now fixes the actual payload by making `JavaScript.js` non-empty so it participates in the existing signing flow.

Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
bschnurr: Agreed; that SWIX change was the wrong direction. I reverted the Web/Django `.swixproj` changes in f428eb0e3. The existing flow signs staged template files and copies them back to `Python\Templates` before SWIX packaging, so the SWIX projects should continue using their source template paths.

Python/Setup/signlayout.proj
bschnurr: Superseded by f428eb0e3. I removed the custom `Get-AuthenticodeSignature` validation target instead of trying to tune it. The PR now avoids adding a second signature validation policy and leaves signing failure detection to the existing signing pipeline.

rchiodo:

Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="SetupProjectBefore.settings" />
 
   <PropertyGroup>
-    <SourcePath>$(BuildRoot)\Python\Templates\Web</SourcePath>
+    <SourcePath>$(BinariesOutputPath)templates\Web</SourcePath>
```
rchiodo: 📍 Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj:6
The Skeptic structurally confirmed (by reading `swix.generate.targets` + `signlayout.proj`) that the VSIX recursively packages everything under `SourcePath`, but `$(BinariesOutputPath)templates\` is populated by `_StageTemplateFiles`, which copies **only `*.js;*.ttf`**. Repointing here drops all `.vstemplate` manifests and content files, so the Web Templates VSIX ships incomplete and the templates won't instantiate in VS — strictly worse than the original unsigned-but-complete state. Either keep `SourcePath = $(BuildRoot)\Python\Templates\Web` and add an enforced dependency so `_CopySignedTemplateFiles` runs before this swixproj packs, or make `_StageTemplateFiles` mirror the entire templates tree (not just js/ttf). The Advocate defended this as the signing fix, but the staging-content analysis shows it breaks completeness.

[verified]

Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="SetupProjectBefore.settings" />
 
   <PropertyGroup>
-    <SourcePath>$(BuildRoot)\Python\Templates\Django</SourcePath>
+    <SourcePath>$(BinariesOutputPath)templates\Django</SourcePath>
```
rchiodo: 📍 Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj:6
Same structurally-confirmed defect as the Web swixproj: `$(BinariesOutputPath)templates\Django` only contains the signed `.js`/`.ttf` files staged by `_StageTemplateFiles`, not the full Django template tree. Packaging from here yields a VSIX missing its `.vstemplate` manifests and content files. Apply the same fix chosen for the Web project (enforced copy-back ordering with the original source path, or full-tree staging).

[verified]

Python/Setup/signlayout.proj
```diff
@@ -161,9 +164,31 @@
...
+           Text="No template JS/TTF files were staged for signing." />
+    <WriteLinesToFile File="$(IntermediateOutputPath)script-template-files-to-validate.txt"
+                      Lines="@(_ScriptTemplateFilesToValidate->'%(FullPath)')"
+                      Overwrite="true" />
+    <Exec Condition="'@(_ScriptTemplateFilesToValidate)' != ''"
+          Command="powershell -NoProfile -ExecutionPolicy Bypass -Command &quot;$files = Get-Content -LiteralPath '$(IntermediateOutputPath)script-template-files-to-validate.txt'; $invalid = @($files | Where-Object { (Get-AuthenticodeSignature -LiteralPath $_).Status -ne 'Valid' }); if ($invalid.Count -gt 0) { Write-Error ('Template JS files were not validly signed: ' + ($invalid -join '; ')); exit 1 }&quot;" />
+    <Exec Condition="'@(_FontTemplateFilesToValidate)' != ''"
+          Command="&quot;$(_SignToolCommand)&quot; verify /pa /all @(_FontTemplateFilesToValidate->'&quot;%(FullPath)&quot;', ' ')" />
+  </Target>
+
```
rchiodo: 📍 Python/Setup/signlayout.proj:167
[unverified] The Skeptic and Architect both flag that `_ValidateSignedTemplateFiles` runs for `SignType=Test`/`test` yet asserts `(Get-AuthenticodeSignature ...).Status -eq 'Valid'` for `.js` and uses `signtool verify /pa /all` (default production policy) for `.ttf`. Test-signing certs frequently don't chain to a trusted root on build agents, so a validly test-signed file can report `UnknownError`/`NotTrusted` and `exit 1` a legitimate Test build. Either restrict the strict validity check to `Real`, or use a test-root-tolerant verification (e.g. `signtool verify /pa` configured for the test root) for Test builds. Confirm on a real test-signing agent before merge.

[verified]

Python/Setup/signlayout.proj
```diff
@@ -107,18 +110,18 @@
...
         <StrongName></StrongName>
         <SignedPath>$(BinariesOutputPath)templates\%(RecursiveDir)%(Filename)%(Extension)</SignedPath>
         <UnsignedPath>$(UnsignedOutputPath)templates\%(RecursiveDir)%(Filename)%(Extension)</UnsignedPath>
       </FilesToSign>
     </ItemGroup>
+    <WriteLinesToFile File="$(IntermediateOutputPath)template-files-to-sign.txt"
+                      Lines="@(TemplateFilesToSign->'%(FullPath)')"
+                      Overwrite="true" />
+    <Exec Condition="'@(TemplateFilesToSign)' != ''"
+          Command="powershell -NoProfile -ExecutionPolicy Bypass -Command &quot;$files = Get-Content -LiteralPath '$(IntermediateOutputPath)template-files-to-sign.txt'; $empty = @($files | Where-Object { (Get-Item -LiteralPath $_).Length -eq 0 }); if ($empty.Count -gt 0) { Write-Error ('Template JS/TTF files must not be empty because signtool cannot sign zero-byte files and unsigned files fail VS insertion SignCheck: ' + ($empty -join '; ')); exit 1 }&quot;" />
```
rchiodo: 📍 Python/Setup/signlayout.proj:113
The new empty-file `Exec` now hard-fails the build for **any** zero-byte `.js`/`.ttf` under `$(BinariesOutputPath)templates\`, replacing the old silent-drop. Only `JavaScript.js` was given content. Both the Advocate and Skeptic note that if any other empty placeholder exists in the templates tree (e.g. a Django placeholder), this converts a previously-green build into a hard failure. Run a tree-wide search for zero-byte `.js`/`.ttf` under `Python\Templates\` to confirm `JavaScript.js` is the only one before merging.

[verified]

Python/Setup/signlayout.proj
```diff
@@ -161,9 +164,31 @@
           OverwriteReadOnlyFiles="true" />
   </Target>
 
+  <Target Name="_ValidateSignedTemplateFiles"
+          Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'real' or '$(SignType)' == 'Test' or '$(SignType)' == 'test'">
+    <PropertyGroup>
+      <_SignToolCommand Condition="'$(_SignToolCommand)' == '' and '$(WindowsSdkBinPath)' != ''">$(WindowsSdkBinPath)signtool.exe</_SignToolCommand>
+      <_SignToolCommand Condition="'$(_SignToolCommand)' == ''">signtool</_SignToolCommand>
```
rchiodo: 📍 Python/Setup/signlayout.proj:167
The Skeptic notes `$(WindowsSdkBinPath)signtool.exe` is an unguarded string concat assuming a trailing backslash. If `WindowsSdkBinPath` lacks the trailing `\`, this resolves to `...binsigntool.exe`, the font-verify `Exec` can't find signtool, and the build fails. Normalize the path (e.g. ensure a trailing separator) or use a path-combine.

[verified]

Python/Setup/signlayout.proj
```diff
@@ -49,6 +49,9 @@
     <ManagedFiles Include="
         Microsoft.PythonTools.Debugger.VCLauncher.dll;
         " Condition="$(IncludeVCDebugLauncher)" />
+    <ManagedFiles Include="Newtonsoft.Json.dll">
+      <SkipStrongName>true</SkipStrongName>
+    </ManagedFiles>
```
rchiodo, line 54: 📍 Python/Setup/signlayout.proj:52
The Architect raises a follow-up worth a one-line answer: now that the VSIX reads from `$(BinariesOutputPath)templates\`, does `_CopySignedTemplateFiles` (which copies signed files back into the `$(BuildRoot)\Python\Templates\` source tree) still have a consumer? If the VSIX was its only consumer, that target is now dead code mutating tracked source as a side effect and should be removed; if other packaging still reads `BuildRoot`, that's fine but should be made explicit.

[verified]

rchiodo:
Main blocker: the VSIX SourcePath repoint (Web + Django) packages from $(BinariesOutputPath)templates\, but staging only copies *.js/*.ttf — the .vstemplate manifests and other content files would be missing, producing broken template VSIXes. Please either keep the original source path with enforced signed copy-back ordering, or make staging mirror the full templates tree. A few lower-severity notes on the new signing/validation targets are also worth a look.

bschnurr:
Follow-up to the later review comments: f428eb0e3 intentionally backs out the SWIX `SourcePath` change and the custom JS/TTF validation target.

Current cumulative PR diff is only:

- `Newtonsoft.Json.dll` is included for Authenticode signing, with strong-name signing skipped.
- `JavaScript.js` is made non-empty so the existing template signing flow signs it instead of skipping it as a zero-byte placeholder.

That means the Web/Django template VSIX projects keep packaging from `$(BuildRoot)\Python\Templates\...`; `_CopySignedTemplateFiles` remains the mechanism that copies signed staged JS/TTF files back into that tree before SWIX packaging.

Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
bschnurr: Agreed. Reverted in f428eb0e3. The Web template `.swixproj` no longer points at `$(BinariesOutputPath)templates\Web`; it is back to `$(BuildRoot)\Python\Templates\Web`, so it continues packaging the full template tree.

Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj
bschnurr: Agreed. Reverted in f428eb0e3. The Django template `.swixproj` is back to `$(BuildRoot)\Python\Templates\Django`, so it continues packaging the full Django template tree.

Python/Setup/signlayout.proj
bschnurr: Superseded by f428eb0e3. `_ValidateSignedTemplateFiles` was removed, so there is no new `Test`/`Real` signature validation behavior in this PR.

Python/Setup/signlayout.proj
bschnurr: Superseded by f428eb0e3. The new hard-failing empty-file `Exec` was removed. The existing zero-byte skip remains, and the PR fixes the specific payload by making `JavaScript.js` non-empty so it participates in the existing signing flow.

Python/Setup/signlayout.proj
bschnurr: Superseded by f428eb0e3. The new signtool command was removed along with `_ValidateSignedTemplateFiles`, so there is no `WindowsSdkBinPath` concatenation in the final PR diff.

Python/Setup/signlayout.proj
bschnurr: Superseded by f428eb0e3. Since the SWIX source-path change was reverted, Web/Django template packaging still reads from `$(BuildRoot)\Python\Templates\...`; `_CopySignedTemplateFiles` remains the path that puts signed staged JS/TTF files back where SWIX packages from.

rchiodo, line 54: 📍 Python/Setup/signlayout.proj:52
The PR description and Validation section describe changes that are not in this diff: the swixproj `SourcePath` repoint, the `_EmptyTemplateFiles` hard-fail, and a `_ValidateSignedTemplateFiles` target. The current head is only two hunks (this `Newtonsoft.Json.dll` entry + `JavaScript.js`). Reconcile the description and Validation notes with what actually ships, and confirm whether pyrx#8701's scope (template VSIX payload signing) is fully addressed by this reduced changeset before merge.

rchiodo (approved):
Approved via Review Center.

rchiodo:

PR #8567, opened 2026-07-01
Fix property page content clipping at high OS text scaling (accessibility)

rchiodo:
## Summary

Project property pages did not reflow or scroll when the OS text size was increased (e.g. 200%), so labels, inputs, and controls were clipped with no way to reach them. This is a **MAS 1.4.4 "Resize Text"** accessibility violation (AzDO work item 3026378).

## Root cause

Each page's top-level `UserControl` had `AutoScroll` disabled while its main container (`tableLayout` / `tableLayoutPanel1`) was docked `Fill` — a WinForms combination that suppresses the scrollbar. Content taller than the fixed property-page area (as happens at high text scaling) was clipped with no scrollbar.

## Fix

- Enable `AutoScroll` on the page `UserControl`.
- Dock the main container `Top` (with `AutoSize`) so its true scaled height propagates upward and a vertical scrollbar appears on demand.
- Horizontal reflow already works via percent-width columns and anchored fields, so it is preserved.
- The WPF **Test** page's root `StackPanel` is wrapped in a `ScrollViewer` (vertical auto, horizontal disabled so the `*` column keeps reflowing).

## Pages fixed

Debug, General, Web, Django, Publish, and Test.

## Validation

- Builds clean locally (VS 2022 IntPreview / VSTarget 18.0): `PythonTools.csproj` and `Django.csproj` both 0 warnings / 0 errors.
- Manual verification at 200% OS text size recommended during review.

### DIFF
--- a/Python/Product/PythonTools/PythonTools/Project/PythonGeneralPropertyPageControl.Designer.cs
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonGeneralPropertyPageControl.Designer.cs
@@ -115,6 +115,7 @@ namespace Microsoft.PythonTools.Project {
             // 
             resources.ApplyResources(this, "$this");
             this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
+            this.AutoScroll = true;
             this.Controls.Add(this.tableLayoutPanel1);
             this.Name = "PythonGeneralPropertyPageControl";
             this._applicationGroup.ResumeLayout(false);
--- a/Python/Product/PythonTools/PythonTools/Project/PythonTestPropertyPageView.xaml
+++ b/Python/Product/PythonTools/PythonTools/Project/PythonTestPropertyPageView.xaml
@@ -183,54 +183,54 @@
     </UserControl.Resources>
 
     <ScrollViewer VerticalScrollBarVisibility="Auto" HorizontalScrollBarVisibility="Disabled">
-    <StackPanel Grid.IsSharedSizeScope="True">
-        <Grid>
-            <Grid.ColumnDefinitions>
-                <ColumnDefinition Width="Auto" SharedSizeGroup="LabelSizeGroup"/>
-                <ColumnDefinition Width="*"/>
-            </Grid.ColumnDefinitions>
-            <Grid.RowDefinitions>
-                <RowDefinition/>
-                <RowDefinition/>
-            </Grid.RowDefinitions>
-            <Label
-                Name="FrameworkLabel"
-                Grid.Row="0"
-                Grid.Column="0"
-                Content="{x:Static common:Strings.PythonTestPropertyPageFrameworkLabel}"/>
-            <ComboBox
-                Grid.Row="0"
-                Grid.Column="1"
-                MinWidth="100"
-                HorizontalAlignment="Left"
-                ItemsSource="{Binding Path=Frameworks}"
-                SelectedItem="{Binding Path=SelectedFramework}"
-                AutomationProperties.LabeledBy="{Binding ElementName=FrameworkLabel}"
-                AutomationProperties.HelpText="{x:Static common:Strings.PythonTestPropertyPageFrameworkHelpText}"/>
-        </Grid>
+        <StackPanel Grid.IsSharedSizeScope="True">
+            <Grid>
+                <Grid.ColumnDefinitions>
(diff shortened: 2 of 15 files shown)


rchiodo:
I tested this one manually to make sure it works. It fixes the font sizes on the property pages and makes the property windows scrollable.

heejaechang:

Python/Product/PythonTools/PythonTools/Project/PythonTestPropertyPageView.xaml
edvilme, line 1: Maybe fix identation

edvilme (approved):
LGTM. Left a linting comment, and am curious about the Fill -> Top change

Python/Product/PythonTools/PythonTools/Project/PythonGeneralPropertyPageControl.Designer.cs
```diff
@@ -115,6 +115,7 @@ private void InitializeComponent() {
             // 
             resources.ApplyResources(this, "$this");
             this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
+            this.AutoScroll = true;
```
heejaechang, line 118: This `UserControl` sets `AutoScroll=true` here while its `.resx` keeps `$this.AutoSize=True` (GrowAndShrink) — a contradictory WinForms combination. `AutoSize` grows the control to fit content (never scrolls), whereas `AutoScroll` only engages when the control is constrained smaller than its content. The Debug page is the only one without `$this.AutoSize`, which is the self-consistent scroll-container shape. Recommended fix: remove `$this.AutoSize`/`$this.AutoSizeMode` to match the Debug page's proven pattern, or explicitly verify a scrollbar (not clipping) appears at 200% OS text size. The same concern applies to the Web, Publish, and Django pages, which share this AutoSize+AutoScroll combination.

heejaechang:
Overall the accessibility fix looks sound and idiomatic. The one worthwhile follow-up is the AutoSize vs AutoScroll inconsistency: the Debug page uses a clean scroll-container shape (no `$this.AutoSize`) while General/Web/Publish/Django keep `$this.AutoSize=True` alongside the new `AutoScroll=true`. Please confirm those pages actually scroll (not grow-and-clip) at 200% OS text size, or align them with the Debug page's pattern. The remaining first-pass notes (font re-apply on live setting change, font ownership/clone, broad catch, cosmetic min-width scrollbar) are low/no-impact and not worth blocking on.

heejaechang (approved):
Approved via Review Center.

Python/Product/PythonTools/PythonTools/Project/PythonGeneralPropertyPageControl.Designer.cs
rchiodo: Good catch — this was the core issue. I removed `$this.AutoSize`/`$this.AutoSizeMode` from the General, Web, Publish, and Django page controls so the top-level UserControl is a pure scroll container (matching the Debug page, which never had `$this.AutoSize`). With AutoSize gone, the host sets the control's fixed size, the inner Dock=Top panel grows to its scaled content height, and AutoScroll shows a scrollbar. I also added the missing `tableLayoutPanel1.AutoSizeMode=GrowAndShrink` to the General page's inner panel so all four now match Debug exactly.

Python/Product/PythonTools/PythonTools/Project/PythonTestPropertyPageView.xaml
rchiodo: Fixed — re-indented the `StackPanel` block so it nests properly inside the `ScrollViewer`.

rchiodo:
@edvilme re: your question about the `Dock=Fill` → `Dock=Top` change on the inner panels —

A child docked `Fill` is stretched to exactly the parent's client area, so it can never be *taller* than the visible region. That means the parent's `AutoScroll` has nothing to scroll to and the content just gets clipped at large text sizes. Switching the inner panel to `Dock=Top` + `AutoSize=True` lets it report its true (scaled) height; when that exceeds the UserControl's client area, `AutoScroll` kicks in and shows a vertical scrollbar. This mirrors what the Debug page already did.

heejaechang (approved):
Approved via Review Center.

PR #8485, opened 2026-04-20
Sign Django and Web template JS/TTF files with 3PartyScriptsSHA2

jesuszarate:
## Summary

The VS signing scan flags unsigned files in two PTVS template payloads:
- `mspythontoolsdjangotemplatesvsix170260206neutral`: 15 files (14 JS + 1 TTF)
- `mspythontoolswebtemplatesvsix170260206neutral`: 45 files (42 JS + 3 TTF)

All are 3rd-party OSS files (jQuery 1.10.2, Bootstrap 3.x, Modernizr 2.6.2, Respond.js, Glyphicons) in the StarterDjangoProject and Bottle/Flask web project templates.

## Root Cause

`Python/Setup/signlayout.proj` signs Pylance JS via `_AddPylanceFiles` but does not cover the Django or Web template directories. The swixprojs source from `Python/Templates/Django/` and `Python/Templates/Web/` respectively, so these files flow into the VSIXes unsigned.

## Fix

Add 3 MSBuild targets to `signlayout.proj`:

1. **`_StageTemplateFiles`** - globs JS/TTF from `Python/Templates/` (preserving Django/Web subdirectory structure) and copies to a signing staging area
2. **`_AddTemplateFiles`** - registers them for MicroBuild signing with `3PartyScriptsSHA2` certificate
3. **`_CopySignedTemplateFiles`** - copies signed files back to the source tree for SWIX packaging

## Verification

After merge + insertion, both payloads should show 0 unsigned files in the VS signing scan (60 files resolved total).

## Related

- AzDO bug [2949498](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2949498) (Django templates - 15 files)
- AzDO bug [2951790](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2951790) (Web templates - 45 files)

PR #8570, opened 2026-07-03
Sign staged Web and Django template fonts

bschnurr:
## Summary
- follow-up to microsoft/PTVS#8485, which added staging/signing for Django/Web template `.js` and `.ttf` files
- stage full Web/Django template package trees so their SWIX VSIX payloads can be generated from the staged signing area
- keep actual signing scoped to staged `.js` and `.ttf` files
- make Web/Django SWIX packages prefer staged template trees when available

## Why
PR #8485 added the signing list and copied signed files back to `Python\Templates` before SWIX packaging. The latest SignCheck failure still reports the four Glyphicons `.ttf` files inside the final Web/Django template VSIX payloads, so this PR removes the dependence on source-tree copy-back for those packages and packages Web/Django directly from the staged tree used by signing.

## Validation
- `MSBuild.exe Python\Setup\signlayout.proj /t:ListFiles /p:VSTarget=18.0 /p:Configuration=Release /p:SignType=none /p:BUILD_BINARIESDIRECTORY=C:\dev\PTVS\BuildOutput\validation /v:minimal /nologo`
- `MSBuild.exe Python\Setup\swix\Packages\Microsoft.PythonTools.Web.Templates.Vsix.swixproj /t:_GenerateSwrMetadata;_GenerateSwrFromSources ...`
- `MSBuild.exe Python\Setup\swix\Packages\Microsoft.PythonTools.Django.Templates.Vsix.swixproj /t:_GenerateSwRMetadata;_GenerateSwrFromSources ...`
- confirmed the four Glyphicons `.ttf` files are in `FilesToSign` and generated Web/Django SWR metadata with no duplicate staged Glyphicons paths

Real signing/SignCheck still requires the pipeline signing environment.

### DIFF
--- a/Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj
+++ b/Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj
@@ -3,7 +3,9 @@
   <Import Project="SetupProjectBefore.settings" />
 
   <PropertyGroup>
-    <SourcePath>$(BuildRoot)\Python\Templates\Django</SourcePath>
+    <!-- Prefer the staged template tree created by signlayout.proj so packaged VSIX contents use signed files. -->
+    <SourcePath Condition="'$(SourcePath)' == '' and Exists('$(BinariesOutputPath)templates\Django')">$(BinariesOutputPath)templates\Django</SourcePath>
+    <SourcePath Condition="'$(SourcePath)' == ''">$(BuildRoot)\Python\Templates\Django</SourcePath>
     <BaseInstall>InstallDir:\Common7\IDE</BaseInstall>
   </PropertyGroup>
 
--- a/Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
+++ b/Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
@@ -3,7 +3,9 @@
   <Import Project="SetupProjectBefore.settings" />
 
   <PropertyGroup>
-    <SourcePath>$(BuildRoot)\Python\Templates\Web</SourcePath>
+    <!-- Prefer the staged template tree created by signlayout.proj so packaged VSIX contents use signed files. -->
+    <SourcePath Condition="'$(SourcePath)' == '' and Exists('$(BinariesOutputPath)templates\Web')">$(BinariesOutputPath)templates\Web</SourcePath>
+    <SourcePath Condition="'$(SourcePath)' == ''">$(BuildRoot)\Python\Templates\Web</SourcePath>
     <BaseInstall>InstallDir:\Common7\IDE</BaseInstall>
   </PropertyGroup>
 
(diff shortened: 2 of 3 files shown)


rchiodo:

Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj
```diff
@@ -3,7 +3,9 @@
   <Import Project="SetupProjectBefore.settings" />
 
   <PropertyGroup>
-    <SourcePath>$(BuildRoot)\Python\Templates\Django</SourcePath>
+    <!-- Prefer the staged template tree created by signlayout.proj so packaged VSIX contents use signed files. -->
+    <SourcePath Condition="'$(SourcePath)' == '' and Exists('$(BinariesOutputPath)templates\Django')">$(BinariesOutputPath)templates\Django</SourcePath>
+    <SourcePath Condition="'$(SourcePath)' == ''">$(BuildRoot)\Python\Templates\Django</SourcePath>
```
rchiodo, line 8: The staged-tree preference relies on `Exists(...)`, which becomes true after *staging*, not after *signing*. If this `.swixproj` packages before the signing step completes (or the staged tree is only partially populated), the second condition silently falls back to the unsigned `$(BuildRoot)\Python\Templates\Django`, reproducing the exact SignCheck failure with no error. Consider gating on the signing step — e.g. emit an `<Error>` when `SignType != none` and the staged tree is missing — and confirm the pipeline guarantees signing precedes packaging.

Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
```diff
@@ -3,7 +3,9 @@
   <Import Project="SetupProjectBefore.settings" />
 
   <PropertyGroup>
-    <SourcePath>$(BuildRoot)\Python\Templates\Web</SourcePath>
+    <!-- Prefer the staged template tree created by signlayout.proj so packaged VSIX contents use signed files. -->
+    <SourcePath Condition="'$(SourcePath)' == '' and Exists('$(BinariesOutputPath)templates\Web')">$(BinariesOutputPath)templates\Web</SourcePath>
+    <SourcePath Condition="'$(SourcePath)' == ''">$(BuildRoot)\Python\Templates\Web</SourcePath>
```
rchiodo, line 8: Same silent-fallback concern as the Django swixproj: `Exists()` only proves the directory was staged, not that its `.js`/`.ttf` files are signed. A partial or out-of-order sign leaves a present-but-unsigned staged tree that this project will happily package. Prefer gating on the signing step rather than directory presence, and confirm signing runs before Web packaging.

rchiodo:
Overall the staging + staged-tree packaging approach looks correct. The one thing worth double-checking before merge: the swixproj projects prefer the staged tree based on directory existence, which only proves staging ran — not that signing completed. Please confirm the pipeline guarantees signing precedes Web/Django packaging so the fallback to the unsigned source tree can't silently re-introduce the SignCheck failure.

rchiodo (approved):
Approved via Review Center.

PR #8571, opened 2026-07-03
Fix template font signing source

bschnurr:
## Summary
- keep template `.ttf` files in the `3PartyScriptsSHA2` signing bucket, matching DevDiv examples for third-party font/script assets
- split template script and font signing items so zero-byte JS filtering does not apply to fonts
- require signed Web/Django template VSIX builds to package from `$(BinariesOutputPath)templates\...`, the staged tree produced by `signlayout.proj`

## Why this fixes SignCheck
The insertion failure reported unsigned `glyphicons-halflings-regular.ttf` files inside the Web and Django template VSIX payloads. The prior change staged and signed template fonts, but the VSIX projects could still fall back to `$(BuildRoot)\Python\Templates\...` if the staged tree was missing or not selected, which packages the original unsigned source `.ttf` files.

This change makes signed builds (`SignType=Real` or `SignType=Test`) fail fast unless the template VSIXs use the staged signed template tree. That prevents the final VSIX payload from being built from unsigned source templates. The `.ttf` files remain signed with `3PartyScriptsSHA2`, consistent with the DevDiv `.ttf` signing examples.

## Validation
- Parsed the changed MSBuild XML files successfully locally
- Full signing build was not run locally because Visual Studio MSBuild is not installed in this environment

### DIFF
--- a/Python/Setup/signlayout.proj
+++ b/Python/Setup/signlayout.proj
@@ -9,6 +9,7 @@
   <PropertyGroup>
     <OutputPath>$(BinariesOutputPath)</OutputPath>
     <OutDir>$(BinariesOutputPath)</OutDir>
+    <SignedTemplateSentinelPath>$(BinariesOutputPath)templates\SignedTemplateFiles.marker</SignedTemplateSentinelPath>
   </PropertyGroup>
   
   <ItemDefinitionGroup>
@@ -123,6 +124,8 @@
            which would otherwise break the signing step. -->
       <TemplateScriptFilesToSign Remove="@(TemplateScriptFilesToSign)"
                                 Condition="'$([System.IO.File]::ReadAllText(%(TemplateScriptFilesToSign.FullPath)))' == ''" />
+      <TemplateFontFilesToSign Remove="@(TemplateFontFilesToSign)"
+                               Condition="'$([System.IO.File]::ReadAllText(%(TemplateFontFilesToSign.FullPath)))' == ''" />
       <FilesToSign Include="@(TemplateScriptFilesToSign)">
         <Authenticode>3PartyScriptsSHA2</Authenticode>
         <StrongName></StrongName>
@@ -183,6 +186,7 @@
           DependsOnTargets="_GetBinariesInLayout;_GetJsFilesInLayout;_CopySignedTemplateFiles"
           AfterTargets="SignFiles">
     <Copy SourceFiles="%(_BinariesToCopy.FullPath)" DestinationFiles="%(_BinariesToCopy.TargetPath)" />
+    <Touch Files="$(SignedTemplateSentinelPath)" AlwaysCreate="true" />
   </Target>
 
   <Import Project="$(TargetsPath)\MicroBuild.targets" />
--- a/Python/Setup/signlayout.proj
+++ b/Python/Setup/signlayout.proj
@@ -116,13 +116,20 @@
   <Target Name="_AddTemplateFiles" BeforeTargets="_PreserveUnsigned;ListFiles">
     <ItemGroup>
       <!-- Only sign file types that SignCheck validates inside the final template VSIX payloads. -->
-      <TemplateFilesToSign Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
+      <TemplateScriptFilesToSign Include="$(BinariesOutputPath)templates\**\*.js" />
+      <TemplateFontFilesToSign Include="$(BinariesOutputPath)templates\**\*.ttf" />
       <!-- Drop zero-byte files (e.g. the empty 'Add New Item -> JavaScript' template placeholder).
            signtool cannot map a 0-byte file (CreateFileMapping fails with ERROR_FILE_INVALID 0x3EE),
            which would otherwise break the signing step. -->
-      <TemplateFilesToSign Remove="@(TemplateFilesToSign)"
-                           Condition="'$([System.IO.File]::ReadAllText(%(TemplateFilesToSign.FullPath)))' == ''" />
-      <FilesToSign Include="@(TemplateFilesToSign)">
+      <TemplateScriptFilesToSign Remove="@(TemplateScriptFilesToSign)"
+                                Condition="'$([System.IO.File]::ReadAllText(%(TemplateScriptFilesToSign.FullPath)))' == ''" />
+      <FilesToSign Include="@(TemplateScriptFilesToSign)">
+        <Authenticode>3PartyScriptsSHA2</Authenticode>
+        <StrongName></StrongName>
+        <SignedPath>$(BinariesOutputPath)templates\%(RecursiveDir)%(Filename)%(Extension)</SignedPath>
+        <UnsignedPath>$(UnsignedOutputPath)templates\%(RecursiveDir)%(Filename)%(Extension)</UnsignedPath>
+      </FilesToSign>
+      <FilesToSign Include="@(TemplateFontFilesToSign)">
         <Authenticode>3PartyScriptsSHA2</Authenticode>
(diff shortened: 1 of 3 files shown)


rchiodo:

Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
```diff
@@ -14,6 +15,12 @@
     <Dependency Include="Microsoft.PythonTools.Core.Vsix" />
   </ItemGroup>
 
+  <Target Name="_RequireSignedTemplateSource" BeforeTargets="_GeneratePackage" Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'Test'">
+    <Error Text="Signed template source path '$(SignedTemplateSourcePath)' was not found. Build signlayout.proj before packaging Microsoft.PythonTools.Web.Templates.Vsix."
+           Condition="!Exists('$(SignedTemplateSourcePath)')" />
+    <Error Text="Signed builds must package Microsoft.PythonTools.Web.Templates.Vsix from '$(SignedTemplateSourcePath)', not '$(SourcePath)'."
+           Condition="'$(SourcePath)' != '$(SignedTemplateSourcePath)'" />
+  </Target>
```
rchiodo, line 27: MSBuild silently ignores a `BeforeTargets` that names a non-existent target. If the MicroBuild.Swix SDK's package-generation target is not actually named `_GeneratePackage`, `_RequireSignedTemplateSource` never runs and this PR provides zero protection — a signed build would package unsigned fonts exactly as before. Confirm the real target name (it's imported via `SetupProjectAfter.settings`), and add a `<Message Importance="high" Text="_RequireSignedTemplateSource executed"/>` inside the target so a signed build log positively proves it fired.

Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj
```diff
@@ -14,6 +15,12 @@
     <Dependency Include="Microsoft.PythonTools.Django.Vsix" />
   </ItemGroup>
 
+  <Target Name="_RequireSignedTemplateSource" BeforeTargets="_GeneratePackage" Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'Test'">
+    <Error Text="Signed template source path '$(SignedTemplateSourcePath)' was not found. Build signlayout.proj before packaging Microsoft.PythonTools.Django.Templates.Vsix."
+           Condition="!Exists('$(SignedTemplateSourcePath)')" />
+    <Error Text="Signed builds must package Microsoft.PythonTools.Django.Templates.Vsix from '$(SignedTemplateSourcePath)', not '$(SourcePath)'."
+           Condition="'$(SourcePath)' != '$(SignedTemplateSourcePath)'" />
+  </Target>
```
rchiodo, line 27: Same `BeforeTargets="_GeneratePackage"` concern as the Web project — the guard is silently dead if that target name is wrong. Because this block is byte-identical to the Web swixproj, any correction to the hook target must be applied in both files. Confirm the target name and add a positive-execution `Message` here too.

Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
```diff
@@ -14,6 +15,12 @@
     <Dependency Include="Microsoft.PythonTools.Core.Vsix" />
   </ItemGroup>
 
+  <Target Name="_RequireSignedTemplateSource" BeforeTargets="_GeneratePackage" Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'Test'">
+    <Error Text="Signed template source path '$(SignedTemplateSourcePath)' was not found. Build signlayout.proj before packaging Microsoft.PythonTools.Web.Templates.Vsix."
+           Condition="!Exists('$(SignedTemplateSourcePath)')" />
```
rchiodo, line 22: `Exists('$(SignedTemplateSourcePath)')` only proves the staged directory is present, not that its `.ttf`/`.js` contents were actually signed. In `signlayout.proj`, staging copies unsigned files into `templates\Web` before `SignFiles` runs; if the swix package build runs after staging but not strictly after signing (parallel graph, or swix built without signlayout as a hard predecessor), the guard passes while packaging unsigned fonts — the exact failure this PR targets. Enforce swix-runs-after-signlayout's signing step explicitly (project dependency / sequencing) or gate on a signing-completion sentinel rather than mere directory existence. At minimum, document the required build ordering in the swixproj comment.

Python/Setup/signlayout.proj
```diff
@@ -116,13 +116,20 @@
   <Target Name="_AddTemplateFiles" BeforeTargets="_PreserveUnsigned;ListFiles">
     <ItemGroup>
       <!-- Only sign file types that SignCheck validates inside the final template VSIX payloads. -->
-      <TemplateFilesToSign Include="$(BinariesOutputPath)templates\**\*.js;$(BinariesOutputPath)templates\**\*.ttf" />
+      <TemplateScriptFilesToSign Include="$(BinariesOutputPath)templates\**\*.js" />
+      <TemplateFontFilesToSign Include="$(BinariesOutputPath)templates\**\*.ttf" />
       <!-- Drop zero-byte files (e.g. the empty 'Add New Item -> JavaScript' template placeholder).
```
rchiodo, line 122: `TemplateFontFilesToSign` has no zero-byte `Remove` guard, so the safety net that protects `.js` was removed for fonts. A 0-byte `.ttf` would flow into `FilesToSign` and cause signtool's `CreateFileMapping` to fail with `ERROR_FILE_INVALID (0x3EE)` — the same failure the JS filter exists to prevent — breaking the whole signing step. Either apply the same zero-byte guard to fonts or add a comment justifying why template `.ttf` files can never be zero bytes. (Splitting `.ttf` out of the `ReadAllText` text comparison is itself correct — reading a binary font as text never matched `== ''`.)

rchiodo:
The staging/guard approach is reasonable, but please verify the guard actually fires in a real signed build (the BeforeTargets hook and Exists-only check are the crux risks) and reinstate zero-byte protection for fonts before merging.

Python/Setup/swix/Packages/Microsoft.PythonTools.Web.Templates.Vsix.swixproj
```diff
@@ -14,6 +16,15 @@
     <Dependency Include="Microsoft.PythonTools.Core.Vsix" />
   </ItemGroup>
 
+  <Target Name="_RequireSignedTemplateSource" BeforeTargets="_GeneratePackage" Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'Test'">
```
rchiodo, line 19: The `Message Importance="high"` was added (good), but it only prints if the target actually fires, and MSBuild silently ignores a `BeforeTargets` that names a non-existent target. This is load-bearing: if `_GeneratePackage` is not the real MicroBuild.Swix package target (imported via `SetupProjectAfter.settings`), the guard is dead and a signed build with a missing staged tree leaves `SourcePath` empty (the fallback is gated to `SignType != Real/Test`), producing a broken/empty VSIX instead of a clear error. Please confirm the target name by observing this message in a real signed-build log before merge.

rchiodo:
Overall this looks sound and is safe to merge once two things are confirmed: (1) that `_RequireSignedTemplateSource` actually fires in a real signed build (verify the `Message Importance="high"` line appears in the log, since MSBuild silently ignores a `BeforeTargets` naming a non-existent target), and (2) that template `.ttf` files can never be zero bytes now that the zero-byte filter no longer covers fonts.

rchiodo (approved):
Approved via Review Center.

PR #8572, opened 2026-07-04
Fix template font signing catalogs

bschnurr:
## Summary
- generate signed catalogs for the Django/Web template glyphicons TTF payloads
- sign those catalogs with Microsoft400 instead of attempting direct TTF signing
- require the staged signed template tree and matching catalog when packaging signed template VSIXes

## Validation
- parsed changed MSBuild project files as XML
- ran targeted VS MSBuild ListFiles validation for signlayout.proj with SignType=Test
- verified SWIX source generation includes the catalogs alongside the four glyphicons TTF payloads

### DIFF
--- a/Python/Setup/signlayout.proj
+++ b/Python/Setup/signlayout.proj
@@ -10,6 +10,13 @@
     <OutputPath>$(BinariesOutputPath)</OutputPath>
     <OutDir>$(BinariesOutputPath)</OutDir>
     <SignedTemplateSentinelPath>$(BinariesOutputPath)templates\SignedTemplateFiles.marker</SignedTemplateSentinelPath>
+    <WindowsKitsRoot Condition="'$(WindowsKitsRoot)' == ''">$([MSBuild]::GetRegistryValueFromView('HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots', 'KitsRoot10', null, RegistryView.Registry32, RegistryView.Default))</WindowsKitsRoot>
+    <MakeCatExe Condition="'$(MakeCatExe)' == '' and '$(WindowsSdkVerBinPath)' != '' and Exists('$(WindowsSdkVerBinPath)x86\makecat.exe')">$(WindowsSdkVerBinPath)x86\makecat.exe</MakeCatExe>
+    <MakeCatExe Condition="'$(MakeCatExe)' == '' and '$(WindowsKitsRoot)' != '' and '$(TargetPlatformSdkVersion)' != '' and Exists('$(WindowsKitsRoot)bin\$(TargetPlatformSdkVersion)\x86\makecat.exe')">$(WindowsKitsRoot)bin\$(TargetPlatformSdkVersion)\x86\makecat.exe</MakeCatExe>
+    <MakeCatExe Condition="'$(MakeCatExe)' == '' and '$(WindowsKitsRoot)' != '' and Exists('$(WindowsKitsRoot)bin\10.0.26100.0\x86\makecat.exe')">$(WindowsKitsRoot)bin\10.0.26100.0\x86\makecat.exe</MakeCatExe>
+    <MakeCatExe Condition="'$(MakeCatExe)' == ''">makecat.exe</MakeCatExe>
+    <DjangoTemplateFontCatalogPath>$(BinariesOutputPath)templates\Django\Microsoft.PythonTools.Django.Templates.Fonts.cat</DjangoTemplateFontCatalogPath>
+    <WebTemplateFontCatalogPath>$(BinariesOutputPath)templates\Web\Microsoft.PythonTools.Web.Templates.Fonts.cat</WebTemplateFontCatalogPath>
   </PropertyGroup>
   
   <ItemDefinitionGroup>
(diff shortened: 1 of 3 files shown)


rchiodo:

Python/Setup/signlayout.proj
```diff
@@ -114,29 +121,58 @@
...
-      <TemplateFontFilesToSign Include="$(BinariesOutputPath)templates\**\*.ttf" />
+      <TemplateFontCatalogFilesToSign Include="$(DjangoTemplateFontCatalogPath);$(WebTemplateFontCatalogPath)" />
       <!-- Drop zero-byte files (e.g. the empty 'Add New Item -> JavaScript' template placeholder).
            signtool cannot map a 0-byte file (CreateFileMapping fails with ERROR_FILE_INVALID 0x3EE),
            which would otherwise break the signing step. -->
       <TemplateScriptFilesToSign Remove="@(TemplateScriptFilesToSign)"
                                 Condition="'$([System.IO.File]::ReadAllText(%(TemplateScriptFilesToSign.FullPath)))' == ''" />
-      <TemplateFontFilesToSign Remove="@(TemplateFontFilesToSign)"
-                               Condition="'$([System.IO.File]::ReadAllText(%(TemplateFontFilesToSign.FullPath)))' == ''" />
       <FilesToSign Include="@(TemplateScriptFilesToSign)">
```
rchiodo, line 165: 📍 Python/Setup/signlayout.proj:135
[unverified] Load-bearing premise, flagged Critical by the Architect: the entire approach assumes an in-package `.cat` sitting beside the TTFs makes those TTFs count as validly signed to SignCheck. A security catalog normally vouches for a file only when installed into the system catalog store or when the verifier is explicitly catalog-aware. The PR's Validation proves the catalog is *generated and packaged*, but not that it *satisfies signing validation* — different claims. Please run the real SignCheck pass over the produced signed template VSIX and confirm the glyphicons TTFs report as validly signed *via* the catalog (not skipped). If SignCheck does not honor the in-package catalog, this is a lateral move and the fix is incomplete.

[verified]

Python/Setup/signlayout.proj
```diff
@@ -114,29 +121,58 @@
...
   </Target>
+
+  <Target Name="_GenerateTemplateFontCatalogs"
+          DependsOnTargets="_StageTemplateFiles"
+          BeforeTargets="_AddTemplateFiles"
+          Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'Test'">
+    <PropertyGroup>
+      <DjangoTemplateFontCatalogCdfPath>$(IntermediateOutputPath)Microsoft.PythonTools.Django.Templates.Fonts.cdf</DjangoTemplateFontCatalogCdfPath>
+      <WebTemplateFontCatalogCdfPath>$(IntermediateOutputPath)Microsoft.PythonTools.Web.Templates.Fonts.cdf</WebTemplateFontCatalogCdfPath>
+    </PropertyGroup>
```
rchiodo, line 132: 📍 Python/Setup/signlayout.proj:130
[unverified] Signing coverage narrowed from `templates\**\*.ttf` (every font in the staged tree except Core/NativeDevelopment) to a single hardcoded `glyphicons-halflings-regular.ttf` under only `Django`/`Web`. The Skeptic and Advocate both flag this: any TTF that is differently named, or lives outside these two trees, is now neither Authenticode-signed nor covered by a signed catalog — a silent SignCheck-coverage regression. Please confirm glyphicons is genuinely the only shipped template font, or broaden the globs to `templates\Django\**\*.ttf` / `templates\Web\**\*.ttf` (plus any other font-bearing template dir), or add a build assertion that fails if any `templates\**\*.ttf` is not covered by a generated catalog.

[verified]

Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj
```diff
@@ -16,12 +17,14 @@
     <Dependency Include="Microsoft.PythonTools.Django.Vsix" />
   </ItemGroup>
 
-  <Target Name="_RequireSignedTemplateSource" BeforeTargets="_GeneratePackage" Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'Test'">
+  <Target Name="_RequireSignedTemplateSource" BeforeTargets="_GeneratePackage" Condition="'$(SignType)' == 'Real' or '$(SignType)' == 'Test' or Exists('$(SignedTemplateSentinelPath)')">
```
rchiodo, line 20: 📍 Python/Setup/swix/Packages/Microsoft.PythonTools.Django.Templates.Vsix.swixproj:20
[unverified] The gate condition was broadened to `... or Exists('$(SignedTemplateSentinelPath)')`. The sentinel is produced by signed builds, but `_GenerateTemplateFontCatalogs` only runs for `SignType=Real/Test`. A later non-signed packaging build in an output dir that still has a stale `SignedTemplateFiles.marker` now runs `_RequireSignedTemplateSource` and hard-errors on the missing `*.Fonts.cat` that a non-signed build never regenerates. Fail-loud is defensible, but this couples routine local packaging to residual filesystem state (Skeptic + Advocate + Architect agree). Prefer gating on the build's signing *intent* rather than a leftover marker, or tie the sentinel's lifetime to the binaries it guards. Same applies to the Web swixproj.

[verified]

Python/Setup/signlayout.proj
```diff
@@ -10,6 +10,13 @@
     <OutputPath>$(BinariesOutputPath)</OutputPath>
     <OutDir>$(BinariesOutputPath)</OutDir>
     <SignedTemplateSentinelPath>$(BinariesOutputPath)templates\SignedTemplateFiles.marker</SignedTemplateSentinelPath>
+    <WindowsKitsRoot Condition="'$(WindowsKitsRoot)' == ''">$([MSBuild]::GetRegistryValueFromView('HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots', 'KitsRoot10', null, RegistryView.Registry32, RegistryView.Default))</WindowsKitsRoot>
+    <MakeCatExe Condition="'$(MakeCatExe)' == '' and '$(WindowsSdkVerBinPath)' != '' and Exists('$(WindowsSdkVerBinPath)x86\makecat.exe')">$(WindowsSdkVerBinPath)x86\makecat.exe</MakeCatExe>
+    <MakeCatExe Condition="'$(MakeCatExe)' == '' and '$(WindowsKitsRoot)' != '' and '$(TargetPlatformSdkVersion)' != '' and Exists('$(WindowsKitsRoot)bin\$(TargetPlatformSdkVersion)\x86\makecat.exe')">$(WindowsKitsRoot)bin\$(TargetPlatformSdkVersion)\x86\makecat.exe</MakeCatExe>
+    <MakeCatExe Condition="'$(MakeCatExe)' == '' and '$(WindowsKitsRoot)' != '' and Exists('$(WindowsKitsRoot)bin\10.0.26100.0\x86\makecat.exe')">$(WindowsKitsRoot)bin\10.0.26100.0\x86\makecat.exe</MakeCatExe>
```
rchiodo, line 16: 📍 Python/Setup/signlayout.proj:15
[unverified] The `MakeCatExe` discovery chain hardcodes SDK build `10.0.26100.0` and then falls back to a bare `makecat.exe` PATH lookup. The magic version will silently rot, and when every dynamic probe misses, resolution depends on `makecat.exe` being on `PATH` — a failure surfaces deep inside `Exec` rather than fast with a clear "SDK not found" message. Consider deriving the version from an SDK property and/or adding an explicit `<Error>` when no `makecat.exe` was resolved, so signed builds fail fast with an actionable diagnostic.

[verified]

Python/Setup/signlayout.proj
```diff
@@ -10,6 +10,13 @@
...
+    <MakeCatExe Condition="'$(MakeCatExe)' == ''">makecat.exe</MakeCatExe>
+    <DjangoTemplateFontCatalogPath>$(BinariesOutputPath)templates\Django\Microsoft.PythonTools.Django.Templates.Fonts.cat</DjangoTemplateFontCatalogPath>
+    <WebTemplateFontCatalogPath>$(BinariesOutputPath)templates\Web\Microsoft.PythonTools.Web.Templates.Fonts.cat</WebTemplateFontCatalogPath>
   </PropertyGroup>
```
rchiodo, line 20: 📍 Python/Setup/signlayout.proj:19
Single-source-of-truth violation (Architect): the catalog filenames `Microsoft.PythonTools.{Django,Web}.Templates.Fonts.cat` are defined here and then independently reconstructed in both swixproj require-checks (`SignedTemplateFontCatalogPath`). A rename must touch ≥3 files or the require-gate silently looks for the wrong name. Hoist the catalog name/path into one shared property imported by both the producer (`signlayout.proj`) and the swixproj consumers.

[verified]

rchiodo:
Overall the catalog-based approach is reasonable. The main things worth confirming before merge: (1) that SignCheck genuinely treats the in-package `.cat` as valid signing coverage for the glyphicons TTFs, and (2) that narrowing from `templates\**\*.ttf` to a hardcoded glyphicons filename under only Django/Web doesn't silently drop coverage for any other shipped template font.

rchiodo (approved):
Approved via Review Center.

PR #8576, opened 2026-07-13
Route Python packages through Azure Artifacts

StellaHuang95:
## Summary

Route all pipeline Python package downloads through the team-owned `DevDiv/Pylance_PublicPackages` Azure Artifacts feed to resolve the SFI-ES4.2.4 public PyPI violation.

- authenticate the build and Glass test jobs with `PipAuthenticate@1`
- replace direct `pypi.org` JSON and `files.pythonhosted.org` requests with the authenticated PEP 503 feed
- preserve the existing merged-wheel installation for debugpy and etwtrace
- reject extra/public indexes and public PyPI redirects
- validate wheel names and verify SHA-256 hashes when supplied
- support both Azure Artifacts hostname formats emitted by local tooling and hosted agents

## Validation

- PTVS-Build run 26195.3 succeeded
- debugpy 1.8.21 and etwtrace 0.1b9 installed through `Pylance_PublicPackages`
- product, tests, installer, bootstrapper, signing, SBoM, security scans, and artifact publication succeeded
- CodeQL and SonarCloud passed

### DIFF
--- a/Build/install_pypi_package.py
+++ b/Build/install_pypi_package.py
@@ -1,22 +1,26 @@
 # This is a modified version of https://github.com/microsoft/vscode-python/blob/main/pythonFiles/install_debugpy.py
 
-# It downloads and extracts a specified package version from PyPi to a specified directory.
+# It downloads and extracts a specified package version from an authenticated Azure Artifacts feed.
 # Doing a `pip install` will only install the package for the python version that is running the script,
 # but since we don't know what version of python the user will be running, we want to download ALL the wheels and
 # perform a "fat install" by merging all the wheels into the same directory.
 
 import argparse
+import base64
+import hashlib
+from html.parser import HTMLParser
 import io
-import json
 import os
+import re
+import ssl
+import urllib.parse
 import urllib.request as url_lib
 import zipfile
-import urllib.parse
-import hashlib
-import ssl
+
 import certifi
 
-# if this import fails, run PreBuild.ps1, which will pip install packaging
+# If this import fails, run PreBuild.ps1, which installs packaging from the authenticated feed.
+from packaging.utils import canonicalize_name, parse_wheel_filename
 from packaging.version import parse as version_parser
 
 # only install wheels for these python versions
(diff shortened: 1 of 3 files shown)


rchiodo:

heejaechang:

Build/install_pypi_package.py
```diff
@@ -25,67 +29,183 @@
...
-            try:
-                with open(cache_path, "wb") as f:
-                    f.write(data)
-            except Exception:
-                pass
+            with open(cache_path, "wb") as f:
+                f.write(data)
+
+    fragment = urllib.parse.parse_qs(urllib.parse.urlsplit(url).fragment)
+    expected_hashes = fragment.get("sha256")
```
rchiodo, line 206: The optional wheel cache lost all of its error handling. Previously `os.makedirs`, the cache read, and the cache write were each wrapped in `try/except`, so a broken or opt-in `PTVS_PYPI_WHEEL_CACHE_DIR` (read-only mount, permission error, full disk, concurrent job) could never fail the install. Now any of those conditions will hard-fail the entire build, turning an optional accelerator into a load-bearing dependency. Restore best-effort semantics (degrade to a direct download on cache failure), or explicitly document that the cache is mandatory-when-configured.

Build/install_pypi_package.py
```diff
@@ -25,67 +29,183 @@
...
         if cache_path:
-            try:
-                with open(cache_path, "wb") as f:
-                    f.write(data)
-            except Exception:
-                pass
+            with open(cache_path, "wb") as f:
+                f.write(data)
+
+    fragment = urllib.parse.parse_qs(urllib.parse.urlsplit(url).fragment)
```
rchiodo, line 205: The cache write is non-atomic and happens *before* SHA-256 verification. A truncated/corrupt download gets persisted to `cache_path` before the hash check runs, so every subsequent run reads the poisoned file, re-verifies, and fails again — a persistent build break until the cache is manually purged. With the shared-cache parallelism this feed change enables, a partial `f.write()` can also be read by another job. Verify the SHA-256 on the downloaded bytes before writing, and write via a temp file + `os.replace` atomic rename.

Build/install_pypi_package.py
```diff
@@ -25,67 +29,183 @@
...
+    for href, yanked in parser.links:
+        wheel_url = urllib.parse.urljoin(package_url, href)
+        filename = os.path.basename(urllib.parse.urlsplit(wheel_url).path)
+        if not filename.endswith(".whl"):
+            continue
+
+        wheel_name, version, _, _ = parse_wheel_filename(filename)
+        if canonicalize_name(wheel_name) != expected_name:
+            continue
+
```
rchiodo, line 159: `parse_wheel_filename` can raise `InvalidWheelFilename` (or `InvalidVersion` on a non-PEP440 version), and neither is caught in `get_package_data`. A single malformed `.whl`-suffixed link on the feed's simple-index page for the requested package would abort all builds — a hard-failure the old JSON path didn't have. Wrap the call in `try/except (InvalidWheelFilename, InvalidVersion): continue` to skip unparseable links.

Build/install_pypi_package.py
```diff
@@ -25,67 +29,183 @@
...
+        if allowed.startswith("."):
+            if hostname.endswith(allowed):
+                return True
+        elif hostname == allowed:
+            return True
+    return False
+
+
+class _SfiRedirectHandler(url_lib.HTTPRedirectHandler):
+    # Azure Artifacts responds with a 303 pointing at a SAS-signed blob URL;
```
rchiodo, line 66: The `_SfiRedirectHandler` comment claims it refuses "any redirect that leaves Microsoft-owned infrastructure," but the implementation only raises for the 5-host `FORBIDDEN_HOSTS` deny-list — a redirect to any other non-forbidden host would still be followed. (The Basic credential is correctly not resent, so no secret leaks.) Update the comment to describe the actual deny-list behavior so a future maintainer doesn't 'tighten' it based on the misleading description and break legitimate `*.blob.core.windows.net` redirects.

rchiodo:
Solid, security-focused change that meets the SFI goal. A few robustness edge cases around the optional wheel cache and feed-link parsing are worth tightening before merge, but none are blocking given the passing pipeline validation.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

bschnurr (approved):

PR #8582, opened 2026-07-21
Fix Python 3.13 mixed-mode debugger crash by computing line numbers in-process

rchiodo:
## Problem

Mixed-mode (Python/Native) debugging crashes Visual Studio when debugging a C++ app that embeds CPython 3.11+ (reported against 3.13). When a breakpoint is hit in a GUI app that pumps messages, VS becomes unresponsive and then crashes. A console build of the same app doesn't crash but stepping from Python into native code misbehaves. Python 3.9 worked previously.

## Root cause

When building the call stack, `PyFrameObject.ComputeLineNumber` needs the current line for each Python frame. On 3.11+ the frame's `f_lineno` is `0` during execution (the line is computed lazily), so the code fell back to a **per-frame func-eval in the debuggee** on every stop:

```
((int(*)(void*))PyFrame_GetLineNumber)(frameAddr)
```

Executing debuggee code (a func-eval) while constructing the call stack is unsafe in a GUI process: the func-eval runs the interpreter, which pumps messages / re-enters, causing reentrancy and deadlock that hangs and crashes VS. It also perturbs the step engine, which explains the broken Python→native stepping in the console case. On 3.9 the line number was read directly from the frame, so no func-eval happened and it worked.

## Fix

Compute the line number **in-process using memory reads only**, mirroring CPython's `PyCode_Addr2Line` / `_PyInterpreterFrame_GetLine`:

- Add the interpreter frame's instruction pointer to the `PyInterpreterFrame` proxy (`instr_ptr` on 3.13+, `prev_instr` on 3.11/3.12).
- Derive the bytecode offset as `instr_ptr - co_code_adaptive` and walk `co_linetable` (a managed port of CPython's location-table decoder) to get the line.
- `PyFrameObject.ComputeLineNumber` now returns `f_lineno` when it's already set, otherwise decodes the line table in-process, and only falls back to the old func-eval as a last resort.

This path only feeds `CallStackFilter` (display), never step decisions, so stepping behavior is unaffected — but it removes the debuggee func-eval that caused the crash.

## Files

- `Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs` — add `prev_instr`/`instr_ptr` fields and `ComputeLineNumber()` (Addr2Line decoder).
- `Python/Product/Debugger.Concord/Proxies/Structs/PyCodeObject311.cs` — add `co_linetable`.
- `Python/Product/Debugger.Concord/Proxies/Structs/PyFrameObject.cs` — prefer `f_lineno`, then in-process decode, then legacy func-eval fallback.
- `Python/Product/Debugger.Concord/Proxies/Structs/PyFrameObject311.cs` — override to decode from the interpreter frame.

## Testing

Manually verified against a C++ GUI app embedding CPython 3.13 (the `Examples/PythonNative` sample) in VS 18 (2026 preview): breakpoints in both Python and native code are hit, stepping between Python and native works, and Visual Studio no longer hangs or crashes.

## Notes

- This addresses the Issue-1 GUI crash and the console stepping side-effect. The separate report about combining native debugging + debugpy attach in the *same* VS instance is architectural (the native debugger freezes the process debugpy needs running) and is not changed here.

### DIFF
--- a/Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs
+++ b/Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs
@@ -35,6 +35,13 @@ namespace Microsoft.PythonTools.Debugger.Concord.Proxies.Structs {
             public StructField<PointerProxy<PyInterpreterFrame>> previous;
             public StructField<ArrayProxy<PointerProxy<PyObject>>> localsplus;
             public StructField<CharProxy> owner;
+            // Pointer to the current bytecode instruction. Renamed from prev_instr to
+            // instr_ptr in 3.13 (and it now points at the current instruction rather
+            // than the previous one), but the offset math for the line table is identical.
+            [FieldProxy(MaxVersion = PythonLanguageVersion.V312)]
+            public StructField<PointerProxy> prev_instr;
+            [FieldProxy(MinVersion = PythonLanguageVersion.V313)]
+            public StructField<PointerProxy> instr_ptr;
         }
 
         private const int FRAME_OWNED_BY_THREAD = 0;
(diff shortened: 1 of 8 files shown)


heejaechang:

Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs
```diff
@@ -77,6 +84,119 @@ public PointerProxy<PyFrameObject> frame_obj {
...
+                shift += 6;
+                val |= (read & 63) << shift;
+            }
+            return (int)val;
+        }
+
+        private static int ReadSignedVarint(byte[] data, ref int p) {
+            uint uval = (uint)ReadVarint(data, ref p);
+            return ((uval & 1) != 0) ? -(int)(uval >> 1) : (int)(uval >> 1);
+        }
```
heejaechang, line 134: The new location-table decoder is deterministic but has no automated coverage. Please add table-driven cases using recorded `co_linetable` bytes from supported 3.11–3.13 runtimes, including `NONE` entries, to lock down the CPython-format port and prevent regressions.

Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs
```diff
@@ -77,6 +84,119 @@ public PointerProxy<PyFrameObject> frame_obj {
...
+
+            // _PyCode_CODE(co) == &co->co_code_adaptive[0] - the start of the bytecode.
+            ulong codeStart = code.co_code_adaptive.Address;
+
+            // The current instruction pointer: prev_instr (3.11/3.12) or instr_ptr (3.13+).
+            ulong instrPtr = _fields.instr_ptr.Process != null
+                ? GetFieldProxy(_fields.instr_ptr).Read()
+                : GetFieldProxy(_fields.prev_instr).Read();
+            if (instrPtr == 0 || instrPtr < codeStart) {
+                return 0;
```
heejaechang, line 112: For 3.11/3.12, `prev_instr` can be just before `co_code_adaptive` before the first bytecode instruction. CPython handles the resulting negative offset by returning `co_firstlineno`, but this guard returns 0 and falls back to the func-eval this change is intended to avoid. Handle that entry-frame case by returning the code object's first line.

heejaechang (approved):
Approved via Review Center.

Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs
rchiodo: Good catch. Fixed in 46ce5053d: `ComputeLineNumber` now computes a signed `addrq` and returns `co_firstlineno` when it's negative — the 3.11/3.12 entry-frame case where `prev_instr` points one `_Py_CODEUNIT` before the first instruction — matching `PyCode_Addr2Line` instead of returning 0 and falling back to the func-eval.

Python/Product/Debugger.Concord/Proxies/Structs/PyInterpreterFrame.cs
rchiodo: Added table-driven coverage in `Python/Tests/DebuggerTests/PyLineTableTests.cs`. I pulled the decoder out into an `internal static PyLineTable` class and lock it down against `co_linetable` bytes recorded from real 3.11.9, 3.12.7 and 3.13.14 runtimes, asserting every bytecode offset maps to the line `code.co_lines()` reports — including `NONE` (no-line) markers, negative/entry-frame offsets, past-end offsets, and a null table. All 5 tests pass locally.

bschnurr (approved):

heejaechang (approved):
Approved via Review Center.

StellaHuang95 (approved):

PR #8583, opened 2026-07-22
Opt out of IntelliSense-ready deferral for the Python profiling session editor

davkean:
The .pyperf profiling session editor doesn't use CodeModel, so it has no reason to wait for IntelliSense before opening. Switching from ProvideEditorExtension to ProvideEditorExtension2 adds DeferUntilIntellisenseIsReady=dword:00000000 under the Editors\{factory} key, opting the editor out of the deferral so it opens immediately.

With the arguments used here the two attributes produce identical registration; the only difference is that added value. The code paths where they diverge (params extensions, template-dir escaping, linked-editor GUID, common view attributes) aren't exercised.

*Cowritten with Copilot (Claude Opus 4.8)*


heejaechang:

heejaechang (approved):
Approved via Review Center.

davkean:
@heejaechang  Would you be able to merge this on completion, I don't have permissions.

davkean:
@heejaechang Actually, the thing is blocked because I don't have permissions to run the build.

PR #8587, opened 2026-07-22
Fix Pylance settings handling on Dev18

### LINKED ISSUE #8578: Checking "Automatically add brackets to functions" has no effect.
**Environment**
Build of Visual Studio: 18.10.0 Canary [12017.272.main]
Build of Python Package: 18.0.26200.1
Version of Pylance: 2025.12.104

**Steps to Reproduce**
1.Check "Automatically add brackets for functions" (Tools/Options/Text Editor/Python/Advanced)
2.Add "import os" to python file.
3.Write "os.getc", you'll see autocomplete for os.getcwd, selecting it, ensure:

**Expected behavior**
Auto-complete adds os.getcwd() to the source code.

Only **os.getcwd** is added to the file.

image

StellaHuang95:
## Summary

Fixes #8578.

Checking **Automatically add brackets to functions** persisted correctly in Visual Studio, but Dev18 did not deliver the enabled setting to Pylance. As a result, Pylance continued returning plain completion text such as `getcwd` instead of a snippet such as `getcwd($0)`.

## Root cause

Dev18 uses System.Text.Json for eligible language clients, exposing three serialization and binding mismatches in the PTVS/Pylance configuration path:

- Pylance sends `workspace/configuration` parameters as one object (`{ "items": [...] }`), but the custom StreamJsonRpc handler did not opt into whole-object parameter binding, so the request was rejected with `-32602`.
- PTVS's outbound settings DTO used public fields. System.Text.Json does not serialize those fields by default, so the settings payload became `{}` even when the option was enabled.
- The fallback `workspace/didChangeConfiguration` payload placed analysis settings directly under `settings.analysis`, while Pylance expects the section under `settings.python.analysis`.

Latest `main` also introduced a `JToken` signature-help middle layer. On Dev18, that type selects the Newtonsoft formatter for the whole language client, which conflicts with the STJ-compatible path needed for PTVS's initialize interception. The Dev18 implementation therefore uses `JsonDocument`, while the existing Dev17 `JToken` implementation remains unchanged.

## Fix

- Enable whole-object binding for object-shaped custom RPC messages, including `workspace/configuration`.
- Deserialize custom payloads through a serializer-neutral helper that accepts both Newtonsoft and Dev18 STJ values.
- Convert outbound Pylance settings fields to properties so both serializers emit the same JSON contract.
- Send fallback configuration under the required `python` section.
- Keep Dev18 signature-help interception on the System.Text.Json middle-layer contract while preserving the prior Dev17 path.
- Align the custom Pylance progress notification handler signatures with their protocol shapes.

## Risk

Risk is low because the changes are limited to the Pylance language-client serialization and custom RPC boundary:

- Wire names and setting values are unchanged; the settings that were previously omitted are now serialized.
- Whole-object binding is applied only to methods whose protocol parameters are object-shaped.
- The fallback envelope now matches Pylance's documented section lookup.
- Dev17 retains its existing signature-help implementation through conditional compilation.
- Completion insertion behavior is not changed; this only allows Pylance to receive the existing option and return the intended snippet.

## Validation

The fix was manually verified on Dev18: after enabling the option, function completions included parentheses as expected. No tests were added for this change.

### DIFF
--- a/Build/18.0/packages.config
+++ b/Build/18.0/packages.config
@@ -85,6 +85,7 @@
   <package id="Microsoft.Extensions.Logging.Abstractions" version = "8.0.2" /> <!-- 5.0.0 -> 8.0.2 -->
   <package id="System.Threading.Tasks.Extensions" version="4.6.0"/>
   <package id="System.Memory" version="4.6.0"/>
+  <package id="System.Text.Json" version="8.0.5"/>
   <package id="Microsoft.VisualStudio.RpcContracts" version = "17.15.25-pre" /> <!-- 17.15.26-pre -->
   <package id="Microsoft.DiaSymReader.Pdb2Pdb" version = "1.1.0-beta2-26280-01" /> <!-- 1.1.0-beta2-21181-01 -> 1.1.0-beta2-24172-02 -->
   <package id="Microsoft.TestPlatform.ObjectModel" version = "17.14.1" /> <!-- 17.4.0-preview-20221003-03 -> 17.14.1 -->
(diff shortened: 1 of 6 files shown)


heejaechang:

heejaechang (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

PR #8589, opened 2026-07-24
Update Microsoft repository policy files

StellaHuang95:
## Summary

- Update `SECURITY.md` from Microsoft's V0.0.3 block to the current V1.0.0 shared template.
- Add the standard standalone Microsoft `CODE_OF_CONDUCT.md`.
- Preserve PTVS's existing Apache 2.0 license.

## Context

Related to #8568 and #8569.

Both policy-service issues report files as missing, but PTVS already had a root `LICENSE` and a root `SECURITY.md` when the issues were created. GitHub recognizes the existing license as Apache 2.0, and changing it to Microsoft's default MIT template would be a substantive relicensing operation rather than a missing-file fix.

The existing `SECURITY.md` used an older Microsoft V0.0.3 block. This PR updates it to the current official template and adds the standalone code-of-conduct file called for by Microsoft's repository guidance. The policy-service presence findings should still be rerun or investigated separately.

`SUPPORT.md` is not included because the guidance treats it as optional and its template requires project-specific support commitments.

## Validation

- Confirmed both files match the current blobs in `microsoft/repo-templates`.
- `git diff --check`
- No product build or test run; this PR changes repository policy documentation only.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,10 @@
+# Microsoft Open Source Code of Conduct
+
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/).
+
+Resources:
+
+- [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/)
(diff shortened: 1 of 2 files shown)


heejaechang:

heejaechang (approved):
Approved via Review Center.

PR #8591, opened 2026-07-28
Enable Dev18 Python profiling through Diagnostics Hub

StellaHuang95:
## Context

Python profiling in Dev18 is currently unavailable. The legacy PTVS path depends on `VSPerfMon.exe` and `VSPerfCmd.exe`, which were removed from the Dev18 shipping manifest, and its native bridge only understands CPython layouts through Python 3.10. This is the PTVS half of moving basic function instrumentation to the existing Diagnostics Hub Python target and `microsoft/python-etwtrace` pipeline. It addresses the user-visible failure tracked by #8276.

This PR is **not standalone**. It requires companion Diagnostics Hub changes to:

- load the PTVS profiling command service without the failing dynamic-MEF handoff;
- treat `TargetProperties.Python` as script instrumentation instead of opening the native executable/PDB planner;
- resolve script module paths and function IDs correctly in DataWarehouse.

Those companion changes are currently on the DiagnosticsHub branch `feature/ptvs-python-profiling`.

## PTVS changes

- Route **Debug > Launch Python Profiling** to the Dev18 Diagnostics Hub launcher while preserving the legacy behavior for older VS targets.
- Register the currently hidden `PythonProfilingEnabled` Diagnostics Hub feature flag for Dev18.
- Expose a package-independent `PythonProfilerCommandService` that Diagnostics Hub can construct dynamically with the global VS service provider.
- Rework command construction so Diagnostics Hub owns the session/report while PTVS supplies the selected interpreter, script, arguments, working directory, and environment.
- Add `diaghub_profile.py`, which:
  - launches a dedicated profiled child process to satisfy Diagnostics Hub's child-attach contract;
  - activates `etwtrace` for the lifetime of the interpreter, including worker threads;
  - supports both script and `-m` execution;
  - preserves target arguments and import behavior;
  - reports unsupported interpreter/wheel combinations clearly.
- Package coherent `etwtrace` payloads:
  - `0.1b8` for CPython 3.9-3.11;
  - `0.1b9` for CPython 3.12-3.14.
- Include CPython 3.14 wheels in the fat package restore.
- Force the intended Dev18 `System.Text.Json` assembly reference to remove the existing MSB3277 conflict warning.

## Runtime flow

```text
PTVS target dialog
  -> Diagnostics Hub PythonTarget
  -> Instrumentation collector starts a root Python bootstrap
  -> bootstrap launches a profiled child Python process
  -> etwtrace attaches through DiagnosticsHub.InstrumentationCollector.dll
[...]

PR #8598, opened 2026-07-30
Limit Dev18 Python profiling to Python 3.12-3.14

StellaHuang95:
## Context

Python profiling is currently unavailable in Dev18 because the legacy PTVS path depends on `VSPerfMon.exe` and `VSPerfCmd.exe`, which are no longer shipped. #8591 began an exploratory effort to make profiling work again by routing PTVS through the Diagnostics Hub Python target and `microsoft/python-etwtrace`.

This is exploratory enablement rather than a regression in a working Dev18 user scenario. We have no known user complaints about Dev18 Python profiling, and the feature was already broken before this work. The priority is therefore to establish a reliable, supportable path for current Python versions rather than restore every historical interpreter at substantially higher release and compliance cost.

## Why PTVS originally included two etwtrace versions

`etwtrace` contains native `.pyd` extensions compiled for specific CPython ABIs. A `cp312` extension cannot load into Python 3.11, for example.

- `etwtrace 0.1b8` is the last release that contains wheels for Python 3.9, 3.10, and 3.11. PTVS installed it into a separate `etwtrace_legacy` directory and used extra collector-loading compatibility code for those interpreters.
- `etwtrace 0.1b9` contains the current Python 3.12, 3.13, and 3.14 wheels and is the payload used by the new Diagnostics Hub path.

The bootstrap selected the legacy directory for Python 3.9-3.11 and the current directory for Python 3.12-3.14.

## Why the insertion failed

The Dev18 insertion SymbolCheck found 24 newly introduced native binaries under `etwtrace_legacy`:

- Eight `etwtrace 0.1b8` Windows wheels were merged into the legacy payload.
- Each wheel contains three native extensions: `_etwtrace`, `_etwinstrument`, and `_vsinstrument`.
- The wheels contain no matching PDB files.
- PTVS's symbol archival step only stages explicitly listed PTVS build outputs; it never staged these prebuilt dependency binaries or their symbols.
- The Windows PDB identities embedded in the 24 `.pyd` files are not archived on MSDL.

As a result, the PTVS build and release could succeed while the Visual Studio insertion correctly failed its symbol policy. This is not a transient policy issue or a portable-PDB false positive; the exact Windows PDBs required by the shipped binaries are unavailable to the insertion.

## What would be required to keep Python 3.9-3.11 profiling

Adding `*.pyd` to PTVS's symbol publishing list is not sufficient because the matching PDBs are absent. A compliant legacy payload would require one of the following:

1. Recover the exact PDBs produced alongside the published `0.1b8` wheels and archive each matching `.pyd`/PDB pair to MSDL.
2. If those artifacts no longer exist, produce a new legacy `etwtrace` release from the `0.1b8` line:
   - Build the required CPython 3.9-3.11 architecture matrix in an official pipeline.
   - Retain the exact Windows PDB generated for every `.pyd`.
   - Sign and publish the new wheel binaries.
   - Archive the matching binary/PDB pairs before insertion.
   - Publish the new version to the approved internal feed and update PTVS to consume it.
   - Filter the legacy restore so it does not also package the unnecessary Python 3.12 and 3.13 wheels from `0.1b8`.

That is primarily a producer/release-pipeline investment, not a small PTVS packaging fix. It also preserves a second native payload and compatibility path indefinitely.

## Changes

- Remove the `etwtrace 0.1b8` restore and the `etwtrace_legacy` VSIX payload.
- Package only `etwtrace 0.1b9` for Python 3.12-3.14.
- Remove the legacy collector compatibility path from `diaghub_profile.py`.
- Make the Python bootstrap explicitly reject versions outside 3.12-3.14.
- Probe the selected interpreter version before returning a command to Diagnostics Hub.
  - Python 3.9-3.11 now receives a clear warning with the detected version and supported range.
  - PTVS returns no launch command, so Diagnostics Hub does not create an empty session.
  - An invalid or unreadable interpreter also produces an actionable error instead of starting a session that cannot collect data.
- Keep the Python-side version check as a defense-in-depth guard.

## Why this is the best fix for the exploratory item

- It restores a reliable path for the currently supported Python 3.12-3.14 matrix.
- It avoids bypassing a valid symbol-compliance failure.
- It removes 24 unarchived native binaries and the old collector compatibility path.
- It avoids investing in reconstruction of a 2024 dependency release and its missing symbol-production pipeline for a Dev18 feature that was already unavailable and has no known user complaints.
- It makes the unsupported behavior explicit before collection rather than producing an empty `.diagsession`.
- Legacy support can still be reconsidered independently if usage data or customer demand justifies the upstream build, signing, and symbol archival work.

## User impact

- Dev18 Python profiling supports Python 3.12, 3.13, and 3.14.
- Python 3.9, 3.10, and 3.11 users receive a clear warning before a profiling session starts.
- Editing, running, and debugging older Python versions are unaffected.
- The legacy profiler behavior in older Visual Studio releases is unaffected.

## Manual validation

- Built and deployed PTVS to the Dev18 Canary experimental hive.
- Confirmed an older interpreter is rejected with the new warning and no empty session is created.
- Confirmed Python 3.14 proceeds through profiling and reports user functions when the target defines and calls functions.
- Confirmed the profiling VSIX contains the Python 3.12-3.14 `etwtrace` binaries and no `etwtrace_legacy` entries.

### DIFF
--- a/Python/Product/Profiling/diaghub_profile.py
+++ b/Python/Product/Profiling/diaghub_profile.py
@@ -12,35 +12,8 @@ _CHILD_MARKER = "PTVS_DIAGHUB_PROFILE_CHILD"
 _TARGET_ARGUMENTS = "PTVS_DIAGHUB_TARGET_ARGUMENTS"
 
 
-def _load_legacy_collector():
-    import ctypes
-
-    collector_root = os.getenv("DIAGHUB_INSTR_COLLECTOR_ROOT")
-    runtime_name = os.getenv("DIAGHUB_INSTR_RUNTIME_NAME")
-    if not collector_root or not runtime_name:
-        raise RuntimeError(
-            "Python profiling must be launched from the Visual Studio "
-            "Performance Profiler."
-        )
-
-    if sys.winver.endswith("-32"):
-        architecture = "x86"
-    elif sys.winver.endswith("-arm64"):
-        architecture = "arm64"
-    else:
-        architecture = "amd64"
-
-    collector_path = os.path.join(collector_root, architecture, runtime_name)
-    collector = ctypes.WinDLL(collector_path)
-    collector.ChildAttach.argtypes = []
-    collector.ChildAttach.restype = ctypes.c_int
-    if not collector.ChildAttach():
-        raise RuntimeError("Failed to attach Python to the profiling session.")
-    return collector
-
-
 def _is_supported_version(version):
-    return (3, 9) <= tuple(version[:2]) <= (3, 14)
+    return (3, 12) <= tuple(version[:2]) <= (3, 14)
 
 
 def _has_valid_target(arguments):
@@ -82,7 +55,7 @@ def _run_target(arguments):
 def main():
     if not _is_supported_version(sys.version_info):
         print(
-            "Visual Studio Python profiling supports Python 3.9 through 3.14.",
+            "Visual Studio Python profiling supports Python 3.12 through 3.14.",
             file=sys.stderr,
         )
         return 2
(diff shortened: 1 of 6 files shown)


Python/Product/Profiling/diaghub_profile.py
```diff
@@ -82,7 +55,7 @@ def _run_target(arguments):
 def main():
     if not _is_supported_version(sys.version_info):
         print(
-            "Visual Studio Python profiling supports Python 3.9 through 3.14.",
+            "Visual Studio Python profiling supports Python 3.12 through 3.14.",
```
bschnurr, line 58: should be a loc string

bschnurr (approved):
just the loc string

Python/Product/Profiling/diaghub_profile.py
StellaHuang95: Its printed message is a fallback diagnostic. The normal user-facing unsupported-version message does come from PTVS before the session starts and uses localized `Strings.resx`.

PR #8601, opened 2026-08-04
Restore Pylance document synchronization on Dev18

### LINKED ISSUE #8592: Some modules cannot be found during testing.
**Environment**
Build of Visual Studio: 18.10.0 Canary [12024.283.main]
Build of Python Package: 18.0.26207.1
Version of Pylance: 2026.2.109

**Steps to Reproduce**
1. Create a new python application
2. Input 'import parse'

**Expected behavior**
It should be found 'argparse' in the completion list.

Some modules cannot be found.
Image

Additionally, these are not included in the list either.

Image

Image

Image

Image

### LINKED ISSUE #8593: Quick Info is not working.
**Environment**
Build of Visual Studio: 18.10.0 Canary [12024.283.main]
Build of Python Package: 18.0.26207.1
Version of Pylance: 2026.2.109

**Steps to Reproduce**
1. Create a new python application
2. Input some codes
```
from math import cos, radians
	
# Create a string with spaces proportional to a cosine of x in degrees
def make_dot_string(x):
	rad = radians(x)                             # cos works with radians
	numspaces = int(20 * cos(radians(x)) + 20)   # scale to 0-40 spaces
	st = ' ' * numspaces + 'o'                   # place 'o' after the spaces
	return st
	
def main():
	for i in range(0, 1800, 12):
		s = make_dot_string(i)
		print(s)
	
main()
```
3. Hovering the mouse pointer over an identifier 

**Expected behavior**
It should display a Quick Info tooltip.

Image

Quick Info is not working.

Image

### LINKED ISSUE #8594: Go to Definition is not working.
**Environment**
Build of Visual Studio: 18.10.0 Canary [12024.283.main]
Build of Python Package: 18.0.26207.1
Version of Pylance: 2026.2.109

**Steps to Reproduce**
1. Create a new python application
2. Input some codes
```
from math import cos, radians
	
# Create a string with spaces proportional to a cosine of x in degrees
def make_dot_string(x):
	rad = radians(x)                             # cos works with radians
	numspaces = int(20 * cos(radians(x)) + 20)   # scale to 0-40 spaces
	st = ' ' * numspaces + 'o'                   # place 'o' after the spaces
	return st
	
def main():
	for i in range(0, 1800, 12):
		s = make_dot_string(i)
		print(s)
	
main()
```
3. Right click ‘make_dot_string’， choose go to definition.

**Expected behavior**
It should jump to the definition.

No response after click go to definition.

Image
**Sample Code**

### LINKED ISSUE #8596: Error list is not working.
**Environment**
Build of Visual Studio: 18.10.0 Canary [12024.283.main]
Build of Python Package: 18.0.26207.1
Version of Pylance: 2026.2.109

**Steps to Reproduce**
1. Create a python application project
2. Type import xxxxxxxx which doesn’t exist in default opened file PythonApplication12.py 
3. Green wave-underline is expected under xxxxxxxx 
4. Check that there is a warning about no such module in Error List Window

**Expected behavior**
Green wave-underline is expected under xxxxxxxx and there is a warning about no such module in Error List Window.

There is no warning on error list.
Image

StellaHuang95:
## Summary

- Normalize Pylance's numeric `textDocumentSync` capability into the equivalent options object for affected Dev18 clients.
- Limit the workaround to Pylance and Dev18; other language servers, Visual Studio versions, and post-initialization traffic are unchanged.
- Restore completion, diagnostics, Quick Info, and definition navigation while preserving the System.Text.Json migration from #8587 and automatic completion parentheses.

## Root cause

Pylance returns the valid LSP shorthand `"textDocumentSync": 2`, meaning incremental synchronization. It has used this representation for years, so the recent Pylance update and PTVS #8587 were not the source of the regression.

The regression was introduced by the VSLanguageServerClient generated-capability migration in [commit d97ea9c64](https://devdiv.visualstudio.com/DevDiv/_git/VSLanguageServerClient/commit/d97ea9c64fdae1183d779c5029e311e8e8e4cd70) ([PR 760545](https://devdiv.visualstudio.com/DevDiv/_git/VSLanguageServerClient/pullrequest/760545)). The generated union retains the numeric value as `TextDocumentSyncKind.Incremental`. The `didChange` path accepts either the enum or `TextDocumentSyncOptions`, but the `didOpen`, `didClose`, and `didSave` paths only inspect `TextDocumentSyncOptions`.

Consequently, VSL does not register `didOpen` for Pylance. Because the document is never marked open, subsequent `didChange` notifications are discarded as well. Pylance then has no current editor contents:

- completion requests are answered from stale/incomplete document state, so `import sys` omits the `sys` module;
- diagnostics configured for open files never run, so unresolved imports produce neither squiggles nor Error List entries;
- hover responses are computed against stale document state and their ranges cannot be mapped reliably to the current editor snapshot, so Quick Info is absent or VSL throws an invalid-range exception;
- definition requests are evaluated without the current document, so local symbols may return no usable navigation target.

VSL has an equivalent normalization in [commit ac604b1b8](https://devdiv.visualstudio.com/DevDiv/_git/VSLanguageServerClient/commit/ac604b1b8cf82b698ca0c175cf9de814ee897247), but at the time of investigation that commit was only on `feature/dynamic-registration-redesign`, not `main` or `release/18.10`.

## PTVS workaround

VSL performs the initialize exchange directly, before PTVS language-client middle layers can intercept the response. This change therefore wraps only the Pylance stdout stream on Dev18 and, for the initialize response only, expands:

```json
"textDocumentSync": 2
```

into the semantically equivalent form:

```json
"textDocumentSync": {
  "openClose": true,
  "change": 2,
  "save": { "includeText": false }
}
```

The stream handles normal LSP framing, preserves any notifications preceding the initialize response, recalculates the UTF-8 `Content-Length`, and switches to direct pass-through immediately after initialization. It does not inspect or alter completions, diagnostics, hover or definition responses, document updates, or any later server traffic.

## Compatibility with the VSL fix

The two fixes can safely coexist. Once VSL's fix ships, it will receive the already-expanded, valid options object and use it directly; its numeric-to-object normalization is not applied a second time. This does not create duplicate registrations or change synchronization behavior. The PTVS workaround can be removed later after fixed VSL builds are the only supported Dev18 clients.

## Validation

Validated in the Dev18 Experimental Instance with Pylance `2026.2.109`:

- `import sys` offers `sys` in completion results;
- `import xxxx` produces an editor squiggle and an Error List diagnostic;
- Quick Info appears for Python symbols without an invalid-range exception;
- F12 and **Go To Definition** navigate to local Python definitions;
- automatic completion parentheses still work;
- LSP tracing contains `didOpen`, `didChange`, and `publishDiagnostics` traffic.

No tests were added or modified.

Fixes #8592
Fixes #8593
Fixes #8594
Fixes #8596

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/LanguageServer.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/LanguageServer.cs
@@ -95,7 +95,7 @@ namespace Microsoft.PythonTools.LanguageServerClient {
                 }
 
                 if (!process.HasExited) {
-                    // Create a connection where we wrap the stdin stream so that we can intercept all messages
+                    // Intercept outgoing messages and, on Dev18, normalize the Pylance initialize response.
 #if DEV18
                     var serverOutput = new PylanceInitializeResponseStream(process.StandardOutput.BaseStream);
 #else
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/LanguageServer.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/LanguageServer.cs
@@ -96,8 +96,13 @@ namespace Microsoft.PythonTools.LanguageServerClient {
 
                 if (!process.HasExited) {
                     // Create a connection where we wrap the stdin stream so that we can intercept all messages
+#if DEV18
+                    var serverOutput = new PylanceInitializeResponseStream(process.StandardOutput.BaseStream);
+#else
+                    var serverOutput = process.StandardOutput.BaseStream;
+#endif
                     return new Connection(
-                        process.StandardOutput.BaseStream,
+                        serverOutput,
                         new StreamIntercepter(process.StandardInput.BaseStream, _serverSendHandler, (a) => { }));
                 }
             }
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/PylanceInitializeResponseStream.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/PylanceInitializeResponseStream.cs
@@ -1,0 +1,238 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION, ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for the specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.PythonTools.LanguageServerClient.StreamHacking {
+    /// <summary>
+    /// Normalizes Pylance's numeric text synchronization capability for affected Dev18 LSP clients.
+    /// All server output is passed through unchanged after the initialize response.
+    /// </summary>
+    internal sealed class PylanceInitializeResponseStream : Stream {
+        private const string ContentLengthHeader = "Content-Length:";
+        private const int MaxHeaderLength = 16 * 1024;
+
+        private readonly Stream _baseStream;
+        private readonly List<byte> _inputBuffer = new List<byte>();
+        private readonly SemaphoreSlim _readLock = new SemaphoreSlim(1, 1);
+        private byte[] _pendingOutput = Array.Empty<byte>();
+        private int _pendingOutputOffset;
+        private int _disposed;
+        private bool _initializeResponseProcessed;
+
+        public PylanceInitializeResponseStream(Stream baseStream) {
+            _baseStream = baseStream ?? throw new ArgumentNullException(nameof(baseStream));
+        }
+
+        public override bool CanRead => true;
+        public override bool CanSeek => false;
+        public override bool CanWrite => false;
+        public override long Length => throw new NotSupportedException();
+        public override long Position {
+            get => throw new NotSupportedException();
+            set => throw new NotSupportedException();
+        }
+
+        public override int Read(byte[] buffer, int offset, int count) =>
+            ReadAsync(buffer, offset, count, CancellationToken.None).GetAwaiter().GetResult();
+
+        public override async Task<int> ReadAsync(
+            byte[] buffer,
+            int offset,
+            int count,
+            CancellationToken cancellationToken) {
+            ValidateReadArguments(buffer, offset, count);
+            if (count == 0) {
+                return 0;
+            }
+            ThrowIfDisposed();
+
+            await _readLock.WaitAsync(cancellationToken).ConfigureAwait(false);
+            try {
+                ThrowIfDisposed();
+                if (_pendingOutputOffset >= _pendingOutput.Length) {
+                    if (_initializeResponseProcessed && _inputBuffer.Count == 0) {
+                        return await _baseStream.ReadAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
+                    }
+
+                    _pendingOutput = await ReadNextFrameAsync(cancellationToken).ConfigureAwait(false);
+                    _pendingOutputOffset = 0;
+                    if (!_initializeResponseProcessed) {
+                        _pendingOutput = NormalizeInitializeResponse(_pendingOutput);
+                    }
+                }
+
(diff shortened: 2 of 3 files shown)


Python/Product/PythonTools/PythonTools/LanguageServerClient/LanguageServer.cs
```diff
@@ -96,8 +96,13 @@ public async Task<Connection> ActivateAsync() {
 
                 if (!process.HasExited) {
                     // Create a connection where we wrap the stdin stream so that we can intercept all messages
```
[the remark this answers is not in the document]
StellaHuang95: Agreed and addressed in 2ec41cd44. The comment now describes both directions accurately: outgoing messages are intercepted by the existing stdin wrapper, while Dev18 additionally normalizes the Pylance initialize response on stdout.

Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/PylanceInitializeResponseStream.cs
```diff
@@ -0,0 +1,238 @@
...
+            var message = JObject.Parse(messageJson);
+            if (message["id"] == null || (message["result"] == null && message["error"] == null)) {
+                return frame;
+            }
+
+            _initializeResponseProcessed = true;
+            var textDocumentSync = message["result"]?["capabilities"]?["textDocumentSync"];
+            if (textDocumentSync?.Type != JTokenType.Integer) {
+                return frame;
+            }
```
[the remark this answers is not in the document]
StellaHuang95: Agreed and addressed in 2ec41cd44. The stream now marks initialization processed only after observing a response with `result.capabilities`, the identifying shape of a successful initialize response. Notifications, server requests, and unrelated responses remain byte-for-byte unchanged and no longer stop the search. Once initialize is observed, the stream still switches immediately to direct pass-through, including when `textDocumentSync` is already in object form.

StellaHuang95:
Manual follow-up confirmed that #8593 has the same synchronization root cause. With this PR alone (and without the earlier plaintext-hover experiment), Quick Info displays correctly and the `Invalid range provided by language server` exception no longer occurs. The PR description now includes this behavior and closes #8593.

StellaHuang95:
Manual isolation testing confirmed #8594 has the same document-synchronization root cause. In an Exp instance containing this PR only—and without the separate serializer experiment—F12 and **Go To Definition** both navigate correctly for local Python functions. The PR now documents and closes #8594.

rchiodo:

Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/PylanceInitializeResponseStream.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            var header = Encoding.ASCII.GetString(headerBytes);
+            foreach (var line in header.Split(new[] { "\r\n" }, StringSplitOptions.RemoveEmptyEntries)) {
+                if (line.StartsWith(ContentLengthHeader, StringComparison.OrdinalIgnoreCase) &&
+                    int.TryParse(line.Substring(ContentLengthHeader.Length).Trim(), out var contentLength) &&
+                    contentLength >= 0) {
+                    return contentLength;
+                }
+            }
```
rchiodo, line 211: > **Warning** · Non-blocking recommendation

Please add a focused unit test that passes a framed initialize response through this stream and verifies the rewritten body plus its UTF-8 `Content-Length`. This new framing and rewrite path is self-contained and otherwise has no automated regression coverage.

Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/PylanceInitializeResponseStream.cs
```diff
@@ -0,0 +1,239 @@
...
+        private const string ContentLengthHeader = "Content-Length:";
+        private const int MaxHeaderLength = 16 * 1024;
+
+        private readonly Stream _baseStream;
+        private readonly List<byte> _inputBuffer = new List<byte>();
+        private readonly SemaphoreSlim _readLock = new SemaphoreSlim(1, 1);
+        private byte[] _pendingOutput = Array.Empty<byte>();
+        private int _pendingOutputOffset;
+        private int _disposed;
+        private bool _initializeResponseProcessed;
```
rchiodo, line 40: > **Warning** · Non-blocking recommendation

This introduces a second implementation of LSP framing and `Content-Length` parsing alongside `MessageParser`, with slightly different header handling. Please extract the shared framing rules into a common stream-safe helper so future protocol fixes do not need to be kept in sync manually.

Python/Product/PythonTools/PythonTools/LanguageServerClient/StreamHacking/PylanceInitializeResponseStream.cs
```diff
@@ -0,0 +1,239 @@
...
+                headerBytes[i] = frame[i];
+            }
+
+            var header = Encoding.ASCII.GetString(headerBytes);
+            foreach (var line in header.Split(new[] { "\r\n" }, StringSplitOptions.RemoveEmptyEntries)) {
+                if (line.StartsWith(ContentLengthHeader, StringComparison.OrdinalIgnoreCase) &&
+                    int.TryParse(line.Substring(ContentLengthHeader.Length).Trim(), out var contentLength) &&
+                    contentLength >= 0) {
+                    return contentLength;
+                }
```
rchiodo, line 210: > **Info** · Optional note

Guard this rewrite to the supported `Full` and `Incremental` values. A numeric `textDocumentSync: 0` currently becomes `openClose: true`, which changes the negotiated behavior for a server that explicitly advertised no document synchronization.

rchiodo (approved):
Approved via Review Center.

PR #8603, opened 2026-08-04
Synchronize Pylance project contexts

### LINKED ISSUE #8597: Add folder and Zip Archive to Search Path are not working.
**Environment**
Build of Visual Studio: 18.10.0 Canary [12027.280.main]
Build of Python Package: 18.0.26209.1
Version of Pylance: 2026.2.109

**Steps to Reproduce**
1. Create a new folder and add a new file named Module2.py
2. Input code in Module2.py
```
def summizing(title: str, mock: bool):
    pass

def addcross(a: int, b: int):
    a = a+1
    b = b+1
    return a * b
```
3. Open a new python application and right click Search Paths, choose "Add Folder to Search Paths"
4. Choose the folder which include Module2.py

**Expected behavior**
Type "from", the module2 should in completion list.

Module2 is not in completion list.
Image
Add to Zip Archive to Search Path is also not working.

StellaHuang95:
## Summary

- Keep Pylance client contexts synchronized with the current Python project or workspace lifecycle.
- Refresh `python.analysis.extraPaths` when project search paths change.
- Safely handle project additions/removals, workspace transitions, interpreter changes, and background reanalysis.

## Root cause

PTVS uses `PythonLanguageClientContextProject` and `PythonLanguageClientContextWorkspace` objects to bridge project/workspace state into Pylance configuration. These contexts subscribe to interpreter and search-path changes and supply values such as `python.analysis.extraPaths`.

Before this change, `_clientContexts` was populated once when the language client activated. Later project lifecycle events updated LSP workspace folders, but did not keep `_clientContexts` synchronized. A project added after activation therefore had no `PythonLanguageClientContextProject` subscription. Adding a folder or ZIP archive under that project's **Search Paths** node changed the project, but nothing notified Pylance or refreshed its configuration, so the new module remained unresolved and absent from completion until the language client was recreated.

The context list also crossed UI lifecycle events and background workspace-reanalysis callbacks without a consistent synchronization strategy. Simply adding contexts in one event handler would leave races with reads, removals, disposal, and timer-driven reanalysis.

## Fix

- Give project and workspace contexts stable identities so the active set can be compared and deduplicated.
- Reconcile `_clientContexts` whenever project/workspace lifecycle changes occur rather than only during initial activation.
- Subscribe newly added contexts and dispose/unsubscribe stale contexts.
- Take locked snapshots for readers and mutate the context set under a dedicated lock.
- Perform event unsubscription and disposal outside the lock to avoid blocking or reentrancy while synchronized.
- Refresh Pylance configuration after context reconciliation and search-path changes.
- Preserve workspace-folder updates and handle response cardinality and temporarily unresolved interpreters safely.

The change is PTVS-local and does not modify Pylance's import resolution or Visual Studio's shared language-server client.

## Validation

Validated in the Dev18 Experimental Instance with Pylance `2026.2.109`:

1. Added `Z:\source\PTVS8597SearchPath` through **Add Folder to Search Path** while VS and Pylance remained running.
   - `folder_module` immediately appeared in import completion.
   - `folder_sum` and `FOLDER_SENTINEL` appeared in member completion.
   - The import had no unresolved-import diagnostic.
2. Added `Z:\source\PTVS8597ZipPath.zip` through **Add Zip Archive to Search Path** without restarting.
   - The root-level `zip_module` immediately appeared in import completion.
   - `zip_product` and `ZIP_SENTINEL` appeared in member completion.
   - The import had no unresolved-import diagnostic.
3. The Dev18 Core VSIX build succeeded.

No tests were added or modified.

Fixes #8597

### DIFF
--- a/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
+++ b/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
@@ -86,7 +86,8 @@ namespace Microsoft.PythonTools.LanguageServerClient {
         public static Task ReadyTask => _readyTcs.Task;
 
         private readonly Common.Core.Disposables.DisposableBag _disposables;
-        private List<IPythonLanguageClientContext> _clientContexts = new List<IPythonLanguageClientContext>();
+        private readonly object _clientContextsLock = new object();
+        private readonly List<IPythonLanguageClientContext> _clientContexts = new List<IPythonLanguageClientContext>();
         private PythonAnalysisOptions _analysisOptions;
         private PythonAdvancedEditorOptions _advancedEditorOptions;
         private ITaskList _taskListService;
@@ -100,6 +101,7 @@ namespace Microsoft.PythonTools.LanguageServerClient {
         private static TaskCompletionSource<int> _readyTcs = new TaskCompletionSource<int>();
         private readonly PythonSignatureHelpMiddleLayer _signatureHelpMiddleLayer;
         private bool _loaded = false;
+        private bool _solutionClosing;
         // Set by the dispose action before any other cleanup; consulted by
         // TriggerWorkspaceUpdateConfig and GetSettings so teardown cannot
         // dereference cleared state.
@@ -138,6 +140,11 @@ namespace Microsoft.PythonTools.LanguageServerClient {
                 return null;
             }
             await JoinableTaskContext.Factory.SwitchToMainThreadAsync();
+
+            ProjectContextProvider.ProjectsChanged += OnProjectsChanged;
+            PythonWorkspaceContextProvider.WorkspaceOpening += OnPythonWorkspaceOpening;
+            PythonWorkspaceContextProvider.WorkspaceInitialized += OnPythonWorkspaceInitialized;
+            PythonWorkspaceContextProvider.WorkspaceClosed += OnPythonWorkspaceClosed;
             CreateClientContexts();
 
             _analysisOptions = Site.GetPythonToolsService().AnalysisOptions;
@@ -159,11 +166,10 @@ namespace Microsoft.PythonTools.LanguageServerClient {
                 // tear down fields.
                 _disposed = true;
 
-                _clientContexts.ForEach(c => {
-                    c.InterpreterChanged -= OnInterpreterChanged;
-                    c.SearchPathsChanged -= OnSettingsChanged;
-                    c.ReanalyzeChanged -= OnReanalyzeChanged;
-                });
+                ProjectContextProvider.ProjectsChanged -= OnProjectsChanged;
+                PythonWorkspaceContextProvider.WorkspaceOpening -= OnPythonWorkspaceOpening;
+                PythonWorkspaceContextProvider.WorkspaceInitialized -= OnPythonWorkspaceInitialized;
+                PythonWorkspaceContextProvider.WorkspaceClosed -= OnPythonWorkspaceClosed;
                 _analysisOptions.Changed -= OnSettingsChanged;
                 _advancedEditorOptions.Changed -= OnSettingsChanged;
 
@@ -173,8 +179,9 @@ namespace Microsoft.PythonTools.LanguageServerClient {
                 } catch (ServiceUnavailableException) {
                 }
 
-                _clientContexts.ForEach(c => c.Dispose());
-                _clientContexts.Clear();
+                foreach (var context in GetClientContextsSnapshot()) {
+                    RemoveClientContext(context);
+                }
                 solutionEvents.Opened -= OnSolutionOpened;
                 solutionEvents.ProjectAdded -= OnProjectAdded;
                 solutionEvents.ProjectRemoved -= OnProjectRemoved;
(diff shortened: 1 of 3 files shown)


rchiodo:

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -576,15 +615,114 @@ private void WatchedFilesRegistered(object sender, DidChangeWatchedFilesRegistra
...
+                    changed = true;
                 }
             }
+
+            foreach (var project in projectNodes.Where(p => !existingProjects.Contains(p))) {
+                AddClientContext(new PythonLanguageClientContextProject(project));
+                changed = true;
+            }
+
+            return changed;
```
rchiodo, line 648: > **Warning** · Non-blocking recommendation

Could a queued full synchronization run while `PythonWorkspaceContextProvider.Workspace` is still null during `WorkspaceOpening`? If so, it removes the context just added for `e.Workspace` and leaves configuration empty until initialization. Please ensure reconciliation uses the opening workspace or waits until the provider reflects it.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -576,15 +615,114 @@ private void WatchedFilesRegistered(object sender, DidChangeWatchedFilesRegistra
...
+                await JoinableTaskContext.Factory.SwitchToMainThreadAsync();
+                if (_disposed) {
+                    return;
+                }
+
+                if (SynchronizeClientContexts() && IsInitialized) {
+                    await TriggerWorkspaceUpdateConfig();
+                }
+            }).Task.DoNotWait();
+        }
```
rchiodo, line 694: > **Warning** · Non-blocking recommendation

This fire-and-forget synchronization task has no failure observation at this boundary. A failure after partial reconciliation or while refreshing configuration would be difficult to diagnose and could leave contexts stale. Please ensure failures from the queued task are surfaced consistently with other background operations.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs:L729](https://github.com/microsoft/ptvs/blob/0a030923df86843cafb5c8a6de0c29d40fc91076/Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs#L729).

> **Warning** · Non-blocking recommendation

After closing a workspace, this removes only workspace contexts and does not reconcile project contexts. If a folder-to-project transition can occur without a later `ProjectsChanged` or `SolutionOpened`, no project context is restored. Please confirm that lifecycle guarantee or trigger reconciliation here.

rchiodo (approved):
Approved via Review Center.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -576,15 +615,114 @@ private void WatchedFilesRegistered(object sender, DidChangeWatchedFilesRegistra
...
-                    AddClientContext(n);
+            SynchronizeClientContexts();
+        }
+
+        private bool SynchronizeClientContexts() {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            if (_disposed) {
+                return false;
+            }
```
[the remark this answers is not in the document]
StellaHuang95: Thanks. I agree that automated coverage would be valuable if we had a reliable maintained integration harness for this lifecycle. Given the current PTVS UI/integration-test infrastructure, I validated the end-to-end behavior manually in the Dev18 Experimental Instance instead: adding both a folder and a ZIP search path while Pylance remained running immediately updated import/member completions and removed unresolved-import diagnostics. I am keeping this PR free of new test coverage for that reason.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
StellaHuang95: Thanks for checking this ordering. `PythonWorkspaceContextProvider.InitializeCurrentContext` raises `WorkspaceOpening` synchronously on the UI thread, then assigns `_currentContext` immediately after the event returns; there is no `await` between those operations. `OnPythonWorkspaceOpening` is also synchronous. A queued reconciliation must switch to the same UI thread, so it cannot interleave between adding the `e.Workspace` context and the provider assigning that workspace. If it runs before `WorkspaceOpening`, there is no newly added context to remove; if it runs afterward, the provider already returns the new workspace. Therefore no code change is needed here.

Python/Product/PythonTools/PythonTools/LanguageServerClient/PythonLanguageClient.cs
StellaHuang95: Thanks. The task is observed at this boundary through PTVS's `TaskExtensions.DoNotWait()`. That helper attaches a continuation, observes `Task.Exception`, and rethrows unexpected faults through the captured synchronization context (or the thread pool when there is no specialized context). It suppresses only cancellation and the explicitly recognized teardown failures (`ObjectDisposedException` and `ConnectionLostException`). This is the existing PTVS fire-and-forget error-reporting pattern, so adding another observer or catch here would be redundant and could change failure semantics.

StellaHuang95:
Regarding the workspace-close transition feedback: I confirmed that project contexts are restored through the existing authoritative lifecycle events. A folder-to-solution transition subsequently raises `SolutionOpened`, `ProjectAdded`, and/or `VsProjectContextProvider.ProjectsChanged`; each queues full reconciliation from the provider's current project-node snapshot. `WorkspaceClosed` only removes the matching workspace context because the provider still exposes the closing context while raising that event. Queuing unconditional reconciliation there would be undesirable during a folder-to-folder switch: it could run while the provider is between contexts and transiently create project contexts before the replacement `WorkspaceOpening` removes them. The current event division avoids that intermediate state, so I am leaving the code unchanged.

PR #8611, opened 2026-08-13
Add CPython 3.15 support to mixed-mode debugging

rchiodo:
## Summary

Adds CPython 3.15 support to the mixed-mode (Python/Native) debugger, building on the 3.14 work in #8586. It makes the `_Py_DebugOffsets` reader **version-aware** so it can parse both the 3.14 and 3.15 offset-table layouts, and adds the `V315` language version so the 3.14+ code paths (`_PyStackRef` masking, in-process line-number decoding) fire for 3.15.

## Background

CPython 3.15 keeps the self-describing `_Py_DebugOffsets` table introduced/formalized in 3.14, but the **flat layout of the table grew**, which breaks a parser that assumes the fixed 3.14 layout:

- `thread_state` gained 7 fields (3 inserted after `current_frame`: `base_frame`, `last_profiled_frame`, `last_profiled_frame_seq`; 4 appended after `status`: `holds_gil`, `gil_requested`, `current_exception`, `exc_state`) — 9 → 16 fields.
- A new `err_stackitem` group (`exc_value`) is inserted between `thread_state` and `interpreter_frame`.
- `type_object` gained `tp_basicsize` / `tp_dictoffset`; a new `heap_type_object` group (`size`, `ht_cached_keys`) follows it.
- `unicode_object` gained `compactunicodeobject_size`; `gc` gained `frame`, `generation_stats_size`, `generation_stats`.

Because groups after `thread_state` shift, the flat offsets of `interpreter_frame` / `code_object` in the table differ between 3.14 (table size 760) and 3.15 (table size 888). The parser must select the layout by version.

The interpreter-frame field offsets, `_PyStackRef` decode (low-bit tag mask), and the `co_linetable` format are **unchanged** in production 3.15 builds, so the existing 3.14+ masking and line-number code applies to 3.15 as-is.

## What this does

**Version-aware `_Py_DebugOffsets` reader:**
- `PyDebugOffsets` now parses the header/version first, then selects the matching layout (`Layout314`, `Layout315`) built from shared field arrays plus version-specific arrays, and reads that layout's table size. Unknown versions are rejected.
- `Is314` is joined by `Is315`, and callers gate on the new `IsSupported` (3.14 **or** 3.15) instead of `Is314`. `TableSizeFor(major, minor)` replaces the single static table size; `TryRead` reads the larger `MaxTableSize`.

**Language-version wiring (required):**
- `PythonLanguageVersion.V315 = 0x030f` and `GetPythonLanguageVersion` handles `"315"`. Without this, `LanguageVersion` would be `None`, the `>= V314` localsplus / `f_executable` stackref masks would never fire, and `Debug.Assert(version != None)` would trip — producing empty locals / broken frames exactly like pre-3.14.

**No hot-path map change:**
- `DebugOffsetsFieldProvider` needs no mapping change: every hot-path struct/field it reads (`_PyInterpreterFrame`, `PyCodeObject`, `_ts`) exists in both layouts; only its doc comment was refreshed.

## Files

- `Proxies/Structs/PyDebugOffsets.cs` — version-aware layout (`Layout314` 760B / `Layout315` 888B), `GetLayout`, `Is315`, `IsSupported`, `TableSizeFor`, version-selecting `TryParse`, `TryRead` reads `MaxTableSize`.
- `PythonRuntimeInfo.cs` — `case "315"` → `V315`; `StructFieldOffsetProvider` gates on `IsSupported`.
- `Common/Parsing/PythonLanguageVersion.cs` — `V315 = 0x030f`.
- `Proxies/Structs/DebugOffsetsFieldProvider.cs` — doc comment only (map unchanged).
- `Tests/DebuggerTests/PyDebugOffsetsTests.cs`, `PyDebugOffsetsProviderTests.cs` — `TableSizeFor`; **real recorded 3.15.0rc1 vectors** (standard + free-threaded) with a full expected-offset table; a synthetic 3.15 ordinal test retained as an independent structural check; unknown-version rejection.

## Testing

- Unit tests pass (26/26 parser + provider tests).
- The 3.15 coverage runs against **real `_Py_DebugOffsets` tables recorded in-process from CPython 3.15.0rc1** — both the standard `python315.dll` and the free-threaded `python315t.dll` build (888-byte tables, version `0x30f00c1`). The vectors were captured with `ctypes` `id()`-based semantic cross-checks (e.g. `ob_type`, `co_linetable`, `co_firstlineno`, `tuple` length, `str` length) so the recorded offsets are verified against live objects, not just self-consistent. The reader then parses those bytes back to the exact offsets the interpreter reports.
- The free-threaded vector demonstrates the expected layout shifts the table is designed to absorb: `co_linetable` 136→152, `co_firstlineno` 68→84, `ob_type` 8→24, and the TLBC fields (`tlbc_index`, `co_tlbc`, `tlbc_generation`) non-zero only in the free-threaded build.
- Manually verified end-to-end against the `Examples/PythonNative` C++ app embedding CPython 3.15 in VS 18: call stack, line numbers, stepping between Python and native code, and the Locals/Autos/Watch/globals windows (including nested frames) all work.

## Notes

- As with 3.14, only the mixed-mode hot path is driven by the table; struct sizes and all non-hot-path fields still come from the PDB. Free-threaded (3.15t) layout shifts are covered by the table automatically.
- A synthetic 3.15 ordinal test is retained alongside the real vectors: it pins each group/field to the exact ordinal the CPython 3.15 header dictates, independent of the recorded values.

### DIFF
--- a/Python/Product/Debugger.Concord/Proxies/Structs/PyDebugOffsets.cs
+++ b/Python/Product/Debugger.Concord/Proxies/Structs/PyDebugOffsets.cs
@@ -21,16 +21,19 @@ using Microsoft.VisualStudio.Debugger.Native;
 namespace Microsoft.PythonTools.Debugger.Concord.Proxies.Structs {
     /// <summary>
     /// Managed reader for CPython's self-describing <c>_Py_DebugOffsets</c> table
-    /// (CPython 3.14, <c>Include/internal/pycore_debug_offsets.h</c>). The table is placed
+    /// (<c>Include/internal/pycore_debug_offsets.h</c>). The table is placed
     /// at offset 0 of the exported <c>_PyRuntime</c> global so out-of-process debuggers can
     /// discover the byte offset of interesting fields (frames, code objects, thread state,
     /// builtin object layouts, ...) without relying on the interpreter's PDB.
     ///
     /// The table opens with an 8-byte <c>"xdebugpy"</c> cookie, a <c>PY_VERSION_HEX</c>
     /// version word and a <c>free_threaded</c> flag, followed by a flat run of little-endian
-    /// <c>uint64_t</c> offsets grouped by struct. This class only knows the 3.14 layout; the
-    /// header explicitly warns the layout is not stable across minor versions, so callers must
-    /// gate on <see cref="Is314"/> before trusting anything but the cookie/version prefix.
+    /// <c>uint64_t</c> offsets grouped by struct. There is no per-group length in the blob, so the
+    /// byte position of every field is implied purely by the ordered list of groups/fields for that
+    /// interpreter version. CPython grows this table between minor versions (adding fields and whole
+    /// sub-structs), so this class keeps one ordered <see cref="Layout"/> per version it understands
+    /// (3.14 and 3.15) and selects the matching one from the recorded version word. Callers should
+    /// gate on <see cref="IsSupported"/> before trusting any offset.
     /// </summary>
     internal sealed class PyDebugOffsets {
         public const string CookieString = "xdebugpy";
@@ -39,9 +42,12 @@ namespace Microsoft.PythonTools.Debugger.Concord.Proxies.Structs {
         private const int CookieSize = 8;
         private const int HeaderSize = CookieSize + sizeof(ulong) + sizeof(ulong);
 
-        // The 3.14 layout, mirroring pycore_debug_offsets.h field-for-field (every entry is
-        // a uint64 offset). Keeping the groups/fields in exact source order is what makes the
-        // byte positions correct by construction; do not reorder without matching the header.
+        // _Py_DebugOffsets is a flat run of grouped uint64 offsets with no per-group length, so the
+        // byte position of every field is implied by the ordered list of groups/fields. These lists
+        // must therefore mirror pycore_debug_offsets.h field-for-field, in exact source order, for
+        // each interpreter version we support. Do not reorder without matching the header.
+
+        // Groups whose field set is identical across every layout we support (3.14 and 3.15).
         private static readonly string[] RuntimeStateFields = {
             "size", "finalizing", "interpreters_head",
         };
@@ -51,10 +57,6 @@ namespace Microsoft.PythonTools.Debugger.Concord.Proxies.Structs {
             "gil_runtime_state", "gil_runtime_state_enabled",
             "gil_runtime_state_locked", "gil_runtime_state_holder",
             "code_object_generation", "tlbc_generation",
-        };
-        private static readonly string[] ThreadStateFields = {
-            "size", "prev", "next", "interp", "current_frame", "thread_id",
-            "native_thread_id", "datastack_chunk", "status",
         };
         private static readonly string[] InterpreterFrameFields = {
             "size", "previous", "executable", "instr_ptr", "localsplus", "owner",
@@ -67,9 +69,6 @@ namespace Microsoft.PythonTools.Debugger.Concord.Proxies.Structs {
         };
         private static readonly string[] PyObjectFields = {
             "size", "ob_type",
-        };
-        private static readonly string[] TypeObjectFields = {
-            "size", "tp_name", "tp_repr", "tp_flags",
         };
         private static readonly string[] TupleObjectFields = {
             "size", "ob_item", "ob_size",
@@ -92,12 +91,6 @@ namespace Microsoft.PythonTools.Debugger.Concord.Proxies.Structs {
         private static readonly string[] BytesObjectFields = {
             "size", "ob_size", "ob_sval",
         };
-        private static readonly string[] UnicodeObjectFields = {
-            "size", "state", "length", "asciiobject_size",
-        };
-        private static readonly string[] GcFields = {
-            "size", "collecting",
-        };
         private static readonly string[] GenObjectFields = {
             "size", "gi_name", "gi_iframe", "gi_frame_state",
         };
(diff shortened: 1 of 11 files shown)


Python/Product/Debugger.Concord/PythonRuntimeInfo.cs
```diff
@@ -80,6 +80,7 @@ public static PythonLanguageVersion GetPythonLanguageVersion(DkmNativeModuleInst
                 case "312": return PythonLanguageVersion.V312;
                 case "313": return PythonLanguageVersion.V313;
                 case "314": return PythonLanguageVersion.V314;
+                case "315": return PythonLanguageVersion.V315;
```
heejaechang: > **Warning** · Non-blocking recommendation

Add a regression test for `python315.dll` version detection and the resulting 3.14+ stack-reference path; the synthetic offset-table tests do not exercise that runtime wiring. [verified]

heejaechang (approved):
Approved via Review Center.

Python/Product/Debugger.Concord/PythonRuntimeInfo.cs
heejaechang: > **Issue** · Please address or respond

The `V315` case remains unreachable for free-threaded builds because the module-name regex rejects `python315t.dll` and `python315t_d.dll`. Accept the `t` suffix and add detection coverage for standard, debug, free-threaded, and free-threaded-debug names.

Python/Product/Debugger.Concord/Proxies/Structs/PyDebugOffsets.cs
```diff
@@ -144,18 +242,26 @@ internal sealed class PyDebugOffsets {
         public int Minor => (int)((Version >> 16) & 0xFF);
```
heejaechang, line 250: > **Warning** · Non-blocking recommendation

Derive `MaxTableSize` from `KnownLayouts`; the explicit `Math.Max(Layout314.TableSize, Layout315.TableSize)` creates a second registry that can silently truncate live reads when another larger layout is added.

Python/Tests/DebuggerTests/PyDebugOffsetsProviderTests.cs
```diff
@@ -117,5 +119,98 @@ public void ReturnsFalse_ForUnmappedFieldsAndStructs() {
...
+        [TestMethod, Priority(0)]
+        public void Standard315_MapsCodeObjectHotPath() {
+            var provider = Provider(PyDebugOffsetsTests.RawV315);
+            Assert.AreEqual(112L, Offset(provider, "PyCodeObject", "co_filename"));
+            Assert.AreEqual(120L, Offset(provider, "PyCodeObject", "co_name"));
+            Assert.AreEqual(68L, Offset(provider, "PyCodeObject", "co_firstlineno"));
+            Assert.AreEqual(96L, Offset(provider, "PyCodeObject", "co_localsplusnames"));
+            Assert.AreEqual(104L, Offset(provider, "PyCodeObject", "co_localspluskinds"));
+            Assert.AreEqual(208L, Offset(provider, "PyCodeObject", "co_code_adaptive"));
+            Assert.AreEqual(136L, Offset(provider, "PyCodeObject", "co_linetable"));
```
heejaechang, line 142: > **Info** · Optional note

This duplicates the cookie, version, header-size, and ordinal-payload construction in `BuildSynthetic315`. Move the synthetic-table construction into a shared test helper so the parser and provider tests cannot drift.

rchiodo: Fixed in a76e643d8 — the provider test now reuses `PyDebugOffsetsTests.BuildSynthetic315()` rather than reconstructing the cookie, version, header, and ordinal payload independently.

Python/Product/Debugger.Concord/Proxies/Structs/PyDebugOffsets.cs
rchiodo: Fixed in a76e643d8 — `MaxTableSize` is now computed by iterating `KnownLayouts`, so adding a larger layout cannot leave live reads undersized.

Python/Tests/DebuggerTests/PythonRuntimeInfoTests.cs
```diff
@@ -0,0 +1,51 @@
...
+                "python315.dll",
+                "python315_d.dll",
+                "python315t.dll",
+                "python315t_d.dll",
+            };
+
+            foreach (var moduleName in moduleNames) {
+                var version = PythonDLLs.GetPythonLanguageVersion(moduleName);
+                Assert.AreEqual(PythonLanguageVersion.V315, version, moduleName);
+                Assert.IsTrue(version >= PythonLanguageVersion.V314, moduleName);
```
heejaechang: > **Warning** · Non-blocking recommendation

📍 Python/Tests/DebuggerTests/PythonRuntimeInfoTests.cs:36
The prior integration-coverage recommendation remains partially unresolved: this assertion repeats the `>= V314` policy but does not exercise a stack-reference consumer. Add a test using a tagged 3.15 `f_executable` pointer and verify that the consumer clears the `0x3` tag bits.

[verified]

heejaechang (approved):

StellaHuang95 (approved):
Approved via Review Center.

Python/Tests/DebuggerTests/PythonRuntimeInfoTests.cs
rchiodo: Partially addressed in ea9aa1ce9 — the test no longer repeats only the `>= V314` predicate. `PyInterpreterFrame.f_code` and the locals consumer now share `GetStackReferenceTagMask`, and coverage feeds tagged/untagged values through the same production bit-removal helper for 3.13, 3.14, and 3.15. I did not add a live `PointerProxy.Read()`/`f_code` dereference test because `DebuggerTests` has no `DkmProcess` memory harness; that boundary requires a Glass integration test rather than a deterministic unit test.

Python/Tests/DebuggerTests/PythonRuntimeInfoTests.cs
```diff
@@ -0,0 +1,67 @@
...
+                PointerProxy.RemoveTagBits(taggedExecutable,
+                    PyInterpreterFrame.GetStackReferenceTagMask(PythonLanguageVersion.V313)));
+            Assert.AreEqual(0x12345678UL,
+                PointerProxy.RemoveTagBits(taggedExecutable,
+                    PyInterpreterFrame.GetStackReferenceTagMask(PythonLanguageVersion.V314)));
+            Assert.AreEqual(0x12345678UL,
+                PointerProxy.RemoveTagBits(taggedExecutable,
+                    PyInterpreterFrame.GetStackReferenceTagMask(PythonLanguageVersion.V315)));
+        }
+
```
heejaechang, line 55: > **Warning** · Non-blocking recommendation

This verifies the extracted helpers directly, so it would still pass if `f_code`, tagged pointer reads, or localsplus stopped applying them. Add the closest feasible consumer-level or Glass integration test that feeds a tagged 3.15 value through the production dereference path.

[verified]

heejaechang (approved):

StellaHuang95 (approved):

PR #8613, opened 2026-08-18
Fix DebuggerTests engine reference

bschnurr:
## Summary
- add a direct `Microsoft.VisualStudio.Debugger.Engine` reference to `DebuggerTests`
- resolve CS0012 for `DkmNativeModuleInstance` exposed by the `PythonDLLs.GetPythonLanguageVersion` API

## Validation
- `git diff --check`
- targeted VS 18 build progressed without the reported CS0012, but the local dependency build is blocked earlier by an unrelated missing `System.Text.Json` reference in `PythonTools`


StellaHuang95 (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

PR #8616, opened 2026-08-25
Fix NullReferenceException crash in Concord debugger's PyFrameObject.IsInEvalFrame

rchiodo:
## Problem

Visual Studio (`devenv.exe`) crashes during mixed-mode Python debugging with a `System.NullReferenceException` in the Concord debugger, surfacing as an access violation (`c0000005`). The crash occurs while walking the call stack, so it takes down the IDE as soon as the user breaks in or opens the Call Stack window.

Stack from the crash dump:

```
Microsoft.PythonTools.Debugger.Concord.Proxies.Structs.PyFrameObject.IsInEvalFrame+0x8f
Microsoft.PythonTools.Debugger.Concord.Proxies.Structs.PyFrameObject.TryCreate+0xb2
Microsoft.PythonTools.Debugger.Concord.CallStackFilter.FilterNextFrame+0x30e
Microsoft.PythonTools.Debugger.Concord.LocalComponent.IDkmCallStackFilter.FilterNextFrame+0x53
Microsoft.VisualStudio.Debugger.EntryPoint.IDkmCallStackFilter_FilterNextFrame+0x9a
```

## Root cause

The faulting instruction pair makes the cause unambiguous:

```asm
00007ffd`5495a46b  mov rcx, qword ptr [rsi+48h]   ; rcx = frame.BasicSymbolInfo -> NULL
00007ffd`5495a46f  mov rcx, qword ptr [rcx+8]     ; ...MethodName -> reads [0x8] => AV
```

The exception record's `Parameter[1]: 0x8` matches the field offset off a null base exactly, and `READ_ADDRESS: 0000000000000008` confirms it.

`DkmStackWalkFrame.BasicSymbolInfo` is nullable — it is null whenever no symbol information can be resolved for a frame (modules without PDBs, stripped or optimized native modules, and synthesized/unwound frames). The previous code dereferenced it unconditionally:

```csharp
var addressMatch = frame.InstructionAddress.IsInSameFunction(...);
var nameMatch = frame.BasicSymbolInfo.MethodName == name;   // unguarded
return addressMatch || nameMatch;
```

Because both operands were eagerly computed into locals *before* the `||`, short-circuiting never protected the dereference — the crash happened even when `addressMatch` was already `true`. Any Python stack walk containing a symbol-less native frame (e.g. a third-party C extension without PDBs) crashes VS.

## Fix

Return early on the address match — which also avoids needless symbol work in the common case — and null-guard the method name comparison.

## Tests

The `Dkm*` SDK types are sealed with no public constructors, and `Microsoft.VisualStudio.Debugger.Engine.dll` in this repo is a reference assembly that cannot be loaded for execution, so `IsInEvalFrame` itself cannot be unit tested directly. (This is why the existing `DebuggerTests` Concord tests only cover pure static helpers.)

To get real regression coverage without introducing a mocking framework, the version-dependent name logic is extracted into two pure `internal static` helpers, `GetEvalFrameFunctionName` and `IsEvalFrameName`, which `IsInEvalFrame` now delegates to. `Debugger.Concord` already grants `InternalsVisibleTo("DebuggerTests")`, so no project changes were needed.

Three tests added to `PythonRuntimeInfoTests`:

- `IsEvalFrameNameHandlesFramesWithoutSymbols` — the regression guard: a null method name returns `false` instead of throwing.
- `IsEvalFrameNameMatchesEvalFrameFunction` — positive matches, and pins that the names are version-specific (`_PyEval_EvalFrameDefault` must not match under 3.5, and vice versa).
- `EvalFrameFunctionNameDependsOnLanguageVersion` — locks the 3.5/3.6 boundary.

All 6 tests in `PythonRuntimeInfoTests` pass, including the 3 pre-existing ones:

```
Passed RecognizesPython315ModuleNames
Passed StackReferenceHelpersClearTagBitsByVersion
Passed RecognizesFreeThreadedModuleNamesSincePython313
Passed EvalFrameFunctionNameDependsOnLanguageVersion
Passed IsEvalFrameNameMatchesEvalFrameFunction
Passed IsEvalFrameNameHandlesFramesWithoutSymbols

Total tests: 6  Passed: 6
```

## Notes for reviewers

- Behavior is otherwise unchanged: `addressMatch || nameMatch` semantics are preserved, just with genuine short-circuiting.
- The null check is intentionally belt-and-braces — `IsEvalFrameName` checks for null internally *and* the call site uses `?.` — so the helper stays safe regardless of how future callers use it.
- The fix was verified against the crash dump rather than a live repro. Reproducing this manually requires a narrow window: Python's own symbols must load (otherwise the Concord runtime never initializes and `GetFunctionAddress` returns 0, bailing out early), while the *specific frame* being filtered must lack symbol info. A third-party C extension without PDBs is the natural trigger.


StellaHuang95 (approved):
Approved via Review Center.

bschnurr (approved):

